PR #1, opened 2025-06-01
[Build] Upgrade cmake minimum version to be compatible with cmake 4.x

hughperkins:
Pulling these commits in from https://github.com/taichi-dev/taichi/pull/8703

### DIFF
--- a/cmake/TaichiCAPI.cmake
+++ b/cmake/TaichiCAPI.cmake
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.17)
+cmake_minimum_required(VERSION 3.11)
 
 # This function creates a static target from OBJECT_TARGET, then link TARGET with the static target
 #
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ The language has broad applications spanning real-time physical simulation, nume
   - Apple Metal
   - WebAssembly (experiemental)
 - Other packages:
-  - cmake >= 3.17.1
+  - cmake >= 3.11.0
  </details>
 
 Use Python's package installer **pip** to install Taichi Lang:
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -1,0 +1,1 @@
+We are setting cmake minimum version to 3.11, because it's the minimum version that is compatible with the cmake 4.0.1 that Mac OS X brew brings down by default.
--- a/cmake/TaichiCAPITests.cmake
+++ b/cmake/TaichiCAPITests.cmake
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.17)
+cmake_minimum_required(VERSION 3.11)
 
 set(C_API_TESTS_NAME taichi_c_api_tests)
 if (WIN32)
--- a/cmake/TaichiExamples.cmake
+++ b/cmake/TaichiExamples.cmake
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.17)
+cmake_minimum_required(VERSION 3.11)
 
 function(add_taichi_example NAME)
 set(TARGET_NAME "cpp_examples_${NAME}")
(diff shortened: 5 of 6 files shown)


cmake/TaichiCAPI.cmake
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.17)
```
duburcqa: I know you are against comments, but I really feel it would be helpful to document why this specific version is being enforced. At least a like pointing to the original PR?

hughperkins: I don't know why this specific vesion is being enforced, but the benefit of upgrading is that if we have minimum 3.0, then cmake refuses to run on machines with cmake 4.0. I'm not 100% clear that upgrdaing to 3.17 is sufficient to fix that issue (havent tested to be honest). Happy to check that point, and report that here.

duburcqa: Yes, I found this information in the original issue, but I feel like it should be mentioned somewhere in the code to prevent forgetting about it.

hughperkins: with 3.0, it fails, running with 4.0.1:
Screenshot 2025-06-02 at 7 21 09 AM

with 3.17, it passes, running with 4.0.1:
Screenshot 2025-06-02 at 7 21 26 AM

I guess I can add a comment about that

hughperkins: do we really need a comment about every upgrade though? Like, why would someone ever regress the version? We'll have similar issues in our pyproject.toml/requriements.txt files I feel?

hughperkins: You can easily find why the version was changed by using `git blame`, then that will link directly to this PR. I feel that is sufficient?

hughperkins: Example of git blame https://github.com/genesis-company/taichi/blame/b9ba2f41955b942fff15c1fd26ce273c52d5b75c/cmake/TaichiTests.cmake

hughperkins: Screenshot 2025-06-02 at 7 25 53 AM

duburcqa: > do we really need a comment about every upgrade though?

No necessarily. It is up to you. Just stating that the only known reason for this lower bound is supporting both CMake 3 and Cmake 4 is fine for me. I like to keep track of the changes but it is just a hobby I would say. I like knowing the history of the projects I'm relying on (see this [example](https://github.com/duburcqa/jiminy/blob/accdf3629d9d8c08bb97f2905f010f1616d2168a/python/jiminy_py/setup.py#L92)).

duburcqa: Having to use `git blame` and go through the whole history of changes to determine whether a given line of code is associated with a well-motivated PRs or just some random noise sounds painful, even if doable of course.

hughperkins: Would prefer to keep the code clean please.

Realistically, the only reason we'd want to know the history of a dependency in pyproject.toml, or this version number, is if it breaks stuff in the first few weeks after the change, and we want to know things like, why was it changed, and what was the previous version number.

I doubt we're ever going to want to look at the previous previous version, and once the version is in prod for a month or two, we aren't going to want to know what was the previous version either I feel.

duburcqa: This is definitely not how open source works. You need to keep requirements as lax as possible while maintaining correctness to make sure it fits everyone’s need. There is no other way to do this than keeping track of why did you pick given lower and upper bounds for each dependency.

hughperkins: Looks like putting minimum version 3.11 is compatiable with 4.0.1. I'll update to 3.11.

hughperkins: Modified to 3.11, and added a readme into cmake folder explaining why taht version.

PR #3, opened 2025-06-02
[Build] Linux x86 runner

hughperkins:
Issue: #

### Brief Summary

Add linux x86 runner
- runs on github standard runner
- runs on any branch
- for now it is 'run on demand', though I could update it to run on every commit, to a PR, eg could update that in this PR itself if we want

Notes:
- Currently this doesnt include changes for tests to run cleanly (though I realized just now that the opengl changes are needed only for being able to run tests withou seg faulting I think, so we could kick that change ouf of this PR if we want?)
- I found I needed to update changelog.py, in order to not crash when not making a release.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/docs/lang/articles/reference/operator.md
+++ b/docs/lang/articles/reference/operator.md
@@ -56,9 +56,8 @@ Python3 distinguishes `/` (true division) and `//` (floor division), e.g., `1.0 
   operands to the default integral type.
 
 To avoid such implicit casting, you can manually cast your operands to
-desired types, using `ti.cast`. Please see
-[Default precisions](#default-precisions) for more details on
-default numerical types.
+desired types, using `ti.cast`. You can control yourself the precision instead of using the default, see
+[Default precisions](#default-precisions).
 
 Taichi also provides `ti.raw_div` function which performs true division if one of the operands is floating point type
 and performs floor division if both operands are integral types.


.github/workflows/scripts/ti_build/dep.py
```diff
@@ -150,9 +150,9 @@ def download_dep(url, outdir, *, strip=0, force=False, args=None, plain=False, e
     if name.endswith(".zip"):
         outdir.mkdir(parents=True, exist_ok=True)
         unzip(local_cached, outdir, strip=strip)
-    elif name.endswith(".tar.gz") or name.endswith(".tgz"):
+    elif name.endswith(".tar.gz") or name.endswith(".tgz") or name.endswith(".xz") or name.endswith(".tar.bz2"):
```
duburcqa: may be worth doing `any(name.endswith(ext) for ext in (".tar.gz", ".tgz", ".xz", ".tar.bz2"))` at this point.

.github/workflows/scripts/ti_build/dep.py
```diff
@@ -150,9 +150,9 @@ def download_dep(url, outdir, *, strip=0, force=False, args=None, plain=False, e
...
         outdir.mkdir(parents=True, exist_ok=True)
-        tar("-xzf", local_cached, "-C", outdir, f"--strip-components={strip}")
+        tar("-xf", local_cached, "-C", outdir, f"--strip-components={strip}")
```
duburcqa, line 155: why changing this?

cherry_linx86.sh
```diff
@@ -0,0 +1,10 @@
+set -ex
```
duburcqa: What is this? What are these commits? What are you trying to achieve? Documentation is missing.

hughperkins: whoops. Removed. Thanks :)

.github/workflows/scripts/ti_build/dep.py
hughperkins: The formats are given by letters:
- `-z` means `gzip format`
- `-j` => .bz2
- `-J` => .xz

On my mac, apparently `tar -xzf` does work on a `tar.xz` file. However, on linux this is not the case:
```
hugh@hugh-Intel-R-Core-TM-i9-14900KF:/tmp/foo$ tar -xzf workflows.tar.xz

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
hugh@hugh-Intel-R-Core-TM-i9-14900KF:/tmp/foo$ tar -xf workflows.tar.xz
hugh@hugh-Intel-R-Core-TM-i9-14900KF:/tmp/foo$
```

Without the option, it auto-detects, so works no matter which format. If we tell it 'this is defintiely a .gz file' then it fails on an `.xz` file.

.github/workflows/scripts/ti_build/dep.py
hughperkins: addressed in a00aa84

.github/workflows/linux_x86.yml
```diff
@@ -0,0 +1,14 @@
...
+      - main
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Linux x86 Build
+        run: |
+          bash .github/workflows/scripts_new/linux_x86.sh
```
duburcqa: I do get the point of putting the CI script in a dedicated file instead of just pasting itself contain here. If the script is only used for CI, avoid unnecessary layers for indirection that are just obfuscating the logic.

.github/workflows/scripts/ti_build/dep.py
```diff
@@ -150,9 +150,9 @@ def download_dep(url, outdir, *, strip=0, force=False, args=None, plain=False, e
     if name.endswith(".zip"):
         outdir.mkdir(parents=True, exist_ok=True)
         unzip(local_cached, outdir, strip=strip)
-    elif name.endswith(".tar.gz") or name.endswith(".tgz"):
+    elif any(name.endswith(ext) for ext in [".tar.gz", ".tgz", ".xz", ".tar.bz2"]):
```
duburcqa, line 153: No need to change here, but I recommend immutable tuple instead of mutable list if mutability is irrelevant. better use the minimal required container, this makes the code more self-explanatory and may avoid silly mistakes (i.e. default values for method in Python).

.github/workflows/linux_x86.yml
CharlesMasson: It would also allow splitting it into multiple steps or jobs for better modularity and readability of failures.

hughperkins: Ok, so my understanding of the desires here are:
- nice to see multiple steps
- nice to be able to track progress across multiple steps in UI, eg something like https://github.com/duburcqa/jiminy/actions/runs/14775364707/job/41482561535

I in my turn personally like to factorize code out of the runner configuration because:
- decouples code from runner configuration
- means I can run/test the code easily, on arbitrary platforms, without needing to somehow trigger full CI flow etc
    - eg spin up a container, try running the script
    - run on my mac
    - etc ...
- personally, I also find it easier to format code in standalone bash scripts, without needing to indent, pay attention to various yaml quirks etc
- lastly, I feel that we shouldn't require potential developers to be able to read github runner config files, when setting up their development environment
    - yet I feel that the scripts used by CI are the best 'golden source' for what works today, when someone wants to install

To comply with all the above 😓 , what I have done is:
- split the runner configuration into 4 steps:
    - prerequisites
    - build
    - install
    - test
- split the script corresondingly into 4 parts, one part for each step
- added documentation to the dev install page, linking to each of these steps, as a reference

Now I admit that we should ideally turn on the markup link checker, to run across all .md files, systematically, even if they didn't change, since someone could rename one of these scripts, and break the build page 😓  Well, baby steps. We do at least have the link checker running, and we'll turn it on, to run across all .md files, systematically, in a bit.

.github/workflows/linux_x86.yml
```diff
@@ -0,0 +1,23 @@
...
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Linux x86 prerequisites
+        run: |
+          bash .github/workflows/scripts_new/linux_x86/1_prerequisites.sh
+      - name: Linux x86 build
+        run: |
+          bash .github/workflows/scripts_new/linux_x86/2_build.sh
```
duburcqa, line 17: You should be able to remove `bash` prefix for all these lines. You are already running in bash.

tests/python/test_ad_ndarray.py
```diff
@@ -685,6 +685,7 @@ def test_stacked_inner_loops_local_variable(
     assert x.grad[None] == 38.0
 
 
+@pytest.mark.flaky(reruns=5, delay=0.1)
```
duburcqa: same, delay is necessary?

.github/workflows/linux_x86.yml
hughperkins: I have a preference to keep the `bash`:
- it reduces the number of failure modes due to permissions not being set, for various reasons
- it makes it very clear we are using bash (and not sh, or something else)

I would say it's analagous to our using `python` to run python scripts. We could use a shebang on those too. And occasionally we do, when we want it to run from the PATH, as a command. But in general we use `python` to run `python` scripts.

Note that in either case, a new `bash` process is launched: there is no benefits in speed, or memory, in not adding `bash` prefix.

What benefits do you perceive in removing the `bash` prefix?

tests/python/test_ad_ndarray.py
hughperkins: I'm not sure. Some of them have delay, some of them have not. We are randomly sampling configurations, and we can see which ones are more robust. Since there are only typically ~3 failures per 15 minute CO run, it's only adding ~0.3 seconds or so, which seems not a huge deal?

.github/workflows/linux_x86.yml
hughperkins: (note: chagned "strong preference" to "preference". If you have a strong preference to remove it, I can remove it)

.github/workflows/scripts/ti_build/dep.py
hughperkins: Personally, I feel that list is more intuitive to read. `tuple` on the whole I feel represents something more analogous to a struct, even though it's technically iterable. I don't have a strong preference to keep it as a list, but I do have a gentle preference.

Default values for a method, totally agree with using tuple, instead of list, for the reason you allude to: that parameters to methods should not be mutaable, otherwise if you modify them inside the method, then you just modified thd efault parametrsr 😓

hughperkins: I guess stachexchange tends to align with your own pov on this: when not mutable, use a tuple https://stackoverflow.com/questions/1708510/list-vs-tuple-when-to-use-each. i'll change to tuple

hughperkins: actually, I might ask in #code

.github/workflows/linux_x86.yml
duburcqa: > it reduces the number of failure modes due to permissions not being set, for various reasons

This is a non-issue. If it works once, then you are good. It is not going to break for no reason. And as soon as it is no longer the case, you could enable back `bash` anytime if necessary

> it makes it very clear we are using bash (and not sh, or something else)

You can explicitly select bash in you workflow file. Actually this is what I'm doing:
```
    defaults:
      run:
        shell: bash -ieo pipefail {0}
```

> What benefits do you perceive in removing the bash prefix?

I just don't like adding stuffs that are not necessary. It makes other people think it is necessary why actually it does nothing, which is confusing.

tests/python/test_ad_ndarray.py
duburcqa: It is not about whether it is a big deal or not, but rather doing what is necessary to make test pass. Knowing whether this delay is necessary to help understanding what could be the root cause.

hughperkins: Ok. So you would prefer I remove it, and we see if it still fails? That makes sense.

.github/workflows/linux_x86.yml
hughperkins: > This is a non-issue. If it works once, then you are good. It is not going to break for no reason. And as soon as it is no longer the case, you could enable back bash anytime if necessary

Let me rephrase: I would prefer to not have to set permissions, which I feel tend not to be highly visible, and often fail to get copied around the place. They do show up in github PRs, but not in a form that is highly intuitive I feel. Permissions don't work well, when cloning a repo onto windows for example.

tests/python/test_ad_ndarray.py
hughperkins: remoed

.github/workflows/linux_x86.yml
duburcqa: I'm not following you. Which permission are you talking about? You mean you don't with to have to do this?
```
sudo env "PATH=$PATH" .github/workflows/scripts_new/linux_x86/2_build.sh
```

hughperkins: You need `chmod +x` to run the bash script, without prefixing with `bash` or similar, I think?

duburcqa: You can commit these permissions I think. Or may be I'm wrong? If it is possible you should do it since you want to run them locally.

hughperkins: You can commit them. However, if you clone the repo onto windows, windows doesnt handle permissions well I feel. And generally, I just prefer to avoid having to set permisisons on things in github repos, on the whole.

duburcqa: I don't get the argument. Windows does not have file mode, so what? It is a shell script anyway. If it is a shell script, anybody would expect the file to avoid run mode enabled. It is not like we are doing it behind the back of the user.

hughperkins: Clone the repo on windows
- update teh linux script (which you are testing on a linux builder)
- push the change
- permissions often get updated at the same time

hughperkins: There's just a whole bunch of 'gotchas' I feel, that can be avoided by simply writing `bash`, just as we often write `python`, e.g. https://github.com/duburcqa/jiminy/blob/accdf3629d9d8c08bb97f2905f010f1616d2168a/.github/workflows/macos.yml#L218 (you could make `acro_ppo.py` have a shebang `#!/usr/bin/env python` )

duburcqa: Alright let's move on x)

PR #4, opened 2025-06-02
[Build] Add Windows github runner

hughperkins:
Issue: #

### Brief Summary

Add Windows github runner, using github standard runner, with windows server core 2025

copilot:summary

### Walkthrough

copilot:walkthrough


.github/workflows/win_ondemand.yml
```diff
@@ -0,0 +1,28 @@
...
+          Get-ChildItem
+          python -V
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Windows 2025 Build
+        run: |
+          pwd
+          & .\.github\workflows\scripts_new\win.ps1
```
duburcqa: Same as for the other PR. I would just move the contain of `win.ps1` here.

.github/workflows/win_ondemand.yml
```diff
@@ -0,0 +1,28 @@
...
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: windows-2025
+    steps:
+      - name: Python check
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: explore env
```
duburcqa: I think this step is irrelevant but it is up to you.

tests/python/test_ad_ndarray.py
```diff
@@ -1159,6 +1160,7 @@ def func(x: ti.types.ndarray(), y: ti.types.ndarray()):
     assert x.grad[1] == -0.25
 
 
+@pytest.mark.skipif(platform.system() == "Windows", reason="ad_ndarray failing sporadically on Windows")
```
duburcqa: You should rather use `@pytest.mark.xfail(reason="ad_ndarray failing sporadically on Windows.")`

Or yet another option, add `pytest-retry` plus the decorator `@pytest.mark.flaky(retries=2, only_on=(ValueError, IndexError), condition=(platform.system() == "Windows"))`

.github/workflows/scripts_new/win.ps1
```diff
@@ -0,0 +1,26 @@
...
+systeminfo
+Get-Location
+Get-ChildItem
+python -V
+python --version
+
+git submodule update --init --recursive
+
+Start-Process -NoNewWindow -FilePath "python" -ArgumentList "build.py" -ErrorAction SilentlyContinue -Wait
+# try {
```
duburcqa: No dead code unless a comment clearly state why it should be kept, which some FIXME / TODO prefix.

tests/python/test_sparse_matrix.py
```diff
@@ -355,6 +355,7 @@ def fill(
     ],
 )
 @test_utils.test(arch=ti.cpu)
+@pytest.mark.flaky(retries=5, delay=0.1)
```
duburcqa: Do you need this delay? This is surprising.

.github/workflows/win_ondemand.yml
hughperkins: removed

tests/python/test_sparse_matrix.py
hughperkins: (answered this elsewhere, let's discuss it there?).

.github/workflows/scripts_new/win.ps1
hughperkins: Good spot. Thanks! :) (removed)

tests/python/test_sparse_matrix.py
hughperkins: (removed)

.github/workflows/win_ondemand.yml
hughperkins: addressed in 07ad02462 and ff3217fac3

tests/python/test_ad_ndarray.py
```diff
@@ -1160,13 +1161,15 @@ def func(x: ti.types.ndarray(), y: ti.types.ndarray()):
     assert x.grad[1] == -0.25
 
 
+@pytest.mark.flaky(retries=5, only_on=(ValueError, IndexError))
```
duburcqa: `only_on=(ValueError, IndexError)` probably does not makes sense here x) it was just an example. I don't know which type of exception is being raises in case of failure in your case.

tests/python/test_ad_ndarray.py
```diff
@@ -1160,13 +1161,15 @@ def func(x: ti.types.ndarray(), y: ti.types.ndarray()):
     assert x.grad[1] == -0.25
 
 
+@pytest.mark.flaky(retries=5)
```
CharlesMasson, line 1164: Should we add comments to explain why those are needed?

hughperkins: I think it's self documenting? "these tests are flaky. reason unclear"?

CharlesMasson: Yes, I understand this part, but I lack context about whether this is expected and we're fine with it, whether someone has additional context about it, whether it's meant to be investigated and fixed, etc.

hughperkins: It's not expected, and we're not fine with it, but we don't know how to fix it right now.

hughperkins: (and there are some much bigger things we want to fix first, like:
- not having to recompile every time we modify the batch size
- recompilations not taking 3 minutes
- start from warm cache not taking 20 seconds)

PR #5, opened 2025-06-03
[Build] Add mac build for Mac OS 14 and 15

hughperkins:
Issue: #

### Brief Summary

Add mac build for Mac OS 14 and 15

Note that this will fail without a ton of additional changes.
- open question whether we want to merge this in (with it failing), and then merge in additional things, bit by bit, until the mac build passes?
- or create one 'big bang' pr
- I feel that the first approach will be much easier, and cleaner?

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/scripts/ti_build/sccache.py
+++ b/.github/workflows/scripts/ti_build/sccache.py
@@ -59,6 +59,8 @@ def setup_sccache() -> Command:
     cache.mkdir(parents=True, exist_ok=True)
     os.environ["SCCACHE_DIR"] = str(cache)
     os.environ["SCCACHE_CACHE_SIZE"] = "40G"
+    os.environ["SCCACHE_STARTUP_TIMEOUT"] = "30"
+    os.environ["SCCACHE_IDLE_TIMEOUT"] = "0"
     # </LocalCache>
 
     return sh.bake(str(exe))
--- a/.github/workflows/scripts_new/linters.sh
+++ b/.github/workflows/scripts_new/linters.sh
@@ -7,25 +7,5 @@ pwd
 ls
 uname -a
 
-# python + C++
-# =============
-
 pip install pre-commit
 pre-commit run -a --show-diff
-
-# python
-# ======
-
-pip install pyright
-# Need to deal with C++ linkage issues first (and possibly
-# some other things), before we can turn on pyright
-
-pip install isort
-# TODO: run isort on all python files, and commit those, then
-# uncomment the following line:
-# isort --check-only --diff python
-
-# C++
-# ===
-
-# TODO: figure out how to run clang-tidy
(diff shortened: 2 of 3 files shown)


.github/workflows/mac_ondemand.yml
```diff
@@ -0,0 +1,21 @@
...
+      - main
+  workflow_dispatch:
+jobs:
+  build-15:
+    runs-on: macos-15
+    steps:
+      - uses: actions/checkout@v4
+      - name: Mac Build
+        run: |
+          bash .github/workflows/scripts_new/macos_15.sh
```
duburcqa: Same as windows and linux.

.github/workflows/mac_ondemand.yml
```diff
@@ -0,0 +1,21 @@
+name: Mac OS 15 on-demand
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+jobs:
+  build-15:
+    runs-on: macos-15
```
duburcqa: Use a matrix job instead of multiple explicit `runs-on` that prevent factorizing the code:
```
    strategy:
      fail-fast: false
      matrix:
        OS: ['macos-14', 'macos-15']

    runs-on: ${{ matrix.OS }}
```

hughperkins: oh nice. Thanks! 🙌

.github/workflows/mac_ondemand.yml
hughperkins: addressed in 26563e0cb9e8648523446e15e87a780a2f896e21 and 6fa7479cfb4aaec00142a7108e172d80d853413d

tests/python/test_matrix.py
```diff
@@ -1135,8 +1136,18 @@ def test():
 
 
 @test_utils.test(debug=True)
+@pytest.mark.skipif(
```
duburcqa, line 1139: This is concerning but OK. Better be in a clean state and repair stuffs after.

tests/python/test_matrix.py
```diff
@@ -1135,8 +1136,18 @@ def test():
...
+        "segfaults on Mac with multiprocess. Runs ok with -t 1 "
+        "SHOULD PASS. Created "
+        "https://linear.app/genesis-ai-company/issue/CMP-31/"
+        "fix-failing-test-cross-scope-matrix-atomic-ops-on-mac-in-multiprocess"
+        " to track"
+    ),
+)
 def test_cross_scope_matrix_atomic_ops():
-    n = 128
+    n = 64
```
duburcqa: Is this change still necessary?

hughperkins: fair point

PR #6, opened 2025-06-03
[Build] Add linters

hughperkins:
Issue: #

### Brief Summary

Add linters

- these are basically the same linters that upstream uses
- but I'm running pre-commit from a script, rather than using precommit.ci, since precommit.ci doesn't work on private repos (at least, not without paying money; and it's so easy to simply run from a script anyway...)
- for now it is missing clang-tidy. I need to figure out how to add that (upstream runs clang-tidy, so I can basically just copy that)
    - I figure we can merge as-is, and then I can work on clang-tidy in parallel?

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -6,6 +6,7 @@ on:
       - main
 jobs:
   build:
+    name: Linters
     runs-on: ubuntu-22.04
     steps:
       - name: Python check


.github/workflows/linters.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Python check
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - uses: actions/checkout@v4
+      - name: Run linters
+        run: |
+          bash .github/workflows/scripts_new/linters.sh
```
duburcqa, line 19: Same again.

.pre-commit-config.yaml
```diff
@@ -3,15 +3,15 @@ ci:
   autoupdate_commit_msg: '[misc] Update pre-commit hooks'
 
 default_language_version:
-    python: python3.12
+    python: python3.10
```
duburcqa, line 6: Why is that?

hughperkins: - I want to set allowable versions to 3.10-3.12
- for testing stuff, I find that scripts tends to be very forwards compatible, but very not backwards compatibel
    - therefore strong preference to use the lowest supported python version for our CI

hughperkins: (rewrote that comment a bit, you might want to refresh)

hughperkins: I suppose missing "because if CI runs on the lowest supported python version, then we have a reasonable certainty it will run on later versions of python. But the opposite is not true."

duburcqa: I can buy that. Good point!

python/taichi/tools/vtk.py
```diff
@@ -22,6 +22,8 @@ def write_vtk(scalar_field, filename):
         zcoords = np.array([0, 1])
     elif dimensions == 3:
         zcoords = np.arange(0, field_shape[2])
+    else:
```
duburcqa, line 25: Unrelated but OK.

hughperkins: it was causing one of hte linters to fail. So I'd either have to add an exclusion, or fix it. Fixing it seemed potentailly 'lowest hanging fruit', in this case.

duburcqa: I see. Make sense. Thank you for the feedback.

.github/workflows/scripts_new/linters.sh
```diff
@@ -0,0 +1,31 @@
...
+pre-commit run -a --show-diff
+
+# python
+# ======
+
+pip install pyright
+# Need to deal with C++ linkage issues first (and possibly
+# some other things), before we can turn on pyright
+
+pip install isort
```
CharlesMasson, line 23: What do you think of using ruff? We already use it in `gs-core` and it's the most robust and complete. We can only enable isort rules for now if we want to limit ourself to it.

.github/workflows/linters.yml
CharlesMasson: Ruff for instance has an official well maintained github action. I think it'd make sense to use it, so that what we have to maintain is minimal.

.github/workflows/scripts_new/linters.sh
hughperkins: I like ruff

hughperkins: I think we should merge all CI PRs first , before running a `ruff` pass, since that will conflict with the whole world...

hughperkins: (possibly excluding this one I suppose 🤔 )

hughperkins: Well... I suppose most of the changes are to either C++ or CI, which are both orthogonal to ruff 🤔

hughperkins: So, I tried running `ruff check --fix`, and a looottttt of unfixable
Screenshot 2025-06-05 at 9 07 16 AM
 errors

Maybe best to merge this PR in as-is, and then look at migrating to ruff later? (maybe using isort initially, and adding ruff only on modified files?)

CharlesMasson: I think we could start using ruff now and only enable isort rules instead of migrating later, but up to you.

CharlesMasson: One advantage I see with ruff is that it has well-maintained pre-commit hooks and Github actions, so I think it'd allow removing some code in this PR, especially since it would also allow removing black and pylint in addition to isort.

hughperkins: oh interesting. didnt realize `ruff check --select I` was an option. Looking into that. Thanks!

CharlesMasson: Yes, but I'd favor configuring in `pyproject.toml` or `ruff.toml`, so that configuration is done in a single place and more transparently.

hughperkins: Yup, good point. thanks! 🙌

PR #14, opened 2025-06-05
[Build] Run pyright against modified files

hughperkins:
Issue: #

### Brief Summary

Run pyright against modified files
- first runs pybind11-stubgen
- for detecting modified files, uses an AI generated script, in python/tools/run_modified_files.py

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/scripts_new/linux_x86/4_test.sh
+++ b/.github/workflows/scripts_new/linux_x86/4_test.sh
@@ -3,4 +3,4 @@
 set -ex
 
 pip3 install -r requirements_test.txt
-python3.10 tests/run_tests.py -v
+python3.10 tests/run_tests.py -v -r 3
--- a/.github/workflows/scripts_new/win/4_test.ps1
+++ b/.github/workflows/scripts_new/win/4_test.ps1
@@ -4,4 +4,4 @@ trap { Write-Error $_; exit 1 }
 
 python -c 'import taichi as ti; ti.init();'
 pip install -r requirements_test.txt
-python .\tests\run_tests.py -v
+python .\tests\run_tests.py -v -r 3


.github/workflows/scripts_new/pyright_linter/5_pyright.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+set -ex
+
+python python/tools/run_modified_files.py --include 'python/*.py' -- pyright
```
CharlesMasson: We can have errors in unmodified files due to modifications elsewhere, can't we?
Is it particularly slow to run everywhere? Otherwise, I'd keep things simple and run it everywhere, or would parallelize instead.

.github/workflows/scripts_new/pyright_linter/1_prerequisites.sh
```diff
@@ -0,0 +1,36 @@
...
+    libclang-cpp15 \
+    libclang1-15 \
+    cmake \
+    ninja-build \
+    python3-dev \
+    python3-pip
+
+pip3 install scikit-build
+
+pip install pyright pybind11-stubgen
```
CharlesMasson: Can we configure this in a `pyproject.toml` with a dev dependency group so that everyone can use it to set up local dev as well?

python/tools/run_modified_files.py
```diff
@@ -0,0 +1,72 @@
...
+import subprocess
+import re
+import sys
+import argparse
+from typing import List
+from pathlib import Path
+import fnmatch
+
+
+def get_changed_files(include_pattern: str) -> List[str]:
```
CharlesMasson: I have mixed feelings about adding this complexity. If it's too slow to run everywhere, couldn't we scope actions (e.g., by directory) and trigger only on directory changes. This would seem a more robust and transparent solution to me.

.github/workflows/scripts_new/pyright_linter/5_pyright.sh
hughperkins: Its' super fast. Speed is not an issue. The 1500 errors that pop up when we run it are.

See https://genesis-ai-company.slack.com/archives/C08JT0YHM2A/p1748974351128939 for more context

python/tools/run_modified_files.py
hughperkins: See other thread.

.github/workflows/scripts_new/pyright_linter/1_prerequisites.sh
hughperkins: I dont think we can use pyproject.toml, though perhaps you know how to configure taichi so we can? Would love to migraate to pyproject.toml

But yes, should probalby add to requirements_dev.txt. I'll do that

hughperkins: Added to reuiqrements_dev.txt

hughperkins: Hmmm, I guess since I'm running a full build anyway, not much excuse for not removing it from the pyright script. I'll try doing that.

.github/workflows/scripts_new/pyright_linter/5_pyright.sh
CharlesMasson: Does it keep checking files that have been changed previously, therefore have been fixed, but have not been changed again in the current commit/PR?

hughperkins: it checks files that have been modified relative to main.

it won't check files that have been modified previously in main, but not since.

hughperkins: basically, it proects against breaking new files, and will gradually enforce pyright on any files, as we modify them (and once a file is fixed, the only way to unfix it would be to modify it again, I think?)

hughperkins: What I think will happen in practice:
- first few files will be painful to fix
- then we will gradually learn how to fix them, and can figure out a way to fix 'en masse'
- ('we' here likely meaning 'I' 😅)

CharlesMasson: I think that you can cause typing errors in that file if you modify a dependency in another file (e.g., a type that it uses that is defined in another file). And once a file is fixed, I think we should keep running pyright on it.

An alternative may be to maintain a list of exclusions in the pyright configuration from which we'd gradually remove files, [as done in gs-core](https://github.com/genesis-company/gs-core/blob/main/pyproject.toml#L8).

hughperkins: right now, that would be a list of like 1500 files... Baby steps. We don't even know we can fix even a single file yet.

.github/workflows/pyright_linter.yml
```diff
@@ -0,0 +1,37 @@
...
+    branches:
+      - main
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+      # looks like we need to first build the dll...
+      # easiest way to do that is to build the wheel
+      # at some point, we should probably have a separate wheel building thing
```
duburcqa, line 14: I did this for Jiminy in the past. You can find how to do this [here](https://github.com/duburcqa/jiminy/blob/accdf3629d9d8c08bb97f2905f010f1616d2168a/.github/workflows/manylinux.yml#L135).

.github/workflows/pyright_linter.yml
```diff
@@ -0,0 +1,37 @@
...
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+      # looks like we need to first build the dll...
```
duburcqa, line 12: Actually you should build the DLL, then generate the stubs to get typing working. This is basically what you did actually, and yes it is absolutely necessary. Note that you should bundle the stubs with the wheel for the end-user (actually a developper using taichi) to have proper typing experience out-of-the-box. I did this before for Jiminy [here](https://github.com/duburcqa/jiminy/blob/accdf3629d9d8c08bb97f2905f010f1616d2168a/.github/workflows/manylinux.yml#L115).

.github/workflows/scripts_new/pyright_linter/5_pyright.sh
duburcqa: I agree with Charles. Not a huge fan of the approach of only checking modified files. But maybe we should remove all this confusing / obsfucated logic once Taichi repo is finally clean.

.github/workflows/pyright_linter.yml
hughperkins: Ok, so the pyi bit should be part of the wheel build? Sounds good. Thanks!

.github/workflows/scripts_new/pyright_linter/5_pyright.sh
hughperkins: For sure yeah. I just can't do everything all at once :)

tests/python/test_tools_run_modified_files_integration.py
```diff
@@ -0,0 +1,106 @@
...
+from pathlib import Path
+
+import pytest
+
+from tools import run_modified_files
+
+
+@pytest.fixture
+def git_repo(tmp_path):
+    cwd = os.getcwd()
```
duburcqa: I would rather just pass `cwd` argument to `subprocess.run`. Changing the current directory in Python code is a bad practice.

tests/python/test_tools_run_modified_files_integration.py
```diff
@@ -0,0 +1,106 @@
...
+    subprocess.run(["git", "remote", "add", "origin", "."], check=True)
+    yield tmp_path
+    os.chdir(cwd)
+
+
+def test_get_changed_files_added_file(git_repo):
+    # Create a file and add it
+    f = git_repo / "foo.py"
+    f.write_text("print('hi')")
+    subprocess.run(["git", "add", "foo.py"], check=True)
```
duburcqa: I would specify `cwd` for all the subprocess commands into the whole file.

.github/workflows/pyright_linter.yml
duburcqa: Yes!

.github/workflows/pyright_linter.yml
hughperkins: yeah, lets have that as a separate pr please. One step at a time.

tests/python/test_tools_run_modified_files_integration.py
hughperkins: agree. addressed in 27b3d4bb2

tests/python/test_tools_run_modified_files_integration.py
hughperkins: addressed in 27b3d4bb2

.github/workflows/scripts_new/pyright_linter/5_pyright.sh
CharlesMasson: >that would be a list of like 1500 files...

We can also add full directory exclusions, if that is more manageable.

.github/workflows/pyright_linter.yml
duburcqa: No problem!

PR #15, opened 2025-06-05
[Misc] Add TAICHI_DUMP_IR and TAICHI_LOAD_IR

hughperkins:
Issue: #

### Brief Summary

Add TAICHI_DUMP_IR and TAICHI_LOAD_IR

Allows you to dump IR from various stages in the pipeline, by defining TAICHI_DUMP_IR (eg to '1').
- the IR will be dumped to files in /tmp/ir folder

You can also load from the IR using TAICHI_LOAD_IR
- this will load from the same folder
- in the case of CUDA, it will load the IR just before we drop into PTX

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/taichi/runtime/cuda/jit_cuda.cpp
+++ b/taichi/runtime/cuda/jit_cuda.cpp
@@ -87,7 +87,7 @@ JITModule *JITSessionCUDA::add_module(std::unique_ptr<llvm::Module> M,
     } else {
       std::cout << "problem dumping file " << filename << ": " << EC.message()
                 << std::endl;
-      TI_ERROR("Failed to dump LLVM IR to file: {}", filename);
+      TI_ERROR("Failed to dump LLVM IR to file: {}", filename.c_str());
     }
   }
 


taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2746,6 +2748,31 @@ LLVMCompiledTask TaskCodeGenLLVM::run_compilation() {
       tlctx->mark_function_as_amdgpu_kernel(func);
     }
   }
+  const char *dump_ir_env = std::getenv("TAICHI_DUMP_IR");
+  const std::string dumpOutDir = "/tmp/ir/";
```
duburcqa: Taichi does not have its own temporary folder where everything is stored? That would be surprising a something doing JIT compilation.

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2746,6 +2748,31 @@ LLVMCompiledTask TaskCodeGenLLVM::run_compilation() {
...
+  if (dump_ir_env != nullptr) {
+    std::filesystem::create_directories(dumpOutDir);
+
+    std::string filename = dumpOutDir + "/" + kernel->name + "_llvm.ll";
```
duburcqa: You should leverage C++ `std::filesystem` capability for doing this:
```c++
std::string filename = dumpOutDir / (kernel->name + "_llvm.ll");
```

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2746,6 +2748,31 @@ LLVMCompiledTask TaskCodeGenLLVM::run_compilation() {
...
+    std::error_code EC;
+    llvm::raw_fd_ostream dest_file(filename, EC);
+    if (!EC) {
+      module->print(dest_file, nullptr);
+    }
+  }
+
+  const char *load_ir_env = std::getenv("TAICHI_LOAD_IR");
+  if (load_ir_env != nullptr) {
+    std::string filename = dumpOutDir + "/" + kernel->name + "_llvm.ll";
```
duburcqa: Same.

taichi/codegen/llvm/codegen_llvm.cpp
hughperkins: everything is in memory AFAIK. But I could be wrong 🤔  But for example:
```
(taichi) system ~/git/taichi-play (main|…4) $  python resizable_fields/resizable_prot1vec3b.py --B 5
[Taichi] version 1.8.0, llvm 15.0.7, commit 1e3c8362, osx, python 3.10.16
arch Arch.cuda
[W 06/09/25 13:53:27.961 10851080] [misc.py:adaptive_arch_select@758] Arch=[<Arch.cuda: 3>] is not supported, falling back to CPU
[Taichi] Starting on arch=arm64
*** create a ndarray
kernel_keyT575fa18129530640f192cb6163814f4cc6253f75b2ae0dd9a2a345a94d27bb4c
COMPILING KERNEL: ndarray_matrix_to_ext_arr_c10_0
kernel_keyTa07cbc9ebadf52d1c1cf9d14ba4d633c848c44352a2727070b4b9735fb782894
COMPILING KERNEL: ext_arr_to_ndarray_matrix_c26_0
*** create b ndarray
*** creating field a (B, 4)
*** a_field snode id 3
kernel_keyT3232184207c3a3f5fa6618a5fccdec03f9f3b9e17b4829110d307fcf106328e6
COMPILING KERNEL: ext_arr_to_matrix_c30_0
*** a_field snode id 3
*** creating field b (B + 2, 7)
kernel_keyT0e7f21bfa9e7350df874cfd7b3767022901495c1067fed5457fbff563e1201ab
COMPILING KERNEL: ext_arr_to_matrix_c30_1
*** field template
kernel_keyT6e47b046063bfe78193dc16d33da0e51ea1eec85c8f73526da46d27b3923e5e4
COMPILING KERNEL: template_2d_c82_0
0 0 [0, 1, 2]
0 1 [100, 101, 102]
0 2 [200, 201, 202]
0 3 [300, 301, 302]
1 0 [10000, 10001, 10002]
1 1 [10100, 10101, 10102]
1 2 [10200, 10201, 10202]
1 3 [10300, 10301, 10302]
2 0 [20000, 20001, 20002]
2 1 [20100, 20101, 20102]
2 2 [20200, 20201, 20202]
2 3 [20300, 20301, 20302]
3 0 [30000, 30001, 30002]
3 1 [30100, 30101, 30102]
3 2 [30200, 30201, 30202]
3 3 [30300, 30301, 30302]
4 0 [40000, 40001, 40002]
4 1 [40100, 40101, 40102]
4 2 [40200, 40201, 40202]
4 3 [40300, 40301, 40302]
***
***
(taichi) system ~/git/taichi-play (main|…4) $ find . -cmin -1
./.git
(taichi) system ~/git/taichi-play (main|…4) $ find ~/.cache -cmin -1
(taichi) system ~/git/taichi-play (main|…4) $ find ~/git/taichi -cmin -1
/Users/hugh/git/taichi/.git
/Users/hugh/git/taichi/.git/modules/external/SPIRV-Cross
/Users/hugh/git/taichi/.git/modules/external/volk
/Users/hugh/git/taichi/.git/modules/external/PicoSHA2
/Users/hugh/git/taichi/.git/modules/external/SPIRV-Tools
/Users/hugh/git/taichi/.git/modules/external/imgui
/Users/hugh/git/taichi/.git/modules/external/eigen
/Users/hugh/git/taichi/.git/modules/external/SPIRV-Reflect
/Users/hugh/git/taichi/.git/modules/external/SPIRV-Reflect/modules/third_party/googletest
/Users/hugh/git/taichi/.git/modules/external/spdlog
/Users/hugh/git/taichi/.git/modules/external/VulkanMemoryAllocator
/Users/hugh/git/taichi/.git/modules/external/glad
/Users/hugh/git/taichi/.git/modules/external/DirectX-Headers
/Users/hugh/git/taichi/.git/modules/external/googletest
/Users/hugh/git/taichi/.git/modules/external/SPIRV-Headers
/Users/hugh/git/taichi/.git/modules/external/backward_cpp
/Users/hugh/git/taichi/.git/modules/external/glfw
/Users/hugh/git/taichi/.git/modules/external/Vulkan-Headers
/Users/hugh/git/taichi/.git/modules/external/assets
/Users/hugh/git/taichi/.git/modules/external/glm
/Users/hugh/git/taichi/.git/modules/external/FP16
```

taichi/codegen/llvm/struct_llvm.cpp
```diff
@@ -263,6 +263,20 @@ void StructCompilerLLVM::run(SNode &root) {
     writer.write(module.get());
   }
 
+  const char *dump_ir_env = std::getenv("TAICHI_DUMP_IR");
+  const std::string dumpOutDir = "/tmp/ir/";
+  if (dump_ir_env != nullptr) {
+    std::filesystem::create_directories(dumpOutDir);
+
+    std::string filename =
+        dumpOutDir + "/" + std::string(module->getName()) + "_llvm.ll";
```
duburcqa: same. What is the type of `module->getName()` to require casting to string ?

taichi/codegen/llvm/struct_llvm.cpp
```diff
@@ -263,6 +263,20 @@ void StructCompilerLLVM::run(SNode &root) {
     writer.write(module.get());
   }
 
+  const char *dump_ir_env = std::getenv("TAICHI_DUMP_IR");
+  const std::string dumpOutDir = "/tmp/ir/";
```
duburcqa: Same.

taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2715,6 +2715,25 @@ KernelCodegen::KernelCodegen(const Params &params)
 void KernelCodegen::run(TaichiKernelAttributes &kernel_attribs,
                         std::vector<std::vector<uint32_t>> &generated_spirv) {
   auto *root = params_.ir_root->as<Block>();
+
+  {
+    const char *dump_ir_env = std::getenv("TAICHI_DUMP_IR");
+    const std::string dumpOutDir = "/tmp/ir/";
```
duburcqa: same.

taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2715,6 +2715,25 @@ KernelCodegen::KernelCodegen(const Params &params)
...
+    if (dump_ir_env != nullptr) {
+      std::filesystem::create_directories(dumpOutDir);
+
+      std::string filename =
+          dumpOutDir + "/" + params_.ti_kernel_name + "_before_final_spirv.ll";
```
duburcqa: same.

taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2754,6 +2773,23 @@ void KernelCodegen::run(TaichiKernelAttributes &kernel_attribs,
     TI_TRACE("SPIRV-Tools-opt: binary size, before={}, after={}",
              task_res.spirv_code.size(), optimized_spv.size());
 
+    {
+      const char *dump_ir_env = std::getenv("TAICHI_DUMP_SPIRV");
```
duburcqa: Do you really a flag for a these debug things? A general `TAICHI_DUMP_IR` is not enough?

taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2754,6 +2773,23 @@ void KernelCodegen::run(TaichiKernelAttributes &kernel_attribs,
...
+      const std::string dumpOutDir = "/tmp/spirv/";
+      if (dump_ir_env != nullptr) {
+        std::filesystem::create_directories(dumpOutDir);
+        std::string spirv_asm;
+        spirv_tools_->Disassemble(optimized_spv, &spirv_asm);
+        auto kernel_name = tp.ti_kernel_name;
+        std::string filename = std::filesystem::path(dumpOutDir) / (kernel_name + ".spirv");
```
duburcqa: Just change the type of `dumpOutDir` to `std::filesystem::path` instead of string for the start.

taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -5,14 +5,90 @@ namespace taichi::lang {
...
+  for (auto &func : function_list) {
+    if (func.getName() == "runtime_initialize") {
+      return true;
+    }
+  }
+  return false;
+}
+
+std::string moduleToDumpName(llvm::Module *M) {
+  std::string dumpName(M->getName().begin(), M->getName().end());
```
duburcqa: `std::string dumpName(M->getName().str()))` should work I guess?

taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -5,14 +5,90 @@ namespace taichi::lang {
...
+  std::cout << "module get function list len:" << M->getFunctionList().size()
+            << std::endl;
+  auto func0 = M->getFunctionList().begin();
+  std::cout << "function 0 name: " << func0->getName().str() << std::endl;
```
duburcqa: `if func0 != nullptr` ?

taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -5,14 +5,90 @@ namespace taichi::lang {
...
+  if (!module_has_runtime_initialize(M->getFunctionList())) {
+    dumpName = std::string(func0->getName().begin(), func0->getName().end());
```
duburcqa: `dumpName = func0->getName().str()`

taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -5,14 +5,90 @@ namespace taichi::lang {
...
+  }
+  return dumpName;
+}
+
 JITModule *JITSessionCUDA ::add_module(std::unique_ptr<llvm::Module> M,
                                        int max_reg) {
+  const char *dump_ir_env = std::getenv("TAICHI_DUMP_IR");
+  if (dump_ir_env != nullptr) {
+    const std::string dumpOutDir = "/tmp/ir/";
```
duburcqa: same.

taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -5,14 +5,90 @@ namespace taichi::lang {
...
+    std::filesystem::create_directories(dumpOutDir);
+    std::string dumpName = moduleToDumpName(M.get());
+    std::string filename = dumpOutDir + "/" + dumpName + "_before_ptx.ll";
```
duburcqa: same.

taichi/codegen/llvm/struct_llvm.cpp
hughperkins: llvm::StringRef

taichi/codegen/spirv/spirv_codegen.cpp
hughperkins: Fine...

taichi/codegen/llvm/codegen_llvm.cpp
duburcqa: I see. Fine then.

taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2746,6 +2748,31 @@ LLVMCompiledTask TaskCodeGenLLVM::run_compilation() {
       tlctx->mark_function_as_amdgpu_kernel(func);
     }
   }
+  const char *dump_ir_env = std::getenv("TAICHI_DUMP_IR");
```
bo-rc: a little concerned about hard-coded argument. Better to create named variables:

```cpp
constexpr std::string_view DUMP_IR_ENV = "TAICHI_DUMP_IR";
constexpr std::string_view LOAD_IR_ENV = "TAICHI_LOAD_IR";
constexpr std::string_view IR_OUTPUT_DIR = "/tmp/ir/";
```

taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2715,6 +2715,25 @@ KernelCodegen::KernelCodegen(const Params &params)
...
+    const auto dumpOutDir = std::filesystem::path("/tmp/ir/");
+    if (dump_ir_env != nullptr) {
+      std::filesystem::create_directories(dumpOutDir);
+
+      std::string filename =
+          dumpOutDir / (params_.ti_kernel_name + "_before_final_spirv.ll");
+      std::ofstream out_file(filename);
+      if (out_file.is_open()) {
+        std::string outString;
+        irpass::print((IRNode *)(params_.ir_root), &outString);
```
bo-rc:
```cpp
irpass::print(const_cast<IRNode*>(ir_root), &out_string);
```

taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2715,6 +2715,25 @@ KernelCodegen::KernelCodegen(const Params &params)
...
+        out_file << outString;
+        out_file.close();
+      }
+    }
+  }
+
   auto &tasks = root->statements;
```
bo-rc, line 2735: What is the style guidance of this repo? I tend to comment the following:

What is the type of `root->statements`? prefer to explicit type rather than using `auto` if writing out the type won't be too verbose.

So this is a style thing.

taichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2754,6 +2773,23 @@ void KernelCodegen::run(TaichiKernelAttributes &kernel_attribs,
...
+        std::filesystem::create_directories(dumpOutDir);
+        std::string spirv_asm;
+        spirv_tools_->Disassemble(optimized_spv, &spirv_asm);
+        auto kernel_name = tp.ti_kernel_name;
+        std::string filename = dumpOutDir / (kernel_name + ".spirv");
+        std::ofstream out_file(filename);
+        if (out_file.is_open()) {
+          out_file.write(spirv_asm.c_str(), spirv_asm.size());
+          out_file.close();
+        }
```
bo-rc, line 2784: Prefer to use RAII (scoped initialization in modern C++) here:

```cpp
const std::string filename = dumpOutDir / (kernel_name + ".spirv");

// 'out_file' automatically closed after the code block
if (std::ofstream out_file(filename); out_file) {
    out_file.write(spirv_asm.data(), spirv_asm.size());
} else {
    // Handle the error
}
```

taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -5,14 +5,89 @@ namespace taichi::lang {
...
+    llvm::Module::FunctionListType &function_list) {
+  for (auto &func : function_list) {
+    if (func.getName() == "runtime_initialize") {
+      return true;
+    }
+  }
+  return false;
+}
+
+std::string moduleToDumpName(llvm::Module *M) {
```
bo-rc: It looks to me that in this repo using raw pointers is OK? i.e. not dis-encouraged.

bo-rc: `const llvm::Module *M`

taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -5,14 +5,89 @@ namespace taichi::lang {
...
+  std::string dumpName(M->getName().str());
+  auto func0 = M->getFunctionList().begin();
+  if(func0 != M->getFunctionList().end()) {
```
bo-rc: This is a C-style checking if the list is empty. C++ prefers to use `.empty()`

taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -5,14 +5,89 @@ namespace taichi::lang {
...
+    if (!module_has_runtime_initialize(M->getFunctionList())) {
```
bo-rc: nested `if` can be flatten

taichi/runtime/cuda/jit_cuda.cpp
```diff
@@ -5,14 +5,89 @@ namespace taichi::lang {
...
+      dumpName = func0->getName().str();
+    }
+  }
+  return dumpName;
+}
```
bo-rc: How about:

```cpp
std::string moduleToDumpName(const llvm::Module* M) {
    // Store a reference to the list to avoid repeated calls.
    const auto& function_list = M->getFunctionList();

    // The logic is now a clear, single statement.
    if (!function_list.empty() && !module_has_runtime_initialize(function_list)) {
        // If conditions are met, use the name of the first function.
        return function_list.front().getName().str();
    }

    // Otherwise, return the this as a default.
    return M->getName().str();
}
```

taichi/codegen/spirv/spirv_codegen.cpp
hughperkins: That's a great question!

Technically it is our repo now, so we can do what we like.

However, looking around, just now, I found:

```
### C++ style guide

We generally follow [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html). One major exception is the naming convention of functions: Taichi adopts the snake case for function naming, as opposed to the camel case [suggested in Google's style](https://google.github.io/styleguide/cppguide.html#Function_Names), e.g. `this_is_a_taichi_function()`.
```

Screenshot 2025-06-10 at 2 32 07 PM

So, the style guide is here: https://google.github.io/styleguide/cppguide.html#C++_Version

Looking at this style guide it says:

```
The fundamental rule is: use type deduction only to make the code clearer or safer, and do not use it merely to avoid the inconvenience of writing an explicit type. When judging whether the code is clearer, keep in mind that your readers are not necessarily on your team, or familiar with your project, so types that you and your reviewer experience as unnecessary clutter will very often provide useful information to others. For example, you can assume that the return type of make_unique<Foo>() is obvious, but the return type of MyWidgetFactory() probably isn'
```

So this aligns with your own opinion on this subject :) Will update.

taichi/runtime/cuda/jit_cuda.cpp
hughperkins: happy to migrate to a safer form, as long as the scope of the change is limited approximately-ish to the code that I'm modifying. (cf adding const to one variable, and then having to change entire repo to have consts throughout, for example). What is your suggesiton for a safer form? And to what extent can we keep the scope of the change to just this code I'm adding?

taichi/codegen/spirv/spirv_codegen.cpp
duburcqa: This pattern is a bit confusing and not so common but why not.

taichi/runtime/cuda/jit_cuda.cpp
duburcqa: This is nice I think!

taichi/codegen/spirv/spirv_codegen.cpp
hughperkins: Oh, that bit of code wasn't added in this PR right? We can gradually migrate to this style over time I feel.

taichi/runtime/cuda/jit_cuda.cpp
hughperkins: `M->getFunctionList()` needs M to be non-const:
Screenshot 2025-06-10 at 3 30 48 PM

taichi/runtime/cuda/jit_cuda.cpp
hughperkins: We're going to have to get `func0` anyway. Are you saying, if we first check M->getFunctionList().empty(), then we don't have to do any checking on func0? (There's also `func0.isEnd`?)

taichi/codegen/spirv/spirv_codegen.cpp
bo-rc: Scoped initialization is introduced in C++17 and beyond. A much safer approach (resource creation, initialization, destruction within a given scope, out_file cannot be accidentally used out of the code scope) that also expresses the intent (out_file is only used within this given local scope / code block).

taichi/runtime/cuda/jit_cuda.cpp
bo-rc: In general, modern C++ recommends using smart pointers. but for legacy repo, the best strategy is to stick to its original style for small changes like this one. Migration to modern approach if necessary and make a plan do this in phases.

bo-rc: OH! I think I was not careful. It should be:

```cpp
llvm::Module * const M
```

because M is a const pointer in this code block.

taichi/runtime/cuda/jit_cuda.cpp
bo-rc: The logic is not to use `func0` so we should write our intend to express we want to check emptiness of a list. If `func0` would be used. This is better:

```cpp
auto func0 = M->getFunctionList().front();
// use func0 below 
``` 

Yes there is a `.back()` builtin function for C++ containers.

See my comments below for a suggestion to improve this code block.

taichi/runtime/cuda/jit_cuda.cpp
bo-rc:
```cpp
// preferred, but seems that Module cannot be const in the context?
const llvm::Module * const M;

// at least
llvm::Module * const M;
```

It seems to me that the original code is not of good quality in modern standards but we have to deal with it since it is a legacy codebase we need.

hughperkins: I'm fine iwth migrating bit by bit. Not really fine for taking time to migrate the entire codebase.

taichi/runtime/cuda/jit_cuda.cpp
hughperkins: addressed in b050ffa58

PR #17, opened 2025-06-05
[Build] Fix tools test on wheel

hughperkins:
Issue: #

### Brief Summary

Fix tools test on wheel

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -1,0 +1,30 @@
+name: clang-tidy (static analysis)
+on:
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Python check
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - uses: actions/checkout@v4
+      # we need to generate a special file "compile_commands.json"
+      # to do this, we need to run cmake with the following options:
+      # -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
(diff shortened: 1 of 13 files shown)


tests/python/test_tools_markdown_check.py
```diff
@@ -1,7 +1,17 @@
...
+
+from pathlib import Path
+import importlib.util
+
+# I don't want to make python/tools a module, and I don't want to move this tool
+# into `taichi` namespace, so that leaves temporarily importing it somehow
+tools_path = Path(__file__).parent.parent.parent / "python" / "tools" / "markdown_link_check.py"
+spec = importlib.util.spec_from_file_location("markdown_link_check", tools_path)
+markdown_link_check = importlib.util.module_from_spec(spec)
+spec.loader.exec_module(markdown_link_check)
```
CharlesMasson, line 13: Isn't there any way we can solve this declaratively in `pyproject.toml`?

hughperkins: Well, yes, by putting this into `taichi` namespace, but this is strictly for use in CI, not part of Taichi product, so I'm not very keen on doing that. I suppose we could make a new namespace, like `taichi-ci` perhaps 🤔 Thoughts?

hughperkins: (I'm not incredibly keen on adding a new `taichi-ci` namespace to be honest. we could also have `taichi.ci` I suppose 🤔 )

hughperkins: (unless you're aware of some way of handling this in pyproject.toml , that does not involve either adding this to `taichi` namespace or similar?)

PR #27, opened 2025-06-16
[Misc] Remove ui, and related doc and examples

hughperkins:
Issue: #

### Brief Summary

Remove ui, and related doc and examples

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/linux_x86.yml
+++ b/.github/workflows/linux_x86.yml
@@ -4,6 +4,9 @@ on:
     branches:
       - main
   workflow_dispatch:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
 jobs:
   build:
     name: Linux x86 Build
--- a/.github/workflows/macosx.yml
+++ b/.github/workflows/macosx.yml
@@ -4,6 +4,9 @@ on:
     branches:
       - main
   workflow_dispatch:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
 jobs:
   build_mac_14_15:
     strategy:
--- a/.github/workflows/manylinux_wheel.yml
+++ b/.github/workflows/manylinux_wheel.yml
@@ -14,6 +14,9 @@ jobs:
     name: Manylinux wheel Build
     runs-on: ubuntu-22.04
     container: quay.io/pypa/manylinux_2_28_x86_64:latest
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}-${{ matrix.PYTHON_CP_VERSION }}-${{ matrix.PYTHON_CP_VERSION != 'cp310' && 'all' || github.sha }}-build
+      cancel-in-progress: true
     strategy:
       matrix:
         PYTHON_CP_VERSION: ['cp310', 'cp311', 'cp312', 'cp313']
@@ -43,6 +46,9 @@ jobs:
     runs-on: ubuntu-22.04
     needs: build_wheel
     environment: aws_storage
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}-${{ matrix.PYTHON_CP_VERSION }}-${{ matrix.PYTHON_CP_VERSION != 'cp310' && 'all' || github.sha }}-upload
+      cancel-in-progress: true
     strategy:
       matrix:
         PYTHON_CP_VERSION: ['cp310', 'cp311', 'cp312', 'cp313']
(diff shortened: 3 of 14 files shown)


README.md
```diff
@@ -77,50 +77,6 @@ pip install --upgrade taichi
 pip install -i https://pypi.taichi.graphics/simple/ taichi-nightly
 ```
 
-### Run your "Hello, world!"
```
duburcqa, line 80: I would rather still have a small basic "Hello, world!" that does not involve GUI, but I can get over it.

docs/lang/articles/basic/field.md
```diff
@@ -188,11 +188,6 @@ def fill_image():
         gray_scale_image[i,j] = ti.random()
 
 fill_image()
-# Creates a GUI of the size of the gray-scale image
```
duburcqa, line 191: `The following code snippet creates and displays a 640&times;480 gray scale image of randomly-generated values:`

This makes no sense anymore.

docs/lang/articles/get-started/cloth_simulation.md
```diff
@@ -1,500 +0,0 @@
----
-sidebar_position: 3
```
duburcqa, line 2: Why is this being removed completely?

docs/lang/articles/get-started/hello_world.md
```diff
@@ -23,53 +23,6 @@ Taichi is available as a PyPI package:
 ```bash
 pip install taichi
 ```
-
```
duburcqa, line 26: I don't think this file makes sense anymore now that you stripped away 90% of it.

python/taichi/examples/algorithm/marching_squares.py
duburcqa, line 1: Deleting all the example algorithms just because you cannot display the result using custom Taichi imgui interface does not sound right.

hughperkins: If you can't see the result, then how can anyone see if it's working?

I don't have a good solution for this, but the short-term solution is to nuke them I feel.

I don't have time to rewrite them to use an alternative renderer.

docs/lang/articles/get-started/cloth_simulation.md
hughperkins: Because cloth simulation without being able to view the cloth is not something I personally would be interested in running I feel? If it's not good enough for ourselves to run, when learning about Taichi, why would we expect anyone else to find it useful? 🤔

README.md
hughperkins: Yeah, we should probalby have one simple example yeah. Maybe the prime numbers example?

duburcqa: Yes, that would be great.

hughperkins: Oh right, fair. will add.

python/taichi/examples/algorithm/marching_squares.py
duburcqa: > If you can't see the result, then how can anyone see if it's working?

True, but in practice, if there are not enough examples, then I will have to go on the original repo to find some, which is less than ideal I think.

hughperkins: Yeah, I guess, viscerally I don't feel a concern with that? I also think that Genesis is by and large the only example we need really?

(Speaking for myself personally, I looked at prime numbers, then I worked through the soft cloth falling over sphere tutorial, and then I didn't really look at any other examples I htink 🤔 )

README.md
hughperkins: addressed in 39af057ea

python/taichi/examples/algorithm/marching_squares.py
duburcqa: Let's do this. But remove everything that does not make sense anymore without UI. Basically the entire `hello_world.md` file at this point.

hughperkins: what do you mean? I htink hello_world.md is ok? 🤔

hughperkins: (at least, skimming, it is devoid of ui things?)

duburcqa: Check yourself the file. There is almost nothing left in this file, and what is left either makes no sense or is irrelevant.

hughperkins: are we looking at the same file? 🤔 

Screenshot 2025-07-11 at 10 51 22 AM
Screenshot 2025-07-11 at 10 51 26 AM
Screenshot 2025-07-11 at 10 51 31 AM
Screenshot 2025-07-11 at 10 51 34 AM

hughperkins: https://github.com/Genesis-Embodied-AI/taichi/blob/hp/remove-ui/docs/lang/articles/get-started/hello_world.md

hughperkins: (if you are talking about marchin_squares.py, I believe it is removed?)

duburcqa: Ok it is fine. I got something different on my side. Must have done something wrong.

PR #37, opened 2025-06-17
[Build] Manylinux wheel

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


.github/workflows/scripts_new/manylinux_wheel/1_prerequisites.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -ex
+
+# yum update
+yum install -y git wget libXrandr-devel libXinerama-devel libXcursor-devel libXi-devel
```
duburcqa, line 6: Could you try without these dependencies? They seem irrelevant.

.github/workflows/scripts/ti_build/llvm.py
```diff
@@ -23,10 +23,6 @@ def setup_llvm() -> None:
         if cmake_args.get_effective("TI_WITH_AMDGPU"):
             out = get_cache_home() / "llvm15-amdgpu-005"
             url = "https://github.com/GaleSeLee/assets/releases/download/v0.0.5/taichi-llvm-15.0.0-linux.zip"
-        elif is_manylinux2014():
```
duburcqa, line 26: Why to remove this? I don't see the point.

.github/workflows/manylinux_wheel.yml
```diff
@@ -0,0 +1,103 @@
...
+          - PYTHON_CP_VERSION: 'cp311'
+            PYTHON_VERSION: '3.11'
+          - PYTHON_CP_VERSION: 'cp312'
+            PYTHON_VERSION: '3.12'
+          - PYTHON_CP_VERSION: 'cp313'
+            PYTHON_VERSION: '3.13'
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4
+      - name: Configure AWS credentials
```
duburcqa, line 82: Why not to use GitHub artefacts for this?

.github/workflows/scripts/ti_build/llvm.py
hughperkins: Otherwise it doesn't run, because the manylinux 2014 version of llvm links to the old version of std::string instead of the newer c++11 std::string.

.github/workflows/manylinux_wheel.yml
hughperkins: I first upload to github artifacts, you can see. Then I download from artifact, and reupload to s3.

The reason is that then this lets me share the built wheel easily across other builds, eg for docs, pyright etc, without having to put everyting into a single .yml file. And also without coupling them together: I'm free to runs just the docs build, without re-running entire wheel build first each time.

.github/workflows/scripts_new/manylinux_wheel/1_prerequisites.sh
hughperkins: I think they are needed until we remove UI. The steps are:
1. get wheel build owrking, so I can use that to run docs build
2. one get docs building, merge docs PR
3. remove UI; merge remove UI PR
4. maybe we can remove these

duburcqa: You are right!

.github/workflows/manylinux_wheel.yml
duburcqa: I get all of this and I have no problem with jointly uploading wheels on S3. 

I'm just wondering why not using GitHub artefacts to download artefacts within the Github workflow.

.github/workflows/scripts/ti_build/llvm.py
duburcqa: But you are not using manylinux 2014 ?

hughperkins: No. we are using 2_28, which is then downgraded to 2_27 by repair.

This aligns with what upstream are using: https://pypi.org/project/taichi/#files

.github/workflows/manylinux_wheel.yml
hughperkins: Because it acts as a unit test that the uploading and downloading bit are working correclty; and means we can simply copy/paste the download part into other jobs.

duburcqa: Ok this is acceptable.

.github/workflows/scripts/ti_build/llvm.py
duburcqa: So why this change is necessary if you are not using manylinux 2024 here ? x)

hughperkins: because the function that detects 'manylinux2014' actually just detects manylinux in general, as far as I can tell.

hughperkins: (I mean, it tests for centos basically)

duburcqa: This is very fragile way of checking manylinux. But the OS is different for almost every manylinux version, so many it is actually able to detection manylinux 2014 in particular. Manylinux 2 28 is using another distribution so it should not be identified as manylinux 2014.

PR #39, opened 2025-06-19
[Test] Fold py38_only.py into appropriate other test scripts

hughperkins:
Issue: #

### Brief Summary

Fold py38_only.py into appropriate other test scripts

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_print.py
```diff
@@ -11,12 +11,46 @@
 # TODO: capfd doesn't function well on CUDA backend on Windows
 cuda_on_windows = (ti.cuda, "Windows")
 
-if sys.version_info >= (3, 8):
```
gasnica, line 14: Also remove `import sys` at the beginning of this file (?)

tests/python/test_print.py
```diff
@@ -11,12 +11,46 @@
...
+        # with self-documenting expressions
+        print(f"{a[0] = :.1f}")
+
+    func()
+    ti.sync()
+
+    out, err = capfd.readouterr()
+    expected_out = """a[0] = 1.0
+"""
+    print("out", out)
```
gasnica: The prints are not in the original version. Do we want to always print?

PR #41, opened 2025-06-20
[Type] Allow ti.template and ti.Template for typing

hughperkins:
Issue: #

### Brief Summary

Allow ti.template and ti.Template for typing

The follwing becomes valid:

```
@ti.kernel
def some_kernel(a: ti.template, b: ti.Template) -> None:
    a[None] = b[None] + 2
```

copilot:summary

### Walkthrough

copilot:walkthrough


python/taichi/lang/kernel_impl.py
```diff
@@ -215,7 +215,7 @@ def __init__(self, _func, _classfunc=False, _pyfunc=False, is_real_function=Fals
         self.extract_arguments()
         self.template_slot_locations = []
         for i, arg in enumerate(self.arguments):
-            if isinstance(arg.annotation, template):
+            if isinstance(arg.annotation, template) or arg.annotation == template:
```
duburcqa: You should flip the order of these two checks, it would be faster.

python/taichi/aot/module.py
```diff
@@ -49,7 +49,7 @@ def instantiate(self, **kwargs):
             template_args[index] = (key, value)
 
         for arg in kernel.arguments:
-            if isinstance(arg.annotation, template):
+            if arg.annotation == template or isinstance(arg.annotation, template):
```
gasnica, line 52: How is the equality operator working differently here from isinstance() ?

tests/python/test_template_typing.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+import taichi as ti
+from tests import test_utils
+
+
+@ti.kernel
+def some_kernel(a: ti.template, b: ti.Template) -> None:
```
gasnica, line 8: Is the lower-case ti.template without parentheses compiling here?

python/taichi/aot/module.py
hughperkins: `ti.template()` is an instance. `ti.template` is the class.

tests/python/test_template_typing.py
hughperkins: yeah, you can see there is a unit test, that is passing.

gasnica: @duburcqa 

Do we want to have this freedom to use either ti.template and ti.Template, and then possibly have inconsistent style in code? Or should we just remove the ti.Template alias?

hughperkins: @duburcqa your thoughts on this? (I can see good/bad points for both approaches, so I'll just go with the consensus basically).

duburcqa: Honestly I don't know. But I would prefer a unique syntax instead of multiple valid options. Still, `ti.template()` must be kept for backward compatibility. I'm inclined to prefer `ti.Template`, because it is more consistent with `typing` conventions across the world including numpy.

PR #42, opened 2025-06-20
[Type] Allow ti.types.NDArray with square brackets as type

hughperkins:
Issue: #

### Brief Summary

allow ti.types.NDArray[.., ...] as type

e.g. the following becomes valid:

```
@ti.kernel
def some_kernel(a: ti.types.NDArray[ti.i32, 2], b: ti.types.NDArray[ti.i32, 2]) -> None:
    for i, j in b:
        a[i, j] = b[i, j] + 2
```

copilot:summary

### Walkthrough

copilot:walkthrough


python/taichi/types/ndarray_type.py
```diff
@@ -124,6 +128,7 @@ def __str__(self):
 
 
 ndarray = NdarrayType
+NDArray = NdarrayType
```
gasnica, line 131: Is NdArray more preferrable than NDArray, to follow PascalCase naming convention for types?

hughperkins: Not sure. @duburcqa any opinions on this?

hughperkins: (Personally I'm fine either way)

duburcqa: Numpy is using NDArray, so let's stick to this. For reference: https://numpy.org/devdocs/reference/typing.html#numpy.typing.NDArray

tests/python/test_ndarray_typing.py
```diff
@@ -0,0 +1,19 @@
+import pytest
+
+import taichi as ti
+from tests import test_utils
+
+
+@ti.kernel
+def some_kernel(a: ti.types.NDArray[ti.i32, 2], b: ti.types.NDArray[ti.i32, 2]) -> None:
```
YilingQiao, line 8: Why do we prefer `ti.types.NDArray[ti.i32, 2]` instead of `ti.types. ndarray[ti.i32, 2]`?

hughperkins: Because non-primitive types tend to be in title case. e.g. numpy uses NDArray, see eg:
- https://medium.com/@goldengrisha/using-numpy-typing-for-type-safe-list-handling-in-python-35f8c99c76ac
- https://www.w3resource.com/numpy/snippet/exploring-numpy-typing.php
- https://jackatkinson.net/post/numpy_typing/

Or for allennlp:
- https://guide.allennlp.org/training-and-prediction#1

Torchtyping:
- https://github.com/patrick-kidger/torchtyping

Intuitively I think the types represent approximately the class of the thing that you are passing, and class names tend to be in title case.

hughperkins: That said:
- I think that we should choose whatever is least confusing for users, conditional on, the notation should be compliant with pyright and mypy linting
- so if you feel taht lowercase is less confusing, then that might be the right approach
- in the case of taichi, the class name is in fact lowercase, so it does make sense for the type to be lowercase too.

YilingQiao: Oh, I think which one to use in Genesis is still TBD, but this question doesn’t block the PR.

duburcqa: Let’s stick to what other people do. It is also nice to distinguiez typing from concrète types when they are not the same.

PR #50, opened 2025-06-30
[Build] Grid of macos python versions and upload to s3

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/scripts/ti_build/entry.py
+++ b/.github/workflows/scripts/ti_build/entry.py
@@ -8,6 +8,8 @@ import platform
 import subprocess
 import sys
 
+import psutil
+
 # -- third party --
 # -- own --
 from . import misc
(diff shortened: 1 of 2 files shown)


.github/workflows/macosx.yml
```diff
@@ -5,13 +5,23 @@ on:
       - main
   workflow_dispatch:
 jobs:
-  build:
+  build_mac_14_15:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-14, macos-15]
```
duburcqa, line 12: What about:
```yaml
matrix:
    MAC_OS_VERSION: [14]
    PYTHON_CP_VERSION: [cp310, cp311, cp312, cp313]
    include:
      - MAC_OS_VERSION: 15
        PYTHON_CP_VERSION: cp310
```

duburcqa: By the way, I think it should be the other way around. You should mainly using MacOS 15. These are the ARM runners.

.github/workflows/macosx.yml
```diff
@@ -26,3 +36,44 @@ jobs:
...
+        with:
+          name: mac_wheel_${{ matrix.MAC_OS_VERSION }}_${{ matrix.PYTHON_CP_VERSION }}
+          path: dist/*.whl
+  upload_to_storage:
+    name: Mac wheel upload
+    runs-on: ubuntu-22.04
+    needs: build_mac_14_15
+    environment: aws_storage
+    strategy:
+      matrix:
```
duburcqa, line 43: same.

.github/workflows/macosx.yml
hughperkins: re: change in syntax: persoally, I don't find this easier to read, because it mixes two different representations. But happy to update to this approach. I agree it has lower Kolmogorov complexity.

hughperkins: you know, I do prefer your suggested syntax. Much cleaner.

hughperkins: alos addressed swapping 14-15

PR #51, opened 2025-07-01
[Build] Win build on matrix, and upload to s3

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,7 @@ requires = [
     "pybind11",
     "pybind11-stubgen",
     "cmake>=3.11",
+    "scikit-build",
     "ninja; platform_system != 'Windows'",
     "psutil",
     "setuptools>=68.0.0",
--- a/.github/workflows/api_doc.yml
+++ b/.github/workflows/api_doc.yml
@@ -1,0 +1,58 @@
+name: api docs build
+
+on:
+  workflow_call:
+    inputs:
+      head_ref:
+        description: 'The head ref (branch name or PR ref)'
+        required: true
+        type: string
+    secrets:
+      AWS_ACCESS_KEY_ID:
+        required: true
+      AWS_SECRET_ACCESS_KEY:
+        required: true
+  workflow_dispatch:
+concurrency:
+  group: gh-pages-deploy-${{ github.event.inputs.head_ref || github.head_ref }}
+  cancel-in-progress: true
+permissions:
+  contents: write
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    environment: aws_storage
+    steps:
+      - uses: actions/checkout@v4
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - name: Download artifact
+        run: |
+          set -x
+          aws s3 sync s3://genesis-ai-compiler/ci/taichi/${GITHUB_SHA:0:7}/cp310/ dist/
+      - name: Python check
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: manylinux install wheel
+        run: |
+          set -x
+          ls -lh dist/
+          pip install dist/*.whl
+          python -c 'import taichi as ti; ti.init(arch=ti.cpu); print(ti.__version__)'
+      - name: Build docs
+        run: |
+          pip install pdoc
+          pdoc --output-dir public taichi
+      - name: Deploy docs
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./public
+          destination_dir: ${{ github.event.inputs.head_ref || github.head_ref }}
+          keep_files: true
+          publish_branch: gh-pages
--- a/.github/workflows/manylinux_wheel.yml
+++ b/.github/workflows/manylinux_wheel.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - main
   workflow_dispatch:
+permissions:
+  contents: write
 jobs:
   build_wheel:
     name: Manylinux wheel Build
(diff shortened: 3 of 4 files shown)


.github/workflows/win.yml
```diff
@@ -5,14 +5,26 @@ on:
...
 jobs:
-  build:
+  win_build:
     name: Windows 2025 Build
     runs-on: windows-2025
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - PYTHON_CP_VERSION: 'cp310'
```
duburcqa: This is redundant. Defining `PYTHON_CP_VERSION` should be sufficient. I don't think it is a good idea to define both.

.github/workflows/win.yml
```diff
@@ -5,14 +5,26 @@ on:
...
+          - PYTHON_CP_VERSION: 'cp312'
+            PYTHON_VERSION: '3.12'
+          - PYTHON_CP_VERSION: 'cp313'
+            PYTHON_VERSION: '3.13'
     steps:
       - name: Python check
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: ${{ MATRIX.PYTHON_VERSION}}
```
duburcqa, line 19: You should be able to use `PYTHON_CP_VERSION` here.

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 requires = [
     "setuptools",
     "wheel",
-    "numpy",
+    "numpy>=2.0.0",
```
duburcqa, line 5: We don't want this. This is not a requirement for Genesis and I don't see why it should be for Taichi.

requirements_test.txt
```diff
@@ -4,7 +4,7 @@ git+https://github.com/taichi-dev/pytest-xdist@a3b5ad3038#egg=pytest-xdist
 pytest-rerunfailures
 pytest-cov
 pytest-retry
-numpy
+numpy>=2.0.0  # otherwise, on windows, tries to install 1.26.4, which has no wheel for python 3.13
```
duburcqa, line 7: > otherwise, on windows, tries to install 1.26.4, which has no wheel for python 3.13

This must be something wrong. This is not how `pip` works. If numpy 1.26 is not available it should try installing 2.0 instead automatically (it is even the other way around for slack requirements).

.github/workflows/win.yml
hughperkins: without it, I need some complicated way to map from 'cp310' => '3.10', or visa versa. Having them listed out seems to me much easier to read than having a complicated mapping script?

.github/workflows/win.yml
hughperkins: oh really? 🤔  Interesting

pyproject.toml
hughperkins: It's used direclty by taichi, e.g. https://github.com/genesis-company/taichi/blob/0b259729546ddec7dc07725bde0f8bad29026078/python/taichi/lang/matrix.py#L134

requirements_test.txt
hughperkins: 1.26 is available - as source. It downloads the numpy source, attempts to compile it, and compilation fails, for various reasons. I'd prefer to avoid having to add lots of machinery to build a very old version of numpy, ideally, when a newer version will work just fine, without having to compile it from scratch.

On python 3.10-3.12, it installs numpy wheels just fine. But not on python 3.13.

.github/workflows/win.yml
hughperkins: using `cp313` etc doesnt appear to work:

Screenshot 2025-07-03 at 7 40 44 AM

requirements_test.txt
duburcqa: Just do `pip install --prefer-binary`

pyproject.toml
duburcqa: I don't get it.

hughperkins: Not sure I follow. Taichi directly uses numpy in its code. Therefore, we need to install numpy when installing taichi. Therefore numpy needs to be in the dependencies for taichi installation.

.github/workflows/win.yml
duburcqa: OK indeed I was confused. Still, I would recommend define one based on the other.

requirements_test.txt
hughperkins: Interesting

.github/workflows/win.yml
hughperkins: would need some gross looking scripting. Happy to do that though 🤔

.github/workflows/win.yml
duburcqa: I would still advocate for a "complicated" mapping.

duburcqa: `python -c "from packaging import tags ; print(next(tags.sys_tags()).interpreter)"`

pyproject.toml
duburcqa: Yes, I'm talking about `>= 2.0` only of course :)

.github/workflows/win.yml
hughperkins: stupid question, but how to get this into:

Screenshot 2025-07-03 at 9 27 50 AM

?

hughperkins: oh, figured it out

.github/workflows/win.yml
```diff
@@ -5,14 +5,24 @@ on:
...
+      - name: Set Python CP version
+        id: set_cp_version
+        run: |
+          PYTHON_CP_VERSION=$(python -c "from packaging import tags ; print(next(tags.sys_tags()).interpreter)")
```
duburcqa: Windows OS - style is like this:
`$PYTHON_CP_VERSION = (python -c "from packaging import tags ; print(next(tags.sys_tags()).interpreter)")`

hughperkins: how to make it work on all 3? (linux + mac + win) ?

hughperkins: oh I guess this is windows specific :)

requirements_test.txt
hughperkins: Tried that. But still unisntalls 2.3.1 on python 3.13, and then fails:

Screenshot 2025-07-03 at 1 46 00 PM

https://github.com/genesis-company/taichi/actions/runs/16059405242/job/45321637365?pr=51

Screenshot 2025-07-03 at 1 47 40 PM

Thoughts?

hughperkins: numpy 1 is EOL from september 2025:

Screenshot 2025-07-03 at 1 52 41 PM

hughperkins: https://scientific-python.org/specs/spec-0000/

PR #52, opened 2025-07-02
[Misc] Add TI_DUMP_AST

hughperkins:
Issue: #

### Brief Summary

Add TI_DUMP_AST=1 env var, to dump ast files into /tmp/ast folder.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/macosx.yml
+++ b/.github/workflows/macosx.yml
@@ -5,13 +5,17 @@ on:
       - main
   workflow_dispatch:
 jobs:
-  build:
+  build_mac_14_15:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-14, macos-15]
+        MAC_OS_VERSION: [15]
+        PYTHON_CP_VERSION: [cp310, cp311, cp312, cp313]
+        include:
+          - MAC_OS_VERSION: 14
+            PYTHON_CP_VERSION: cp310
     name: Mac OS Build
-    runs-on: ${{ matrix.os}}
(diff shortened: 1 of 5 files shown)


python/taichi/lang/kernel_impl.py
```diff
@@ -675,6 +678,38 @@ def taichi_ast_generator(kernel_cxx):
...
+                            "type": node.__class__.__name__,
+                            "fields": fields,
+                            "lineno": getattr(node, "lineno", None),
+                            "col_offset": getattr(node, "col_offset", None),
+                        }
+                    if isinstance(node, list):
+                        return [ast_to_dict(x) for x in node]
+                    return node  # Basic types (str, int, None, etc.)
+
+                if "TAICHI_DUMP_AST" in os.environ:
```
duburcqa: AFAIK, taichi is usually checking the value of environment variables. Typically `TI_OFFLINE_CACHE=1` is different from `TI_OFFLINE_CACHE=0`. I would recommend doing the same. Moreover, mixing `TI_` prefix and `TAICHI_` prefix for environment variables is confusing.

python/taichi/lang/kernel_impl.py
```diff
@@ -675,6 +678,38 @@ def taichi_ast_generator(kernel_cxx):
...
+                    target_dir = "/tmp/ast"
```
duburcqa: You should probably move to `pathlib` at some point.

hughperkins: Interesting

PR #53, opened 2025-07-02
[Misc] Migrate to TI_DUMP_IR and TI_LOAD_IR and compare with 1

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


taichi/codegen/llvm/codegen_llvm.cpp
```diff
@@ -2750,7 +2750,7 @@ LLVMCompiledTask TaskCodeGenLLVM::run_compilation() {
     }
   }
   const char *dump_ir_env = std::getenv(DUMP_IR_ENV.data());
-  if (dump_ir_env != nullptr) {
+  if (dump_ir_env != nullptr && std::string(dump_ir_env) == "1") {
```
bo-rc: nitpicking change to `if (dump_ir_env && std::string(dump_ir_env) == "1") {`

and it seems that this is a recurring check pattern. Does it deserve a helper functions which is properly named so readers understand the intent easily?

hughperkins: sure ok. Thoughts on naming? (naming and off by one errors are the three hardests problems in programming :) )

duburcqa: What about something like `check_env_flag` ?

bo-rc: `const char *dump_ir_env = std::getenv(DUMP_IR_ENV.data());` this line also is repeating. There are lots of code redundancy in the code blocks.

> on checking `=="1"` 

what does this magic number string mean? Is there a more descriptive string to use here?

hughperkins: `1` is easy to write, when using it. easier than 'ON' or 'yes' etc

hughperkins: (but yeah, that is an argument in favor of not checking it, just checking if the env var is defined at all)

hughperkins: Turns out we already have one. Will use that

Screenshot 2025-07-09 at 9 17 46 AM

hughperkins: addressed in 5611537b0

PR #60, opened 2025-07-05
[Type] Bunch of typing added for kernel_impl.py, impl.py, and related

hughperkins:
Issue: #

(note that this contains also changes from
- https://github.com/genesis-company/taichi/pull/57 (add stub post-processing)
- https://github.com/genesis-company/taichi/pull/55 (re-add scikit-build)

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,7 @@ endif()
 
 # No support of Python for Android build; or in any case taichi is integrated
 # in another project as submodule.
+set(PYBIND11_VERSION "2.13.6" CACHE STRING "Pybind11 version to use")
 option(TI_WITH_PYTHON "Build with Python language binding" ON)
 if (TI_WITH_PYTHON AND NOT ANDROID)
     include(cmake/PythonNumpyPybind11.cmake)
--- a/cmake/PythonNumpyPybind11.cmake
+++ b/cmake/PythonNumpyPybind11.cmake
@@ -12,10 +12,15 @@ include_directories(${NUMPY_INCLUDE_DIR})
 
 include(FetchContent)
 
+message("Using pybind11 version: ${PYBIND11_VERSION}")
 FetchContent_Declare(
   pybind11
   GIT_REPOSITORY https://github.com/pybind/pybind11.git
-  GIT_TAG v2.13.6  # Should be less than 3.0.0 for now
+  GIT_TAG v${PYBIND11_VERSION}  # Should be less than 3.0.0 for now
+     # pybind11 3.0.0 caused builds to start failing. This PR downgrades pybind11 to latest <3.0.0 release,
+     # which fixes the issue
+     # whilst we could probably get 3.0.0 working too, I feel that x.0.0 releases, in opensource (and closed source
+     # too often) might not always be very stable, so we might want to wait 3-6 months before trying it?)
 )
 
 FetchContent_MakeAvailable(pybind11)


python/taichi/lang/impl.py
```diff
@@ -324,18 +336,19 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        self.compiling_callable: Function | None = None  # pointer to instance of lang::Kernel/Function
+        self.current_kernel: Kernel | None = None
         self.global_vars = []
         self.grad_vars = []
         self.dual_vars = []
         self.matrix_fields = []
         self.default_fp = f32
         self.default_ip = i32
         self.default_up = u32
+        self.print_full_traceback = False
```
gasnica: : bool 

:-)

python/taichi/lang/impl.py
```diff
@@ -66,20 +72,20 @@
 
 @taichi_scope
 def expr_init_shared_array(shape, element_type):
-    return (
-        get_runtime()
-        .compiling_callable.ast_builder()
-        .expr_alloca_shared_array(shape, element_type, _ti_core.DebugInfo(get_runtime().get_current_src_info()))
+    compiling_callable = get_runtime().compiling_callable
```
gasnica, line 75: why do we only need to check for compiling_callable being not None? while everything else -- get_runtime(), ast_builder(), and others are certain to be not None ?

hughperkins: There's a lot more typing coming. If you prefer you can jump straight to https://github.com/genesis-company/taichi/pull/69 No opinion either way. One giant pr, wiht lots of changes, or smaller incremental prs.

hughperkins: oh, you mean, in this specfic line? Well, because get_runtime() does not return a type that can be None, according to he typing:

Screenshot 2025-07-09 at 10 13 31 AM

hughperkins: if we modified get_runtime() to look like:

```
def get_runtime() -> PyTaichi | None:
    return pytaichi
```
...then the type says that it can be None, so we'd need to check it.

python/taichi/lang/kernel_impl.py
```diff
@@ -234,20 +241,22 @@ def __call__(self, *args, **kwargs):
...
                 raise TaichiSyntaxError("Real function in gradient kernels unsupported.")
             instance_id, arg_features = self.mapper.lookup(args)
             key = _ti_core.FunctionKey(self.func.__name__, self.func_id, instance_id)
-            if self.compiled is None:
-                self.compiled = {}
             if key.instance_id not in self.compiled:
                 self.do_compile(key=key, args=args, arg_features=arg_features)
             return self.func_call_rvalue(key=key, args=args)
+        current_kernel = impl.get_runtime().current_kernel
+        assert current_kernel is not None
```
gasnica: move this current_kernel local variable and the assert to before the previous `if self.is_real_function` to only assert in one place

python/taichi/types/compound_types.py
```diff
@@ -7,7 +7,8 @@
 
 
 class CompoundType:
-    pass
+    def from_kernel_struct_ret(self, launch_ctx, index: tuple):
```
gasnica, line 10: is this used?

python/taichi/lang/kernel_impl.py
```diff
@@ -308,7 +317,9 @@ def do_compile(self, key, args, arg_features):
         tree, ctx = _get_tree_and_ctx(
             self, is_kernel=False, args=args, arg_features=arg_features, is_real_function=self.is_real_function
         )
-        fn = impl.get_runtime().prog.create_function(key)
+        prog = impl.get_runtime().prog
```
gasnica: would we want a helper function get_prog() that does the assert and returns the prog, since we do it so many times?

possibly simiarlly for getting compiling_callable.

hughperkins: Intresting question 🤔  Will ponder what options exist for this.

hughperkins: renamed prog => _prog, and added prog property. thoughts?

python/taichi/lang/_ndarray.py
```diff
@@ -240,15 +240,20 @@ class ScalarNdarray(Ndarray):
...
         )
         self.shape = tuple(self.arr.shape)
         self.element_type = dtype
 
     def __del__(self):
-        if impl is not None and impl.get_runtime() is not None and impl.get_runtime().prog is not None:
-            impl.get_runtime().prog.delete_ndarray(self.arr)
+        if (
+            impl is not None
+            and impl.get_runtime is not None
```
duburcqa, line 252: `impl.get_runtime` may be `None`. What the fuck?

python/taichi/lang/impl.py
```diff
@@ -183,14 +189,16 @@ def validate_subscript_index(value, index):
         validate_subscript_index(value, index.start)
         validate_subscript_index(value, index.stop)
 
-    if isinstance(index, numbers.Number) and index < 0:
+    if isinstance(index, int) and index < 0:
```
duburcqa, line 192: What index is supposed to be if not an integer at this point ? I would check if `isinstance(index, int)`, and check that index is positive, otherwise raise an exception for wrong `index` type.

python/taichi/lang/kernel_impl.py
```diff
@@ -206,16 +213,16 @@ def _process_args(self, args, kwargs):
...
         Func.function_counter += 1
-        self.compiled = None
+        self.compiled = {}
         self.classfunc = _classfunc
         self.pyfunc = _pyfunc
         self.is_real_function = is_real_function
-        self.arguments = []
-        self.return_type = None
+        self.arguments: list[KernelArgument] = []
+        self.return_type: tuple[Type] or None = None
```
duburcqa: `tuple[Type, ...] | None` ?

python/taichi/lang/kernel_impl.py
```diff
@@ -880,24 +888,25 @@ def call_back():
...
+                return float(x)
 
-                def cast_func(x):
-                    if not isinstance(x, (int, float, np.integer, np.floating)):
-                        raise TaichiRuntimeTypeError(
-                            f"Argument {needed.dtype.to_string()} cannot be converted into required type {type(x)}"
-                        )
-                    return float(x)
+            def cast_int(x):
+                if not isinstance(x, (int, np.integer)):
```
duburcqa, line 895: Use `try-catch`, it would be more efficient in Python nowadays:
```
try:
    return int(x)
except TypeError as e:
    raise TaichiRuntimeTypeError(
        f"Argument {needed.dtype.to_string()} cannot be converted into required type {type(x)}"
    ) from e
```

python/taichi/lang/kernel_impl.py
gasnica: Looks neat :-)

Okay, so we sometimes access _prog with asserts, sometimes without.

I'm not asking for anything more here :-) 

But to share my own style -- I prefer to be more explicit in code -- _prog and prog are effectively the same name. I'd name the property with assert e.g. prog_checked. Or actually I tend to have longer names for unsafe methods, so I'd have 2 properties prog and prog_unchecked, and not access _prog directly. 

But that's just my idea/preference. The code looks grand as-is :-)

python/taichi/lang/kernel_impl.py
hughperkins: try...catch is faster than isinstance? 🤔  Are you sure about that? My understanding, in general, is that exception handling is always a much slower path?

python/taichi/lang/kernel_impl.py
hughperkins: good spot.

python/taichi/lang/impl.py
hughperkins: For whateve reason, some combinatino of pylance and/or pyright would not allow comarison with Number, only int.

anyway, responding to your suggestino, you mean somethign like? :
```
def validate_subscript_index(value, index):
    if isinstance(value, Field):
        # field supports negative indices
        return
    elif isinstance(index, Expr):
        return
    elif isinstance(index, Iterable):
        for ind in index:
            validate_subscript_index(value, ind)
    elif isinstance(index, slice):
        validate_subscript_index(value, index.start)
        validate_subscript_index(value, index.stop)
    elif isinstance(index, int) 
        if index < 0:
            raise TaichiSyntaxError("Negative indices are not supported in Taichi kernels.")
    else:
        raise TaichiSyntaxError(f"Invalid index type {type(index)}.")
```
?

python/taichi/lang/_ndarray.py
hughperkins: Yes, this is not a typing issue. This is a practical real thing that I experienced systematically.

What happens is taht when the program finishes, stuff starts being destroyed.

At the time taht __del__ is called on the ndarray:
- misc.__name__ is None
- misc. (some other thing I forget, that also shows up when you do str(misc)) was also None
- imp.get_runtime was None
- ... and an ugly exception appears

By adding this conitioanl, we elmiinate the ugly exception message.

python/taichi/lang/kernel_impl.py
hughperkins: oh, well, I suppose if we assume it usually is an int, and we dont use the xcept, then that makes sense. Empircally though isinstance is slightly faster:

```
import time

num_its = 10000000


for val_str in ["5", "x"]:
    print("")
    print("==========")
    print(val_str)
    start = time.time()
    for it in range(num_its):
        if isinstance(val_str, int):
            val = int(val_str)

    end = time.time()
    print('elapsed', end - start)


    start = time.time()
    val_str = "5"
    for it in range(num_its):
        try:
            val = int(val_str)
        except:
            pass

    end = time.time()
    print('elapsed', end - start)

```
(taichi) system ~/git/taichi (hp/typing-2-b|…4△1) $ python /tmp/test.py

==========
5
elapsed 0.517409086227417
elapsed 0.6724801063537598

==========
x
elapsed 0.5407359600067139
elapsed 0.6832132339477539
```

(Also, cleaner/easeier to write I feel: fewer lines)

Comment (hughperkins) [reply python/taichi/lang/kernel_impl.py:895]: wait a bug in my code:
- for 5, isinstanae islighlty faster
- for x, isinstance is *much* faster:

```
import time

num_its = 10000000


for val_str in ["5", "x"]:
    print("")
    print("==========")
    print(val_str)
    start = time.time()
    for it in range(num_its):
        if isinstance(val_str, int):
            val = int(val_str)

    end = time.time()
    print('elapsed', end - start)


    start = time.time()
    # val_str = "5"
    for it in range(num_its):
        try:
            val = int(val_str)
        except:
            pass

    end = time.time()
    print('elapsed', end - start)

```

```
(taichi) system ~/git/taichi (hp/typing-2-b|…4△1) $ python /tmp/test.py

==========
5
elapsed 0.5399680137634277
elapsed 0.6968648433685303

==========
x
elapsed 0.5562381744384766
elapsed 3.135991096496582
```

File: python/taichi/lang/impl.py (hunk 7 above)
Comment (hughperkins) [reply python/taichi/lang/impl.py:192]: or possibly:
```
def validate_subscript_index(value, index):
    if isinstance(value, Field):
        # field supports negative indices
        return

    if isinstance(index, Expr):
        return

    if isinstance(index, Iterable):
        for ind in index:
            validate_subscript_index(value, ind)
        return

    if isinstance(index, slice):
        validate_subscript_index(value, index.start)
        validate_subscript_index(value, index.stop)
        return

    if isinstance(index, int):
        if index < 0:
            raise TaichiSyntaxError("Negative indices are not supported in Taichi kernels.")
        return

    raise TaichiSyntaxError(f"Invalid index type {type(index)}.")
```
?

PR #67, opened 2025-07-06
[Type] Typing batch 4, including handling kernel/func/real_func wrapper

hughperkins:
Issue: #

### Brief Summary

Note: depends on previous PRs (batch 2b, 2c, 2d, 3...)

copilot:summary

### Walkthrough

copilot:walkthrough


python/taichi/lang/_ndarray.py
```diff
@@ -27,7 +32,7 @@ def __init__(self):
         self.dtype = None
         self.arr = None
         self.layout = Layout.AOS
-        self.grad = None
+        self.grad: "ScalarNdarray" | "VectorNdarray" | "MatrixNdarray" | None = None
```
duburcqa: Having some AnyNdarray type that would corresponds to `"ScalarNdarray" | "VectorNdarray" | "MatrixNdarray"` may be useful to hide underlying complexity?

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,54 @@
...
+        return self.wrapper.__call__(*args, **kwargs)
+
+    def __get__(self, instance, owner):
+        if instance is None:
+            return self
+        return BoundFunc(self.wrapper, instance, self)
+
 
-def func(fn: Callable, is_real_function: bool = False):
+def func(fn: Callable, is_real_function=False) -> TaichiCallable:
```
gasnica: removed 'bool' arg type

python/taichi/lang/_ndarray.py
gasnica: is 'ScalarNdarray' imported in this file?

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -16,7 +16,15 @@
 import numpy as np
 
 from taichi._lib import core as _ti_core
-from taichi.lang import _ndarray, any_array, expr, impl, kernel_arguments, matrix, mesh
+from taichi.lang import (
```
gasnica: that's the same list. Why change to multiline?

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -628,13 +636,15 @@ def build_Call(ctx: ASTTransformerContext, node: ast.Call):
...
             return node.ptr
+
         ASTTransformer.warn_if_is_external_func(ctx, node)
         try:
             node.ptr = func(*args, **keywords)
         except TypeError as e:
             module = inspect.getmodule(func)
             error_msg = re.sub(r"\bExpr\b", "Taichi Expression", str(e))
-            msg = f"TypeError when calling `{func.__name__}`: {error_msg}."
+            func_name = getattr(func, "__name__", func.__class__.__name__)
```
gasnica, line 638: so this is a change to handle both object- and class methods? why is that needed?

python/taichi/lang/_ndarray.py
hughperkins: It doesnt appear to be, but for some reason neither pylance nor pyright complain, and it shows up in green in vs code 🤔 
Screenshot 2025-07-21 at 4 22 12 PM
Screenshot 2025-07-21 at 4 22 42 PM

hughperkins: because its in the same file :)

hughperkins: Screenshot 2025-07-21 at 4 23 52 PM

python/taichi/lang/ast/ast_transformer.py
hughperkins: I assume pre-commit/pylint changed it.

hughperkins: hmm, not sure. can revert

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,54 @@
 from taichi.types.enums import AutodiffMode, Layout
 from taichi.types.utils import is_signed
 
+CompiledKernelKeyType = tuple[Callable, int, AutodiffMode]
+
+
+class BoundFunc:
```
gasnica: What's the purpose of this. Would a docstring be useful here?

python/taichi/lang/ast/ast_transformer.py
hughperkins: Honestly,  Idon't remember. But I'm pretty sure that without this, you. get really ugly error messages sometimes. Our tests on error messages are ~vanishingly few.

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,54 @@
...
+        return res
+
+    def __setattr__(self, k: str, v: Any) -> None:
+        if k in ("fn", "instance", "taichi_callable"):
+            object.__setattr__(self, k, v)
+        else:
+            setattr(self.taichi_callable, k, v)
+
+
+class TaichiCallable:
```
gasnica, line 75: similarly. a docstring maybe?

So this class now is used in stead of @functools.wraps decorator? Why are we doing that?

python/taichi/lang/kernel_impl.py
```diff
@@ -123,33 +172,31 @@ def pyfunc(fn: Callable):
...
+    return taichi_callable
 
 
 def _get_tree_and_ctx(
     self: "Func | Kernel",
+    args: tuple[Any, ...],
     excluded_parameters=(),
     is_kernel: bool = True,
     arg_features=None,
-    args=None,
```
gasnica, line 142: possibly reorder args_features param too to match is_kernel, args, arg_features order used in do_compile() ... (?)

python/taichi/lang/kernel_impl.py
hughperkins: Added. see updated code.

python/taichi/lang/kernel_impl.py
hughperkins: See docstring for `BoundFunc`. Added docstring to TaichiCallable to refer to BoudnFunc docstirng.

python/taichi/lang/kernel_impl.py
hughperkins: args_features is not part of this PR, and I'd prefer to avoid reordering arguments on the whole, without a strong reason, becasue stuff can break, when argumetns are passed by position.

python/taichi/lang/_ndarray.py
```diff
@@ -27,7 +30,7 @@ def __init__(self):
         self.dtype = None
         self.arr = None
         self.layout = Layout.AOS
-        self.grad = None
+        self.grad: "ScalarNdarray | VectorNdarray | MatrixNdarray | None" = None
```
duburcqa: Still no `AnyArray` type ? This long types are districting. Hiding this complexity would be better.

python/taichi/lang/ast/ast_transformer.py
duburcqa: > Honestly, Idon't remember.

And this is why I recommend adding a comment when doing this type of changes. Because someone will change it again at some point, thinking rightfully that it is overly complicated for no reason, and best case scenario a few hours of work will be lost. But we already had this discussion x)

python/taichi/lang/impl.py
```diff
@@ -1162,7 +1160,8 @@ def static(x, *xs) -> Any:
         return x
     if isinstance(x, Field):
         return x
-    if isinstance(x, (FunctionType, MethodType)):
+    if isinstance(x, (FunctionType, MethodType, BoundFunc, TaichiCallable)):
+        print(" is instanace FunctionType or MethodType", x, type(x))
```
duburcqa: What is this half-baked sentence ? (also spelling mistake 'instanace')

stub_replacements_global.yaml
```diff
@@ -0,0 +1,8 @@
+# This file is used for post-processing of the pybind11-stubgen generated stubs.
+# The file is a yaml file where:
+# - the key is text that will be matched (pure string match, no wildcards/regular expressions), and
+# - the value is the replacement text.
+#
+# This will be used in the first instance to replace all list[int] with tuple[int, ...]
+# rather than using `stub_replacmeent_funcs.yaml` on every function that takes a vector<int>.
+"list[int]": "tuple[int, ...]"
```
duburcqa, line 8: Are you sure this type is valid? All your functions only takes tuple as input arguments? What about using `typing.Sequence[int]` or even `typing.Iterable[int]` to be more generic?

tests/python/test_ptr_assign.py
```diff
@@ -71,37 +71,48 @@ def func(x: ti.f32, y: ti.f32):
 
 @test_utils.test()
 def test_ptr_func():
-    a = ti.field(dtype=ti.f32, shape=())
+    a = ti.field(dtype=ti.f32, shape=(3))
```
duburcqa: `shape=(3,)`

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,115 @@
...
+    def __setattr__(self, k: str, v: Any) -> None:
+        if k in ("fn", "instance", "taichi_callable"):
+            object.__setattr__(self, k, v)
+        else:
+            setattr(self.taichi_callable, k, v)
+
+
+class TaichiCallable:
+    """
+    See docstring for `BoundFunc` for what this is being used for
```
duburcqa: Valid English if possible --> at least a final dot.

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,115 @@
...
+    """
+    def __init__(self, fn: Callable, wrapper: Callable) -> None:
+        # self.func: Func | None = None
+        self.fn = fn
+        self.wrapper = wrapper
+        self._is_real_function = False
```
duburcqa: Why not adding typing directly for new implementations? Otherwise it may never be fully typed :/

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,115 @@
...
+            object.__setattr__(self, k, v)
+        else:
+            setattr(self.taichi_callable, k, v)
+
+
+class TaichiCallable:
+    """
+    See docstring for `BoundFunc` for what this is being used for
+    """
+    def __init__(self, fn: Callable, wrapper: Callable) -> None:
```
duburcqa, line 148: Are these two callable completely independent? Otherwise typing is not right.

python/taichi/lang/kernel_impl.py
```diff
@@ -91,22 +203,20 @@ def func(fn: Callable, is_real_function: bool = False):
...
-
-    @functools.wraps(fn)
-    def decorated(*args, **kwargs):
-        return fun.__call__(*args, **kwargs)
-
-    decorated._is_taichi_function = True
-    decorated._is_real_function = is_real_function
-    decorated.func = fun
-    return decorated
+    taichi_callable = TaichiCallable(
```
duburcqa: Splitting on multiple line should improve readability, which is not the case here: `taichi_callable = TaichiCallable(fn, fun)`

python/taichi/lang/kernel_impl.py
```diff
@@ -123,33 +233,31 @@ def pyfunc(fn: Callable):
...
-
-    @functools.wraps(fn)
-    def decorated(*args, **kwargs):
-        return fun.__call__(*args, **kwargs)
-
-    decorated._is_taichi_function = True
-    decorated._is_real_function = False
-    decorated.func = fun
-    return decorated
+    taichi_callable = TaichiCallable(
```
duburcqa: Same.

python/taichi/lang/kernel_impl.py
duburcqa, line 148: I would rather add `is_real_function`, `is_taichi_function`, `is_wrapped_kernel`, `is_classkernel` as input arguments. That would be more natural. And froze them if possible since it is not supposed to change from what I understand (but not sure it is worth the extra boilerplate).

python/taichi/lang/kernel_impl.py
```diff
@@ -459,6 +570,17 @@ def extract_arguments(self) -> None:
...
 
 
+AnnotationType = Union[
+    template,
+    ArgPackType,
+    "texture_type.TextureType",
+    "texture_type.RWTextureType",
+    ndarray_type.NdarrayType,
+    sparse_matrix_builder,
+    Any,
```
duburcqa, line 583: This is wrong, this should be at most `typing.Hashable`, and maybe `typing.Type` (which is significantly more restrictive).

python/taichi/lang/kernel_impl.py
```diff
@@ -937,10 +1062,9 @@ def call_back():
                         int(v.grad.data_ptr()) if v.grad is not None else 0,
                     )
                 else:
-                    raise TaichiRuntimeTypeError(
-                        f"Argument {needed.to_string()} cannot be converted into required type {v}"
-                    )
+                    raise TaichiRuntimeTypeError(f"Argument {needed} cannot be converted into required type {v}")
             elif has_paddle():
+                # Do we want to continue to support paddle? :thinking_face:
```
duburcqa, line 1071: As I already mentioned, having some tag prefix to easily find this kind of comments in the future would be nice. Anyscale company is using `# (@hughperkins)` as prefix in Ray.

python/taichi/lang/kernel_impl.py
```diff
@@ -1094,7 +1214,7 @@ def recursive_set_args(needed_arg_type, provided_arg_type, v, indices):
             if isinstance(needed_arg_type, StructType):
                 if in_argpack:
                     return 1
-                if not isinstance(v, needed_arg_type):
+                if not isinstance(v, needed_arg_type):  # type: ignore Might be invalid? Maybe should rewrite as: not needed.__instancecheck__(v) ?
```
duburcqa: `not needed.__instancecheck__(v)` is exactly the same as `isinstance(v, needed)`

python/taichi/lang/kernel_impl.py
```diff
@@ -1257,24 +1378,32 @@ def _kernel_impl(_func: Callable, level_of_class_stackframe: int, verbose: bool
...
         @functools.wraps(_func)
-        def wrapped(*args, **kwargs):
+        def wrapped_classkernel(*args, **kwargs):
             # If we reach here (we should never), it means the class is not decorated
             # with @ti.data_oriented, otherwise getattr would have intercepted the call.
             clsobj = type(args[0])
             assert not hasattr(clsobj, "_data_oriented")
             raise TaichiSyntaxError(f"Please decorate class {clsobj.__name__} with @ti.data_oriented")
 
+        wrapped = TaichiCallable(
```
duburcqa: Same. One line is fine.

python/taichi/lang/kernel_impl.py
```diff
@@ -1257,24 +1378,32 @@ def _kernel_impl(_func: Callable, level_of_class_stackframe: int, verbose: bool
...
-        def wrapped(*args, **kwargs):
+        def wrapped_func(*args, **kwargs):
             try:
                 return primal(*args, **kwargs)
             except (TaichiCompilationError, TaichiRuntimeError) as e:
                 if impl.get_runtime().print_full_traceback:
                     raise e
                 raise type(e)("\n" + str(e)) from None
 
+        wrapped = TaichiCallable(
```
duburcqa: Same.

python/taichi/lang/kernel_impl.py
hughperkins: well, `isinstance` fails typing/linting checks. Thoughts?

python/taichi/lang/_ndarray.py
hughperkins: AnyArray is a specific type in Taichi. Different from ndarraytype. I'd prefer not to use a name similar to AnyArray. I already find AnyArray pretty confusing.

I need to fix ndarray typing at some point (currently broken). Maybe I think about this at that point?

python/taichi/lang/_ndarray.py
hughperkins: AnyArray exists, but is a different concept: including external arrays, such as torch, numpy etc. It's a specficic type, and is not an alias of ScalarNdarray etc.

python/taichi/lang/impl.py
hughperkins: scaffolding/spam/scar tissue. removed.

stub_replacements_global.yaml
hughperkins: It's not that they only take tupels as input. It's that the python calling code is all passing tuples as input. And tuple[int] cannot be passed into list[int].

Note: if I change from `tuple[int, ...]` to `Sequence[int]`, then in VSCode, the type shows as "unknown", see screenshots:
with Sequence:
Screenshot 2025-07-22 at 12 13 30 PM
with tuple:
Screenshot 2025-07-22 at 12 17 18 PM

python/taichi/lang/kernel_impl.py
hughperkins: sorry, I'm not sure I understand what you mean by "adding typing directly"?

python/taichi/lang/kernel_impl.py
hughperkins: They are both Callables. They are both functors, which can be executed. What do you mean by 'completely independent'? And how would any dependence (primarily "has-a") affect the typing?

hughperkins: That would make this bit of code longer:
```
    wrapped: TaichiCallable
    if is_classkernel:
        @functools.wraps(_func)
        def wrapped_classkernel(*args, **kwargs):
            clsobj = type(args[0])
            assert not hasattr(clsobj, "_data_oriented")
            raise TaichiSyntaxError(f"Please decorate class {clsobj.__name__} with @ti.data_oriented")
        wrapped = TaichiCallable(
            _func,
            wrapped_classkernel,
        )
    else:

        @functools.wraps(_func)
        def wrapped_func(*args, **kwargs):
            try:
                return primal(*args, **kwargs)
            except (TaichiCompilationError, TaichiRuntimeError) as e:
                if impl.get_runtime().print_full_traceback:
                    raise e
                raise type(e)("\n" + str(e)) from None

        wrapped = TaichiCallable(
            _func,
            wrapped_func,
        )
        wrapped.grad = adjoint

    wrapped._is_wrapped_kernel = True
    wrapped._is_classkernel = is_classkernel
    wrapped._primal = primal
    wrapped._adjoint = adjoint
```

python/taichi/lang/kernel_impl.py
hughperkins: what is the type of `@dataclasses.dataclass`? Similar question for `ti.data_oriented`.

python/taichi/lang/kernel_impl.py
hughperkins: added `#maybeprunable`

python/taichi/lang/kernel_impl.py
gasnica: So, is the wrapper parameter always of type class Func?

It seems so currently -- if I understand well --  and we initialize TaichiCallable in two places in kernel_impl. Should we change wrapper type to Func ?

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
 from taichi.types.enums import AutodiffMode, Layout
 from taichi.types.utils import is_signed
 
+CompiledKernelKeyType = tuple[Callable, int, AutodiffMode]
+
+
+class BoundFunc:
+    """
+    This class is used to enable wrapping a bindable function with a class.
```
gasnica: Multiple tiny requests, after our call, Hugh -- hoping for this to be easier to understand for a new person:

* Move TaichiCallable above BoundFunc, since chronologically that's what we first create. And also BoundFunc references TaichiCallable.
* Consider renaming BoundFunc, to BoundTaichiCallable, because:
   - BoundFunc suggests it's directly related to taichi class Func, 
   - while it's not -- it's related to TaichiCallable
 * I feel this comment belongs in TaichiCallable, not in this BoundFunc/BoundTaichiCallable <-- because it mentions passing wrapped functions around, which we kinda (maybe i'm wrong with my interpretation) do with TaichiCallable, up till the moment it's called and that's when it becomes BoundXxx. So maybe let's have the comment under TaichiCallable.
 * in the docs, let's refer to specific classes by name, rather than using vague names like 'This class is used to enable wrapping a bindable function' --> "TaichCallable wraps taichi's class Func instance." 
 
 I'll add edit suggestions below.

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+
+    We have the following requirements:
+    - pass wrapped functions around
+    - assign attributes to these functions, such as `_if_real_function`, `_primal`, etc
+    - use strong typing, and type checkers, such as pyright/mypy
+    - the wrapped functions should be able to be used with class instances
+
```
gasnica, line 90: My suggestion for edit of the above 6 lines:

    Design requirements for TaichiCallable:
    - wrap/contain a reference to a class Func instance, and allow (the TaichiCallable) being passed around like normal function pointer
       - [we could add a sentence describing class Func]
    - expose attributes of the wrapped class Func, such as `_if_real_function`, `_primal`, etc
    - allow for (now limited) strong typing, and enable type checkers, such as pyright/mypy
       - currently TaichiCallable is a shared type used for all functions marked with @ti.func, @ti.kernel, python functions (?) 
       - note: current type-checking implementation does not distinguish between different type flavors of TaichiCallable, with different values of `_if_real_function`, `_primal`, etc
    - handle not only class-less functions, but also class-instance methods (where determining the `self` reference is a challenge)

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+        @ti.kernel
+        def func(self):
+            a, add_py, add_func = ti.static(self.a, self.add2numbers_py, self.add2numbers_func)
+            a[0] = add_py(2, 3)
+            a[1] = add_func(3, 7)
+
+    (taken from test_ptr_assign.py).
+
+    When @ti.func runs, the function `add2numbers_func` exists, but there is not yet any `self`
```
gasnica: When @ti.func runs --> When the @ti.func decorator is parsed

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+    - it is not possible for the method to be bound, to a `self` instance
+    - however, it is at this point that we create the wrapped function, via the @ti.func annotation,
+      which runs the `func` function
```
gasnica: Above two lines (current phrasing might be misleading, imho)

 - however, the @ti.func annotation, runs the `kernel_imp.py::func` function --- it is at this point that Taichi's original code creates a class Func instance (that wraps the `add2numbers_func`) and immediately we create a TaichiCallable instance that wraps the FuncInstance. 
     - effectively, we have two layers of wrapping TaichiCallable->Func->function pointer (actual function definition)

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+    - later on, when we cann self.add2numbers_py, here:
```
gasnica: What is "cann" here meaning?

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+
+            a, add_py, add_func = ti.static(self.a, self.add2numbers_py, self.add2numbers_func)
+
+      ... we want to call the bound method, `self.add2numbers_py`.
+    - an actual function reference, created by doing somevar = MyClass.add2numbers, can automatically
+      binds to self under, when called from self in this way (remember that add2numbers_py is actually
```
gasnica: what does 'under' mean here?

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+    When @ti.func runs, the function `add2numbers_func` exists, but there is not yet any `self`
+    - it is not possible for the method to be bound, to a `self` instance
+    - however, it is at this point that we create the wrapped function, via the @ti.func annotation,
+      which runs the `func` function
+    - later on, when we cann self.add2numbers_py, here:
+
+            a, add_py, add_func = ti.static(self.a, self.add2numbers_py, self.add2numbers_func)
+
+      ... we want to call the bound method, `self.add2numbers_py`.
+    - an actual function reference, created by doing somevar = MyClass.add2numbers, can automatically
```
gasnica: I'd emphasize with "an actual `pure/python' function reference"

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+      binds to self under, when called from self in this way (remember that add2numbers_py is actually
+      the wrapped function, returned by the `func` function, run by `@ti.func`)
```
gasnica: I'm not getting the message of: 
(remember that add2numbers_py is actually the wrapped function, returned by the `func` function, run by `@ti.func`)
is this clearer?
(however, add2numbers_py is actually a class Func instance, wrapping 'pure/python' function reference -- now also all wrapped by a TaichiCallable instance -- returned by the `kernel_impl.py::func` function, run by `@ti.func`)

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+    - however, in order to be able to add strongly typed attributes to the wrapped function, we need
+      to wrap the wrapped function in a class
```
gasnica: clarify (?):

- however, in order to be able to add strongly typed attributes to [idk: the wrapped 'pure/python' function reference OR the class Func instance], we need to [idk:], again, wrap the class Func instance in an instance of TaichiCallable

[idk] -- i don't know, correct me if i'm wrong

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+    - the wrapped function, wrapped in a class, will NOT automatically bind
```
gasnica: - the wrapping TaichiCallable instance (which is also callable, and executes the underlying double-wrapped 'pure/python' function reference) will NOT automatically bind

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+    - what happens then, is that later when we call the wrapped function, which is unbound, `self`
```
gasnica: "when we invoke TaichiCallable, which is unbound .."

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+      is not automatically passed in, as an argument, and things break
+
+    To address this we need to use the `__get__` method, in our function wrapper, ie TaichiCallable,
+    and have the `__get__` method return this `BoundFunc` object. The `__get__` method handles
+    running the binding for use, and effectively binds `BoundFunc` object to `self` object, by passing
+    in the instance, as an argument into `__init__`.
```
gasnica: Suggstions:

- [assuming this comment will be under TaichiCallable] returns the 'BoundTaichiCallable' object.
- maybe better phrasing for "The `TaichiCallable.__get__` method handles running the binding for use"
- by passing in the 'self' instance, as an argument into 'BoundTaichiCallable.__init__'

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+
+    `BoundFunc` can then be used as a normal bound func - even though it's just an object instance -
+    using its `__call__` method.
+    """
+
+    def __init__(self, fn: Callable, instance: Any, taichi_callable: "TaichiCallable"):
+        self.fn = fn
```
gasnica: can we remove the first param, and just initialize:
self.fn = taichi_callable.fn ?

gasnica: not sure if __setattr__ below needs to be adjusted then?

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+
+
+class TaichiCallable:
+    """
+    See docstring for `BoundFunc` for what this is being used for.
+    """
+
+    def __init__(self, fn: Callable, wrapper: Callable) -> None:
+        # self.func: Func | None = None
+        self.fn = fn
```
gasnica: Same suggestion as with BoundFunc:
- can we remove the first param, and instead initalize:
  self.fn = wrapper.func # given wrapper is of class Func type

python/taichi/lang/kernel_impl.py
```diff
@@ -64,8 +69,117 @@
...
+      is not automatically passed in, as an argument, and things break
+
+    To address this we need to use the `__get__` method, in our function wrapper, ie TaichiCallable,
+    and have the `__get__` method return this `BoundFunc` object. The `__get__` method handles
+    running the binding for use, and effectively binds `BoundFunc` object to `self` object, by passing
+    in the instance, as an argument into `__init__`.
+
+    `BoundFunc` can then be used as a normal bound func - even though it's just an object instance -
+    using its `__call__` method.
+    """
```
gasnica, line 146: Add: effectively, at the time of actually invoking the underlying 'pure/python' function, we have 3 layers of wrapper instances:
 BoundTaichiCallabe -> TaichiCallable -> Func -> 'pure/python' function reference/definition

tests/python/test_ptr_assign.py
```diff
@@ -71,37 +71,48 @@ def func(x: ti.f32, y: ti.f32):
...
         def __init__(self):
-            self.a = ti.field(dtype=ti.f32, shape=())
+            self.a = ti.field(dtype=ti.f32, shape=(3))
+
+        def add2numbers_py(self, x, y):
+            return x + y
 
         @ti.func
-        def add2numbers(self, x, y):
+        def add2numbers_func(self, x, y):
```
gasnica, line 106: add comment: here @ti.func creates the wrapping TaichiCallable instance, wrapping also a newly created class Func instance, referencing the `add2numbers_func` 'pure/python' function reference/definition

tests/python/test_ptr_assign.py
```diff
@@ -71,37 +71,48 @@ def func(x: ti.f32, y: ti.f32):
...
             return x + y
 
         @ti.kernel
         def func(self):
-            a, add = ti.static(self.a, self.add2numbers)
-            a[None] = add(2, 3)
+            a, add_py, add_func = ti.static(self.a, self.add2numbers_py, self.add2numbers_func)
```
gasnica, line 111: comment, to be verified: this does not create any new wrapping instances?

Question -- actually why do we need the ti.static here?

gasnica: oh, what happens with add2numbers_py here -- it doesn't have the @ti.func decorator. Is it also wrapped actually somehow? does the python function need the ti.static?

tests/python/test_ptr_assign.py
```diff
@@ -71,37 +71,48 @@ def func(x: ti.f32, y: ti.f32):
...
+            a[0] = add_py(2, 3)
+            a[1] = add_func(3, 7)
```
gasnica, line 113: add comment: this is where BoundTaichiCallable is created, and immediately invoked

tests/python/test_ptr_assign.py
```diff
@@ -71,37 +71,48 @@ def func(x: ti.f32, y: ti.f32):
...
-        def add2numbers(self, x, y):
+        def add2numbers_func(self, x, y):
             return x + y
 
         @ti.kernel
         def func(self):
-            a, add = ti.static(self.a, self.add2numbers)
-            a[None] = add(2, 3)
+            a, add_py, add_func = ti.static(self.a, self.add2numbers_py, self.add2numbers_func)
+            a[0] = add_py(2, 3)
```
gasnica, line 112: question -- i actually don't know what happens here: do we have the wrappers here?

python/taichi/lang/kernel_impl.py
hughperkins: Thanks! Note: personal preference: separate comment per "thing". This way I can tick them off as I go :) (Feel free to ignore this, but just in case you are specifically attempting to group things in a single comment, because you think I prefer :) )

python/taichi/lang/kernel_impl.py
hughperkins: `call` :)

python/taichi/lang/kernel_impl.py
hughperkins: removed

python/taichi/lang/kernel_impl.py
hughperkins: note: a "pure function" is a technical term, which we are going to start using. It means a function with no side effects.

python/taichi/lang/kernel_impl.py
hughperkins: Don't think that would be correct. See:
```
        @functools.wraps(_func)
        def wrapped_func(*args, **kwargs):
            try:
                return primal(*args, **kwargs)
            except (TaichiCompilationError, TaichiRuntimeError) as e:
                if impl.get_runtime().print_full_traceback:
                    raise e
                raise type(e)("\n" + str(e)) from None

        wrapped = TaichiCallable(_func, wrapped_func)
        wrapped.grad = adjoint
```

tests/python/test_ptr_assign.py
hughperkins: That's an implementation detail, and I'm not sure we should be making these 'end to end' integration tests have implementation details in the comments?

tests/python/test_ptr_assign.py
hughperkins: yes, ti.static means the function is executed at compile time, as a normal python function.

tests/python/test_ptr_assign.py
hughperkins: implementation detail. Should not be part of these integration tests I feel?

tests/python/test_ptr_assign.py
hughperkins: we do yes. super complicated at this point... Took me hours to get this test working... over 2-3 days.

python/taichi/lang/_ndarray.py
duburcqa: In that case, `TensorNdarray` or just `Ndarray`, but you get my point I guess.

python/taichi/lang/kernel_impl.py
duburcqa: I see. This pattern is weird but I can understand how you got there.

python/taichi/lang/kernel_impl.py
duburcqa: "directly" = right away (as opposed to postponing to a later time)

python/taichi/lang/kernel_impl.py
duburcqa: This is weird. Can you share the traceback?

hughperkins: Screenshot 2025-07-24 at 9 43 30 AM

python/taichi/lang/kernel_impl.py
hughperkins: what typing do you feel is missing?

python/taichi/lang/kernel_impl.py
gasnica: bumping this. Also -- if possible, let's remove the fn param, and init self.fn = wrapper.func

python/taichi/lang/_ndarray.py
```diff
@@ -27,7 +34,7 @@ def __init__(self):
         self.dtype = None
         self.arr = None
         self.layout = Layout.AOS
-        self.grad = None
+        self.grad: "TensorNdarray | None" = None
```
duburcqa, line 37: Should be `TensorNdarray | None`

python/taichi/lang/_ndarray.py
```diff
@@ -187,7 +194,7 @@ def copy_from(self, other):
         ndarray_to_ndarray(self, other)
         impl.get_runtime().sync()
 
-    def _set_grad(self, grad):
+    def _set_grad(self, grad: "ScalarNdarray | VectorNdarray | MatrixNdarray"):
```
duburcqa: `: TensorNdarray`

python/taichi/lang/_ndarray.py
hughperkins: it needs the quotes, because otherwise there is a circula dependency

stub_replacements_global.yaml
duburcqa: There must be something wrong with your stub. Is `Sequence` imported in the first place ? Because of course replacing `list` by `Sequence` will not work otherwise.

duburcqa: There must be something wrong with your stub. Is `Sequence` imported in the first place ? Because of course replacing `list` by `Sequence` will not work otherwise.

hughperkins: Yes, I imagine that's the reason. How to add imports into the stubs?

python/taichi/lang/kernel_impl.py
duburcqa: > what is the type of @dataclasses.dataclass? Similar question for ti.data_oriented.

I would just ignore this concern for now. For dataclass it is painful to do as it requires implementing some protocol. See https://stackoverflow.com/questions/54668000/type-hint-for-an-instance-of-a-non-specific-dataclass. Still, I would recommend to use a type that is more specific, and at the very least that supports hashing since this property is used at some places.

duburcqa: Since `ti.data_oriented` wraps a class and returns a class, and dataclass is also a class, `typing.Type` should do just fine in this context apparently.

hughperkins: you mean, put just `Type`, and that will handle dataclass too?

hughperkins: hashable | type works. but just Type gives:

hughperkins: Screenshot 2025-07-28 at 3 09 22 PM

hughperkins: Screenshot 2025-07-28 at 3 09 08 PM

python/taichi/lang/kernel_impl.py
duburcqa: This is a legitimate concern. There is something strange here. `needed_arg_type` is an instance so you cannot use `instance(v, needed_arg_type)`. It should rather be `isinstance(v, type(needed_arg_type))`.

hughperkins: But I dont think needed_arg_type is an instance? I think it is a type? 🤔

python/taichi/lang/kernel_impl.py
duburcqa: OK you are technically right, it is not necessary as long as the type is properly extrapolated from the default value. Actually I like to be pedantic and specify the desired type for class and instance attributes, since extrapolation is often not right and it enhances readability. But it is up to you since it is not part of any standard AFAIK.

python/taichi/lang/_ndarray.py
duburcqa: I see. You are right.

python/taichi/lang/kernel_impl.py
hughperkins: added

python/taichi/lang/kernel_impl.py
duburcqa: This error is legitimate actually. It seems to are missing some assert to make sure that `field.type` is not a string, or you should rather add `str | type` instead of `hashable | type` (which is less specific).

python/taichi/lang/kernel_impl.py
duburcqa: It is an instance, you checked that right above (3 lines before)

stub_replacements_global.yaml
duburcqa: Good question, but I think it would be nice to find out. Do you have a stub file to share ?

python/taichi/lang/kernel_impl.py
hughperkins: Following discussions offline, added comments around StructType implementing instancecheck, and StructType not making instancecheck a classmethod.

PR #68, opened 2025-07-06
[Test] Make test_print no longer break if stdout output

hughperkins:
Issue: #

### Brief Summary

Make test_print no longer break if stdout output

Previously, if I add debug code to print to stdout, this broke the unit tests, when running locally. This fixes this (in conjunection with https://github.com/genesis-company/taichi/pull/62 )

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_print.py
```diff
@@ -10,6 +10,14 @@
 cuda_on_windows = (ti.cuda, "Windows")
 
 
+def grep(target: str, pattern: str) -> str:
```
duburcqa: I'm not convinced by the name of this method and the name of the input arguments. Based on what it does, it should rather be `def filter_lines(target: str, prefix: str) -> str:`

You should also add a one-line docstring, that would be helpful.

hughperkins: it's not actuallly a prefix, it's a pattern. Not sure if this changes your thinking on the naming?

hughperkins: (at least, it has to match anywhere in the string, not necessarily at the start. Note that grep doesnt do regular expression matches by default. You need `egrep` for regular expressions I think?)

hughperkins: I suppose, yes, it doesn't support eg `foo*bar`. Will update

hughperkins: addressed in [f2cc42e](https://github.com/genesis-company/taichi/pull/68/commits/f2cc42e923a2a7ae5d3ca672522258270d9526d3)

PR #69, opened 2025-07-07
[Type] Typing 5: Rename imported cpp classes to end in Cxx

hughperkins:
Issue: #

### Brief Summary

Rename imported cpp classes to end in Cxx, when there are python classes otherwise with identical name.

This PR is based on earlier PRs (typing-4 etc)

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/taichi/lang/_ndarray.py
+++ b/python/taichi/lang/_ndarray.py
@@ -254,13 +254,10 @@ class ScalarNdarray(Ndarray):
         self.element_type = dtype
 
     def __del__(self):
-        if (
-            impl is not None
-            and impl.get_runtime is not None
-            and impl.get_runtime() is not None
-            and impl.get_runtime().prog is not None
-        ):
-            impl.get_runtime().prog.delete_ndarray(self.arr)
+        if impl is not None and impl.get_runtime is not None and impl.get_runtime() is not None:
+            prog = impl.get_runtime()._prog
+            if prog is not None:
+                prog.delete_ndarray(self.arr)
 
     @property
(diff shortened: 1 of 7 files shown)


python/taichi/lang/_ndarray.py
```diff
@@ -254,13 +254,12 @@ def __init__(self, dtype, arr_shape):
...
 
     def __del__(self):
-        if (
-            impl is not None
-            and impl.get_runtime is not None
-            and impl.get_runtime() is not None
-            and impl.get_runtime().prog is not None
-        ):
-            impl.get_runtime().prog.delete_ndarray(self.arr)
+        if impl is not None:
```
hughperkins: In case you're wondering, I was gettign. tons of spammy messages without checking these things carefully, e.g. on https://github.com/Genesis-Embodied-AI/gstaichi/pull/114 . basically, when the process shuts down, impl is basically empty, even though it exists, and impl.get_runtime is None. But _prog can also exist, but be None...

hughperkins: (Like, I've seen _prog exist and be none)
Oh whoops, I'm accidnetlaly checking prog intea of _prog. let me fix that)

hughperkins: oh wait, no its correct

hughperkins: (I suppose I could simply revert this bugfix to prog for now, and leave that to a different pr, eg 114? No strong opoiion either way)

PR #70, opened 2025-07-07
[Build] Run on merge to main

hughperkins:
Issue: #

### Brief Summary

run on merge to main, so that docs are published to `main` folder at that point.

copilot:summary

### Walkthrough

copilot:walkthrough


.github/workflows/manylinux_wheel.yml
```diff
@@ -107,7 +110,7 @@ jobs:
     uses: ./.github/workflows/api_doc.yml
     needs: upload_to_storage
     with:
-      head_ref: ${{ github.head_ref }}
+      head_ref: ${{ github.event_name == 'push' && github.ref_name == 'main' && 'main' || github.head_ref }}
```
duburcqa: Why not just add `if: github.event_name == 'push'` instead of changing `head_ref` ?

hughperkins: what do you mean by "changing" head_ref?

duburcqa: You replaced `head_ref: ${{ github.head_ref }}` by `head_ref: ${{ github.event_name == 'push' && github.ref_name == 'main' && 'main' || github.head_ref }}`.

duburcqa: But I guess you should still do: `head_ref: ${{ github.ref_name }}` to get the branch name instead of the commit hash (in addition to adding `if: github.event_name == 'push'` condition)

hughperkins: Yeah. Also, I'd like to make it clear that this is *only* for pushes to main.

duburcqa: This is already clear from the rules that triggers the workflow I think.

hughperkins: That requires looking in a different part of the yaml file. My brain is too small to be able to contain the entire yaml file at once; and I would prefer to make such things painfully obvious to my brain. Do you see any downsides to making it obvious that it only doesnt use head_ref when it is a push to `main`?

hughperkins: (well... I suppose one variation on the rule is:
"use head ref, unless it doesnt exist, in which case, use ref_name"?
Perhasp this is what you are saying?

duburcqa: > Do you see any downsides to making it obvious that it only doesnt use head_ref when it is a push to main?

Yes, it gives the feeling that this case may be possible in the first place, which was confusing for me.

duburcqa: I was suggesting to do this:
```yaml
  api_docs:
    if: github.event_name == 'push'             # <------ FIRST CHANGE
    uses: ./.github/workflows/api_doc.yml
    needs: upload_to_storage
    with:
      head_ref: ${{ github.ref_name }}.         # <------ SECOND CHANGE
    secrets:
      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```

hughperkins: The ref_name should be the primary I feel. We can do simply:

```
head_ref: ${{ github.head_ref || github.ref_name }}
```
I think?

hughperkins: Possibly new informaion: I'm pushing to a directory named after the branch. This lets me test the docs are publishing correctly in PRs. So it should be running:
- on PRs, for a folder with the same name as the PR branch
- on merge to main, for the folder `main`

hughperkins: updated in c1ba2e542

duburcqa: Ok I see. This makes sense.

PR #73, opened 2025-07-09
[Type] Add ndarray struct

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_ndarray_struct.py
```diff
@@ -0,0 +1,139 @@
...
+    e = ti.ndarray(ti.i32, shape=(227,))
+
+    @dataclass
+    class MyStruct:
+        a: ti.types.NDArray[ti.i32, 1]
+        b: ti.types.NDArray[ti.i32, 1]
+        c: ti.types.NDArray[ti.i32, 1]
+
+    @ti.func
+    def s2(z3: ti.types.NDArray[ti.i32, 1], my_struct3: MyStruct, bar3: ti.types.NDArray[ti.i32, 1]) -> None:
```
YilingQiao: Does it work only for level 2, or for any level recursively?

hughperkins: I've only tested what you see, but I *expect* it to work indefinitely.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -555,6 +556,31 @@ def canonicalize_formatted_string(raw_string: str, *raw_args: list, **raw_keywor
         args.insert(0, re.sub(r"{.*?}", "{}", raw_string))
         return args
 
+    @staticmethod
+    def expand_node_args_dataclasses(args: tuple[ast.AST]) -> tuple[ast.AST]:
```
duburcqa: Without signature: `def expand_node_args_dataclasses(args: tuple[ast.AST, ...]) -> tuple[ast.AST, ...]:`

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -555,6 +556,31 @@ def canonicalize_formatted_string(raw_string: str, *raw_args: list, **raw_keywor
...
+        args_new = []
+        for i, arg in enumerate(args):
```
duburcqa: `for arg in args:`

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -555,6 +556,31 @@ def canonicalize_formatted_string(raw_string: str, *raw_args: list, **raw_keywor
...
+            val = arg.ptr
+            if dataclasses.is_dataclass(val):
+                dataclass_type = val
+                for field_idx, field in enumerate(dataclasses.fields(dataclass_type)):
```
duburcqa: `for field in dataclasses.fields(dataclass_type):`

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -555,6 +556,31 @@ def canonicalize_formatted_string(raw_string: str, *raw_args: list, **raw_keywor
...
+                    field_name = field.name
+                    field_type = field.type
```
duburcqa: Unused variable.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -555,6 +556,31 @@ def canonicalize_formatted_string(raw_string: str, *raw_args: list, **raw_keywor
...
 
+    @staticmethod
+    def expand_node_args_dataclasses(args: tuple[ast.AST]) -> tuple[ast.AST]:
+        args_new = []
+        for i, arg in enumerate(args):
+            val = arg.ptr
+            if dataclasses.is_dataclass(val):
+                dataclass_type = val
+                for field_idx, field in enumerate(dataclasses.fields(dataclass_type)):
+                    field_name = field.name
```
duburcqa: Unless variable. Just call `field.name` in `child_name` directly

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -565,6 +591,8 @@ def build_Call(ctx: ASTTransformerContext, node: ast.Call):
         else:
             build_stmt(ctx, node.func)
             build_stmts(ctx, node.args)
+            node.args = ASTTransformer.expand_node_args_dataclasses(node.args)
+            build_stmts(ctx, node.args)
```
duburcqa, line 591: Is it relevant to overwrite `node.args` instead of doing `args = ASTTransformer.expand_node_args_dataclasses(node.args)`

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -724,6 +752,22 @@ def transform_as_kernel():
                     for item in items_to_put_in_dict:
                         invoke_later_dict[item[0]] = argpack, item[1], *item[2]
                     create_variable_later[arg.arg] = argpack
+                elif dataclasses.is_dataclass(ctx.func.arguments[i].annotation):
+                    dataclass_type = ctx.func.arguments[i].annotation
+                    arg_features = ctx.arg_features[i]
+                    ctx.create_variable(ctx.func.arguments[i].name, dataclass_type)
+                    for field_idx, field in enumerate(dataclasses.fields(dataclass_type)):
+                        field_name = field.name
```
duburcqa: Useless variable. Just call `field.name` in `new_field_name` directly.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -724,6 +752,22 @@ def transform_as_kernel():
...
+                        new_field_name = f"__ti_{ctx.func.arguments[i].name}_{field_name}"
```
duburcqa: `new_field_name` is not a great variable name.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -724,6 +752,22 @@ def transform_as_kernel():
...
+                        result, obj = decl_and_create_variable(
+                            field.type,
+                            new_field_name,
+                            arg_features[field_idx],
+                            invoke_later_dict,
+                            "",
+                            0,
+                        )
+                        ctx.create_variable(new_field_name, obj if result else obj[0](*obj[1]))
```
duburcqa: This is not clear enough.
```python
if result:
    ctx.create_variable(new_field_name, obj)
else:
    decl_type_func, type_args = obj
    type = decl_type_func(*type_args)
    ctx.create_variable(new_field_name, type)
```

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -724,6 +752,22 @@ def transform_as_kernel():
                     for item in items_to_put_in_dict:
                         invoke_later_dict[item[0]] = argpack, item[1], *item[2]
                     create_variable_later[arg.arg] = argpack
+                elif dataclasses.is_dataclass(ctx.func.arguments[i].annotation):
```
duburcqa: You should probably define `arg = ctx.func.arguments[i]` on top of the for-loop at this point.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -751,15 +795,47 @@ def transform_as_kernel():
...
                 transform_as_kernel()
             else:
-                assert len(args.args) == len(ctx.argument_data)
-                for i, (arg, data) in enumerate(zip(args.args, ctx.argument_data)):
-                    # Template arguments are passed by reference.
-                    if isinstance(ctx.func.arguments[i].annotation, annotations.template):
-                        ctx.create_variable(ctx.func.arguments[i].name, data)
+                args_offset = 0
+                for data_i, data in enumerate(ctx.argument_data):
+                    annotation = ctx.func.arguments[data_i].annotation
```
duburcqa: define a variable `arg = ctx.func.arguments[data_i]`

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -751,15 +795,47 @@ def transform_as_kernel():
...
+                    if isinstance(ctx.func.arguments[data_i].annotation, annotations.template):
+                        ctx.create_variable(ctx.func.arguments[data_i].name, data)
+                        continue
+
+                    elif dataclasses.is_dataclass(ctx.func.arguments[data_i].annotation):
+                        dataclass_type = ctx.func.arguments[data_i].annotation
+                        for field_idx, field in enumerate(dataclasses.fields(dataclass_type)):
+                            field_name = field.name
```
duburcqa: I would not define these 2 temporaries. They have no added value.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -751,15 +795,47 @@ def transform_as_kernel():
...
+                            field_type = field.type
+                            new_field_name = f"__ti_{ctx.func.arguments[data_i].name}_{field_name}"
```
duburcqa: I don't find the name `new_field_name` very informative.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -751,15 +795,47 @@ def transform_as_kernel():
...
+                            data_child = getattr(data, field_name)
+                            if not isinstance(
+                                data_child,
+                                (
+                                    _ndarray.ScalarNdarray,
+                                    matrix.VectorNdarray,
+                                    matrix.MatrixNdarray,
+                                    any_array.AnyArray,
```
duburcqa, line 819: I would rather keep them on a single line.

python/taichi/lang/kernel_impl.py
```diff
@@ -166,32 +168,51 @@ def _get_tree_and_ctx(
     )
 
 
-def _process_args(self, args, kwargs):
-    ret = [argument.default for argument in self.arguments]
+def expand_func_arguments(arguments: list[KernelArgument]) -> list[KernelArgument]:
+    new_arguments = []
+    for i, argument in enumerate(arguments):
```
duburcqa: `for argument in arguments:`

python/taichi/lang/kernel_impl.py
```diff
@@ -166,32 +168,51 @@ def _get_tree_and_ctx(
...
+        if dataclasses.is_dataclass(argument.annotation):
+            for field in dataclasses.fields(argument.annotation):
+                field_name = field.name
```
duburcqa: These 2 temporaries are useless I think.

python/taichi/lang/kernel_impl.py
```diff
@@ -166,32 +168,51 @@ def _get_tree_and_ctx(
...
+                new_arguments.append(new_argument)
+        else:
+            new_arguments.append(argument)
+    return new_arguments
+
+
+def _process_args(self, is_func: bool, args, kwargs):
+    if is_func:
+        self.arguments = expand_func_arguments(self.arguments)
+    fused_args = [argument.default for argument in self.arguments]
```
duburcqa, line 195: What does `fused_args` is supposed to mean?

python/taichi/lang/kernel_impl.py
```diff
@@ -200,7 +221,27 @@ def _process_args(self, args, kwargs):
...
+def unpack_ndarray_struct(tree: ast.Module, struct_locals: set[str]) -> ast.Module:
+    class AttributeToNameTransformer(ast.NodeTransformer):
+        def visit_Attribute(self, node):
+            if isinstance(node.value, ast.Attribute):
+                return node
+            if not isinstance(node.value, ast.Name):
+                return node
+            base_id = node.value.id
+            attr_name = node.attr
+            new_id = "__ti_" + base_id + "_" + attr_name
```
duburcqa: what about f-string for this?

python/taichi/lang/kernel_impl.py
```diff
@@ -250,6 +292,19 @@ def __call__(self, *args, **kwargs):
...
         )
+
+        assert ctx.func is not None
+        sig = inspect.signature(ctx.func.func)
+        parameters = sig.parameters
+        struct_locals = set()
+        for param_name, parameter in parameters.items():
+            if dataclasses.is_dataclass(parameter.annotation):
+                for field in dataclasses.fields(parameter.annotation):
+                    field_name = field.name
```
duburcqa: same.

python/taichi/lang/kernel_impl.py
```diff
@@ -376,9 +431,12 @@ def extract_arguments(self):
                     pass
                 elif isinstance(annotation, primitive_types.RefType):
                     pass
+                elif isinstance(annotation, type) and hasattr(annotation, "__dataclass_fields__"):
```
duburcqa: What are you trying to achieve here ? Using `hasattr` is usually a bad practice

python/taichi/lang/kernel_impl.py
```diff
@@ -424,6 +482,17 @@ def extract_arg(arg, anno, arg_name):
                 TaichiCallableTemplateMapper.extract_arg(arg[name], dtype, arg_name)
                 for index, (name, dtype) in enumerate(anno.members.items())
             )
+        if dataclasses.is_dataclass(anno):
+            dataclass_type = anno
+            _res_l = []
+            for field in dataclasses.fields(dataclass_type):
+                field_name = field.name
```
duburcqa: same.

python/taichi/lang/kernel_impl.py
```diff
@@ -424,6 +482,17 @@ def extract_arg(arg, anno, arg_name):
                 TaichiCallableTemplateMapper.extract_arg(arg[name], dtype, arg_name)
                 for index, (name, dtype) in enumerate(anno.members.items())
             )
+        if dataclasses.is_dataclass(anno):
```
duburcqa: I would recommend using `annotation` instead of `anno` as variable name.

python/taichi/lang/kernel_impl.py
```diff
@@ -424,6 +482,17 @@ def extract_arg(arg, anno, arg_name):
...
+            dataclass_type = anno
```
duburcqa: Why not just replace the variable name `anno` by `dataclass_type` if it makes more sense? Because here you have just renaming a variable, which is weird.

python/taichi/lang/kernel_impl.py
```diff
@@ -576,6 +645,26 @@ def reset(self):
         self.runtime = impl.get_runtime()
         self.compiled_kernels = {}
 
+    def expand_dataclasses(self, params: dict[str, Any]) -> dict[str, Any]:
+        new_params = {}
+        arg_names = params.keys()
+        for i, arg_name in enumerate(arg_names):
```
duburcqa: `for arg_name in arg_names:`

python/taichi/lang/kernel_impl.py
```diff
@@ -576,6 +645,26 @@ def reset(self):
...
+            param = params[arg_name]
+            annotation = param.annotation
+            if isinstance(annotation, type) and hasattr(annotation, "__dataclass_fields__"):
```
duburcqa: same.

python/taichi/lang/kernel_impl.py
```diff
@@ -576,6 +645,26 @@ def reset(self):
         self.runtime = impl.get_runtime()
         self.compiled_kernels = {}
 
+    def expand_dataclasses(self, params: dict[str, Any]) -> dict[str, Any]:
```
duburcqa: This method is not called anywhere?

python/taichi/lang/kernel_impl.py
```diff
@@ -637,6 +726,8 @@ def extract_arguments(self):
                     pass
                 elif annotation == template:
                     pass
+                elif isinstance(annotation, type) and hasattr(annotation, "__dataclass_fields__"):
```
duburcqa: same.

python/taichi/lang/kernel_impl.py
```diff
@@ -713,6 +805,17 @@ def ast_to_dict(node):
...
                         json.dumps({"elapsed_txt": elapsed_txt, "elapsed_json": elapsed_json}, indent=2)
                     )
+                assert ctx.func is not None
+                sig = inspect.signature(ctx.func.func)
+                parameters = sig.parameters
+                struct_locals = set()
+                for param_name, parameter in parameters.items():
+                    if dataclasses.is_dataclass(parameter.annotation):
+                        for field in dataclasses.fields(parameter.annotation):
+                            field_name = field.name
```
duburcqa: same.

python/taichi/lang/kernel_impl.py
```diff
@@ -250,6 +292,19 @@ def __call__(self, *args, **kwargs):
             ast_builder=impl.get_runtime().current_kernel.ast_builder(),
             is_real_function=self.is_real_function,
         )
+
+        assert ctx.func is not None
```
duburcqa: This piece of code is duplicated wrt `python/taichi/lang/kernel_impl.py:808`

python/taichi/lang/kernel_impl.py
```diff
@@ -713,6 +805,17 @@ def ast_to_dict(node):
                     output_file.write_text(
                         json.dumps({"elapsed_txt": elapsed_txt, "elapsed_json": elapsed_json}, indent=2)
                     )
+                assert ctx.func is not None
```
duburcqa: duplicated.

python/taichi/lang/kernel_impl.py
```diff
@@ -953,6 +1056,16 @@ def recursive_set_args(needed, provided, v, indices):
                     return 0
                 set_arg_sparse_matrix_builder(indices, v)
                 return 1
+            if dataclasses.is_dataclass(needed):
```
duburcqa: `needed` and `provided` are not good variable names

python/taichi/lang/kernel_impl.py
```diff
@@ -953,6 +1056,16 @@ def recursive_set_args(needed, provided, v, indices):
...
+                assert provided == needed
+                dataclass_type = needed
```
duburcqa: Not a fan of just renaming variables.

python/taichi/lang/kernel_impl.py
```diff
@@ -953,6 +1056,16 @@ def recursive_set_args(needed, provided, v, indices):
...
+                for j, field in enumerate(dataclasses.fields(dataclass_type)):
+                    field_name = field.name
```
duburcqa: same

tests/python/test_ndarray_struct.py
```diff
@@ -0,0 +1,215 @@
...
+
+    assert d[24] == 89
+    assert a[46] == 32
+    assert b[48] == 33
+    assert c[42] == 34
+    assert e[112] == 125
+
+
+@test_utils.test()
+def test_ndarray_struct_l4():
```
duburcqa: This single test seems to be sufficient and cover all the other cases? If so, I would keep only this one.

tests/python/test_ndarray_struct.py
hughperkins: The reason being by the way, there are three cases AFAIK:
- calling a kernel function, from python
- kernel function calls an inline function
- inline function calls an inline function

python/taichi/lang/kernel_impl.py
hughperkins: those are already existing, not created in this PR (and I actually find them not too terrible tbh)

python/taichi/lang/kernel_impl.py
hughperkins: where?  (I searched around for `func` but didnt find where `assert ctx.func is not None` is duplicated?)

tests/python/test_ndarray_struct.py
hughperkins: fair :)

python/taichi/lang/ast/ast_transformer.py
hughperkins: Yeah, we want to overwrite teh old nodes with the expanded nodes.

python/taichi/lang/ast/ast_transformer.py
hughperkins: oh taht's much easier to read. Thanks!

python/taichi/lang/ast/ast_transformer.py
hughperkins: flat_name ? unpacked_property_name?

python/taichi/lang/ast/ast_transformer.py
hughperkins: wider than 80 columns, wider than my ide. admitteldy because of the crazy indentation level here. And the fact that we have coments like "# ti.func" suggests this should be broken up into sub funcs. (I actually sort of feel that each of the node-specific functions, like build_Call, build_FunctionDef, might be best off going in its own file perhaps 🤔)

python/taichi/lang/kernel_impl.py
hughperkins: combines both args and kwargs. open to a better name. The old name of 'args' I found confusing, because it didn't convey to me that it combines both args and kwargs.

python/taichi/lang/kernel_impl.py
hughperkins: I found this function is superfluous actually 😂

python/taichi/lang/kernel_impl.py
hughperkins: removed function

python/taichi/lang/kernel_impl.py
hughperkins: yeah, right :)

python/taichi/lang/kernel_impl.py
hughperkins: Not sure where this was, but I searched "field_name", and didnt find, so I assume its handled.

python/taichi/lang/kernel_impl.py
hughperkins: Yes, good spot. factorized.

python/taichi/lang/kernel_impl.py
hughperkins: factorized out now.

python/taichi/lang/kernel_impl.py
duburcqa: I do think it is a bad name because I don't even understand what these variables are supposed to store x)

I think they are bad because `needed` and `provided` means nothing on their own. I would much prefer `arg_needed`, `arg_provided` if it is about `arg` so that we know what is supposed to be provided or needed. Otherwise, it feels like it is a boolean flag `is_provided`, `is_needed`.

python/taichi/lang/ast/ast_transformer.py
duburcqa: Then why to do this here and not where `node.args` is first created?

python/taichi/lang/ast/ast_transformer.py
duburcqa: I like `flat_name` because it is what it is. The fact that it will be used as name for the unpacked property is nice but the rest of the code will speak for itself, so less critical.

python/taichi/lang/ast/ast_transformer.py
duburcqa: No problem.

python/taichi/lang/kernel_impl.py
duburcqa: What? This is really what this line does ?! Using args and kwargs ?

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -787,40 +859,40 @@ def transform_as_kernel():
...
+                    if id(argument.annotation) in primitive_types.type_ids:
+                        var_name = argument.name
+                        ctx.create_variable(var_name, impl.expr_init_func(ti_ops.cast(data, argument.annotation)))
                         continue
                     # Create a copy for non-template arguments,
                     # so that they are passed by value.
-                    ctx.create_variable(arg.arg, impl.expr_init_func(data))
+                    var_name = argument.name
+                    ctx.create_variable(var_name, impl.expr_init_func(data))
+                for v in ctx.func.orig_arguments:
```
duburcqa, line 891: what about `arg_orig` instead of `v` ?

python/taichi/lang/kernel_impl.py
```diff
@@ -992,12 +1090,13 @@ def recursive_set_args(needed, provided, v, indices):
             raise ValueError(f"Argument type mismatch. Expecting {needed}, got {type(v)}.")
 
         template_num = 0
+        skip = 0
```
duburcqa: `skipped_num` ?

python/taichi/lang/ast/ast_transformer.py
hughperkins: node.args is first created by running `ast.parse` (a built-in python function) on the python source code, where the argument is a dataclass

duburcqa: I see. I'm fine with the current design then. And you really need to call `build_stmts` both before and after post-processing to handle dataclasses?

python/taichi/lang/kernel_impl.py
hughperkins: I feel like the datatype conveys it is a number, even an integer. How about `skip_offset` or `skip_count`?

python/taichi/lang/ast/ast_transformer.py
hughperkins: for some reason I'm not seeing your reply here, though I see it in email. I guess I can add some comments. Basically:

```
      build_stmts(ctx, node.args)   # creates variable for the dataclass itself
                                                        # necessary for calling further child functions
      node.args = ASTTransformer.expand_node_args_dataclasses(node.args)        
      build_stmts(ctx, node.args)    # creates variables for dataclass members
```

python/taichi/lang/kernel_impl.py
duburcqa: > I feel like the datatype conveys it is a number, even an integer. 

I don't like this approach. For me the variable anemone itself must be sufficient to have a good guess of its contain without looking at its initial value.

> How about skip_offset or skip_count?

Both are fine!

hughperkins: > I don't like this approach. For me the variable anemone itself must be sufficient to have a good guess of its contain without looking at its initial value.

I kind of know what you mean, but note that there is some nuance here. In the world before python typing, I used to give all variables a suffix to indicate the type, like `_l` for lists` `_i` for integers, `_s` for string, etc. I know what you mean though.

python/taichi/lang/kernel_impl.py
hughperkins: kwargs are added here: 
```
    for key, value in kwargs.items():
        found = False
        for i, arg in enumerate(self.arguments):
            if key == arg.name:
                if i < len_args:
                    raise TaichiSyntaxError(f"Multiple values for argument '{key}'.")
                fused_args[i] = value
                found = True
                break
```

python/taichi/lang/kernel_impl.py
hughperkins: How about `needed_type` and `provided_type`? (trying to keep this concise, yet convey that it's a type, not a boolean)

python/taichi/lang/ast/ast_transformer.py
hughperkins: args is so ambiguous to be honest. If we are going to rename, would be good to say what kind fo arg... It's hard to tell from this non-typed branch (I have a much more strongly typed branch, which helps a lot), but I'm guessing these are the original python objects (I mean, everything is a python object, but,  I guess you know what I mean?).

So maybe arg_py or py_arg?

You might say that "orig" aligns with my saying "original python objects", but I also use "original" to mean "not yet unpacked from being a dataclass", so "orig" would be ambiguous too.

python/taichi/lang/kernel_impl.py
duburcqa: I'm still doing this from time to time, especially when the container matters: `sensor_data_map` for instance.

python/taichi/lang/ast/ast_transformer.py
duburcqa: > args is so ambiguous to be honest. If we are going to rename, would be good to say what kind fo arg... It's hard to tell from this non-typed branch

I know, right?

`arg_py` is soooo much better than `arg` or even ``arg_orig` indeed!

hughperkins: actualy, just realized, these are `KernelArgument`s
```
self.orig_arguments: list[KernelArgument] = []

class KernelArgument:
    def __init__(self, _annotation, _name, _default=inspect.Parameter.empty):
        self.annotation = _annotation
        self.name = _name
        self.default = _default
```
"KernelArgument" seems like not the best name either tbh 🤔 

Seems like this is metadata for an argument?

hughperkins: Probably we dont want to factorize everything all at the same time 😅   But maybe we use `arg_meta` for the name of the variable holding the "Kernel Argument"?

duburcqa: yes it is good enough!

python/taichi/lang/kernel_impl.py
```diff
@@ -914,90 +1005,100 @@ def set_arg_sparse_matrix_builder(indices, v):
...
 
         template_num = 0
+        skip_offset = 0
         for i, val in enumerate(args):
             needed_ = self.arguments[i].annotation
             if needed_ == template or isinstance(needed_, template):
                 template_num += 1
                 continue
-            recursive_set_args(needed_, type(val), val, (i - template_num,))
+            skip_offset += recursive_set_args(needed_, type(val), val, (skip_offset + i - template_num,)) - 1
```
gasnica: How does this work?

I understand we want to offset the index, for our expanded array_structs. 

But then some existing types above: Ndarraytype, Texture, sparse matrix builder, return 0. Which will decrement effectively the skip_offset. -- is that a problem?

gasnica: or are those cases somehow handled by the `if .. then continue` on the previous line

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -555,6 +556,29 @@ def canonicalize_formatted_string(raw_string: str, *raw_args: list, **raw_keywor
         args.insert(0, re.sub(r"{.*?}", "{}", raw_string))
         return args
 
+    @staticmethod
+    def expand_node_args_dataclasses(args: tuple[ast.AST, ...]) -> tuple[ast.AST, ...]:
+        args_new = []
+        for arg in args:
+            val = arg.ptr
+            if dataclasses.is_dataclass(val):
```
gasnica, line 559: Do we effectively expand all python dataclasses (but not ti.dataclasses)?

And that's not a problem, because we would not pass any non-array_struct python dataclasses to kernels otherwise?

python/taichi/lang/kernel_impl.py
hughperkins: the skip offset only increases. What do you mean by 'decrement effectively'?

gasnica: recursive_set_args(..) returns 0 in some cases. eg. Ln 1044, 1057, 1063. So:

skip_offset += recursive_set_args(...) - 1

could decrease the value by 1, no ?

python/taichi/lang/ast/ast_transformer.py
hughperkins: we do nothing with ti.dataclasses. They are completely separate.

python/taichi/lang/kernel_impl.py
hughperkins: I see what you are saying. Those `0` cases are all for argpack stuff (that I want to remove), and textures (which we dont use, and should probably also remove). The maths works for ndarrays, fields, dataclasses AFAIK, since tests pass, on a variety of sets of parameter types, taht I've tried.

What is your suggestion to how you would prefer to restructure the maths, bearing in mind that it works as-is, currently AFAIK ? (intuitively, the return type of recursive_set_args is: the number of parameters handled.

Another option would be to decrement i, or to not increment it, but that seems hard to do?

gasnica: Great, I'm good with the math as-is, would just add an assert to check at runtime that recursive_set_args() does not return 0.

gasnica: Since we're not 100% sure if asserts are being always processed -- should we throw an exception, when recursive_set_args() returns zero ---  just so that it immediately crashes, instead of potentially quietly passing causing (very?) hard to trace issues?

@duburcqa

duburcqa: For me there are 3 options there:
* Making sure that supported arguments cannot be passed by design 
* Having unit tests that provides strong guarantees that supported arguments are never passed in practice
* Add an assert to catch this kind of potential issues

All these options are fine for me.

hughperkins: Ok. And what are we asserting on? It is ok for recursive_set_args to return 0 I think 🤔 Or.. what situation are we trying to avoid?

hughperkins: I suppose another option could be just to have some index number ,rather than using the i from enumerate. And then it simply wouldn't inrement. Perhaps this could be more intuitive?

hughperkins: Updated to replace `skip_args` with `i_out`. Thoughts?

gasnica: This looks great. And preserves the original logic. Resolving

hughperkins: Great! 🙌

PR #74, opened 2025-07-09
[Build] Add gc before each unit test, to prevent ndarray issues

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/conftest.py
+++ b/tests/python/conftest.py
@@ -14,7 +14,7 @@ pytest_rerunfailures.works_with_current_xdist = lambda: True
 
 
 @pytest.fixture(autouse=True)
-def run_gc_before_test():
+def run_gc_after_test():
     """
     This is necessary to prevent random test failures when testing with ndarray.
 
@@ -44,8 +44,9 @@ def run_gc_before_test():
         - causing the new ndarray-cpp to be deleted
         - at this point => crash bug
 
-    By calling gc.collect before each test, we avoid this issue.
+    By calling gc.collect after each test, we avoid this issue.
     """
+    yield
     gc.collect()
     gc.collect()
 


tests/python/conftest.py
```diff
@@ -12,6 +13,43 @@
...
+      to where the ndarray-cpp used to be
+    - on mac os x, it regularly happens, as an artifact of how memory management works, that new
+      ndarray-cpps are allocated with the exact same address as the old one
+    - when garbage collection runs, __del__ is called on the old ndarray-py
+        - causing the new ndarray-cpp to be deleted
+        - at this point => crash bug
+
+    By calling gc.collect before each test, we avoid this issue.
+    """
+    gc.collect()
```
duburcqa, line 50: why are you doing it twice? What about doing it after and before systematically?
```python
gc.collect()
yield
gc.collect()
```

hughperkins: because you need to wipe out both generations, though googling now, it seems maybe once is enough 🤔 
Screenshot 2025-07-11 at 12 45 59 PM

last time I looked was for python 2.x. Maybe is diffrerent in 3.x 🤔

hughperkins: well deepseek says when using __del__ might need two passes:
Screenshot 2025-07-11 at 12 47 59 PM

hughperkins: Screenshot 2025-07-11 at 12 48 23 PM

hughperkins: Generally, I find the second gc.collect is super cheap, so I dont think there is any harm in calling it twice. It is the first on that is slow, because it has to do some work.

duburcqa: Yes, and speed is not a big deal. I just wanted to learn something new.

duburcqa: But it looks like it is rather for extra safety than based on some accurate understanding of how Python and garbage collection works in the end :p

duburcqa: So, in the end, what about doing garbage collection before and after running each test systematically? Or only after if you prefer doing garbage collection only once? I think it most natural the leave the environment in a clean state than making sure it is clean before starting.

PR #75, opened 2025-07-10
[Build] Reduce concurrency

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/macosx.yml
+++ b/.github/workflows/macosx.yml
@@ -4,6 +4,9 @@ on:
     branches:
       - main
   workflow_dispatch:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
 jobs:
   build_mac_14_15:
     strategy:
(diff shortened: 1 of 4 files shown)


.github/workflows/macosx.yml
```diff
@@ -16,6 +16,9 @@ jobs:
             PYTHON_CP_VERSION: cp310
     name: Mac OS Build
     runs-on: macos-${{ matrix.MAC_OS_VERSION }}
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}-${{matrix.MAC_OS_VERSION}}-${{ matrix.PYTHON_CP_VERSION }}-build
```
duburcqa: Add OS and CP in group name should not be necessary. See https://github.com/Genesis-Embodied-AI/Genesis/blob/d386d0dfd06d2991d58cc61b0f3ce6c4638c0750/.github/workflows/generic.yml#L9

hughperkins: oh yeah, perhaps not if moved into workflow global location. will be necessary for manylinux though, fairly sure.

hughperkins: updated

PR #76, opened 2025-07-11
[Type] Add fields to dataclasses.dataclass support

hughperkins:
Issue: #

### Brief Summary

Note: depends on dataclasses.dataclass previous PR

copilot:summary

### Walkthrough

copilot:walkthrough


python/taichi/lang/kernel_impl.py
```diff
@@ -166,32 +175,49 @@ def _get_tree_and_ctx(
...
+                    _annotation=field.type,
+                    _name=f"__ti_{argument.name}_{field.name}",
+                )
+                new_arguments.append(new_argument)
+        else:
+            new_arguments.append(argument)
+    return new_arguments
+
+
+def _process_args(self, is_func: bool, args, kwargs):
```
duburcqa: It would be great to change all these function names because they are not explicit enough I think. But this is not related to this PR.

python/taichi/lang/kernel_impl.py
```diff
@@ -200,7 +226,47 @@ def _process_args(self, args, kwargs):
                     f"Parameter `{self.arguments[i].name} : {self.arguments[i].annotation}` missing."
                 )
 
-    return ret
+    return tuple(fused_args)
+
+
+def unpack_ndarray_struct(tree: ast.Module, struct_locals: set[str]) -> ast.Module:
```
duburcqa: I like the name of these new functions. It is very clear.

python/taichi/lang/kernel_impl.py
```diff
@@ -200,7 +226,47 @@ def _process_args(self, args, kwargs):
...
+    class AttributeToNameTransformer(ast.NodeTransformer):
+        def visit_Attribute(self, node: ast.AST):
+            if isinstance(node.value, ast.Attribute):
+                return node
+            if not isinstance(node.value, ast.Name):
+                return node
+            base_id = node.value.id
+            attr_name = node.attr
+            new_id = f"__ti_{base_id}_{attr_name}"
```
duburcqa: Since you are using this pattern at many places, and I guess this is important, It may be better to define some helper that would take `base_id` and `attr_name` as input argument, that would could just call like this:
```python
new_id = get_node_id(base_id=node.value.id, attr_name=node.attr)
```

python/taichi/lang/kernel_impl.py
```diff
@@ -200,7 +226,47 @@ def _process_args(self, args, kwargs):
...
+
+    transformer = AttributeToNameTransformer()
+    new_tree = transformer.visit(tree)
+    ast.fix_missing_locations(new_tree)
+    return new_tree
+
+
+def extract_struct_locals_from_context(ctx: ASTTransformerContext):
+    """
+    Uses ctx.func.func to get the function signature
```
duburcqa: I would recommend skipping one line between short description and details. Plus making sure the whole documentation has proper punctuation and grammar (incl. starting each sentence with uppercase and final dot).

python/taichi/lang/kernel_impl.py
hughperkins: Yes, agree.

Doing a bunch of refactorization in my current branch, https://github.com/Genesis-Embodied-AI/taichi/compare/main...hp/ndarray-nested1-scaff (though I don't think I rename this specific function, for now).

python/taichi/lang/kernel_impl.py
hughperkins: we used to have a Slack emoji "not sure if". I think that's my thought here :)

I rename some of thse in https://github.com/Genesis-Embodied-AI/taichi/compare/main...hp/ndarray-nested1-scaff , though I don't remember if I renamed this specific one yet.

python/taichi/lang/kernel_impl.py
hughperkins: yes, good point. Can we revisit this after I do nested structs? Whilst this pattern essentially stays, conceptually, there is some nuance. (eg I have to do more like:

```
new_id = f"{base_id}_{attr_name}"
if not new_id.startswith("__ti_"):
    new_id = f"__ti_{new_id}"
```
(though you might figure out a less ugly way to do this, but, like I say, maybe we discuss this point in the nested structs PR?)

python/taichi/lang/kernel_impl.py
hughperkins: this isn't actually a short description. refromatted to be a bullet point, to make it clear that it's not a short description.

tests/python/test_ndarray_struct.py
```diff
@@ -122,6 +122,123 @@ def k1(z: ti.types.NDArray[ti.i32, 1], my_struct: MyStruct, bar: ti.types.NDArra
     assert d[20] == 14
 
 
+@test_utils.test()
+def test_ndarray_struct_field_l4():
+    gc.collect()
+    gc.collect()
+    a = ti.field(ti.i32, shape=(55,))
```
gasnica: we don't test for n-dimensional fields, but that's not a problem?

python/taichi/lang/kernel_impl.py
```diff
@@ -1057,11 +1065,12 @@ def recursive_set_args(needed_arg_type, provided_arg_type, v, indices):
                 return 1
             if dataclasses.is_dataclass(needed_arg_type):
                 assert provided_arg_type == needed_arg_type
+                idx = 0
                 for j, field in enumerate(dataclasses.fields(needed_arg_type)):
                     assert not isinstance(field.type, str)
                     field_value = getattr(v, field.name)
-                    recursive_set_args(field.type, field.type, field_value, (indices[0] + j,))
-                return len(dataclasses.fields(needed_arg_type))
+                    idx += recursive_set_args(field.type, field.type, field_value, (indices[0] + idx,))
```
gasnica, line 1072: what are indices?
also, is there a better name for v ? arg? arg_value?

python/taichi/lang/kernel_impl.py
```diff
@@ -158,6 +158,14 @@ def _get_tree_and_ctx(
         for i in self.template_slot_locations:
             template_var_name = self.arguments[i].name
             global_vars[template_var_name] = args[i]
+        parameters = inspect.signature(self.func).parameters
+        for arg_i, (name, param) in enumerate(parameters.items()):
```
gasnica: 1. so arg_i refers to _get_tree_and_ctx() input param 'args'. How is that related to params?

2. name -> param_name?

python/taichi/lang/kernel_impl.py
```diff
@@ -158,6 +158,14 @@ def _get_tree_and_ctx(
...
+            if dataclasses.is_dataclass(param.annotation):
+                for field in dataclasses.fields(param.annotation):
+                    field_name = field.name
+                    child_value = getattr(args[arg_i], field.name)
```
gasnica: if this is child_value, then maybe rename field -> child_field and field_name -> child_field_name

python/taichi/lang/kernel_impl.py
```diff
@@ -158,6 +158,14 @@ def _get_tree_and_ctx(
...
+                    flat_name = f"__ti_{name}_{field.name}"
+                    global_vars[flat_name] = child_value
```
gasnica, line 167: _get_global_vars() is a existing taichi function/concept?

oh, and we're just piggy-backing on it, adding our expanded param names to it?

python/taichi/lang/kernel_impl.py
```diff
@@ -1100,6 +1109,8 @@ def recursive_set_args(needed_arg_type, provided_arg_type, v, indices):
                     )
                 needed_arg_type.set_kernel_struct_args(v, launch_ctx, indices)
                 return 1
+            if isinstance(needed_arg_type, template):
+                return 0
```
gasnica, line 1113: 1. can we, please, add a docstring to this recursive_set_args() to explain when we return zero?

I have a vague intuition, that fields and some other types are not passed as params, but accessed as global vars? 

2. Why wasn't this line needed before?

python/taichi/lang/kernel_impl.py
```diff
@@ -158,6 +158,14 @@ def _get_tree_and_ctx(
         for i in self.template_slot_locations:
             template_var_name = self.arguments[i].name
             global_vars[template_var_name] = args[i]
+        parameters = inspect.signature(self.func).parameters
```
gasnica, line 161: general question: why is this processing of fields not located somewhere closer to where we process ndarrays in structs?

hughperkins: which bit are you referring to when you say 'process ndarrays in structs'?

tests/python/test_ndarray_struct.py
hughperkins: added multimieniaonl tests.

python/taichi/lang/kernel_impl.py
hughperkins: Honestly I'm not entirely clear. Empricially, in the case of ndarrays, it appears to be a single number (wrapped in a list), which must increase by exactly 1 for each argument sent to the cuda api for kernel launch.

This is one of these cases where I'm trying to miimize changes to existing code, adn just ensure there are sufficient tests to exercise my new changes.

hughperkins: For v, yeah, maybe py_obj? I'm not 100% sure if this PR is the best PR to change it in, but can be persuaded to do so. I'll probably get conflicts no matter where/when I cahnge it.

python/taichi/lang/kernel_impl.py
hughperkins: arg_i is from enumerating paraetmers which is from the function siganture. So, it's the index of a function parameter.

hughperkins: is there any benefit in making the name longer? I dont see any other `name`s around?

python/taichi/lang/kernel_impl.py
hughperkins: field_name is actually superfluous, I'm removing it.

hughperkins: `field -> child_field` => makes everything longer to read/write. I feel like one of the hallmarks of python is using relatively short names for things?

python/taichi/lang/kernel_impl.py
hughperkins: yeah. it's already (ab)used for templates. I'm using continuing to (ab)use it for structs containing tempaltes.

hughperkins: I can be persuaded to not store stuff in global_vars, but maybe that should be another pr down teh line? (andmaybe not super high pri for now?)

python/taichi/lang/kernel_impl.py
hughperkins: re: 1 => 👍

hughperkins: re: 2, because non-struct templates are handled in the bit of code at line 1127:
```
            if needed_ == template or isinstance(needed_, template):
                template_num += 1
                i_out += 1
                continue
```

hughperkins: You could argue, perhaps correctly, taht we might be able to remove this bit of extra code at line 1127 🤔

hughperkins: oh, except, would need me to figure out hwat `template_num` is for/doing.

gasnica: Cool. I'm happy with it as-is. Please, don't dig :-)

python/taichi/lang/kernel_impl.py
gasnica: Good question -- I was thinking about recursive_set_args() that we call from .. launch_kernel

So i guess that's a very different moment in time (?) ... So this code here, happens earlier during compilation (?) 

--
Possibly i also misunderstood --- I thought we're processing ti.fields here, but we're processing all member_fields. Can we rename field in the lines/loop below to either member_field or attribute? (and the related xx_name, child_xx_value variables too?)

python/taichi/lang/kernel_impl.py
gasnica: > arg_i is from enumerating parameters which is from the function signature. So, it's the index of a function parameter.

Hmm, but then inline 166: child_value = getattr(args[arg_i], field.name)

We use that index, to reference ordered () args passed into _get_tree_and_ctx() as its args param? .... How is it enforced that the enumerated function-signature params match the set and order of values in _get_tree_and_ctx()'s 'arg' param ?

gasnica: > is there any benefit in making the name longer? I dont see any other names around?

I'm very happy making commonly-used member-field names short. But for local vars which are typed twice, or three times, -- imho --the name should be self explanatory without the need to look up the var's creation ..

so e.g. I needed to use time to decode line 167: "flat_name = f"__ti_{name}_{field.name}" -- cos we have 2 'qualified/semi-qualified' names, and one name-name

python/taichi/lang/kernel_impl.py
hughperkins: we are looping over dataclasses in various dsitributed places in the code. It is, I agree, rather thard to keep track of. At some point in the future, I'd like to think about how we could refactor this, but I I don't know the code well enough at this point to do that.

That said, I do agree that we should probaly run inspect.signature in only one place, and fully expected to be called out on that. So feel free to call me out on that...

python/taichi/lang/kernel_impl.py
hughperkins: but would param_name make that any clearer? 🤔

```
"flat_name = f"_ti{param_name}_{field.name}"
```

I feel that what you want is to rename field to child_field perhaps?

```
"flat_name = f"_ti{name}_{child_field.name}"
```

hughperkins: (or I suppose we could do both 🤔 )

gasnica: yeah, maybe -- since we're talking about it?

PR #91, opened 2025-07-18
[Type] Add nested py dataclasses and enable .shape for members

hughperkins:
Issue: #

### Brief Summary

- Add nested py dataclasses
- enable .shape for members
- enable kwargs

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/ast/ast_transformer.py
+++ b/python/gstaichi/lang/ast/ast_transformer.py
@@ -1312,10 +1312,8 @@ build_stmt = ASTTransformer()
 
 
 def build_stmts(ctx: ASTTransformerContext, stmts: list[ast.stmt]):
-    """
-    TODO: Should we just make this part of ASTTransformer? Then, easier to pass around (just
-    pass the ASTTransformer object around)
-    """
+    # TODO: Should we just make this part of ASTTransformer? Then, easier to pass around (just
+    # pass the ASTTransformer object around)
     with ctx.variable_scope_guard():
         for stmt in stmts:
             if ctx.returned != ReturnStatus.NoReturn or ctx.loop_status() != LoopStatus.Normal:


python/taichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,218 @@
...
+
+def expand_func_arguments(arguments: list[ArgMetadata]) -> list[ArgMetadata]:
+    """
+    Used to expand arguments for @ti.func
+    """
+    new_arguments = []
+    for i, argument in enumerate(arguments):
+        if dataclasses.is_dataclass(argument.annotation):
+            for field in dataclasses.fields(argument.annotation):
+                child_name = f"{argument.name}__ti_{field.name}"
```
gasnica: in what case does argument.name start with "__ti_" ??

python/taichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,218 @@
...
+    - __ti_struct_ab__ti_b
+    - __ti_struct_ab__ti_struct_cd__ti_c
+    - __ti_struct_ab__ti_struct_cd__ti_d
+    - __ti_struct_ab__ti_struct_cd__ti_struct_ef__ti_e
+    - __ti_struct_ab__ti_struct_cd__ti_struct_ef__ti_f
+
+
+    """
+    for field in dataclasses.fields(struct_type):
+        child_name = f"{basename}__ti_{field.name}"
```
gasnica: this name creation is done 10 times in this PR -- with or without the check for basename staring with '__ti_' 

do we want a helper function for it?

python/taichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -150,16 +200,21 @@ def _transform_kernel_arg(
...
+        assert compiling_callable is not None
+        compiling_callable.finalize_rets()
 
-        invoke_later_dict: dict[str, tuple[Any, str, Any]] = dict()
-        create_variable_later = dict()
+        invoke_later_dict: dict[str, tuple[Any, str, Callable, list[Any]]] = dict()
+        create_variable_later: dict[str, Any] = dict()
         for i, arg in enumerate(args.args):
-            argument = ctx.func.arguments[i]
+            argument = ctx.func.arg_metas[i]
```
gasnica: inconsistent local var name --> argument or arg_meta?

python/taichi/lang/kernel_impl.py
```diff
@@ -290,31 +308,20 @@ def _get_tree_and_ctx(
...
+    used for both Func and Kernel
+    """
+
     if is_func:
-        self.arguments = expand_func_arguments(self.arguments)
-    fused_args = [argument.default for argument in self.arguments]
-    ret: list[Any] = [argument.default for argument in self.arguments]
+        self.arg_metas = _kernel_impl_dataclass.expand_func_arguments(self.arg_metas)
+
+    fused_args: list[Any] = [argument.default for argument in self.arg_metas]
```
gasnica: again, we now have inconsistent naming after changing arg_metas ... not sure if we're happy with it

python/taichi/lang/kernel_impl.py
```diff
@@ -633,7 +622,7 @@ def extract_arguments(self) -> None:
             for return_type in self.return_type:
                 if return_type is Ellipsis:
                     raise TaichiSyntaxError("Ellipsis is not supported in return type annotations")
-        params = sig.parameters
+        params = dict(sig.parameters)
```
gasnica: if this is just for typing, can we do params: dict = sig.parameters instead ?

python/taichi/lang/kernel_impl.py
```diff
@@ -755,8 +744,8 @@ def ast_to_dict(node: ast.AST | list | primitive_types._python_primitive_types):
                     output_file.write_text(
                         json.dumps({"elapsed_txt": elapsed_txt, "elapsed_json": elapsed_json}, indent=2)
                     )
-                struct_locals = extract_struct_locals_from_context(ctx)
-                tree = unpack_ndarray_struct(tree, struct_locals=struct_locals)
+                struct_locals = _kernel_impl_dataclass.populate_struct_locals(ctx)
```
gasnica: like we spoke, I like the old name, but either works.

python/taichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,218 @@
...
+                    )
+                    new_arguments.append(new_argument)
+        else:
+            new_arguments.append(argument)
+    return new_arguments
+
+
+def _flatten_attribute_name(node: ast.Attribute) -> str | None:
+    """
+    see unpack_ndarray_struct docstring for more explanation
```
gasnica: unpack_ndarray_struct is gone now

python/taichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,218 @@
...
+                        _annotation=field.type,
+                        _name=child_name,
+                    )
+                    new_arguments.append(new_argument)
+        else:
+            new_arguments.append(argument)
+    return new_arguments
+
+
+def _flatten_attribute_name(node: ast.Attribute) -> str | None:
```
gasnica: this seems to be only called within AttributeToNameTransformer, do we want to move it inside it?

python/taichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -222,14 +287,13 @@ def _transform_func_arg(
                     any_array.AnyArray,
                 ),
             ):
-                raise TaichiSyntaxError(f"Argument {arg.arg} of type {argument_type} is not recognized.")
+                raise TaichiSyntaxError(f"Argument {argument_name} of type {argument_type} is not recognized.")
```
gasnica: does argument_name have the same meaning as arg_name elesewhere? do we want to be consistent with naming?

python/taichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,218 @@
...
+    if isinstance(node.value, ast.Name):
+        return f"__ti_{node.value.id}__ti_{node.attr}"
+    if isinstance(node.value, ast.Attribute):
+        child_flat_name = _flatten_attribute_name(node.value)
+        integrated_flat_name = f"{child_flat_name}__ti_{node.attr}"
+        return integrated_flat_name
+    return None
+
+
+def unpack_ast_struct_expressions(tree: ast.Module, struct_locals: set[str]) -> ast.Module:
```
gasnica: would we want to limit name variations, or aliases on functionality? 

if this just creates the Attribute To Name Transfomer, which essentially calls _flatten_attribute_name, 

then would we rather have:
- FlattenAttributeNameTransformer
- and this function should be a @methodclass of the above Transformer maybe?

That would maybe nicer encapsulate this piece of logic?

tests/python/test_ndarray_struct.py
```diff
@@ -30,6 +31,93 @@ def __getitem__(self, _):
     return TiTemplateBuilder()
 
 
+@test_utils.test()
+def test_ndarray_struct_l4_kwargs():
```
gasnica: what does 'l4' stand for ?

python/taichi/lang/_kernel_impl_dataclass.py
hughperkins: when we recurse, see line 106

python/taichi/lang/_kernel_impl_dataclass.py
hughperkins: that makes for a very long calling string..
Screenshot 2025-08-05 at 1 02 58 PM

Personally, I feel the length adn style of this line kind of looks very Java to me.

Python aims to have a rather terse syntax.

I would prefer to remove the class name, which has no reason to be visible to the person calling this I feel?

python/taichi/lang/ast/ast_transformers/function_def_transformer.py
hughperkins: where do you see `arg_name`?

python/taichi/lang/kernel_impl.py
hughperkins: no because sig.parametrs is not a dict. It's a `MappingProxyType`.

tests/python/test_ndarray_struct.py
hughperkins: level 4. ie 4 levels of nesting

python/taichi/lang/_kernel_impl_dataclass.py
hughperkins: created _dataclass_util.create_flat_name

python/taichi/lang/ast/ast_transformers/function_def_transformer.py
gasnica: I guess mainly in TemplateMapper.extract_arg(). So a bit further away .. but still .. 

Nothing to worry about though

python/taichi/lang/_kernel_impl_dataclass.py
gasnica: Then you can just rename 'unpack_ast_structure_express' to 'execute' or 'process', cos there's no need for repeating the action with different wording .. 

That was just a suggestion though.

tests/python/test_ndarray_struct.py
gasnica: cool, as you guess, I'd use a long descriptive name -- cos it serves as description, and is used nowhere else, I guess. 

just a suggestion though :-)

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+    - struct_ab.struct_cd.strucdt_ef.f
+
+    And the members of struct_locals should be:
+    - __ti_struct_ab__ti_a
+    - __ti_struct_ab__ti_b
+    - __ti_struct_ab__ti_struct_cd__ti_c
+    - __ti_struct_ab__ti_struct_cd__ti_d
+    - __ti_struct_ab__ti_struct_cd__ti_struct_ef__ti_e
+    - __ti_struct_ab__ti_struct_cd__ti_struct_ef__ti_f
+
```
duburcqa: Remove extra spaces at the end.

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+
+    """
+    for field in dataclasses.fields(struct_type):
+        child_name = f"{basename}__ti_{field.name}"
```
duburcqa: Why not use `create_flat_name` ? Because there is no `__ti_` prefix at the beginning? Maybe add some optional argument `no_prefix: bool = False` ?

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+          {"__ti_my_struct_a", "__ti_my_struct_b, "__ti_my_struct_c"}
+    """
+    struct_locals = set()
+    assert ctx.func is not None
+    sig = inspect.signature(ctx.func.func)
+    parameters = sig.parameters
+    for param_name, parameter in parameters.items():
+        if dataclasses.is_dataclass(parameter.annotation):
+            for field in dataclasses.fields(parameter.annotation):
+                child_name = f"__ti_{param_name}__ti_{field.name}"
```
duburcqa: same, it could use `create_flat_name`

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+                    child_args = expand_func_arguments(
+                        [
+                            ArgMetadata(
+                                field.type,
+                                child_name,
+                                argument.default,
+                            )
+                        ]
+                    )
+                    new_arguments.extend(child_args)
```
duburcqa: Do you like this formatting? I think it does not improve readability in this context. Spreading over too many lines. Moreover, could you use keyword arguments ? 

What about:
```python
                    new_argument = ArgMetadata(
                        _annotation=field.type,
                        _name=child_name,
                        _default=argument.default,
                    )
                    child_args = expand_func_arguments([new_argument])
                    new_arguments.extend(child_args)
```

duburcqa: By the way, why use `_` prefix for constructor arguments ? This is weird and it seems not well motivated.

duburcqa: Personally I prefer `+=` over `extend` but it is up to you!

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+                    continue
+                struct_locals.add(child_name)
+    return struct_locals
+
+
+def expand_func_arguments(arguments: list[ArgMetadata]) -> list[ArgMetadata]:
+    """
+    Used to expand arguments for @ti.func
+    """
+    new_arguments = []
```
duburcqa: Do we have a clearer name than `new_` ? Not clear what it means to me...

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+    return new_arguments
+
+
+class FlattenAttributeNameTransformer(ast.NodeTransformer):
+    def __init__(self, struct_locals: set[str]) -> None:
+        self.struct_locals = struct_locals
+
+    def visit_Attribute(self, node):
+        flat_name = FlattenAttributeNameTransformer._flatten_attribute_name(node)
+        if not flat_name or flat_name not in self.struct_locals:
```
duburcqa, line 124: `if flat_name not in self.struct_locals:` is not sufficient ? Can the empty name be in `self.struct_locals` ?

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+            return self.generic_visit(node)
+        return ast.copy_location(ast.Name(id=flat_name, ctx=node.ctx), node)
+
+    @staticmethod
+    def _flatten_attribute_name(node: ast.Attribute) -> str | None:
+        """
+        see unpack_ast_struct_expressions docstring for more explanation
+        """
+        if isinstance(node.value, ast.Name):
+            return f"__ti_{node.value.id}__ti_{node.attr}"
```
duburcqa: `create_flat_name` ?

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+        if isinstance(node.value, ast.Attribute):
+            child_flat_name = FlattenAttributeNameTransformer._flatten_attribute_name(node.value)
+            integrated_flat_name = f"{child_flat_name}__ti_{node.attr}"
```
duburcqa: `create_flat_name(..., no_prefix=True)` ?

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+            return integrated_flat_name
```
duburcqa: Is it really relevant to define a variable name right before returning ? What we are returning is already very clear, not sure if the variable name helps.

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+            populate_global_vars_from_dataclass(
+                param_name=flat_name,
+                param_type=field.type,
+                py_arg=child_value,
+                global_vars=global_vars,
+            )
+        elif isinstance(field.type, (Template,)) or field.type == Template:
+            global_vars[flat_name] = child_value
+        else:
+            pass
```
duburcqa: What is that? I don't get the point.

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,217 @@
...
+        child_value = getattr(py_arg, field.name)
+        flat_name = create_flat_name(param_name, field.name)
+        if dataclasses.is_dataclass(field.type):
+            populate_global_vars_from_dataclass(
+                param_name=flat_name,
+                param_type=field.type,
+                py_arg=child_value,
+                global_vars=global_vars,
+            )
+        elif isinstance(field.type, (Template,)) or field.type == Template:
```
duburcqa: What about some `is_ti_template` helper ? It would make the code cleaner even if it is a one-liner.

python/gstaichi/lang/_template_mapper.py
```diff
@@ -85,15 +89,18 @@ def extract_arg(arg, annotation: AnnotationType, arg_name: str) -> Any:
...
         if dataclasses.is_dataclass(annotation):
             _res_l = []
             for field in dataclasses.fields(annotation):
                 field_value = getattr(arg, field.name)
-                arg_name = f"__ti_{arg_name}_{field.name}"
-                field_extracted = GsTaichiCallableTemplateMapper.extract_arg(field_value, field.type, arg_name)
+                child_name = arg_name
+                if not child_name.startswith("__ti_"):
+                    child_name = f"__ti_{child_name}"
+                child_name = f"{child_name}__ti_{field.name}"
```
duburcqa: Sounds like a job for `create_flat_name` !

python/gstaichi/lang/_dataclass_util.py
```diff
@@ -0,0 +1,5 @@
+def create_flat_name(basename: str, child_name: str) -> str:
```
duburcqa, line 1: Could you add some docstring? Took me a while to understand why this method was written this way. The point is avoiding adding prefix `__ti_` if already including in `basename`, to avoid duplicated "delimiter". This was not clear to me. Otherwise, there are much simpler ways to write this method.

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -620,10 +621,10 @@ def build_Attribute(ctx: ASTTransformerContext, node: ast.Attribute):
                 Matrix._keygroup_to_checker[keygroup](node.value.ptr, node.attr)
                 attr_len = len(node.attr)
                 if attr_len == 1:
+                    compiling_callable = impl.get_runtime().compiling_callable
+                    assert compiling_callable is not None
```
duburcqa: Why `compiling_callable` can be None in the first place? You could add some property that would raise an exception if None, and always return well-defined compiling callable.

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -645,6 +646,10 @@ def build_Attribute(ctx: ASTTransformerContext, node: ast.Attribute):
 
                 node.ptr = getattr(tensor_ops, node.attr)
                 setattr(node, "caller", node.value.ptr)
+        elif dataclasses.is_dataclass(node.value.ptr):
+            type_by_name = {field.name: field.type for field in dataclasses.fields(node.value.ptr)}
+            child_type = type_by_name[node.attr]
+            node.ptr = child_type
```
duburcqa: This is a strange pattern. What about `node.ptr = next(field.type for field in dataclasses.fields(node.value.ptr) if field.name == node.attr)`

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -1309,6 +1314,10 @@ def build_Pass(ctx: ASTTransformerContext, node: ast.Pass) -> None:
 
 
 def build_stmts(ctx: ASTTransformerContext, stmts: list[ast.stmt]):
+    """
+    Should we just make this part of ASTTransformer? Then, easier to pass around (just
+    pass the ASTTransformer object around)
+    """
```
duburcqa: Should rather be a comment with prefix (e.g. TODO or FIXME) than a docstring.

duburcqa: if it makes you life easier, why not to go for it and make this part of `ASTTransformer`. Sounds like a great idea!

python/gstaichi/lang/_dataclass_util.py
duburcqa: By the way, it seems to be more generic than just a dataclass utils. Could be used at many places.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -52,33 +60,64 @@ def _decl_and_create_variable(
...
                 kernel_arguments.decl_sparse_matrix,
                 (
-                    to_gstaichi_type(arg_features),
+                    to_gstaichi_type(this_arg_features),
                     full_name,
                 ),
             )
         if isinstance(annotation, ndarray_type.NdarrayType):
+            assert this_arg_features is not None
             return False, (
```
duburcqa, line 84: What about unpacking `arg_features` with proper names for each element first ? This would clarify what are the various arguments that are post-processed right after, which would be very helpful to understand this code. It would also validate the number of elements in the tuple.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -52,33 +60,64 @@ def _decl_and_create_variable(
...
         if isinstance(annotation, texture_type.TextureType):
-            return False, (kernel_arguments.decl_texture_arg, (arg_features[0], full_name))
+            assert this_arg_features is not None
+            return False, (kernel_arguments.decl_texture_arg, (this_arg_features[0], full_name))
         if isinstance(annotation, texture_type.RWTextureType):
+            assert this_arg_features is not None
             return False, (
                 kernel_arguments.decl_rw_texture_arg,
-                (arg_features[0], arg_features[1], arg_features[2], full_name),
+                (this_arg_features[0], this_arg_features[1], this_arg_features[2], full_name),
```
duburcqa, line 101: Same, what about unpacking with proper names first ? Or at least the extracted subset:
```python
great_amazing_params = this_arg_features[:2]
return False, (
    kernel_arguments.decl_rw_texture_arg,
    (*great_amazing_params, full_name),
)
```

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -52,33 +60,64 @@ def _decl_and_create_variable(
...
+            class ChildStruct:
+                b: ti.types.NDArray[ti.32, 1]
+
+            ... so we'll expand the paramters to:
+            - __ti_some_struct_a: ti.types.NDArray[ti.32, 1]
+            - __ti_some_struct__ti_child__ti_b: ti.types.NDArray[ti.32, 1]
+            (or sometihng smilar ish)
+
+            So... we'll loop over the fields, and send those each to
+            """
```
duburcqa: Wait, there is nothing in this branch ? It is "just" to indicate that the comment holds for data classes only ? This is a bit weird x)

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -112,25 +151,35 @@ def _transform_kernel_arg(
...
+                        invoke_later_dict,
+                        "",
+                        0,
+                    )
+                    if result:
+                        ctx.create_variable(flat_name, obj)
+                    else:
+                        decl_type_func, type_args = obj
+                        obj = decl_type_func(*type_args)
+                        ctx.create_variable(flat_name, obj)
```
duburcqa, line 164: What about:
```python
if not result:
    _decl_type_func, type_args = obj
    obj = decl_type_func(*type_args)
ctx.create_variable(flat_name, obj)
```

duburcqa: Note `_` prefix for unused variables, which clarify that it is unused on purpose but keeps explicit what it stores. This is a common practice.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -150,22 +199,27 @@ def _transform_kernel_arg(
 
     @staticmethod
     def _transform_as_kernel(ctx: ASTTransformerContext, node: ast.FunctionDef, args: ast.arguments) -> None:
+        assert ctx.func is not None
+        assert ctx.arg_features is not None
```
duburcqa, line 183: Would be nice to use properties to make sure these "attributes" are never None by design. Not a big deal though.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -150,22 +199,27 @@ def _transform_kernel_arg(
...
+        assert compiling_callable is not None
+        compiling_callable.finalize_rets()
 
-        invoke_later_dict: dict[str, tuple[Any, str, Any]] = dict()
-        create_variable_later = dict()
-        for i, arg in enumerate(args.args):
-            argument = ctx.func.arguments[i]
+        invoke_later_dict: dict[str, tuple[Any, str, Callable, list[Any]]] = dict()
+        create_variable_later: dict[str, Any] = dict()
+        for i, _ in enumerate(args.args):
```
duburcqa: This pattern is a bit weird. What about `for arg_meta in ctx.func.arg_metas[:len(args.args)]:`, or just `for i in range(len(args.args)):`

duburcqa: besides, prefer `_arg` over just `_` for unused arguments. This may help the reader in complex codebase as taichi.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -203,12 +258,19 @@ def _transform_func_arg(
...
                     ),
                 ):
+                    field.type.check_matched(data_child.get_type(), field.name)
+                    ctx.create_variable(flat_name, data_child)
+                elif dataclasses.is_dataclass(data_child):
+                    FunctionDefTransformer._transform_func_arg(
+                        ctx,
+                        flat_name,
+                        field.type,
+                        getattr(data, field.name),
```
duburcqa, line 248: Either use keyword arguments for clarity, or keep all input arguments in one line if not important.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -274,15 +334,23 @@ def _transform_func_arg(
...
+        assert ctx.argument_data is not None
+        assert ctx.func is not None
         for data_i, data in enumerate(ctx.argument_data):
-            argument = ctx.func.arguments[data_i]
+            argument = ctx.func.arg_metas[data_i]
             FunctionDefTransformer._transform_func_arg(
                 ctx,
                 argument.name,
                 argument.annotation,
                 data,
```
duburcqa: Same, either use keyword arguments for clarity, or keep all input arguments in one line if not important.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -274,15 +334,23 @@ def _transform_func_arg(
...
             )
 
+        # deal with dataclasses
+
+        # pylint: disable=import-outside-toplevel
+        from gstaichi.lang.kernel_impl import Func
```
duburcqa: If you need delayed import, I would recommend to rather move import on top of this method but as you wish.

duburcqa: By the way, just move the assert on the top and remove `assert ctx.func is not None`, it is redundant.

python/gstaichi/lang/kernel_arguments.py
```diff
@@ -15,12 +15,19 @@
 from gstaichi.types.primitive_types import RefType, u64
 
 
-class KernelArgument:
+class ArgMetadata:
+    """
+    Metadata about an argument to a function
+    """
+
     def __init__(self, _annotation, _name, _default=inspect.Parameter.empty):
```
duburcqa: why `_` prefix ? It is weird.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -243,6 +243,28 @@ def pyfunc(fn: Callable) -> GsTaichiCallable:
     return gstaichi_callable
 
 
+def _populate_global_vars_for_templates(
```
duburcqa, line 246: No docstring ? You already have a small comment that could be used as such ?

python/gstaichi/lang/kernel_impl.py
```diff
@@ -290,31 +308,20 @@ def _get_tree_and_ctx(
...
-    ret: list[Any] = [argument.default for argument in self.arguments]
+        self.arg_metas = _kernel_impl_dataclass.expand_func_arguments(self.arg_metas)
+
+    fused_args: list[Any] = [arg_meta.default for arg_meta in self.arg_metas]
     len_args = len(args)
 
     if len_args > len(fused_args):
         arg_str = ", ".join([str(arg) for arg in args])
-        expected_str = ", ".join([f"{arg.name} : {arg.annotation}" for arg in self.arguments])
+        expected_str = ", ".join([f"{arg.name} : {arg.annotation}" for arg in self.arg_metas])
```
duburcqa: `expected_str = ", ".join(f"{arg.name} : {arg.annotation}" for arg in self.arg_metas)`

python/gstaichi/lang/kernel_impl.py
```diff
@@ -323,7 +330,7 @@ def _process_args(self: "Func | Kernel", is_func: bool, args: tuple[Any, ...], k
 
     for key, value in kwargs.items():
         found = False
-        for i, arg in enumerate(self.arguments):
+        for i, arg in enumerate(self.arg_metas):
```
duburcqa, line 334: This is more pythonic:
```python
        for i, arg in enumerate(self.arg_metas):
            if key == arg.name:
                if i < len_args:
                    raise GsTaichiSyntaxError(f"Multiple values for argument '{key}'.")
                fused_args[i] = value
                break
        else:
            raise GsTaichiSyntaxError(f"Unexpected argument '{key}'.")
```

python/gstaichi/lang/kernel_arguments.py
hughperkins: Not part of this PR. But updated it.

python/gstaichi/lang/_kernel_impl_dataclass.py
hughperkins: > By the way, why use _ prefix for constructor arguments ? This is weird and it seems not well motivated.

Not my code... I just renamed the class, but I didn't write the class.

python/gstaichi/lang/_kernel_impl_dataclass.py
hughperkins: _flatten_attribute_name can return None yeah. Notably if it receives anyting other than an Attribute or a Name. Should it throw an Exception instead? Unclear.

Ok, I checked: if we raise exception instead, then some tests fail, e.g. `test_func_of_data_class_as_kernel_arg`:

```
>       raise Exception(f"Unhandled node value type {type(node.value)} in {ast.dump(node)}")
E       Exception: Unhandled node value type <class 'ast.Call'> in Attribute(value=Call(func=Attribute(value=Name(id='x', ctx=Load()), attr='add', ctx=Load()), args=[Name(id='x', ctx=Load())], keywords=[]), attr='x', ctx=Load())

python/gstaichi/lang/_kernel_impl_dataclass.py:140: Exception
```

python/gstaichi/lang/_kernel_impl_dataclass.py
hughperkins: It's kind of 'scar tissue' from development, since it's a pre-requisite for printing out the name, for debugging purposes. Will remove.

python/gstaichi/lang/ast/ast_transformer.py
hughperkins: Yeah, I was expecting adrian to call me out on this.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
hughperkins: I'm going to delete Texture soon anway.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
hughperkins: 🤔 Not sure what this is. Removed it and tests still pass...

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
hughperkins: Which variable is unused? decl_type_func is being called at line 181

hughperkins:
Note: I still have 3 comments to address ,just to be clear.

python/gstaichi/lang/kernel_impl.py
hughperkins: til :)

hughperkins:
addressed all comments now, AFAIK.

hughperkins:
Just realized I forgot to copy the unit tests into this PR. Can you both review the unit tests, and @gasnica can you re-approve after reviewing the unit tests pelase?

python/gstaichi/_test_tools/dataclass_test_tools.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import Any, cast
+
+import gstaichi as ti
+
+
+def build_struct(struct_type: Any) -> Any:
+    member_objects = {}
+    for field in dataclasses.fields(struct_type):
+        if isinstance(field.type, ti.types.NDArray):
+            print("got ndarray")
```
duburcqa: You want to keep this debug print?

python/gstaichi/_test_tools/dataclass_test_tools.py
```diff
@@ -0,0 +1,44 @@
...
+            child_obj = ti.field(ti.i32, (10,))
+        else:
+            raise Exception("unknown type ", field.type)
+        member_objects[field.name] = child_obj
+    # DataclassClass = dataclasses.make_dataclass(struct_name, declaration_type_by_name)
+    dataclass_object = struct_type(**member_objects)
+    return dataclass_object
+
+
+def build_obj_tuple_from_type_dict(name_to_type: dict[str, Any]) -> tuple[Any, ...]:
```
duburcqa, line 31: It would be great to avoid all these code duplication with 'build_struct', but it sounds painful to do for limited advantage here.

python/gstaichi/lang/_kernel_impl_dataclass.py
duburcqa: That is not what I meant. I was saying that if `None` or `""` is guaranteed not to be in `self.struct_locals`, then you don't need to first check whether `not flat_name` is True or not. Nothing more x)

python/gstaichi/lang/_dataclass_util.py
duburcqa, line 1: Is `_dataclass_util` the right name for this file?

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -1309,6 +1312,10 @@ def build_Pass(ctx: ASTTransformerContext, node: ast.Pass) -> None:
 
 
 def build_stmts(ctx: ASTTransformerContext, stmts: list[ast.stmt]):
+    """
+    TODO: Should we just make this part of ASTTransformer? Then, easier to pass around (just
+    pass the ASTTransformer object around)
+    """
```
duburcqa: Strange to use a TODO as docstring

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -29,18 +34,25 @@
 class FunctionDefTransformer:
     @staticmethod
     def _decl_and_create_variable(
-        ctx: ASTTransformerContext, annotation, name, arg_features, invoke_later_dict, prefix_name, arg_depth
+        ctx: ASTTransformerContext,
+        annotation: Any,
+        name: str,
+        this_arg_features: tuple[tuple[Any, ...], ...] | None,
```
duburcqa, line 40: What does 'this_arg_features' means?

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
duburcqa: Ok let's ignore this then.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
duburcqa: > Which variable is unused? decl_type_func is being called at line 181

Ok I thought it was without the full context. My bad. I change I suggest to avoid duplicated `ctx.create_variable(flat_name, obj)` call is still valid.

python/gstaichi/lang/impl.py
```diff
@@ -70,17 +71,16 @@
...
 def expr_init_shared_array(shape, element_type):
-    compiling_callable = get_runtime().compiling_callable
-    assert compiling_callable is not None
-    return compiling_callable.ast_builder().expr_alloca_shared_array(
-        shape, element_type, _ti_core.DebugInfo(get_runtime().get_current_src_info())
+    return (
+        get_runtime()
+        .compiling_callable.ast_builder()
+        .expr_alloca_shared_array(shape, element_type, _ti_core.DebugInfo(get_runtime().get_current_src_info()))
     )
```
duburcqa: What about
```python
ast_builder = get_runtime().compiling_callable.ast_builder()
debug_info = _ti_core.DebugInfo(get_runtime().get_current_src_info())
return ast_builder.expr_alloca_shared_array(shape, element_type, debug_info)
```

python/gstaichi/lang/impl.py
```diff
@@ -962,9 +968,7 @@ def add_separators(_vars):
 
     _vars = add_separators(_vars)
     contents, formats = ti_format_list_to_content_entries(_vars)
-    compiling_callable = get_runtime().compiling_callable
-    assert compiling_callable is not None
-    compiling_callable.ast_builder().create_print(
+    get_runtime().compiling_callable.ast_builder().create_print(
```
duburcqa: Same. Maybe define `ast_builder` and `debug_info` ? It would be more readable.

python/gstaichi/lang/impl.py
```diff
@@ -996,9 +1000,7 @@ def ti_format(*args):
 def ti_assert(cond, msg, extra_args, dbg_info):
     # Mostly a wrapper to help us convert from Expr (defined in Python) to
     # _ti_core.Expr (defined in C++)
-    compiling_callable = get_runtime().compiling_callable
-    assert compiling_callable is not None
-    compiling_callable.ast_builder().create_assert_stmt(Expr(cond).ptr, msg, extra_args, dbg_info)
+    get_runtime().compiling_callable.ast_builder().create_assert_stmt(Expr(cond).ptr, msg, extra_args, dbg_info)
```
duburcqa: Same. Maybe define `ast_builder` ?

python/gstaichi/lang/util.py
```diff
@@ -56,7 +56,9 @@ def has_paddle():
     _env_paddle = os.environ.get("TI_ENABLE_PADDLE", "1")
     if not _env_paddle or int(_env_paddle):
         try:
-            import paddle  # pylint: disable=C0415
+            import paddle  # type: ignore pylint: disable=C0415
+
+            _ = paddle
 
             _has_paddle = True
```
duburcqa, line 63: Let's get rid of paddle sooner than later (but not in this PR).

python/gstaichi/lang/kernel_impl.py
```diff
@@ -290,101 +314,48 @@ def _get_tree_and_ctx(
...
-                new_arguments.append(new_argument)
-        else:
-            new_arguments.append(argument)
-    return new_arguments
-
-
 def _process_args(self: "Func | Kernel", is_func: bool, args: tuple[Any, ...], kwargs) -> tuple[Any, ...]:
+    """
+    used for both Func and Kernel
+    """
```
duburcqa: This is a weird docstring x)

python/gstaichi/lang/kernel_impl.py
```diff
@@ -243,6 +243,34 @@ def pyfunc(fn: Callable) -> GsTaichiCallable:
...
+    fn: Callable,
+    py_args: tuple[Any, ...],
+):
+    """
+    inject template parameters into globals
+
+    globals are being abused to store the python objects associated
+    with templates. we continue this approach, and in addition this function
+    handles injecting expanded python variables from dataclasses
+    """
```
duburcqa: Formatting is not great (Starting with uppercase, end with dot, random linewidth). But I will let that slip in since it is internal documentation.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -290,101 +314,48 @@ def _get_tree_and_ctx(
...
-    ret: list[Any] = [argument.default for argument in self.arguments]
+        self.arg_metas = _kernel_impl_dataclass.expand_func_arguments(self.arg_metas)
+
+    fused_args: list[Any] = [arg_meta.default for arg_meta in self.arg_metas]
     len_args = len(args)
 
     if len_args > len(fused_args):
-        arg_str = ", ".join([str(arg) for arg in args])
-        expected_str = ", ".join([f"{arg.name} : {arg.annotation}" for arg in self.arguments])
+        arg_str = ", ".join(str(arg) for arg in args)
```
duburcqa: If you enjoy Python language, you could go all-in and do: `", ".join(map(str, args))`

python/gstaichi/lang/_dataclass_util.py
hughperkins: sure. What concerns do you have with the name?

python/gstaichi/lang/ast/ast_transformer.py
hughperkins: You mean, change to use `#` comments instead?

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
hughperkins: you mean, rename it as `arg_features`?

python/gstaichi/lang/util.py
hughperkins: added it to Linear

python/gstaichi/lang/kernel_impl.py
hughperkins: you mean, because the type of the first arg suggests this? Removed.

python/gstaichi/lang/ast/ast_transformer.py
duburcqa: Yes.

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
duburcqa: It depends. Does `this_` prefix add any value? Maybe it does, it is just not clear to me what it means in this context.

python/gstaichi/lang/_dataclass_util.py
duburcqa: Because I thought it was used more broadly than just for dataclass handling.

python/gstaichi/lang/ast/ast_transformer.py
hughperkins: addressed

python/gstaichi/_test_tools/dataclass_test_tools.py
```diff
@@ -0,0 +1,36 @@
+import dataclasses
+from typing import Any, cast
+
+import gstaichi as ti
+
+
+def _make_child_obj(obj_type: Any) -> Any:
+    if isinstance(obj_type, ti.types.NDArray):
```
gasnica, line 8: could we add docstrings to all tool/utility functions that are to be reused, e.g. in this file for the start?

python/gstaichi/_test_tools/dataclass_test_tools.py
```diff
@@ -0,0 +1,36 @@
...
+        ndarray_type = cast(ti.types.ndarray, obj_type)
+        assert ndarray_type.ndim is not None
+        shape = tuple([10] * ndarray_type.ndim)
+        child_obj = ti.ndarray(ndarray_type.dtype, shape=shape)
+    elif dataclasses.is_dataclass(obj_type):
+        child_obj = build_struct(obj_type)
+    elif isinstance(obj_type, ti.Template) or obj_type == ti.Template:
+        child_obj = ti.field(ti.i32, (10,))
```
gasnica, line 16: what are we doing here with the seemingly arbitrary field of 10 elems?

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,212 @@
...
+
+from gstaichi.lang import util
+from gstaichi.lang._dataclass_util import create_flat_name
+from gstaichi.lang.ast import (
+    ASTTransformerContext,
+)
+from gstaichi.lang.kernel_arguments import ArgMetadata
+
+
+def _populate_struct_locals_from_params_dict(basename: str, struct_locals, struct_type) -> None:
```
gasnica, line 14: is basename a name_prefix? possibly rename

python/gstaichi/lang/_kernel_impl_dataclass.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    We are populating struct locals from a type included in function parameters, or one of their subtypes
```
gasnica, line 16: what is that type ?? is it the struct_type param ?

python/gstaichi/lang/_kernel_impl_dataclass.py
gasnica: should it be renamed to **_from_type? or **_from_struct_type? -- i'm confused where the params_dict is ... (?)

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -645,6 +646,8 @@ def build_Attribute(ctx: ASTTransformerContext, node: ast.Attribute):
 
                 node.ptr = getattr(tensor_ops, node.attr)
                 setattr(node, "caller", node.value.ptr)
+        elif dataclasses.is_dataclass(node.value.ptr):
+            node.ptr = next(field.type for field in dataclasses.fields(node.value.ptr))
```
gasnica, line 650: what is happening here?
- why are we checking for dataclass here (i actually just don't know)
- what is next() returning some kind of iterator, how does it work, given other if-else paths return an attribute?

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -52,32 +64,41 @@ def _decl_and_create_variable(
...
+            ndim: int
+            needs_grad: bool
+            boundary: BoundaryMode
+            raw_element_type, ndim, needs_grad, boundary = this_arg_features
             return False, (
                 kernel_arguments.decl_ndarray_arg,
                 (
-                    to_gstaichi_type(arg_features[0]),
-                    arg_features[1],
+                    to_gstaichi_type(raw_element_type),
```
gasnica, line 87: what are we doing here? returning some null/uninitialized values?

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -141,31 +172,34 @@ def _transform_kernel_arg(
                 "",
                 0,
             )
-            if result:
-                ctx.create_variable(argument_name, obj)
```
gasnica, line 145: so why are we now not calling ctx.create_variable() ? just asking

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -186,15 +220,16 @@ def _transform_func_arg(
...
             return None
 
         if dataclasses.is_dataclass(argument_type):
-            dataclass_type = argument_type
-            for field in dataclasses.fields(dataclass_type):
+            for field in dataclasses.fields(argument_type):
+                flat_name = create_flat_name(argument_name, field.name)
                 data_child = getattr(data, field.name)
-                if not isinstance(
+                if isinstance(
```
gasnica, line 232: is this negation of condition intentional here?

python/gstaichi/lang/impl.py
```diff
@@ -70,17 +71,14 @@
 
 @gstaichi_scope
 def expr_init_shared_array(shape, element_type):
-    compiling_callable = get_runtime().compiling_callable
-    assert compiling_callable is not None
```
gasnica, line 74: no assert anymore, now?

python/gstaichi/lang/impl.py
```diff
@@ -194,9 +192,7 @@ def validate_subscript_index(value, index):
 @gstaichi_scope
 def subscript(ast_builder, value, *_indices, skip_reordered=False):
     dbg_info = _ti_core.DebugInfo(get_runtime().get_current_src_info())
```
gasnica, line 194: let's use consistent names maybe -- rename this to debug_info?

hughperkins: this pr has been merged ...

hughperkins: Note: if you don't intend me to merge a PR, please dont approve it 🙂

hughperkins: oh, this is an old commment. for some reason I received the notifiation now 🤔  Or maybe I clicked on the wrong email thread or something. anyway, ok :)

PR #94, opened 2025-07-18
[Type] Fix ti.Template on ti.funcs

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_template_typing.py
+++ b/tests/python/test_template_typing.py
@@ -1,5 +1,3 @@
-import pytest
-
 import taichi as ti
 from tests import test_utils
 
@@ -17,6 +15,9 @@ def some_kernel(a: ti.template, b: ti.Template) -> None:
 
 @test_utils.test()
 def test_template_no_braces():
+    """
+    Check that we can use ti.Template as an annotation for kernels and funcs.
+    """
     a = ti.field(int, shape=())
     b = ti.field(int, shape=())
     b[None] = 5


python/taichi/lang/kernel_impl.py
```diff
@@ -436,7 +436,7 @@ def extract_arguments(self) -> None:
                     pass
                 elif type(annotation) == taichi.types.annotations.Template:
                     pass
-                elif isinstance(annotation, template):
+                elif isinstance(annotation, template) or annotation == taichi.types.annotations.Template:
```
gasnica, line 439: 1. I think Alexis once suggested reversed order (?)

2. Why is it needed now? Did we get a crash?

3. Would it be worthwhile to add an introductory sentence or two in PR description, and/or possibly also make the PR title more specific?

gasnica: Like, some context for the issue would ease me into the review. Otherwise, I don't know what we're addressing here.

tests/python/test_template_typing.py
```diff
@@ -4,9 +4,15 @@
 from tests import test_utils
 
 
+@ti.func
+def some_sub_func(a: ti.template, b: ti.Template) -> None:
```
gasnica, line 6: Is it a good idea, that we start adding descriptions to our tests? Even a one liner. Otherwise, it takes time and effort to figure what we may be testing for. 

Like, I understand we "check that ti.Template can be used as param type for ti.functions'?

python/taichi/lang/kernel_impl.py
duburcqa: > I think Alexis once suggested reversed order (?)

Possibly, I'm not sure. The rational is to optimise the speed of this check by leveraging short-circuit mechanism. The optimal order depends on the probably of each condition, plus their respective time. `isinstance` is fairly fast in the context (a single, concrete, type is passed), and I don't know what the equality check on the right hand side is doing under the hood. It may be fast or slow depending on its implementation.

tests/python/test_template_typing.py
duburcqa: > Is it a good idea, that we start adding descriptions to our tests? Even a one liner. Otherwise, it takes time and effort to figure what we may be testing for.

I agree that a good unit test should describe what it is being computed if not directly obvious, and mention which software requirements are being validated even if it is obvious (to ease traceability). I also agree that it is a bit painful to do manually.

PR #96, opened 2025-07-20
[Unittest] Add a test to cover calling class method

YilingQiao:
I got one error in a dev branch, which the current unittests don't cover
```
        self.f1(c, inner=inner)
        ^^^^^^^^^^^^^^^^^^^^^^^
TypeError when calling `f1`: BoundFunc.__call__() got an unexpected keyword argument 'inner'.
```


tests/python/test_class_method.py
```diff
@@ -0,0 +1,44 @@
+import taichi as ti
+from tests import test_utils
+
+
+@test_utils.test()
+def test_class_method():
```
hughperkins: `-> None`

YilingQiao: done

tests/python/test_class_method.py
```diff
@@ -0,0 +1,51 @@
+import pytest
+
+import taichi as ti
+from tests import test_utils
+
+
+@pytest.mark.parametrize(
```
duburcqa: Why are you spreading parameters this much?  This is distracting me from the content that matters.
```python
@pytest.mark.parametrize("use_ndarray", [False, True])
```

tests/python/test_class_method.py
```diff
@@ -0,0 +1,51 @@
...
+    "use_ndarray",
+    [
+        False,
+        True,
+    ],
+)
+@test_utils.test()
+def test_class_method(use_ndarray: bool) -> None:
+    if not use_ndarray:
```
duburcqa: I would suggest adding local fixtures for this:
```python
@pytest.fixture
def ti_type(use_ndarray: bool) -> Any:
    if use_ndarray:
        return ti.ndarray
    return ti.field


@pytest.fixture
def ti_annotation(use_ndarray: bool) -> Any:
    if use_ndarray:
        return ti.types.ndarray()
    return ti.template()


@pytest.mark.parametrize("use_ndarray", [False, True])
@test_utils.test()
def test_class_method(ti_type: Any, ti_annotation: Any) -> None:
    [...]
```

tests/python/test_class_method.py
hughperkins: that came from me. I am used to parametrize blocks being huge things, and they almost always have to be on multiple lines as a consequence. Agree that in this case a single line is cleaner.

YilingQiao: fixed

tests/python/test_class_method.py
YilingQiao: fixed

PR #99, opened 2025-07-21
[Type] Factorization for nested structs

hughperkins:
Issue: #

### Brief Summary

This PR fits in:
- after typing batch 5
- before nested structs

I'm not sure if having this PR inbetween the above two PRs will make my own life easier 🤔 , but I think it should make reviewing easier.

In this PR:
- I move code around, into either new functions, or new modules, or out of being nested
- I make the *minimal* changes required in order to do that
- I try not to make any other code changes at all (so that you know this is basically just ~moves)

Changes were:
- move TaichiCallableTemplateMapper
    - from kernel_impl.py
    - into new file _template_mapper.py
- move get_decorator
    - from ast_transformer.py
    - into ast_transformer_utils.py
- move build_Call things
    - from ast_transformer.py
    - into new file call_transformer.py
- move build_FunctionDef
    - from ast_transformer.py
    - into new file function_def_transformer.py
- in function_def_transformer.py:
    - move decl_and_create_variable
        - from being neseted inside build_FunctionDef
        - to being a top level method
    - move transform_as_kernel
        - from being neseted inside build_FunctionDef
        - to being a top level method
    - factorize out chunk of code
        - into new top level method `_transform_as_func`
        - from inside build_FunctionDef
    - factorize out chunk of code
        - into new top level method `_process_func_arg`
        - from inside `_transform_as_func`
    - factorize out chunk of code
        - into new top level method `_process_kernel_arg`
        - from inside `_tranform_as_kernel`

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/taichi/graph/_graph.py
+++ b/python/taichi/graph/_graph.py
@@ -49,7 +49,7 @@ class Sequential:
 
 class GraphBuilder:
     def __init__(self):
-        self._graph_builder = _ti_core.GraphBuilder()
+        self._graph_builder = _ti_core.GraphBuilderCxx()
 
     def dispatch(self, kernel_fn, *args):
         kernel_cpp = gen_cpp_kernel(kernel_fn, args)


python/taichi/lang/_template_mapper.py
```diff
@@ -0,0 +1,203 @@
...
+    ndarray_type,
+    sparse_matrix_builder,
+    template,
+    texture_type,
+)
+from taichi.types.enums import AutodiffMode
+
+CompiledKernelKeyType = tuple[Callable, int, AutodiffMode]
+
+
```
duburcqa, line 25: I would not skip 2 lines between type declaration. But as you want.

python/taichi/lang/_template_mapper.py
```diff
@@ -0,0 +1,203 @@
...
+    "texture_type.RWTextureType",
+    ndarray_type.NdarrayType,
+    sparse_matrix_builder,
+    Any,
+]
+
+
+class TaichiCallableTemplateMapper:
+    """
+    This should probably be renamed to sometihng like FeatureMapper, or
```
duburcqa, line 39: The way you split lines in docstring seems a bit erratic x)

python/taichi/lang/_template_mapper.py
```diff
@@ -0,0 +1,203 @@
...
+            if isinstance(arg, _ti_core.ExprCxx):
+                return arg.get_underlying_ptr_address()
+            if isinstance(arg, tuple):
+                return tuple(TaichiCallableTemplateMapper.extract_arg(item, annotation, arg_name) for item in arg)
+            if isinstance(arg, taichi.lang._ndarray.Ndarray):
+                raise TaichiRuntimeTypeError(
+                    "Ndarray shouldn't be passed in via `ti.template()`, please annotate your kernel using `ti.types.ndarray(...)` instead"
+                )
+
+            if isinstance(arg, (list, tuple, dict, set)) or hasattr(arg, "_data_oriented"):
```
duburcqa, line 71: `hasattr(arg, "_data_oriented")` should be avoided is possible. Usually it is a bad practice.

python/taichi/lang/_template_mapper.py
```diff
@@ -0,0 +1,203 @@
...
+                # [Composite arguments] Return weak reference to the object
+                # Taichi kernel will cache the extracted arguments, thus we can't simply return the original argument.
+                # Instead, a weak reference to the original value is returned to avoid memory leak.
+
+                # TODO(zhanlue): replacing "tuple(args)" with "hash of argument values"
+                # This can resolve the following issues:
+                # 1. Invalid weak-ref will leave a dead(dangling) entry in both caches: "self.mapping" and "self.compiled_functions"
```
duburcqa, line 78: This specific point can be solved by leveraging the callback capability of weak references. See: https://github.com/duburcqa/jiminy/blob/accdf3629d9d8c08bb97f2905f010f1616d2168a/python/gym_jiminy/common/gym_jiminy/common/bases/quantities.py#L43

python/taichi/lang/_template_mapper.py
```diff
@@ -0,0 +1,203 @@
...
+                raise TaichiRuntimeTypeError(f"Argument {arg_name} must be a texture, got {type(arg)}")
+            if arg.num_dims != annotation.num_dimensions:
+                raise TaichiRuntimeTypeError(
+                    f"RWTextureType dimension mismatch for argument {arg_name}: expected {annotation.num_dimensions}, got {arg.num_dims}"
+                )
+            if arg.fmt != annotation.fmt:
+                raise TaichiRuntimeTypeError(
+                    f"RWTextureType format mismatch for argument {arg_name}: expected {annotation.fmt}, got {arg.fmt}"
+                )
+            # (penguinliong) '0' is the assumed LOD level. We currently don't
```
duburcqa, line 118: Single line comment would be fine.

python/taichi/lang/_template_mapper.py
```diff
@@ -0,0 +1,203 @@
...
+            if shape is None:
+                raise TaichiRuntimeTypeError(f"Invalid type for argument {arg_name}, got {arg}")
+            shape = tuple(shape)
+            element_shape: tuple[int, ...] = ()
+            dtype = to_taichi_type(arg.dtype)
+            if isinstance(annotation.dtype, MatrixType):
+                if annotation.ndim is not None:
+                    if len(shape) != annotation.dtype.ndim + annotation.ndim:
+                        raise ValueError(
+                            f"Invalid value for argument {arg_name} - required array has ndim={annotation.ndim} element_dim={annotation.dtype.ndim}, "
```
duburcqa, line 142: This line is probably too long.

python/taichi/lang/ast/ast_transformer.py
```diff
@@ -1564,10 +1072,10 @@ def build_nested_mesh_for(ctx: ASTTransformerContext, node: ast.For) -> None:
     def build_For(ctx: ASTTransformerContext, node: ast.For) -> None:
         if node.orelse:
             raise TaichiSyntaxError("'else' clause for 'for' not supported in Taichi kernels")
-        decorator = ASTTransformer.get_decorator(ctx, node.iter)
+        decorator = get_decorator(ctx, node.iter)
```
duburcqa, line 1077: `get_decorator` is not a good name anymore, now that it is a standalone method instead of a class method.

python/taichi/lang/ast/ast_transformer_utils.py
```diff
@@ -323,3 +325,17 @@ def gen_line(code: str, hint: str) -> str:
                     hint = ""
                 msg += gen_line(self.src[i], hint)
         return msg
+
+
+def get_decorator(ctx: ASTTransformerContext, node) -> str:
```
duburcqa, line 330: Could you add some docstring? Even if it is one sentence. That would be great, because it is definitely not self-explanatory for me that is not very used to the internals of taichi.

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+from taichi.lang.util import to_taichi_type
+from taichi.types import annotations, ndarray_type, primitive_types, texture_type
+
+
+class FunctionDefTransformer:
+    @staticmethod
+    def decl_and_create_variable(
+        ctx: ASTTransformerContext, annotation, name, arg_features, invoke_later_dict, prefix_name, arg_depth
+    ) -> tuple[bool, Any]:
+        full_name = prefix_name + "_" + name
```
duburcqa: Personally I prefer `"_".join((prefix_name, name))` but it is up to you.

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+            kernel_arguments.push_argpack_arg(name)
+            d = {}
+            items_to_put_in_dict = []
+            for j, (_name, anno) in enumerate(annotation.members.items()):
+                result, obj = FunctionDefTransformer.decl_and_create_variable(
+                    ctx, anno, _name, arg_features[j], invoke_later_dict, full_name, arg_depth + 1
+                )
+                if not result:
+                    d[_name] = None
+                    items_to_put_in_dict.append((full_name + "_" + _name, _name, obj))
```
duburcqa: same.

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+                else:
+                    d[_name] = obj
+            argpack = kernel_arguments.decl_argpack_arg(annotation, d)
+            for item in items_to_put_in_dict:
+                invoke_later_dict[item[0]] = argpack, item[1], *item[2]
```
duburcqa: Use parentheses. I don't like implicit tuple syntax that came with Python 3.

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+        ctx: ASTTransformerContext,
+        invoke_later_dict: dict[str, tuple[Any, str, Callable, list[Any]]],
+        create_variable_later: dict[str, Any],
+        argument_name: str,
+        argument_type: Any,
+        this_arg_features: tuple[Any, ...],
+    ) -> None:
+        if isinstance(argument_type, ArgPackType):
+            kernel_arguments.push_argpack_arg(argument_name)
+            d = {}
```
duburcqa: `d` is not a good name.

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+                    ctx, anno, name, this_arg_features[j], invoke_later_dict, "__argpack_" + name, 1
+                )
+                if not result:
+                    d[name] = None
+                    items_to_put_in_dict.append(("__argpack_" + name, name, obj))
+                else:
+                    d[name] = obj
+            argpack = kernel_arguments.decl_argpack_arg(argument_type, d)
+            for item in items_to_put_in_dict:
+                invoke_later_dict[item[0]] = argpack, item[1], *item[2]
```
duburcqa: Parenthesis.

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+        invoke_later_dict: dict[str, tuple[Any, str, Callable, list[Any]]],
+        create_variable_later: dict[str, Any],
+        argument_name: str,
+        argument_type: Any,
+        this_arg_features: tuple[Any, ...],
+    ) -> None:
+        if isinstance(argument_type, ArgPackType):
+            kernel_arguments.push_argpack_arg(argument_name)
+            d = {}
+            items_to_put_in_dict: list[tuple[str, str, Any]] = []
```
duburcqa: `tuple[str, str, Any]` is not a good type. It would be better to define aliases for each element, or for the tuple itself.

python/taichi/lang/_template_mapper.py
hughperkins: This isnt my code. Im only moving it. It's not something I added in this PR

python/taichi/lang/_template_mapper.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/_template_mapper.py
hughperkins: AGain, This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/ast/function_def_transformer.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/ast/function_def_transformer.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/ast/function_def_transformer.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/ast/function_def_transformer.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/ast/function_def_transformer.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/_template_mapper.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/ast/ast_transformer.py
hughperkins: Ok. Plausibly in scope ofr this PR. Suggestion on new name?

python/taichi/lang/ast/ast_transformer_utils.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/ast/function_def_transformer.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/_template_mapper.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/_template_mapper.py
hughperkins: This isn't my code. I'm only moving it. I didnt write this code in this pr

python/taichi/lang/_template_mapper.py
hughperkins: Actually, I did write this, but not in this PR. Out of scope for this PR please.

python/taichi/lang/ast/function_def_transformer.py
hughperkins: (or, the type migth be my code, but it's not from this pr)

python/taichi/lang/_template_mapper.py
```diff
@@ -0,0 +1,203 @@
...
+    sparse_matrix_builder,
+    template,
+    texture_type,
+)
+from taichi.types.enums import AutodiffMode
+
+CompiledKernelKeyType = tuple[Callable, int, AutodiffMode]
+
+
+AnnotationType = Union[
```
gasnica, line 26: this is copy-paste & duplication from kernel_impl.py Ln570. Can we avoid that ?

gasnica: actually -- is it used in this file (or in the original?)

python/taichi/lang/_template_mapper.py
```diff
@@ -0,0 +1,203 @@
...
+from taichi.lang.util import to_taichi_type
+from taichi.types import (
+    ndarray_type,
+    sparse_matrix_builder,
+    template,
+    texture_type,
+)
+from taichi.types.enums import AutodiffMode
+
+CompiledKernelKeyType = tuple[Callable, int, AutodiffMode]
```
gasnica: This seems to be not used in this file, is it needed ?

python/taichi/lang/ast/ast_transformer.py
gasnica: I had the same inital thought, but then it's already in side ast_transfomer_utils, and it takes the ASTTransfomer context.

but my first thougth was to ask to rename it to ASTTransformer__get_decorator()

gasnica: that will definiltey work properly with the 3 places it's used in the lines below.

python/taichi/lang/ast/call_transformer.py
```diff
@@ -0,0 +1,267 @@
...
+    TaichiSyntaxError,
+    TaichiTypeError,
+)
+from taichi.lang.expr import Expr
+from taichi.lang.matrix import Matrix, Vector
+from taichi.lang.util import is_taichi_class
+from taichi.types import primitive_types
+
+
+class CallTransformer:
```
gasnica: @duburcqa 

Hugh suggested to check it by you -- I'd rather have this named ASTTransformerForCall(Node) -- so that name directly relates to the ASTTransformer that we're splitting from.

Hugh advocates for shorter names, that are a general preference in python. But I'd rather have the name being super long and super explicit, given it's only typed in a couple of places.

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+                (arg_features[0], arg_features[1], arg_features[2], full_name),
+            )
+        if isinstance(annotation, MatrixType):
+            return True, kernel_arguments.decl_matrix_arg(annotation, name, arg_depth)
+        if isinstance(annotation, StructType):
+            return True, kernel_arguments.decl_struct_arg(annotation, name, arg_depth)
+        return True, kernel_arguments.decl_scalar_arg(annotation, name, arg_depth)
+
+    @staticmethod
+    def _process_kernel_arg(
```
gasnica: 1. possibly rename to _transform_xxx
2. do we prefer to have this located under decl_and_create_variable() since that's the order you're read and call the functions?

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+        argument_name: str,
+        argument_type: Any,
+        this_arg_features: tuple[Any, ...],
+    ) -> None:
+        if isinstance(argument_type, ArgPackType):
+            kernel_arguments.push_argpack_arg(argument_name)
+            d = {}
+            items_to_put_in_dict: list[tuple[str, str, Any]] = []
+            for j, (name, anno) in enumerate(argument_type.members.items()):
+                result, obj = FunctionDefTransformer.decl_and_create_variable(
```
gasnica: original code does't use ASTTransformer/FunctionDefTransformer scope. We can drop it if we'd like to reaname the class to ASTTransformer_ForFunctionDefNode, etc.

hughperkins: what do you mean by 'drop it'? Drop waht? how are you imaginging that would look?

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+            argpack[name] = func(*params)
+        for k, v in create_variable_later.items():
+            ctx.create_variable(k, v)
+
+        impl.get_runtime().compiling_callable.finalize_params()
+        # remove original args
+        node.args.args = []
+
+    @staticmethod
+    def _process_func_arg(
```
gasnica: 1. rename to _transform_func_arg
2. move below _transform_as_func

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+        ctx: ASTTransformerContext,
+        argument_name: str,
+        argument_type: Any,
+        data: Any,
+    ) -> None:
+        if isinstance(argument_type, annotations.template):
```
gasnica: possibly use argument_annotation instead of *_type -- only if also ranaming in _transform_kernel_arg

hughperkins: Personally I feel taht argument_type is more intuitive? Took me forever to understand what 'argument annotation' meant.

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+                argument.annotation,
+                data,
+            )
+
+        for v in ctx.func.orig_arguments:
+            if dataclasses.is_dataclass(v.annotation):
+                ctx.create_variable(v.name, v.annotation)
+
+    @staticmethod
+    def build_FunctionDef(
```
gasnica: move it to the top -- since that's the outermost function, and the one public function, and sets context for the other ones ??

python/taichi/lang/ast/function_def_transformer.py
```diff
@@ -0,0 +1,320 @@
...
+)
+from taichi.lang.matrix import MatrixType
+from taichi.lang.struct import StructType
+from taichi.lang.util import to_taichi_type
+from taichi.types import annotations, ndarray_type, primitive_types, texture_type
+
+
+class FunctionDefTransformer:
+    @staticmethod
+    def decl_and_create_variable(
```
gasnica: this is private and should be _underscored ?

python/taichi/lang/ast/function_def_transformer.py
hughperkins: I feel thta I tend to put the functions being called before the fucntions doing the calling 🤔

@duburcqa your thoughts on ordering of python functions?

python/taichi/lang/ast/function_def_transformer.py
hughperkins: In case you mean `drop FunctionDefTransformer`, "beacuse it wasnt present before", thats because decl_and_create_varaible was nested inside the same function scope before, so didn't need a class name.

However, now it is a static class method, and needs to be prefixed by the class name.

python/taichi/lang/_template_mapper.py
hughperkins: Screenshot 2025-07-31 at 5 39 44 PM

python/taichi/lang/ast/call_transformer.py
duburcqa: I'm in favour of explicit names, but the bar minimum that is needed to avoid confusion, because too long names also tend to impede readability. It is a matter of good balance. In this case, what about `ASTCallTransformer`? Sounds like a reasonable compromise.

python/taichi/lang/ast/function_def_transformer.py
duburcqa: I have no preference. It should be intuitive for the majority of contributors. In this case, only @hughperkins ?

hughperkins: @duburcqa tiebreak please?

duburcqa: I would suggest to minimizing changes wrt original taichi if applicable. If not, I prefer to define methods before they are first used. For instance:
```python
def a():
    pass

def b():
    return a()
```
But it is just a matter of taste.

python/taichi/lang/_template_mapper.py
gasnica: Oh, cool -- but I cannot text-search-find it in the original kernel_impl.py, so maybe remove it from there?

python/taichi/lang/ast/call_transformer.py
gasnica: That sounds great to me .. and ASTFunctionDefTransformer then for the other class I guess

hughperkins: Just for the record, I hate that name. Really long, hard to remember, hard to parse with one's eyes. But I'll go with th consensus.

python/taichi/lang/ast/function_def_transformer.py
gasnica: so that vaguely implies putting public interface at the bottom of the class, as it is here .... 

I'm good with it, since you guys agree on it :-)

python/taichi/lang/ast/call_transformer.py
gasnica: Oh -- noo, then stop :-)

gasnica: I'm good with the old name then -- just add a docstring. Optionally can put it in the subfolder as you suggested. But I'm good with things as they are. Didn't mean to forces.

hughperkins: I'm totally up for putting them in a sbufolder, called eg `ast_transformers`.

PR #101, opened 2025-07-22
[Test] Unit tests print full traceback on exception

hughperkins:
Issue: #

### Brief Summary

Unit tests print full traceback on exception (currently not the case, and I've been commetning out the various try, catch blocks during dev).

I dont think it was deliberately supprsed on unit tests. I think it was suppressed in general, so that users dont get shown full staack traces, and no-one ever re-enabled stack traces on unit tests

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/scripts_new/manylinux_wheel/1_prerequisites.sh
+++ b/.github/workflows/scripts_new/manylinux_wheel/1_prerequisites.sh
@@ -5,6 +5,11 @@ set -ex
 # yum update
 yum install -y git wget libXrandr-devel libXinerama-devel libXcursor-devel libXi-devel
 git config --global --add safe.directory /__w/taichi/taichi
+# not sure why this is needed, but breaks without this, after I created a branch that
+# renames from taichi to gs-taichi. I assume some github-site pollution.
+if [[ -d /__w/gs-taichi/gs-taichi ]]; then {
+    git config --global --add safe.directory /__w/gs-taichi/gs-taichi
+} fi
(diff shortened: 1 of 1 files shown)


tests/python/test_assert.py
```diff
@@ -122,7 +122,7 @@ def func():
     func()
 
 
-@test_utils.test(arch=get_host_arch_list())
+@test_utils.test(arch=get_host_arch_list(), print_full_traceback=False)
```
YilingQiao, line 125: why do you pick a subset of functions and set them `print_full_traceback=False`?

tests/python/conftest.py
```diff
@@ -68,6 +68,9 @@ def wanted_arch(request, req_arch, req_options):
             else:
                 # Serial tests run without aggressive resource optimization
                 req_options = {"device_memory_GB": 1, **req_options}
+        if "print_full_traceback" not in req_options:
+            req_options["print_full_traceback"] = True
+        print("req_options", req_options)
```
YilingQiao: Do you intentionally leave this print statement here

tests/python/test_assert.py
hughperkins: These functions all test for specific exceptions.

Enabling traceback modifies the exception.

tests/python/conftest.py
hughperkins: No :). Will fix. Thanks! :)

PR #102, opened 2025-07-23
[Build] Update gitignore for whl, so, stubs, CHANGELOG.md

hughperkins:
Issue: #

Update gitignore for whl, so, stubs, CHANGELOG.md.

Saves me from accidentally committing these files from time to time.

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/conftest.py
+++ b/tests/python/conftest.py
@@ -68,6 +68,8 @@ def wanted_arch(request, req_arch, req_options):
             else:
                 # Serial tests run without aggressive resource optimization
                 req_options = {"device_memory_GB": 1, **req_options}
+        if "print_full_traceback" not in req_options:
+            req_options["print_full_traceback"] = True
         ti.init(arch=req_arch, enable_fallback=False, **req_options)
     yield
     if req_arch is not None:
--- a/tests/python/test_assert.py
+++ b/tests/python/test_assert.py
@@ -122,7 +122,7 @@ def test_assert_with_check_oob():
     func()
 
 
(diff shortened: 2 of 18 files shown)


.gitignore
```diff
@@ -91,3 +91,7 @@ imgui.ini
 *.dylib
 *.ply
 !pyrightconfig.json
+*.whl
+*.so
+stubs/
+CHANGELOG.md
```
YilingQiao, line 97: why put CHANGELOG.md in gitignore?

hughperkins: 1. it's not part of the repo, AFAIK? eg https://github.com/search?q=repo%3AGenesis-Embodied-AI%2Fgs-taichi%20CHANGELOG.md&type=code
2. it gets generated locally, when one runs certain types of builds (maybe wheel builds?), and then it's hard to not acccidentally commit it. How my git looks on other branches:

Screenshot 2025-07-23 at 6 12 56 PM

duburcqa: I don't think it makes sense to ignore `*.whl, *.so, stubs/`.

I think only files that are generated automatically on purpose for various reasons should be ignored automatically (CHANGELOG.md is a good example). Otherwise you are promoting the bad practice of polluting your repository with stuffs that should be stored somewhere else, such as test and build artefacts. But since apparently it was already a huge mess before by looking at the existing .gitignore file, I think this repository is a lost cause on this point :/

YilingQiao: I see, this CHANGELOG.md is generated

PR #114, opened 2025-07-28
[Misc] Increase max kernel args to 512

hughperkins:
Issue: #

### Brief Summary

Increase max kernel args to 512

Implementation details:
- added tests, for ndarray max args 512, and for field max args 512 (using ti.Template)
- modified max args in kernel_impl.py
- modified max num snodes from 1024, to 2048, in order to allow creating up to 512 fields. (probably 1024 + ~2-4 would be sufficient actually, now that I think about it 🤔 )

Following major caveat:
- on metal, maximum num args appears to be less than 30
- opengl and gles also appear to be limited to not many parameters
     - since we are planning to remove opengl and gles, I'm simply excluding them from the new tests

otherwise, tested with matrix over:
- fields vs ndarray (see tests added)
- vulkan, cuda, cpu

Implementation notes:
- for ndarray, the args are actually sent to the kernel, as parameters
- for field, the args are not sent as parameters, but are baked into the kernel
     - for fields, the maximum number of snodes limits possible number of parameters, since limits total number of fields

Note that:
- modified ndarray to not give reaaallllyyy long stdout spam, on shutdown after crash, once per ndarray...
- modified ast_transformer_utils.py to not crash when line >= len(src), which for some reason was happening with num_args = 512

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_ndarray.py
+++ b/tests/python/test_ndarray.py
@@ -1164,8 +1164,7 @@ def test_real_func_write_ndarray_cfg():
 
 
 # exclude metal, because metal limited to < 30 parametrs AFAIK
-# exclude opengl and gles, because they crash
-@test_utils.test(exclude=[ti.opengl, ti.gles, ti.metal])
+@test_utils.test(exclude=[ti.metal])
 def test_ndarray_max_num_args() -> None:
     num_args = 512
     kernel_templ = """


taichi/inc/constants.h
```diff
@@ -9,7 +9,7 @@ constexpr int taichi_max_num_args = 8;
 // TODO: refine argument passing
 constexpr int taichi_max_num_args_total = 64;
 constexpr int taichi_max_num_args_extra = 32;
-constexpr int taichi_max_num_snodes = 1024;
+constexpr int taichi_max_num_snodes = 2048;
```
duburcqa, line 10: This change is not mentioned in the description of the PR and it seems important to me.

tests/python/test_field.py
```diff
@@ -439,3 +440,28 @@ def collide():
...
+@test_utils.test()
+def test_field_max_num_args() -> None:
+    num_args = 512
+    kernel_templ = """
+import taichi as ti
+@ti.kernel
+def my_kernel({args}) -> None:
+{arg_uses}
+"""
+    args_l = []
```
duburcqa, line 454: I like `args_l, arg_uses_l, arg_objs_l = [], [], []` more because they are all describing args, but as you wish.

taichi/inc/constants.h
hughperkins: It's alluded to here:
Screenshot 2025-07-28 at 2 21 43 PM

But I guess I can make it more explicit

hughperkins: added more to the description, in the new first paragraph "imlementation details"

PR #115, opened 2025-07-28
[Misc] Remove opengl and gles

hughperkins:
Issue: #

### Brief Summary

Remove opengl gles
- not used AFAIK (vulkan is used instead, except on very old cards)
- they both add code complexity, test failures; and they make tests run longer, cost more money

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/api_doc.yml
+++ b/.github/workflows/api_doc.yml
@@ -8,9 +8,9 @@ on:
         required: true
         type: string
     secrets:
-      AWS_ACCESS_KEY_ID:
+      AWS_STORAGE_ACCESS_KEY_ID:
         required: true
-      AWS_SECRET_ACCESS_KEY:
+      AWS_STORAGE_SECRET_ACCESS_KEY:
         required: true
   workflow_dispatch:
 concurrency:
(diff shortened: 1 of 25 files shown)


c_api/include/taichi/taichi_core.h
```diff
@@ -368,9 +366,9 @@ typedef enum TiArch {
   // Arm64 native CPU backend.
   TI_ARCH_ARM64 = 5,
   // OpenGL GPU backend.
-  TI_ARCH_OPENGL = 6,
+  // TI_ARCH_OPENGL = 6,
```
duburcqa: If you just comment them out, I would specify that these backend has been removed by PR#115

taichi/python/export_lang.cpp
```diff
@@ -1220,7 +1218,6 @@ void export_lang(py::module &m) {
   m.def("test_logging", [] { TI_INFO("test_logging"); });
   m.def("trigger_crash", [] { *(int *)(1) = 0; });
   m.def("get_max_num_indices", [] { return taichi_max_num_indices; });
-  m.def("get_max_num_args", [] { return taichi_max_num_args; });
```
erizmr, line 1223: just for more context, is this `get_max_num_args` also related to openGL or is it just we found this function not useful?

hughperkins: There was a comment in constants.h saying it is only used for opengl, so I removed it there. Then it refused to compile, because export_lang.cpp was exporting it, so I removed it here too.

hughperkins: Note that in https://github.com/Genesis-Embodied-AI/gstaichi/pull/114, I didnt' find I had to change any of the 'max args' constnats in cosntants.h, leading me to wonder if any of htem are used at all? (I tested on vulkan,cuda,cpu)

PR #118, opened 2025-07-29
[vulkan] Fix test_print on Vulkan

hughperkins:
Issue: #

### Brief Summary

Fix test_print on Vulkan

`print` on vulkan appears to prefix the output with `vkQueueSubmit()`. This PR filters this from the output.

~I fully expect to be asked to factorize out the output and expected comparison to some function `compare_print_output` or simialr, but at least creating the PR for now.~

Update: realized I can make this much more generic, and vulkan-independent, by simply truncating anything before the `match` string.

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
Thanks!

PR #120, opened 2025-07-31
[build] Enable building cpp tests in CI

hughperkins:
Issue: #

### Brief Summary

Enable building cpp tests in CI
- these were not being built till now
- and therefore were not being run either

Note taht this PR builds on https://github.com/Genesis-Embodied-AI/gstaichi/pull/119

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/gstaichi/compilation_manager/kernel_compilation_manager.cpp
+++ b/gstaichi/compilation_manager/kernel_compilation_manager.cpp
@@ -73,9 +73,9 @@ const CompiledKernelData &KernelCompilationManager::load_or_compile(
     const CompileConfig &compile_config,
     const DeviceCapabilityConfig &caps,
     const Kernel &kernel_def) {
-  auto cache_mode = get_cache_mode(compile_config, kernel_def);
+  auto cache_mode = get_cache_mode(compile_config, kernel_def.ir_is_ast());
   const auto kernel_key = make_kernel_key(compile_config, caps, kernel_def);
-  auto cached_kernel = try_load_cached_kernel(kernel_def, kernel_key,
+  auto cached_kernel = try_load_cached_kernel(kernel_def.get_name(), kernel_key,
                                               compile_config.arch, cache_mode);
   return cached_kernel ? *cached_kernel
                        : compile_and_cache_kernel(kernel_key, compile_config,
@@ -197,7 +197,7 @@ std::string KernelCompilationManager::make_kernel_key(
 }
 
 const CompiledKernelData *KernelCompilationManager::try_load_cached_kernel(
-    const Kernel &kernel_def,
+    const std::string &kernel_name,
     const std::string &kernel_key,
     Arch arch,
     CacheData::CacheMode cache_mode) {
@@ -206,7 +206,7 @@ const CompiledKernelData *KernelCompilationManager::try_load_cached_kernel(
     auto iter = kernels.find(kernel_key);
     if (iter != kernels.end()) {
       TI_DEBUG("Create kernel '{}' from in-memory cache (key='{}')",
-               kernel_def.get_name(), kernel_key);
+               kernel_name, kernel_key);
       return iter->second.compiled_kernel_data.get();
     }
   }
@@ -217,12 +217,12 @@ const CompiledKernelData *KernelCompilationManager::try_load_cached_kernel(
     if (iter != kernels.end()) {
       auto &k = iter->second;
       if (k.compiled_kernel_data) {
-        TI_DEBUG("Create kernel '{}' from cache (key='{}')",
-                 kernel_def.get_name(), kernel_key);
+        TI_DEBUG("Create kernel '{}' from cache (key='{}')", kernel_name,
+                 kernel_key);
         return k.compiled_kernel_data.get();
       } else if (auto loaded = load_ckd(kernel_key, arch)) {
-        TI_DEBUG("Create kernel '{}' from cache (key='{}')",
-                 kernel_def.get_name(), kernel_key);
+        TI_DEBUG("Create kernel '{}' from cache (key='{}')", kernel_name,
+                 kernel_key);
         TI_ASSERT(loaded->arch() == arch);
         k.last_used_at = std::time(nullptr);
         k.compiled_kernel_data = std::move(loaded);
@@ -239,7 +239,7 @@ const CompiledKernelData &KernelCompilationManager::compile_and_cache_kernel(
     const CompileConfig &compile_config,
     const DeviceCapabilityConfig &caps,
     const Kernel &kernel_def) {
-  auto cache_mode = get_cache_mode(compile_config, kernel_def);
+  auto cache_mode = get_cache_mode(compile_config, kernel_def.ir_is_ast());
   TI_DEBUG_IF(cache_mode == CacheData::MemAndDiskCache,
               "Cache kernel '{}' (key='{}')", kernel_def.get_name(),
               kernel_key);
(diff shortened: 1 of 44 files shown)


taichi/platform/cuda/detect_cuda.cpp
```diff
@@ -8,7 +8,22 @@ namespace taichi {
 
 bool is_cuda_api_available() {
 #if defined(TI_WITH_CUDA)
-  return lang::CUDADriver::get_instance_without_context().detected();
+  try {
+    auto &instance = lang::CUDADriver::get_instance_without_context();
+    int count;
+    instance.device_get_count(&count);
+    if(count <= 0) {
```
hughperkins: note that this check turned out to be the critical one, that allowed the c api tests to not seg fault, on a machien with no cuda devicse availbale

cmake/GsTaichiTargets.cmake
```diff
@@ -14,14 +14,6 @@ set(CMAKE_IMPORT_FILE_VERSION 1)
 set(_cmake_targets_defined "")
 set(_cmake_targets_not_defined "")
 set(_cmake_expected_targets "")
-foreach(_cmake_expected_target IN ITEMS gstaichi_c_api)
```
hughperkins, line 17: note to reviewers: this is some extraneous c-api related stuff, that we didnt delete before, when we removed c api.

cmake/GsTaichiTargets.cmake
```diff
@@ -51,42 +43,6 @@ if(_IMPORT_PREFIX STREQUAL "/")
   set(_IMPORT_PREFIX "")
 endif()
 
-# - [gstaichi_c_api] -------------------------------------------------------------
-
-add_library(gstaichi_c_api SHARED IMPORTED)
```
hughperkins, line 56: note to reviewers: this is some extraneous c-api related stuff, that we didnt delete before, when we removed c api.

gstaichi/platform/cuda/detect_cuda.cpp
```diff
@@ -8,7 +8,24 @@ namespace gstaichi {
 
 bool is_cuda_api_available() {
 #if defined(TI_WITH_CUDA)
-  return lang::CUDADriver::get_instance_without_context().detected();
+  try {
+    auto &instance = lang::CUDADriver::get_instance_without_context();
```
hughperkins, line 12: this was mostly associated with getting the c api to not seg fault tbh. But still, maybe good to be able to check for cuda devices without causing a seg fault?

python/.gitignore
```diff
@@ -1,10 +1,8 @@
-gstaichi/_lib/c_api
 gstaichi/_lib/runtime
 gstaichi/_lib/core/*.so
 gstaichi/_lib/core/*.pyd
 gstaichi.egg-info
 gstaichi/include
 gstaichi/assets
 gstaichi/tests
-gstaichi/tests38
```
hughperkins, line 9: since we removed tests38 folder, might as well remove this from gitignore (not really realted to this PR, but just happend to notice it au passage)

.github/workflows/scripts_new/linux_x86/4_test.sh
```diff
@@ -2,5 +2,8 @@
 
 set -ex
 
-pip3 install -r requirements_test.txt
-python3.10 tests/run_tests.py -v -r 3
+pip install -r requirements_test.txt
+TI_LIB_DIR=$(python -c "import os; import gstaichi as ti; p = os.path.join(ti.__path__[0], '_lib', 'runtime'); print(p)" | tail -n 1)
```
duburcqa: What about printing directly without temporary variable? 

And may add "_lib/runtime" in bash-scope?
```bash
TI_LIB_DIR="$(python -c 'import os; import gstaichi as ti; print(ti.__path__[0])' | tail -n 1)/_lib/runtime"
```

.github/workflows/scripts_new/macosx/4_test.sh
```diff
@@ -3,4 +3,9 @@
 set -ex
 
 pip install --prefer-binary -r requirements_test.txt
+find . -name '*.bc'
```
duburcqa, line 6: You already added this in part 3.

.github/workflows/scripts_new/macosx/4_test.sh
```diff
@@ -3,4 +3,9 @@
...
+ls -lh build/
+TI_LIB_DIR=$(python -c "import os; import gstaichi as ti; p = os.path.join(ti.__path__[0], '_lib', 'runtime'); print(p)" | tail -n 1)
```
duburcqa: Do you really need to do this systematically? This cannot be shared between parts?

Usually I prefer to export where I define but as you want.

gstaichi/platform/cuda/detect_cuda.cpp
duburcqa: This is nice I think.

tests/python/test_ndarray.py
```diff
@@ -1169,9 +1169,13 @@ def foo(
     assert (a[0] == vec3(3)).all()
 
 
+@pytest.mark.skipif()
 # exclude metal, because metal limited to < 30 parametrs AFAIK
 @test_utils.test(exclude=[ti.metal])
 def test_ndarray_max_num_args() -> None:
+    if platform.system() == "Darwin" and ti.lang.impl.current_cfg().arch == ti.vulkan:
+        # Mac doesn't support so many arguments, on Vulkan
+        return
```
duburcqa, line 1177: This is not the right pattern. Use `pytest.skip("Mac doesn't support so many arguments, on Vulkan")` instead of `return`.

.github/workflows/scripts_new/macosx/4_test.sh
CharlesMasson: If it helps, you may be able to use: `echo "TI_LIB_DIR=..." >> $GITHUB_ENV`

tests/python/test_ndarray.py
CharlesMasson: Specifically:

`@pytest.mark.skipif(sys.platform == "darwin" and ti.lang.impl.current_cfg().arch == ti.vulkan, reason="Mac doesn't support so many arguments, on Vulkan")`

hughperkins: @CharlesMasson  cannot use ti.[blah blah] in annotations, because ti has not been intiailized yet. Any checks we do have to be inside the function.

hughperkins: (this is also why I rarely use pytest.mark.parametrize for taichi tests, by the way, because things like `ti.ndarray` etc also need ti to beinitailzied first)

CharlesMasson: I see. Shouldn't we remove the decorator entirely then, and fully handle conditionally skipping within the function? Or does it still serve some purpose?

hughperkins: leaving aside the weird empty skipif outside the func, which is probably left over from purging opengl and gles, we currently are doing the check insdie the funciton. My understanding is that alexis is propsing I replace `return` with `pytest.skip`, running as a command, inside the function?

hughperkins: (actually, I've no idea why there is an empty skipif annotation, outside the function, probably should be removed...)

.github/workflows/scripts_new/linux_x86/4_test.sh
hughperkins: this actually is copied straight from the output from https://github.com/Genesis-Embodied-AI/gstaichi/blob/b0d25cb3b1eaac636f80687c1054978c0b9bb2ad/gstaichi/util/lang_util.cpp#L38-L43 , but sure.

.github/workflows/scripts_new/macosx/4_test.sh
hughperkins: what do you mean by 'shared between parts'? Which parts are you thinknig of sharing between?

hughperkins: (added export)

tests/python/test_ndarray.py
duburcqa: > I replace return with pytest.skip, running as a command, inside the function?

yes

hughperkins: cool. done that :)

PR #122, opened 2025-08-01
[vulkan] Fix exception for max ndarray args test on mac vulkan

hughperkins:
Issue: #

### Brief Summary

test_ndarray_max_num_args is failing on mac os x 14, for vulkan, eg https://github.com/Genesis-Embodied-AI/gstaichi/actions/runs/16660568675/job/47156487315?pr=113 , presumably because Mac OS X GPU cannot handle so many arguments.

This PR aims to add an exception to the test for mac vulkan.

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_ndarray.py
```diff
@@ -1173,6 +1174,9 @@ def test_ndarray_max_num_args() -> None:
 def my_kernel({args}) -> None:
 {arg_uses}
 """
+    if platform.system() == "Darwin" and ti.lang.impl.get_runtime().prog.config().arch == ti.vulkan:
+        # Mac doesn't support so many arguments, on Vulkan
```
YilingQiao, line 1178: do we use Mac + Vulkan? Is it a common combination?

hughperkins: Mac + Metal is also broken. I'm not sure if it's a Mac thing, or a Metal + Vulkan thing. For now, until proven otherwise, my hypothesis is taht the GPU on an M4 has a hardware limitatino on the number of arguments you can send into a kernel, and taht this affects both Metal and Vulkan.

hughperkins: Basically, we'll probably want to find another approach to py dataclasses with ndarray for Mac, at some point, but for now, on Mac, people can use py dataclasses with fields.

PR #129, opened 2025-08-05
[Build] Fix miscellaneous pypi publish issues

hughperkins:
Issue: #

### Brief Summary

Fix miscellaneous pypi publish issues

Builds were failing, and failing to bpulish
- publish to pypi was not being triggered
- test on mac was failing to run, becasue freetype was being built from source, and then failing
- we were only running on python 3.10, not 3.11, 3.12, 3.13, becuase we wanted to cut number of runs until publicly visible
- getting ambiguous name when downloading from s3, because multiple jobs were running against the same commit:
     - job for merging to main
     - job for releasing from main (from the same commit)

This PR:
- re-enable python 3.11-3.13, and mac 15
- continue to upload artifact to s3 (so that they are easily available to me to check, even when job fails), but use only the artifacts uploaded to github artifact storage, to avoid conflicts
- fix issue with python 3.12 and 3.13 test failing on mac, because freetype is being built from source
    - fix by removing matplotlib, which was used in examples, but no longer appears to be used
 - remove filter on 'created'/'published', which is redundant

Tested by:
- checking this PR builds:
    - tests that python 3.12 and 3.13 are now building ok
    - tests that the new approach to artifacts is working
- created a new pre-release, and checked the artifacts uploaded ok
     - https://pypi.org/project/gstaichi/0.1.25.dev0/#files
     - checks that publishing now correctly triggered again

copilot:summary

### Walkthrough

copilot:walkthrough


CharlesMasson (approved):
I think we could have had one PR per issue/fix to ease review.

PR #130, opened 2025-08-09
[Misc] Add ptx caching

hughperkins:
Issue: #

### Brief Summary

- remove llvm_offline_cache, which is no longer used
- refactorize offline_cache a bit, mostly changing names, and removing aliases, to hopefully facilitate readabilty
- add ptx_cache, in order to cache ptx, given optimized LLVM IR
- moved existing kernel_compilation_manager cache to a subdirectory `kernel_compilatino_manager`
    - and the new ptx cache goes in a peer subfolder `ptx_cache`

copilot:summary

### Walkthrough

copilot:walkthrough


gstaichi/compilation_manager/kernel_compilation_manager.cpp
```diff
@@ -9,10 +9,12 @@ namespace gstaichi::lang {
 
 namespace offline_cache {
 
+constexpr char kTiCacheFilenameExt[] = "tic";
+
 template <>
 struct CacheCleanerUtils<CacheData> {
   using MetadataType = CacheData;
-  using KernelMetaData = typename MetadataType::KernelMetadata;
+  using WrappedData = typename MetadataType::WrappedData;
```
duburcqa: What 'WrappedData' means in this context?

gstaichi/compilation_manager/kernel_compilation_manager.cpp
```diff
@@ -239,21 +245,22 @@ const CompiledKernelData &KernelCompilationManager::compile_and_cache_kernel(
...
+  k.metadata.created_at = k.metadata.last_used_at = std::time(nullptr);
 
   if (get_environ_config("TI_SHOW_COMPILING")) {
     TI_INFO("Compiling kernel '{}'", kernel_def.get_name());
   }
   k.compiled_kernel_data = compile_kernel(compile_config, caps, kernel_def);
-  k.size = 0;  // Populate `size` within the KernelCompilationManager::dump()
-  k.cache_mode = cache_mode;
+  k.metadata.size =
+      0;  // Populate `size` within the KernelCompilationManager::dump()
```
duburcqa: Maybe the comment should not be inlined if it does not fit on one line.

gstaichi/compilation_manager/kernel_compilation_manager.h
```diff
@@ -18,28 +18,30 @@ struct CacheData {
...
     std::size_t size{0};          // byte
     std::time_t created_at{0};    // sec
     std::time_t last_used_at{0};  // sec
 
     // Dump the kernel to disk if `cache_mode` == `MemAndDiskCache`
     CacheMode cache_mode{MemCache};
+    TI_IO_DEF(kernel_key, size, created_at, last_used_at);
+  };
 
+  struct WrappedData {
```
duburcqa: Not clear what this struct is

gstaichi/program/compile_config.cpp
```diff
@@ -72,7 +72,6 @@ void CompileConfig::fit() {
   if (arch_uses_spirv(arch)) {
     demote_dense_struct_fors = true;
   }
-  offline_cache::disable_offline_cache_if_needed(this);
```
duburcqa, line 75: Can this be removed?

gstaichi/program/program.cpp
```diff
@@ -131,6 +131,14 @@ Function *Program::create_function(const FunctionKey &func_key) {
   return functions_.back().get();
 }
 
+Kernel &Program::create_kernel(const std::function<void(Kernel *)> &body,
+                               const std::string &name,
+                               AutodiffMode autodiff_mode) {
+  auto func = std::make_unique<Kernel>(*this, body, name, autodiff_mode);
+  kernels.emplace_back(std::move(func));
```
duburcqa: `push_back` would do fine but no big deal.

gstaichi/program/program_impl.cpp
```diff
@@ -37,4 +37,14 @@ KernelLauncher &ProgramImpl::get_kernel_launcher() {
...
 }
 
+void ProgramImpl::register_needs_finalizing(NeedsFinalizing *needs_finalizing) {
+  need_finalizing_.push_back(needs_finalizing);
+}
+
+void ProgramImpl::run_need_finalizing() {
+  for (auto it = need_finalizing_.begin(); it != need_finalizing_.end(); it++) {
+    (*it)->finalize();
+  }
```
duburcqa: Why not:
```cpp
for (auto& node : need_finalizing_) {
    node.finalize();
}
```

gstaichi/python/export_misc.cpp
```diff
@@ -147,9 +146,6 @@ void export_misc(py::module &m) {
   m.def("with_vulkan", []() { return false; });
 #endif
 
-  m.def("clean_offline_cache_files",
-        lang::offline_cache::clean_offline_cache_files);
-
```
duburcqa, line 152: Why removing this?

gstaichi/runtime/cuda/CMakeLists.txt
```diff
@@ -5,12 +5,14 @@ target_sources(cuda_runtime
...
     jit_cuda.cpp
     kernel_launcher.cpp
+    ptx_cache.cpp
   )
 
 target_include_directories(cuda_runtime
   PRIVATE
     ${PROJECT_SOURCE_DIR}
     ${PROJECT_SOURCE_DIR}/external/eigen
+    ${PROJECT_SOURCE_DIR}/external/PicoSHA2
```
duburcqa, line 15: The standard hash library does not do the job? 
https://en.cppreference.com/w/cpp/utility/hash.html

gstaichi/runtime/cuda/ptx_cache.cpp
```diff
@@ -0,0 +1,249 @@
...
+std::string PtxCache::make_cache_key(const std::string &llvm_ir) const {
+  picosha2::hash256_one_by_one hasher;
+  std::string fast_math_str = compile_config_.fast_math ? "1" : "0";
+  hasher.process(fast_math_str.begin(), fast_math_str.end());
+  hasher.process(llvm_ir.begin(), llvm_ir.end());
+  hasher.finish();
+
+  auto res = picosha2::get_hash_hex_string(hasher);
+  res.insert(res.begin(), 'T');  // The key must start with a letter
+  return res;
```
duburcqa: Sounds like a job that could be done by the standard lib ?
https://en.cppreference.com/w/cpp/string/basic_string/hash

gstaichi/runtime/cuda/ptx_cache.cpp
```diff
@@ -0,0 +1,249 @@
...
+    auto &wrappedDataByKey = cached_all_data_.wrappedDataByKey;
+    auto iter = wrappedDataByKey.find(cache_key);
+    if (iter != wrappedDataByKey.end()) {
+      auto &k = iter->second;
+      TI_DEBUG("Found in cache (key='{}')", cache_key);
+      if (k.ptx.has_value()) {
+        return k.ptx;
+      }
+      // If the PTX is not in memory, try to load it from disk
+      std::string ptx = load_data_from_disk(cache_key);
```
duburcqa: How do you recover from failure?

gstaichi/runtime/cuda/ptx_cache.h
```diff
@@ -0,0 +1,97 @@
...
+  std::vector<WrappedPtx *> updated_data_;
+  const std::string cache_dir_;
+
+  // kernel_compilation_manager:
+  // using KernelCacheData = CacheData::WrappedData;
+  // using CachingKernels = std::unordered_map<std::string, KernelCacheData>;
+
+  // CachingKernels caching_kernels_;
+  // CacheData cached_data_;
+  // std::vector<KernelCacheData *> updated_data_;
```
duburcqa: What are these comments?

gstaichi/runtime/cuda/ptx_cache.h
```diff
@@ -0,0 +1,97 @@
...
+  std::string load_data_from_disk(const std::string &cache_key);
+
+  const Config config_;
+  const CompileConfig &compile_config_;
+  // using WrappedData = WrappedPtx;
+
+  std::unordered_map<std::string, WrappedPtx>
+      wrapped_by_key_;  // caching_kernels_ in kernel_compilation_manager
+  PtxCacheAllData
+      cached_all_data_;  // cached_data_ in kernel_compilation_manager
```
duburcqa: Not a fan of inline comments when they do not fit on the line.

gstaichi/util/offline_cache.h
```diff
@@ -116,42 +107,52 @@ struct CacheCleanerConfig {
...
   // To remove other files except cache files and offline cache metadta files
   static void remove_other_files(const CacheCleanerConfig &config) {
-    TI_NOT_IMPLEMENTED;
   }
 
   // To check if a file is cache file
   static bool is_valid_cache_file(const CacheCleanerConfig &config,
                                   const std::string &name) {
-    TI_NOT_IMPLEMENTED;
+    return true;
```
duburcqa, line 148: Always valid?

tests/python/test_offline_cache.py
```diff
@@ -1,6 +1,7 @@
 import atexit
 import functools
 import math
+import os
 import shutil
 import threading
 from os import listdir, rmdir, stat
```
duburcqa: Using `os.` everywhere would be better.

tests/python/test_offline_cache.py
```diff
@@ -56,9 +60,18 @@ def current_thread_ext_options():
...
 
 
-def cache_files_cnt():
+def cache_files_cnt(folder: str | None = None) -> int:
+    if folder is None:
+        folder = tmp_offline_cache_file_path()
     try:
-        return len(listdir(tmp_offline_cache_file_path()))
+        count = 0
+        for file in os.listdir(folder):
```
duburcqa: I would recommend moving to `pathlib.Path`, and use recursive walk instead of doing recursion manually.

tests/python/test_offline_cache.py
```diff
@@ -31,8 +32,11 @@ def cache_files_size(path):
     files = listdir(path)
     result = 0
     for file in files:
-        if is_offline_cache_file(file):
-            result += stat(join(path, file)).st_size
+        filepath = os.path.join(path, file)
+        if os.path.isdir(filepath):
+            result += cache_files_size(filepath)
+        if is_offline_cache_file(filepath):
+            result += stat(filepath).st_size
```
duburcqa: Same: I would recommend moving to `pathlib.Path`, and use recursive walk instead of doing recursion manually.

gstaichi/compilation_manager/kernel_compilation_manager.cpp
hughperkins: the thing we are caching. Happy to rename somehow if you prefer?

gstaichi/compilation_manager/kernel_compilation_manager.h
hughperkins: Metadata plus the thing we are caching. Happy to rename if you rpefer?

gstaichi/program/compile_config.cpp
hughperkins: I searched the code base and nothing is using it.

gstaichi/util/offline_cache.h
hughperkins: Sure, why not? If want to specify something then can override it. Rather than having to override it.

gstaichi/program/program.cpp
hughperkins: what is the advatnage fo push_back vs emplace_back?

hughperkins: (out of personal curiosity, so I can learn :) )

gstaichi/program/program_impl.cpp
hughperkins: Thanks!

gstaichi/python/export_misc.cpp
hughperkins: not used anywhere inthe code base.

To be fair, it *used* to be used, before we removed cli.

I'm very tempted to put back jsut the cache cleaning bit of the CLI. Thoughts?

gstaichi/runtime/cuda/CMakeLists.txt
hughperkins: std::hash is only 32 bits, more likely to collide:

Screenshot 2025-08-12 at 11 25 12 AM

gstaichi/runtime/cuda/ptx_cache.cpp
hughperkins: see above

gstaichi/runtime/cuda/CMakeLists.txt
hughperkins: (remember that in a standard hashmap and similar, it's ok for values to collide, since they are added to vectors, and checked sequentially. But here, we are assuming that each hash is unique.)

hughperkins: (not actually vecotr,s as in std::vector, but some kind of conceptual list or similar)

gstaichi/runtime/cuda/ptx_cache.cpp
hughperkins: Not sure 🤔. The existing kernel_compilation_manager.cpp also does not have any try... catch. Nor program.cpp that calls it. Nor export_lang.cpp. Perhaps it just manifests as an exception at userspace level? 🤔

gstaichi/runtime/cuda/ptx_cache.h
hughperkins: removed

gstaichi/compilation_manager/kernel_compilation_manager.cpp
duburcqa: What about `CacheData` in that case? Maybe I'm being to naive here x)

gstaichi/program/program.cpp
duburcqa: `emplace_back` is more advanced and tricky compared to `push_back`, so I prefer using `emplace_back` only for the use-case it allows that are not supported by `push_back` or requires extra steps, i.e. "in-place" object construction via perfect forwarding, or moving rvalue automatically for classes not supporting implicit constructor (e.g. `bufferPosition_.emplace_back(telemetry_data_pair_t<Scalar>{valuePtr, positionInBuffer});`). Otherwise one may wonder whether there is some trick behind using `emplace_back`.

gstaichi/python/export_misc.cpp
duburcqa: As you want. I never clean cache. I rather specify some custom temporary folder instead or disable cache completely if necessary.

gstaichi/compilation_manager/kernel_compilation_manager.cpp
hughperkins: CacheData seems reasonable to me. I can cahnge that.

gstaichi/runtime/cuda/CMakeLists.txt
duburcqa: Wait, SHA-256 is also 32 bits.

Is collision really an issue? Do you think there is any likelihood for hash to collide to the point that an outdated cache is being used? It sounds quite unlikely. But no problem with using a more advanced library if really needed.

gstaichi/runtime/cuda/ptx_cache.cpp
duburcqa: This is weird to raise an exception in user space for not being able to load a corrupted cache. As a user I don't care about your cache and I don't want to know anything about it. It should gracefully recover without my help, as we are doing in Genesis simulator currently.

gstaichi/util/offline_cache.h
duburcqa: Not convinced by as you wish.

gstaichi/python/export_misc.cpp
hughperkins: Yeah me too. But I'm not a user.

@zswang666 what are your thoughts/requirements/approaches for cache cleaning, if any?

duburcqa:
One last comment about exception handling that I think is worth discussing before merging.

gstaichi/runtime/cuda/CMakeLists.txt
hughperkins: Yeah, 1 in 4 billion chance is not that low. Occasional collisions will cause weird hard-to-track-down bugs. The hashing is near instantaneous in my benchmarks to date, compard with the python, which is not near instantaneous at all.

hughperkins: sha 256 is 256 bits:

hughperkins: Screenshot 2025-08-12 at 2 30 17 PM

gstaichi/runtime/cuda/ptx_cache.cpp
hughperkins: I mean, I agree, but empirically, this is being used for the existing cache, and I have seen zero issues logged in 2 years on the upstream taichi repo for this, as far as I remember.

hughperkins: Updated to no longer throw exceptions, but return nullopt instead.

But, something that occurs to me... shouldn't this PR have some tests? 🤔

gstaichi/python/export_misc.cpp
zswang666: I've never cleaned cache for taichi-level stuffs. I may only clean cache at Genesis level

hughperkins: Interesting. @duburcqa What does 'clean cache at genesis level' imply as far as Taichi functionality?

gstaichi/compilation_manager/kernel_compilation_manager.cpp
duburcqa: Yay :)

gasnica: -- oh reading this again now -- "CacheData" is already the name of this template's param ??

Anyways, my 2 cents -- what is the meaning of WrappedData -- is it the data only, or data + metadata wrapper? Cos the name is ambiguous ..

gstaichi/compilation_manager/kernel_compilation_manager.cpp
```diff
@@ -30,9 +32,9 @@ struct CacheCleanerUtils<CacheData> {
   // To get cache files name
   static std::vector<std::string> get_cache_files(
       const CacheCleanerConfig &config,
-      const KernelMetaData &kernel_meta) {
+      const WrappedData &wrapped_data) {
     auto fn = fmt::format(KernelCompilationManager::kCacheFilenameFormat,
-                          kernel_meta.kernel_key);
+                          wrapped_data.metadata.kernel_key);
```
gasnica, line 37: just an opinion:
aren't we overusing data in wrapped_data.metadata? could we have some thing like data.descr or data.info or data.meta instead?

Nothing key for this PR though

gstaichi/compilation_manager/kernel_compilation_manager.cpp
```diff
@@ -103,44 +109,45 @@ void KernelCompilationManager::dump() {
   data.version[0] = TI_VERSION_MAJOR;
   data.version[1] = TI_VERSION_MINOR;
   data.version[2] = TI_VERSION_PATCH;
-  auto &kernels = data.kernels;
+  auto &wrappedDataByKey = data.wrappedDataByKey;
```
gasnica: this is a std::*map -- it's always byKey. do we want to make it more specific, or simplify the name. wrappedDatas or wrappedDataMap or cacheKeyToWrappedData?

gstaichi/runtime/cuda/CMakeLists.txt
duburcqa: > The hashing is near instantaneous in my benchmarks to date, compard with the python, which is not near instantaneous at all.

Yeah, it is not about speed / performance, but avoiding additional dependencies that adds up to the maintenance burden if not legitimate.

> sha 256 is 256 bits:

Oh really!? I thought standard hash library was 32 BYTES not 32 BITS, that is why I was confused. Indeed 32 bits sounds fairly small... 

Let's use picosha256 in that case.

gstaichi/compilation_manager/kernel_compilation_manager.cpp
```diff
@@ -103,44 +109,45 @@ void KernelCompilationManager::dump() {
...
   // Load old cached data
   offline_cache::load_metadata_with_checking(data, filepath);
   // Update the cached data
   for (const auto *e : updated_data_) {
-    auto iter = kernels.find(e->kernel_key);
-    if (iter != kernels.end()) {
-      iter->second.last_used_at = e->last_used_at;
+    auto iter = wrappedDataByKey.find(e->metadata.kernel_key);
```
gasnica: are we going to rename kernel_key to something else, given we're trying to make it typeless now?

gstaichi/compilation_manager/kernel_compilation_manager.cpp
gasnica: or can we just call it 'cache'? given old name is 'kernels' ?

gstaichi/python/export_misc.cpp
duburcqa: > What does 'clean cache at genesis level' imply as far as Taichi functionality?

It does not affects taichi at all. These two caches are completely distinct.

gstaichi/compilation_manager/kernel_compilation_manager.cpp
duburcqa: @gasnica In this context, `CacheCleanerUtils<CacheData>` should probably by replaced by `CacheCleanerUtils<CachedData>` actually, or something like that, from what I understand.

gasnica: oh, regarding my last line above -- maybe a less ambiguous name is DataWrapper?

gstaichi/program/program_impl.h
```diff
@@ -32,6 +32,11 @@ class ProgramImpl {
  public:
   explicit ProgramImpl(CompileConfig &config);
 
+  class NeedsFinalizing {
+   public:
```
gasnica, line 36: add a default virtual destructor for an interface ?

gstaichi/compilation_manager/kernel_compilation_manager.cpp
gasnica: .. reading a bit more so it's a CacheData (which is the entire cache manager, afaiu) that contained a map of KernelData. So maybe just renaming it to ItemData for it to be typeless and indicate it's data for a single item.

gstaichi/compilation_manager/kernel_compilation_manager.h
```diff
@@ -18,28 +18,30 @@ struct CacheData {
   };
   using Version = std::uint16_t[3];
 
-  struct KernelData {
+  struct Metadata {
```
gasnica, line 21: I suggest against this name cos we have also kMetadataFilename and such, and this name is overloaded. 

Maybe ItemDesc(ription). 

But yeah, that's just a suggestion.

gstaichi/runtime/cuda/CMakeLists.txt
gasnica: What, whaat ?? I'm confused --- 

Are you saying that if there's a key-collision we'll use a wrong/stale cached data? -- Then it needs to be fixed no matter what probabality, no?

hughperkins: 2^256 is a really big number. Larger than we can comprehend. Still we can do some maths on it.


2^256 is 10^77, ie 1e77.

Lets say we roll a dice every time the clock on your computer ticks. Your computer clock runs at ~3GHz. 3 billion dice rolls a second. Thats 3e9. So, if you roll a number every clock tick of your computer, 1e77 is the following number of seconds:

- 1e77 / 3e9 ~= 3e68

Lets roll those dice for 100 years. How many rolls can we make in 100 years?

3e9 * 60 * 60  * 24 * 365 * 100
= 1e19

Let's say we are trying to hit any of one million possible numbers. So, we'll aproximate that by rolling one million times every clock tick:

1e19 * 1e6 = 1e25

So, if you roll one million times, every clock tick, for 100 years, the chance of a colliion are:

1 in 1e77 / 1e25 = 1 in 1e52

It's still an astronomically tiny probaiblty of happening.

You're more likely to wake up, and that computer program you wrote 10 years ago, and threw on the internet, suddenly went viral overnight, and you became a billionaire, like the Minecraft guy. Waaayyyy more likely. Way way way more likely. Like there's only 6 billoon people in the world. 1 in 1e9.

gstaichi/runtime/cuda/ptx_cache.cpp
```diff
@@ -0,0 +1,249 @@
+// This was copy and hacked from kernel_compilation_manager.cpp
+// It is used to manage the compilation and caching of PTX kernels
```
gasnica, line 2: So with entire ptx_cache basing on copy-pasted kernel_compilation_manager, what is the plan forward. Are we going to remove the latter soon?

gstaichi/compilation_manager/kernel_compilation_manager.cpp
hughperkins: lean slightly towards DataWrapper, since that is what it is I feel. @duburcqa thoughts?

gstaichi/program/program.cpp
hughperkins: updated, though I just realized it's not actually part of the PR. Not sure why it's showing that it is. Maybe I moved the implementation from the .h file? Anyway, updated to push_back.

gstaichi/python/export_misc.cpp
hughperkins: Alright, does anyone have an opinion between these. two options? :
1. put back CLI, just for cache control, put back this function
2. leave CLI deleted, leave this function deleted

We can always undelete stuff later if people ask for it back, so I lean slighlty towards 2.

Thougths?

gstaichi/util/offline_cache.h
hughperkins: Basically I want to have as much boilerplate as possible in the base class ("offline_cache.cpp/.h"), nad keep the derived classes as simple as possible.

gstaichi/compilation_manager/kernel_compilation_manager.cpp
hughperkins: I'm not sure I follow. DataWrapper contains:
- the data
- the metadata

"meta" is greek meaning "about stuff". So "metadata" is "data about data". Metadata is a type of data.

gstaichi/compilation_manager/kernel_compilation_manager.cpp
hughperkins: "Key" here means the "CacheKey". I can write out "CacheKey" in full though if you prefer?

gstaichi/compilation_manager/kernel_compilation_manager.cpp
hughperkins: kernel_compilation_manager.cpp is not part of this PR. I just modified things where I had to so the offline_cache was clean enough that I could make ptx_cache. But the kernel_key is clearly kernel_compilation_manager specific if I remember correctly?

gstaichi/compilation_manager/kernel_compilation_manager.h
hughperkins: Not sure I follow. This struct contains the metadata for the cache entry. And kMetadatFilename is used to store this metadta. In what way do you feel this name is overloaded, and is being used for two different things?

gstaichi/runtime/cuda/ptx_cache.cpp
hughperkins: why would we remove kernel compilation manager? They are two different caches. kernel compilation manager caches LLVM IR. ptx cache caches PTX.

gstaichi/util/offline_cache.h
duburcqa: Yes, but it should not increase the likelihood of "faulty" implementation. I prefer having to explicitly implement this kind of method in derived classes, just to make sure not overwriting the base implementation is not a mistake.

gstaichi/python/export_misc.cpp
duburcqa: I prefer 2.

gstaichi/compilation_manager/kernel_compilation_manager.cpp
duburcqa: Go for it, if it is what it is.

gstaichi/runtime/cuda/ptx_cache.cpp
duburcqa: > Updated to no longer throw exceptions, but return nullopt instead.

This is great, thank you!

> But, something that occurs to me... shouldn't this PR have some tests?

I would appreciate a unit test that check what it does for a valid plus a deliberately corrupt a cache file. But the new implementation looks safe already.

duburcqa (approved):

gstaichi/runtime/cuda/CMakeLists.txt
gasnica: Great. Got also confused above about it being 32-bit .. 

So what happens when there's a conflict? We just waste some time, and eventually purge the cache, and then it works?

gstaichi/compilation_manager/kernel_compilation_manager.cpp
gasnica: I'd defintely prefer that in general. But if it's used a lot, then you decide whether to make it longer.

Also, Is it going to be dataWrappersByCachedKey now?

gstaichi/runtime/cuda/ptx_cache.cpp
gasnica: Oh, can we then avoid-copy pasting code, and make a generic and reusable base implementation? We don't want code duplication, in general, right?

gstaichi/runtime/cuda/CMakeLists.txt
hughperkins: Statistically, it's more likely that your next flight to new york will take a detour via the moon, and you can pick up some moonrocks as souvenirs.

One possible flaw in this is the quality of the hash function, ie it might not be distributed uniformly. But 10^77 is such a huge number that I feel that's unlikely.

But to answer your question: if there is a collision, then yes, the ptx will crash repeatedly, until we added the possiblity of storing multiple ptx files for a single cache value.

This is why we are not going to use 32-bit hashing. We are adding an addiitonal ~220 bits. Thats a lootttt.

gasnica (approved):
approving, to not stall. 

concerned about copy-pasting of the cache functionality, instead of making it generic-enough and reusing it.

gstaichi/runtime/cuda/ptx_cache.cpp
hughperkins: Yup, but not in this PR.

gstaichi/compilation_manager/kernel_compilation_manager.cpp
hughperkins: `CacheKey`, without the `d`. And `dataWrapper`, withou the `s`. But yes.

hughperkins:
Thanks! I might add a cpp unit test or two before merging 🤔

hughperkins:
Added c++ unit test for ptx cache
- note that it doesn't run currently, until we merge https://github.com/Genesis-Embodied-AI/gstaichi/pull/120

Example output for running it manually:
Screenshot 2025-08-15 at 9 11 28 AM

Since I was curious about those things in yellow, I temporarily added a std::cout for each new ptxcache instance:

hughperkins:
Can you both review the unit test please?

hughperkins:
Hmmm, I guess this used to "pass" because we weren't running c++ tests...

tests/cpp/runtime/cuda/ptx_cache_test.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "gstaichi/runtime/cuda/ptx_cache.h"
+
+namespace gstaichi::lang {
+
+TEST(PtxCache, TestBasic) {
+  // char temp_dir[] = "/tmp/test_cache_path_XXXXXX";
+  // if (!mkdtemp(temp_dir)) {
+  //   FAIL() << "Failed to create temporary directory";
+  // }
+
```
duburcqa: You can remove this at this point.

tests/cpp/runtime/cuda/ptx_cache_test.cpp
```diff
@@ -0,0 +1,85 @@
...
+  auto temp_base = std::filesystem::temp_directory_path();
+  auto now = std::chrono::system_clock::now().time_since_epoch().count();
+  std::mt19937_64 rng(now);
+  std::uniform_int_distribution<uint64_t> dist;
+  std::string unique_name = "test_cache_path_" + std::to_string(dist(rng));
```
duburcqa: Using a random number generator does not bring any benefit here since if the time is the same, the value being generated while be the same anyway. I would recommend a non-obfuscated yet more robust pattern:
```cpp
const testing::TestInfo* const test_info = testing::UnitTest::GetInstance()->current_test_info();

auto now = std::chrono::system_clock::now();
auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(now.time_since_epoch()) % 1000;
auto in_time_t = std::chrono::system_clock::to_time_t(now);
std::tm* bt = std::localtime(&in_time_t);
std::ostringstream time_str;
time_str << std::put_time(bt, "%H_%M_%S") << '_' << std::setfill('0') << std::setw(3) << ms.count();

std::string unique_name = "test_cache_path_" + test_info->test_suite_name() + "_" + test_info->name() + "_" + time_str.str();
```

Maybe this snippet could be turned into some "fixture" that you can put in the "header" of each unit test to create isolated cache whenever it is relevant.

tests/cpp/runtime/cuda/ptx_cache_test.cpp
```diff
@@ -0,0 +1,85 @@
...
+    FAIL() << "Failed to create temporary directory";
+  }
+  auto cleanup = [temp_dir]() { std::filesystem::remove_all(temp_dir); };
+
+  struct DirCleaner {
+    std::function<void()> cleaner;
+    ~DirCleaner() {
+      cleaner();
+    }
+  } dir_cleaner{cleanup};
```
duburcqa, line 22: What is the point? Not sure it is any better than calling `std::filesystem::remove_all(temp_dir);` at the end of the test.

tests/cpp/runtime/cuda/ptx_cache_test.cpp
```diff
@@ -0,0 +1,85 @@
...
+  std::string ptx_code2 = "ptx code for test kernel2";
+
+  std::string key_1 = ptx_cache->make_cache_key(llvm_ir1, false);
+  std::string key_1fast = ptx_cache->make_cache_key(llvm_ir1, true);
+  std::string key_2 = ptx_cache->make_cache_key(llvm_ir2, false);
+
+  ASSERT_NE(key_1, key_2);
+  ASSERT_NE(key_1, key_1fast);
+
+  ASSERT_EQ(std::nullopt, ptx_cache->load_ptx(key_1));
```
duburcqa, line 46: There is no way to check if the file has been corrupted? Could you replace the content of the file by some binary data or something? 

But I guess refactoring your implementation to add some proper sanity check that would be able to detect cache tampering, typically by prepending the content with its own hash, would be quite involving on your side

tests/python/test_offline_cache.py
```diff
@@ -43,22 +37,28 @@ def expected_num_cache_files(num_kernels: int = 0) -> int:
     return num_kernels + 1
 
 
-def tmp_offline_cache_file_path():
-    return join(OFFLINE_CACHE_TEMP_DIR, str(threading.current_thread().ident))
+def tmp_offline_cache_file_path() -> pathlib.Path:
+    return OFFLINE_CACHE_TEMP_DIR / str(threading.current_thread().ident)
```
duburcqa: This is not the right way to implement temporary directory with pytest. You should just use what pytest is already providing, i.e. `tmp_path`. This is already "thread safe", in a sense that there would be no collision for unit tests running in parallel. 

If you need a temporary directory that would be share among all the unit tests, then you can use the more generic `tmp_path_factory` fixture. You could take inspiration from what I did for Genesis (see `asset_tmp_path`). I'm using both options depending on the application.

tests/python/test_offline_cache.py
```diff
@@ -43,22 +37,28 @@ def expected_num_cache_files(num_kernels: int = 0) -> int:
...
+        folder = tmp_offline_cache_file_path()
     try:
-        return len(listdir(tmp_offline_cache_file_path()))
+        count = 0
+        for filepath in folder.rglob("*"):
+            if filepath.is_file():
+                count += 1
+        return count
     except FileNotFoundError:
         return 0
```
duburcqa, line 67: This should be moved in some `utils.py` to share boilerplate and focus only on implementing actual unit tests in `test_*.py` files.

tests/python/test_offline_cache.py
```diff
@@ -43,22 +37,28 @@ def expected_num_cache_files(num_kernels: int = 0) -> int:
...
-    return join(OFFLINE_CACHE_TEMP_DIR, str(threading.current_thread().ident))
+def tmp_offline_cache_file_path() -> pathlib.Path:
+    return OFFLINE_CACHE_TEMP_DIR / str(threading.current_thread().ident)
 
 
 def current_thread_ext_options():
     return {
         "offline_cache": True,
-        "offline_cache_file_path": tmp_offline_cache_file_path(),
+        "offline_cache_file_path": str(tmp_offline_cache_file_path()),
```
duburcqa: `tmp_offline_cache_file_path()` should probably return a string from the start.

tests/cpp/runtime/cuda/ptx_cache_test.cpp
hughperkins: simplified a bunch

tests/cpp/runtime/cuda/ptx_cache_test.cpp
hughperkins: I'm not testing if the file has been corrupted. If the file has been corrupted, then the program will crash, and we'll need to wipe the cache. The existing cache does not do such a check, https://github.com/taichi-dev/taichi/blob/ba0e81dce559fb63a5958bf82feb1d00c55c02fe/taichi/compilation_manager/kernel_compilation_manager.cpp#L256-L277 . And there are no github issues concerning such crashes on the existing cache, afaik.

tests/python/test_offline_cache.py
hughperkins: This isnt my code. I only updated it to pathlib. Other than the migraiton to pathlib, It is out of scope for this PR I feel.

tests/python/test_offline_cache.py
hughperkins: This method is only use by test_offline_cache.py I feel. I would prefer to keep it next to the funcitons that are using it, on the whole.

tests/python/test_offline_cache.py
hughperkins: It's returning a pathlib.Path. My understanding is taht you feel that using pathlib.Path is a good approach for paths? I'd prefer to standardize on the whole: either all paths are paths, or all paths are strings. Otherwise it gets confusing knowing which is which I feel.

tests/cpp/runtime/cuda/ptx_cache_test.cpp
hughperkins: It ensures it gets cleaned up even if the test fails/exits early. At least: thats' the intent.

duburcqa: Oh I see! this is nice. I missed that point.

tests/cpp/runtime/cuda/ptx_cache_test.cpp
duburcqa: > I'm not testing if the file has been corrupted. If the file has been corrupted, then the program will crash, and we'll need to wipe the cache. 

Ok I see. I can assure you that automatically ignoring corrupted cache files was very much necessary for genesis. This kind of issue was occurring very frequently, sometimes related to class name changes of stuff like this (because pickle will keep track of the original module path and re-import them on the fly instead of dumping the entire class implementation for efficiency, but then crash if not available anymore). 

Maybe it is a non-issue in your case. But in your case, it may be tricky to identify the root cause of the crash, then come up to the conclusion that cache must be cleaned. What about mentioning in some troubleshooting section of the documentation that one should always clean cache in case of segfault to make sure that corrupted cache is not the root cause?

tests/python/test_offline_cache.py
duburcqa: Ok. But we should keep track of this issue, because this design pattern is wrong.

tests/python/test_offline_cache.py
duburcqa: I get your point. Not very convinced but ok.

tests/python/test_offline_cache.py
duburcqa: Make sense. Let's do this.

tests/python/test_offline_cache.py
hughperkins: https://linear.app/genesis-ai-company/issue/CMP-108/fix-up-design-pattern-in-offline-cache-testcpp

tests/cpp/runtime/cuda/ptx_cache_test.cpp
hughperkins: Added troubleshooting doc, with section on seg faults.

gstaichi/util/offline_cache.h
hughperkins: So the issue is that if the "base class" is just a huge interface, that each "derived class" has to implement, there is no re-use: all the implementation details live in the child classes.

I feel like the child classes for these caches could be considerably smaller, contain currently a lot of code that is common across all caches.

The way this template is constructed currently, you have to either override the entire class, or none of it. There is no in between. So, unless I make it possible for someone to override none of it, currently, all the child classes end up full of implmenetation, which I'm trying to change.

Now maybe this template should be changed so that we don't have 'all or nothing' in the child classes, but in the short-term, this is the reason why I'm trying to make it possibel to not override it.

duburcqa (approved):

hughperkins:
Thanks!

PR #131, opened 2025-08-12
[Misc] Add SRC-LL caching to accelerate warm cache load time

hughperkins:
Issue: #

### Brief Summary

Add SRC-LL caching to accelerate warm cache load time.

We'll hash the source-code, and use that to cache the optimized LLVM-IR.

Manual testing:

Using https://github.com/Genesis-Embodied-AI/Genesis/compare/main...hughperkins:Genesis:hp/add-compilation-timer

Running like:

```
TI_LOG_LEVEL=trace python examples/rigid/single_franka_envs.py -n 2 --horizon 2
```

Looking at the 'compilation time' output:
- Without this PR (and without ptx caching either): 7.1s
- With this PR (but without ptx caching): 2.9s
- With this PR, and with ptx-caching PR too (https://github.com/Genesis-Embodied-AI/gstaichi/pull/130): 0.31s

(Running on a ubuntu box with a 5090 GPU)

copilot:summary

### Walkthrough

copilot:walkthrough


duburcqa:
Could you add some benchmarks in the description of the PR?

gstaichi/program/program.cpp
```diff
@@ -131,6 +149,16 @@ Function *Program::create_function(const FunctionKey &func_key) {
   return functions_.back().get();
 }
 
+Kernel &Program::create_kernel(const std::function<void(Kernel *)> &body,
+                               const std::string &name,
+                               AutodiffMode autodiff_mode) {
+  // Expr::set_allow_store(true);
+  auto func = std::make_unique<Kernel>(*this, body, name, autodiff_mode);
+  // Expr::set_allow_store(false);
```
duburcqa: Add these comments relevant?

gstaichi/program/program.h
```diff
@@ -113,15 +113,22 @@ class TI_DLL_EXPORT Program {
...
-    // Expr::set_allow_store(false);
-    kernels.emplace_back(std::move(func));
-    return *kernels.back();
-  }
+  void dump_cache_data_to_disk();
+
+  void store_fast_cache(const std::string &checksum,
+                        const Kernel &kernel,
+                        const CompileConfig &compile_config,
+                        const DeviceCapabilityConfig &caps,
```
duburcqa: I would prefer `device_caps` but as you wish.

gstaichi/program/program.h
```diff
@@ -113,15 +113,22 @@ class TI_DLL_EXPORT Program {
...
+                        CompiledKernelData &ckd);
```
duburcqa: Not a great fun of this variable name. Too cryptic.

python/gstaichi/_test_tools/__init__.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any
+
+import gstaichi as ti
+
+
+def ti_init_same_arch(options: dict[str, Any] | None = None) -> None:
```
duburcqa: What does this method do?

This method would fail if `arch` is specified in options already. Is this expected? If not, you should rather do:
```
ti.init(**{**options, arch=getattr(ti, ti.cfg.arch.name)})
```
Note that the order matters here.

python/gstaichi/lang/_fast_caching/__init__.py
```diff
@@ -0,0 +1 @@
+FIELD_METADATA_CACHE_VALUE = "add_value_to_cache_key"
```
duburcqa: I would (somehow strongly) recommend not define global variables in `__init__.py` files, exception maybe in the top-level init file of the entire Python project.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -0,0 +1,99 @@
...
+import time
+from typing import Any, Sequence
+
+import numpy as np
+
+from gstaichi.lang._fast_caching import FIELD_METADATA_CACHE_VALUE
+from gstaichi.lang._ndarray import ScalarNdarray
+from gstaichi.lang.field import ScalarField
+from gstaichi.lang.matrix import MatrixField, MatrixNdarray, VectorNdarray
+from gstaichi.lang.util import is_data_oriented
```
duburcqa: I recommend relative import systematically. It clarifies at first glance what are internal dependencies and third-parties. It also helps to avoid circular dependencies and stresses out design flaws with convoluted import schemes.

But as you wish.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -0,0 +1,99 @@
...
+        repr_l.append(full_repr)
+    return "[" + ",".join(repr_l) + "]"
+
+
+def to_representative_str(arg: Any) -> str | None:
+    """
+    string should somehow represent the type of arg. Doesnt have to be hashed, nor does it have
+    to be the actual python type string, just something representative of the type, and won't collide
+    with different (allowed) types
+    """
```
duburcqa: Formatting could be improved. Short description of 1 sentence, skip one line, and details. Starting with uppercase and ending with dot.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -0,0 +1,99 @@
...
+        child_value = getattr(arg, field.name)
+        _repr = to_representative_str(child_value)
+        full_repr = f"{field.name}: ({_repr})"
+        if field.metadata.get(FIELD_METADATA_CACHE_VALUE, False):
+            full_repr += f" = {child_value}"
+        repr_l.append(full_repr)
+    return "[" + ",".join(repr_l) + "]"
+
+
+def to_representative_str(arg: Any) -> str | None:
```
duburcqa: What about `stringify_type` or `stringify_any_type` ?

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -0,0 +1,99 @@
...
+from .hash_utils import hash_string
+
+g_num_calls = 0
+g_num_args = 0
+g_hashing_time = 0
+g_repr_time = 0
+g_num_ignored_calls = 0
+
+
+def dataclass_to_repr(arg: Any) -> str:
```
duburcqa, line 24: Why is this necessary? This is not handled by `to_representative_str` already?

python/gstaichi/lang/_fast_caching/config_hasher.py
```diff
@@ -0,0 +1,16 @@
+from gstaichi.lang import impl
+
+from .hash_utils import hash_string
+
+
+def hash_compile_config() -> str:
```
duburcqa, line 8: One line docstring would be great for a "public" method.

python/gstaichi/lang/_fast_caching/fast_caching_types.py
```diff
@@ -0,0 +1,8 @@
+from pydantic import BaseModel
+
+from gstaichi.lang._wrap_inspect import FunctionSourceInfo
+
+
+class HashedFunctionSourceInfo(BaseModel):
```
duburcqa, line 6: Add one line docstring, and maybe one sentence to explain why this class is useful?

hughperkins:
> Could you add some benchmarks in the description of the PR?

Added. To what extent is that what you are looking for?

python/gstaichi/lang/_fast_caching/hash_utils.py
```diff
@@ -0,0 +1,5 @@
+import hashlib
+
+
+def hash_string(v: str) -> str:
+    return hashlib.sha256(v.encode("utf-8")).hexdigest()
```
duburcqa: From what I understand, you don't have to concatenate the entire string.  Moreover, I'm not sure if you need cryptographic feature. What about just:
```python
def hash_string(data: Iterable[str]) -> str:
    hasher = hashlib.sha256(usedforsecurity=False)
    for v in data:
        hasher.update(v.encode("utf-8"))
    return hasher.hexdigest()
```

python/gstaichi/lang/_fast_caching/function_hasher.py
```diff
@@ -0,0 +1,56 @@
...
+    fn.is_pure = True
+    return fn
+
+
+def _hash_function(function_info: FunctionSourceInfo) -> str:
+    with open(function_info.filepath) as f:
+        contents = f.read().split("\n")
+    lines = contents[function_info.start_lineno : function_info.end_lineno]
+    _hash = hash_string("\n".join(lines))
+    return _hash
```
duburcqa: You should rather do (see revised implementation of `hash_string` below):
```python
from itertools import islice

def _hash_function(function_info: FunctionSourceInfo) -> str:
    with open(function_info.filepath) as f:
        return hash_string(islice(f, function_info.start_lineno, function_info.end_lineno))
```

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -0,0 +1,99 @@
...
+    for arg in args:
+        start = time.time()
+        _hash = to_representative_str(arg)
+        g_repr_time += time.time() - start
+        if not _hash:
+            g_num_ignored_calls += 1
+            return None
+        hash_l.append(_hash)
+    start = time.time()
+    res = hash_string("_".join(hash_l))
```
duburcqa: Based on my implementation of `hash_string`:
```python
res = hash_string(hash_l)
```

python/gstaichi/lang/_fast_caching/config_hasher.py
```diff
@@ -0,0 +1,16 @@
...
+    config = impl.get_runtime().prog.config()
+    config_l = []
+    for k in dir(config):
+        if k.startswith("_"):
+            continue
+        v = getattr(config, k)
+        config_l.append(f"{k}={v}")
+    config_str = "\n".join(config_l)
+    config_hash = hash_string(config_str)
```
duburcqa: See revised implementation of `hash_string`:

`config_hash = hash_string(config_l)`

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -0,0 +1,75 @@
...
+    - cache value arg values
+    - kernel function (but not sub functions)
+    - compilation config (which includes arch, and debug)
+    """
+    args_hash = args_hasher.hash_args(args)
+    if args_hash is None:
+        return None
+    kernel_hash = function_hasher.hash_kernel(kernel_source_info)
+    config_hash = config_hasher.hash_compile_config()
+    cache_key = hash_string(f"{kernel_hash}_{args_hash}_{config_hash}")
```
duburcqa: `cache_key = hash_string((kernel_hash, args_hash , config_hash))`

python/gstaichi/lang/_fast_caching/function_hasher.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def hash_functions(function_infos: Iterable[FunctionSourceInfo]) -> list[HashedFunctionSourceInfo]:
+    results = []
+    for f_info in function_infos:
+        _hash = _hash_function(f_info)
```
duburcqa: I suggest suffix `_` instead of prefix, that is reserved for unused variables.

python/gstaichi/lang/_fast_caching/pyside_cache.py
```diff
@@ -0,0 +1,36 @@
+import os
+
+from gstaichi.lang import impl
+
+
+class PysideCache:
```
duburcqa: Pyside is not a good name, unless you mean "PySide, also known as Qt for Python, a Python library that provides official bindings for the cross-platform GUI toolkit Qt."

python/gstaichi/lang/_fast_caching/pyside_cache.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _get_filepath(self, key: str) -> str:
+        filepath = os.path.join(self.cache_folder, f"{key}.cache.txt")
+        return filepath
+
+    def _touch(self, filepath):
+        """
+        We can then simply remove any files older than ~24 hours or so,
+        to keep the cache automatically clean.
+        No need for metadata and stuff :)
```
duburcqa: I don't like this kind of heuristic but ok.

Still, docstring could be improved with short sentence, skip line, then details.

python/gstaichi/lang/_fast_caching/pyside_cache.py
duburcqa: A short one-line docstring would be great.

tests/python/gstaichi/lang/test_wrap_inspect.py
```diff
@@ -0,0 +1,44 @@
...
+
+    wrap_inspect = importlib.import_module("wrap_inspect")
+    source_info, src = _wrap_inspect.get_source_info_and_src(wrap_inspect.my_func)
+    assert source_info.function_name == "my_func"
+    assert source_info.start_lineno == 2
+    assert source_info.end_lineno == 5
+    assert len(src) == 4
+    assert all([line != "" and len(line) > 2 for line in src])
+    actual_src = "".join(src)
+    expected = """def my_func(a: int, b: int) -> int:
```
duburcqa: You should probably have a look at the standard utility [textwrap.dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent).

tests/python/gstaichi/lang/fast_caching/test_src_hasher.py
```diff
@@ -0,0 +1,145 @@
...
+
+    tmp_path = tmp_path / test_src_hasher_store_validate.__name__
+    tmp_path.mkdir(exist_ok=True)
+
+    offline_cache_path = tmp_path / "cache"
+    temp_import_path = tmp_path / "temp_import"
+    temp_import_path.mkdir(exist_ok=True)
+    print("tmp_path", tmp_path)
+    print("temp_import_path", temp_import_path)
+    print("offline_cache_path", offline_cache_path)
```
duburcqa: I don't like prints in unit tests but OK.

tests/python/gstaichi/lang/fast_caching/test_src_hasher.py
```diff
@@ -0,0 +1,145 @@
...
+    setup_folder("child_diff_same.py")
+    assert function_hasher.validate_hashed_function_infos(hashed_fileinfos)
+
+
+@test_utils.test()
+def test_src_hasher_store_validate(monkeypatch: pytest.MonkeyPatch, tmp_path: pathlib.Path) -> None:
+    test_files_path = pathlib.Path("tests/python/gstaichi/lang/fast_caching/test_files")
+
+    tmp_path = tmp_path / test_src_hasher_store_validate.__name__
+    tmp_path.mkdir(exist_ok=True)
```
duburcqa: `tmp_path` is already specific to every single test. What is your point?

tests/python/gstaichi/lang/fast_caching/test_src_hasher.py
```diff
@@ -0,0 +1,145 @@
...
+    print("temp_import_path", temp_import_path)
+    print("offline_cache_path", offline_cache_path)
+
+    ti_init_same_arch(options={"offline_cache_file_path": str(offline_cache_path)})
+
+    monkeypatch.syspath_prepend(temp_import_path)
+
+    def setup_folder(filename: str) -> None:
+        shutil.copy2(
+            str(test_files_path / filename), str(temp_import_path / "child_diff_test_src_hasher_store_validate.py")
```
duburcqa: `shutil.copy2` supports passing `pathlib.Path`. No need to convert to string.

tests/python/gstaichi/lang/fast_caching/test_src_hasher.py
```diff
@@ -0,0 +1,145 @@
...
+    assert key_base is not None
+    assert key_base != ""
+    assert key_base == key_same
+    assert key_base != key_diff
+
+
+@test_utils.test()
+def test_src_hasher_validate_hashed_function_infos(monkeypatch, tmp_path: pathlib.Path) -> None:
+    test_files_path = pathlib.Path("tests/python/gstaichi/lang/fast_caching/test_files")
+    monkeypatch.syspath_prepend(str(tmp_path))
```
duburcqa: `monkeypatch.syspath_prepend` supports `pathlib.Path`

tests/python/gstaichi/lang/fast_caching/test_files/__init__.py
duburcqa, line 1: Are you sure that you need this file?

tests/python/gstaichi/lang/fast_caching/test_args_hasher.py
```diff
@@ -0,0 +1,243 @@
...
+def _ti_init_same_arch() -> None:
+    assert ti.cfg is not None
+    ti.init(arch=getattr(ti, ti.cfg.arch.name))
+
+
+@test_utils.test()
+def test_args_hasher_field() -> None:
+    """
+    This is trickier than the others, since we need to take into account snode id, and we need to reinitialize
+    taichi each attempt 🤔
```
duburcqa: Not a fan of emoji in doc but ok.

tests/python/gstaichi/lang/fast_caching/test_args_hasher.py
```diff
@@ -0,0 +1,243 @@
...
+import gstaichi as ti
+from gstaichi.lang._fast_caching import FIELD_METADATA_CACHE_VALUE, args_hasher
+
+from tests import test_utils
+
+
+@test_utils.test()
+def test_args_hasher_numeric() -> None:
+    seen = set()
+    for arg in [3, 5.3, np.int32(3), np.int64(5), np.float32(2), np.float64(2)]:
```
duburcqa: As always, I prefer `tuple` over `list` if possible.

tests/python/gstaichi/lang/fast_caching/test_args_hasher.py
```diff
@@ -0,0 +1,243 @@
...
+                for ndim in [0, 1, 2]:
+                    arg = ti.Matrix.ndarray(m, n, dtype, [1] * ndim)
+                    for it in [0, 1]:
+                        hash = args_hasher.hash_args([arg])
+                        assert hash is not None
+                        if it == 0:
+                            assert hash not in seen
+                            seen.add(hash)
+                        else:
+                            assert hash in seen
```
duburcqa, line 82: What about defining some helper to share the logic ? it is always the same.

You could just use `itertools.product(*a)` to generate you sequence of combinations you want to test, and pass it to some helper function that would do the loop and keep track of `seen`.

python/gstaichi/types/ndarray_type.py
```diff
@@ -93,9 +93,9 @@ def check_matched(self, ndarray_type: NdarrayTypeMetadata, arg_name: str):
 
         # Check dtype match
         if isinstance(self.dtype, CompoundType):
-            if not self.dtype.check_matched(ndarray_type.element_type):
+            if not self.dtype.check_matched(ndarray_type.element_type):  # type: ignore
                 raise ValueError(
-                    f"Invalid value for argument {arg_name} - required element type: {self.dtype.to_string()}, but {ndarray_type.element_type.to_string()} is provided"
+                    f"Invalid value for argument {arg_name} - required element type: {self.dtype.to_string()}, but {ndarray_type.element_type} is provided"  # type: ignore
```
duburcqa: I encourage you to follow the same rule than "normal" code lines for strings.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1256,6 +1293,7 @@ def wrapped_func(*args, **kwargs):
     wrapped._is_classkernel = is_classkernel
     wrapped._primal = primal
     wrapped._adjoint = adjoint
+    primal.gstaichi_callable = wrapped
```
duburcqa, line 1212: The standard dunder attribute name for wrapped function is `__wrapped__`

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1170,7 +1207,7 @@ def __call__(self, *args, **kwargs) -> Any:
             _logging.warn("""opt_level = 1 is enforced to enable gradient computation.""")
             impl.current_cfg().opt_level = 1
         key = self.ensure_compiled(*args)
-        kernel_cpp = self.compiled_kernels[key]
+        kernel_cpp = self.materialized_kernels[key]
```
duburcqa, line 1126: Why this change? `materialized` is less clear than `compiled` to me, but maybe it carries a different meaning?

python/gstaichi/lang/kernel_impl.py
```diff
@@ -614,16 +624,19 @@ def __init__(self, _func: Callable, autodiff_mode: AutodiffMode, _classkernel=Fa
         impl.get_runtime().kernels.append(self)
         self.reset()
         self.kernel_cpp = None
-        self.compiled_kernels: dict[CompiledKernelKeyType, KernelCxx] = {}
+        self.materialized_kernels: dict[CompiledKernelKeyType, KernelCxx] = {}
```
duburcqa, line 605: Could you clarify somewhere in the documentation what "materialization" means (as opposed to compiled)? I find this term very confusing.

python/gstaichi/lang/_wrap_inspect.py
```diff
@@ -186,4 +188,30 @@ def getsourcefile(obj):
...
+    end_lineno = start_lineno + len(src) - 1
+    return (
+        FunctionSourceInfo(
+            function_name=name,
+            filepath=file,
+            start_lineno=start_lineno,
+            end_lineno=end_lineno,
+        ),
+        src,
+    )
```
duburcqa: what about:
```
func_info = FunctionSourceInfo(name, file, start_lineno, end_lineno)
return func_info, src
```

python/gstaichi/lang/_wrap_inspect.py
```diff
@@ -19,8 +19,10 @@
 import inspect
 import os
 import tempfile
+from typing import Callable
 
 import dill
+from pydantic import BaseModel
```
duburcqa, line 25: Why using pydantic for this? For stricter enforcement? Because annotated frozen dataclass are supported natively in python.

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -0,0 +1,75 @@
...
+    return cache_key
+
+
+class CacheValue(BaseModel):
```
duburcqa, line 29: What a pydantic model for something that only contains a single attribute? For stricter static typing enforcement? I'm not sure it is worth it for something that has no user-facing interface, but as you want.

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -0,0 +1,75 @@
...
+    hashed_function_source_infos: list[HashedFunctionSourceInfo]
+
+
+def store(cache_key: str, function_source_infos: Iterable[FunctionSourceInfo]) -> None:
+    """
+    Note that unlike other caches, this cache is not going to store the actual value we want. This cache is only used for verification
```
duburcqa: It should start with a one-line short description before details.

I would advise you not to violate linewidth policy in docstring, even if you formatter allows this.

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -0,0 +1,75 @@
...
+        - i.e. is our cache key still valid?
+    - the python side cache contains information we will use to verify that our cache key is valid
+        - ie the list of function source infos
+    """
+    if not cache_key:
+        return
+    cache = PysideCache()
+    hashed_function_source_infos = function_hasher.hash_functions(function_source_infos)
+    cache_value_obj = CacheValue(hashed_function_source_infos=list(hashed_function_source_infos))
+    cache_value_json = cache_value_obj.json()
```
duburcqa: I would avoid this temporary as it is already perfectly explicit, but as you wish.

python/gstaichi/_test_tools/__init__.py
hughperkins: Interesting.

python/gstaichi/lang/_fast_caching/__init__.py
hughperkins: Ok. Thoughts on where we place this instead?

hughperkins: (create a file `constants.py`?)

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: It could be inlined into stringify_obj_type, but would make stringify_obj_type longer, less factorized. Would prefer to extract this method out, idelaly.

python/gstaichi/lang/_fast_caching/hash_utils.py
hughperkins: Where do you see `usedforsecurity`?  https://github.com/okdshin/PicoSHA2/blob/1677374f23352716fc52183255a40c1b8e1d53eb/picosha2.h

hughperkins: oh this is in python 🤔 Ok, let me re-read...

hughperkins: What do you see as an advantage of setting usedforsecurity to False? The doc suggests that it allows access to additional algorithms. What algorithm are you thinking we might consider using, that would be blocked by default vale of usedforsecurity?

Screenshot 2025-08-13 at 8 42 49 AM

python/gstaichi/lang/_fast_caching/function_hasher.py
hughperkins: Are you sure this will gain in speed? My understanding is that each call to a method from python is pretty slow. So, we have two options:
1. keep things as a list, call every method multiple times, once for each item of the list
2. concatenate to a long string (single method call), then operate each methd just once, on this string.

hughperkins: well.... apparently list is faster:
```
import time
import hashlib

a_string = "afargreg erg erg erg erg srg erg erg erg esrg serg ergr g"

N = 1000

string_list = [a_string for i in range(N)]

def run_concatenated(string_list):
	concatenated = "_".join(string_list)
	h = hashlib.sha256()
	for v in string_list:
		h.update(concatenated.encode("utf-8"))


def run_list_direct(string_list):
	h = hashlib.sha256()
	for v in string_list:
		h.update(v.encode("utf-8"))
		h.update("_".encode("utf-8"))


start = time.time()
run_concatenated(string_list)
elapsed = time.time() - start
print('concat time', elapsed)

start = time.time()
run_list_direct(string_list)
elapsed = time.time() - start
print('list time', elapsed)

```

output:
```
(taichi) system ~/git/taichi (hp/src-ll-cache|△10) $ python /tmp/test_speed.py
concat time 0.018529176712036133
list time 0.00013327598571777344
```

hughperkins: (Note: this last quetion and response was intended for the thread about the hashing string/list function.)

tests/python/gstaichi/lang/test_wrap_inspect.py
hughperkins: Interesting. Thanks!

hughperkins: Note that it doesnt actually detect the indentation of the sourcecode, and magically remove just that indentation. It does something more like: find the maximum indenetation that only eats whitespace, and then remove that from every line. So it wont work if one needs all lines to be indented by eg 4 spaces.

tests/python/gstaichi/lang/fast_caching/test_src_hasher.py
hughperkins: removed

tests/python/gstaichi/lang/fast_caching/test_src_hasher.py
hughperkins: The imports are not. But actually this didnt fix the imports. But I found a way to fix the imports now.

hughperkins: So I've removed this

tests/python/gstaichi/lang/fast_caching/test_src_hasher.py
hughperkins: For context, this test was failing a lot. The reason was to do with imports, which I hope are fixed now. But whilst trying to fix the issue, I was doing everything I could to remove anything that mihgt be potentially problematic. hence creating a subfolder, converting to str everywhere, etc.

tests/python/gstaichi/lang/fast_caching/test_files/__init__.py
hughperkins: I used to, I think, becasue I was importing from this folder. But seems the tests pass without it now.

python/gstaichi/types/ndarray_type.py
hughperkins: What rule are you referring to? You mean, for wrapping text?

python/gstaichi/lang/kernel_impl.py
hughperkins: because these kernels haven't been compiled, and it's really REALLY confusing to have kernels that have not been compiled being referred to as 'compiled kernels'.

We also get weirdness wehre there were two functions called "compile_kernel", one which compiles the kernel, and one which ... doesnt.

To be fair, there is some minimal level of compilation in a materialized kernel: it's been converted into ast, and frontend ir, if I remember correctly; but it hasn't been compiled into LLVM IR, which is what I imagine when we say a kernel has been 'compiled'.

Yes, this particular line is adding a compiled kernel in, but there's the exact same line at 775, which does not:

```
        gstaichi_kernel = impl.get_runtime().prog.create_kernel(gstaichi_ast_generator, kernel_name, self.autodiff_mode)
        assert key not in self.materialized_kernels
        self.materialized_kernels[key] = gstaichi_kernel
```

python/gstaichi/lang/_wrap_inspect.py
hughperkins: I'd prefer to keep the keywords on the whole. I believe that your personal constraint in general is:
- either use keywrods
- or put on oneline
?

python/gstaichi/lang/_wrap_inspect.py
hughperkins: because I need a way to parse the json back into objects. I'm simply using what we used to use in my old company. I did check briefly using dataclasses to parse to objects, but looks like it will only parse the top level object, and none of the children? Have to recurse by hand?

python/gstaichi/lang/_fast_caching/src_hasher.py
hughperkins: Originally this contained the kernel source info and some other stuff too, but I stripped that out. I do feel it's nicer to simply store a single 'thing' in the cache, rather than a list or something though. Makes it much easier to parse from the json back into objects I feel.

python/gstaichi/types/ndarray_type.py
duburcqa: Yes, wrapping text based on maximum line width.

python/gstaichi/lang/kernel_impl.py
duburcqa: > because these kernels haven't been compiled, and it's really REALLY confusing to have kernels that have not been compiled being referred to as 'compiled kernels'.

It is, right !?

> Yes, this particular line is adding a compiled kernel in, but there's the exact same line at 775, which does not:

Fuck. I hate it. Could we have the definition of "materialized" that you just give me written somewhere in the code, close to where it is being used for the first time? Moreover, it would be nice to add comments mentioned whether a kernel is supposed to be compiled or not at the exact line where it is being added.

python/gstaichi/lang/_wrap_inspect.py
duburcqa: I would add that I encourage using keyword when your define the variable where it is used, without storing it with an explicit name beforehand, i.e.:
```python
fortune = let_keep_dreaming(tutorial="".join((root_path, "my_rich.mp4")))
more_fortune = make_it_double(fortune)  # instead of make_it_double(fortune=fortune)
```

duburcqa: > I'd prefer to keep the keywords on the whole. 

You can keep the keyword if you like. Still I would suggest to define `func_info = [...]` outside the return because formatting gets messy.

python/gstaichi/lang/_wrap_inspect.py
duburcqa: > because I need a way to parse the json back into objects.

oh I see! Make sense.

> I did check briefly using dataclasses to parse to objects, but looks like it will only parse the top level object, and none of the children? Have to recurse by hand?

I implemented exactly that, because I don't like adding dependencies for limited features like this, but I don't know if it is worth it actually. I posted on stackoverflow some initial implementation. I'm afraid the exact implementation is closed-source but not much more advanced: https://stackoverflow.com/a/76587845/4820605

python/gstaichi/lang/_fast_caching/src_hasher.py
duburcqa: Ok!

tests/python/gstaichi/lang/test_wrap_inspect.py
duburcqa: Hum... that is sad.

python/gstaichi/lang/_fast_caching/args_hasher.py
duburcqa: So basically this method is `stringify_dataclass_type` ? Do as you want, I'm fine with it.

python/gstaichi/lang/_fast_caching/__init__.py
duburcqa: Yes, or in the module it is used if not shared between multiple modules.

python/gstaichi/lang/_fast_caching/hash_utils.py
duburcqa: > What do you see as an advantage of setting usedforsecurity to False? 

I set it to False since we are targeting security applications, without looking for what it was doing. Feel free to keep this flag enabled, but I would suggest to keep its value explicit.

python/gstaichi/lang/_fast_caching/__init__.py
hughperkins: I made it for *you* to use. Otherwise you have to remember this exact
string all the time, and it's hard to remember.

python/gstaichi/lang/_fast_caching/function_hasher.py
duburcqa: List is slower. There is an issue in your benchmark I think:

```python
In [4]: import time
   ...: import hashlib
   ...: 
   ...: a_string = "afargreg erg erg erg erg srg erg erg erg esrg serg ergr g"
   ...: 
   ...: N = 1000
   ...: 
   ...: string_list = [a_string for i in range(N)]
   ...: 
   ...: def run_concatenated(string_list):
   ...:     concatenated = "".join(string_list)
   ...:     return hashlib.sha256(concatenated.encode("utf-8")).hexdigest()
   ...: 
   ...: 
   ...: def run_list_direct(string_list):
   ...:     h = hashlib.sha256()
   ...:     for v in string_list:
   ...:         h.update(v.encode("utf-8"))
   ...:     return h.hexdigest()
   ...: 
   ...: 
   ...: %timeit run_concatenated(string_list)
   ...: %timeit run_list_direct(string_list)
   ...: 
21.1 μs ± 27.6 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
53.3 μs ± 540 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

duburcqa: Still, I would argue that such optimisation (concatenating instead of looping) should be delegated to `hash_string`, not to the user.

gstaichi/program/program.h
duburcqa: You are still using `caps` in many places in newly added code.

gstaichi/program/program.h
duburcqa: Same, still used at many places in newly added code.

python/gstaichi/_test_tools/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def ti_init_same_arch(**options) -> None:
+    """
+    Used in tests to call ti.init, passing in the same arch as currently
+    configured. Since it's fairly fiddly to do that, extracting this out
+    to this helper function.
+    """
+    assert ti.cfg is not None
+    options["arch"] = getattr(ti, ti.cfg.arch.name)
```
duburcqa, line 14: You are modifying the input dictionary in place. Not cool.

python/gstaichi/lang/_fast_caching/hash_utils.py
```diff
@@ -0,0 +1,11 @@
+import hashlib
+from typing import Iterable
+
+
+def hash_iterable_strings(strings: Iterable[str], separator: str = "_") -> str:
```
duburcqa, line 5: Why there should be a separator? This is weird. Just hash the concatenated input no? I don't think there is any risk of things colliding because different lists reduces to the same concatenation.

python/gstaichi/lang/_fast_caching/python_side_cache.py
```diff
@@ -0,0 +1,52 @@
+import os
+
+from .. import impl
+
+
+class PythonSideCache:
+    """
+    Manages a cache that is managed from the python side (we also have c++-side caches)
+
+    The cache is disk-based. When we create teh PythonSideCache object, the cache
```
duburcqa: "teh"

python/gstaichi/lang/_fast_caching/__init__.py
duburcqa: > I made it for you to use. 

I know, but it has to be used somewhere in your code also! Just define it in the module (.py file) it is used on your side, and expose it in `__init__.py` of course, for me to enjoy it!

python/gstaichi/lang/_fast_caching/hash_utils.py
hughperkins: I would prefer to add a separator, and not having to think about the possiblity of adjacent things combining in ways we didn't intend.

python/gstaichi/_test_tools/__init__.py
hughperkins: Fair :)

hughperkins:
Addressed, or responded to, all comments AFAIK?

hughperkins:
Added end-to-end test, including testing pure annotaion works.

duburcqa (approved):

python/gstaichi/lang/_wrap_inspect.py
hughperkins: Yeah, I'd prefer depend on pydantic than have bespoke implementation on the whole. pydantic is widely used. I've seen it used in Genesis company. And we used it in my last job too.

python/gstaichi/lang/kernel_impl.py
hughperkins: I have this at line 601:

Screenshot 2025-08-14 at 5 44 08 PM

Thoughts?

python/gstaichi/lang/kernel_impl.py
hughperkins: You mena, rename `gstaichi_callable` to `__wrapped`? What unintended consequences are possible if I do that?

python/gstaichi/lang/_wrap_inspect.py
duburcqa: As you want. Be bit overkill for your need but it does the job just right.

PR #133, opened 2025-08-14
[Build] Add gpu runner

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


.github/workflows/test_gpu.yml
```diff
@@ -0,0 +1,51 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+      - name: install vulkan validation layer
+        run: |
+          sudo apt-get install -y vulkan-validationlayers
+      - name: install cuda stuff
+        run: |
+          sudo apt-get install -y libcusolver-dev-12-8
+          sudo apt-get install -y libcusparse-dev-12-8
```
duburcqa, line 35: I would suggest cuda 12-4 because we want to support it (but nothing older I think)

duburcqa: But maybe you have no other choice as a limitation of what is pre-installed on the runner?

duburcqa (approved):

hughperkins:
Discussed cuda with @duburcqa :
- cuda 12.4 wont install 
- turns out we need ubuntu 22.04 for 12.4, see:
     - packages for 24.04, has down to 12.5 https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/x86_64/ 
     - packages for 22.04 has 12.4: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/
- looks like the github t4 runners cannot use ubuntu 22.04?
- therefore I will keep with 12.8

PR #134, opened 2025-08-14
[Type] Revise the ndarray annotations on test_ndarray.py

hughperkins:
Issue: #

### Brief Summary

Revise the ndarray annotations on test_ndarray.py, to have fewer yellow squiggles in visual studio code

- also make it possible to use `ti.types.NDArray` instead of `ti.types.ndarray()`, to enable the above task.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -641,6 +641,9 @@ class Kernel:
                     ),
                 ):
                     pass
+                elif annotation in [ndarray_type.NdarrayType]:
+                    # convert from ti.types.NDArray into ti.types.NDArray()
+                    annotation = annotation()
                 elif id(annotation) in primitive_types.type_ids:
                     pass
                 elif isinstance(annotation, sparse_matrix_builder):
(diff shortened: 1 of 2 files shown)


python/gstaichi/lang/kernel_impl.py
```diff
@@ -641,6 +641,9 @@ def extract_arguments(self) -> None:
                     ),
                 ):
                     pass
+                elif annotation in [ndarray_type.NdarrayType]:
```
duburcqa: why not `annotation is ndarray_type.NdarrayType`

yun-long:
So, if you would define constants in the test python file, it should also do the job, right?

e.g.,

`INT32_1D_ARRAY = ti.types.NDArray[ti.i32, 1]`

hughperkins:
> INT32_1D_ARRAY = ti.types.NDArray[ti.i32, 1]

Looks like that works, though personally I would prefer to write out explicitly, since one doesn't save much space with the alias, and with the explicit approach anyone who knows taichi can understand the type, so, easier to read I feel.

hughperkins:
Thanks!

PR #136, opened 2025-08-14
[type] Fix pyright warnings

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/impl.py
+++ b/python/gstaichi/lang/impl.py
@@ -165,7 +165,7 @@ def _calc_slice(index, default_stop):
                 "GsTaichi does not support variables in slice now, please use constant instead of it."
             )
 
-    check_validity(start), check_validity(stop), check_validity(step)
+    _, _, _ = check_validity(start), check_validity(stop), check_validity(step)
     return [_ for _ in range(start, stop, step)]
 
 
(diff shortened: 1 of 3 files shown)


python/gstaichi/lang/impl.py
```diff
@@ -165,7 +165,7 @@ def check_validity(x):
                 "GsTaichi does not support variables in slice now, please use constant instead of it."
             )
 
-    check_validity(start), check_validity(stop), check_validity(step)
+    _, _, _ = check_validity(start), check_validity(stop), check_validity(step)
```
duburcqa: What is this? I don't get it?

hughperkins: Throws an exception if any of start, stop, step are variables rather than constants.

Get some pyright issue without the _ warning we aren't using the return value from check_validity:

Screenshot 2025-08-15 at 9 16 43 AM

Also gives yellow squiggles in VSC, same reason:

Screenshot 2025-08-15 at 9 17 14 AM

YilingQiao (approved):

python/gstaichi/lang/impl.py
duburcqa: Ok I see. strange. You could use a single `_ = ` if you want.

hughperkins:
Thanks!

PR #138, opened 2025-08-18
[Misc] Fix bug with np.bool in args hasher and construct path for debugging

hughperkins:
Issue: #

### Brief Summary

Fix bug with np.bool in args hasher and construct path for debugging
- np.bool is deprecated. We use np.bool_ instead for now.
- also added path for debugging purposes. Perhaps we should also add a user-accessible way to print this when a type/argument is not being hashed?

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/gstaichi/lang/fast_caching/test_function_hasher.py
+++ b/tests/python/gstaichi/lang/fast_caching/test_function_hasher.py
@@ -37,12 +37,12 @@ def test_read_file(tmp_path: pathlib.Path) -> None:
 
 
 @test_utils.test()
-def test_function_hasher_hash_kernel(monkeypatch) -> None:
+def test_function_hasher_hash_kernel(monkeypatch, temporary_module) -> None:
     test_files_path = "tests/python/gstaichi/lang/fast_caching/test_files"
     monkeypatch.syspath_prepend(test_files_path)
 
     def get_hash(name: str) -> _wrap_inspect.FunctionSourceInfo:
-        mod = importlib.import_module(name)
+        mod = temporary_module(name)
         info, _src = _wrap_inspect.get_source_info_and_src(mod.f1.fn)
         hash = function_hasher.hash_kernel(info)
         return hash
@@ -57,12 +57,12 @@ def test_function_hasher_hash_kernel(monkeypatch) -> None:
 
 
 @test_utils.test()
-def test_function_hasher_hash_functions(monkeypatch) -> None:
+def test_function_hasher_hash_functions(monkeypatch, temporary_module) -> None:
     test_files_path = "tests/python/gstaichi/lang/fast_caching/test_files"
     monkeypatch.syspath_prepend(test_files_path)
 
     def get_infos(name: str) -> list[HashedFunctionSourceInfo]:
-        mod = importlib.import_module(name)
+        mod = temporary_module(name)
         info, _src = _wrap_inspect.get_source_info_and_src(mod.f1.fn)
         hashed_infos = function_hasher.hash_functions([info])
         return hashed_infos


python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -21,25 +21,28 @@
...
+def stringify_obj_type(path: tuple[str, ...], obj: Any) -> str | None:
     """
     stringify the type of obj
 
     String should somehow represent the type of obj. Doesnt have to be hashed, nor does it have
     to be the actual python type string, just something representative of the type, and won't collide
     with different (allowed) types.
+
+        path is used during debugging. We should have a way of printing this without
+        having to hack the code really. Using logger perhaps?
```
duburcqa: Bad formatting.

duburcqa: Plus I don't think this comment should be part of the official documentation.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -21,25 +21,28 @@
...
         if field.metadata.get(FIELD_METADATA_CACHE_VALUE, False):
             full_repr += f" = {child_value}"
         repr_l.append(full_repr)
     return "[" + ",".join(repr_l) + "]"
 
 
-def stringify_obj_type(obj: Any) -> str | None:
+def stringify_obj_type(path: tuple[str, ...], obj: Any) -> str | None:
     """
     stringify the type of obj
```
duburcqa: Bad formatting and poorly worded. "stringify" is used in programming but is not an actual English word. "obj" is not an actual word either.

Convert an object into a string representation that only depends on its type.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -21,25 +21,28 @@
...
 
     String should somehow represent the type of obj. Doesnt have to be hashed, nor does it have
     to be the actual python type string, just something representative of the type, and won't collide
```
duburcqa: > nor does it have to be the actual python type string

What? It always return a Python string.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -21,25 +21,28 @@
...
     with different (allowed) types.
```
duburcqa, line 43: Could be nice to mention what is allowed or not but ok.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -57,18 +60,21 @@ def stringify_obj_type(obj: Any) -> str | None:
     if isinstance(obj, MatrixField):
         return f"[fm-{obj.m}-{obj.n}-{obj.snode._id}-{obj.dtype}-{obj.shape}]"
     if dataclasses.is_dataclass(obj):
-        return dataclass_to_repr(obj)
+        return dataclass_to_repr(path, obj)
     if is_data_oriented(obj):
         child_repr_l = []
         for k, v in obj.__dict__.items():
-            _child_repr = stringify_obj_type(v)
+            _child_repr = stringify_obj_type(path + (k,), v)
```
duburcqa: Personally I prefer `(*path, k)` but it is up to you.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -57,18 +60,21 @@ def stringify_obj_type(obj: Any) -> str | None:
...
             if _child_repr is None:
-                print("not representable child", k, type(v))
+                print("not representable child", k, type(v), "path", path)
                 return None
             child_repr_l.append(f"{k}: {_child_repr}")
         return ", ".join(child_repr_l)
-    if arg_type in [int, float, np.float32, np.float64, np.int32, np.int64, bool, np.bool]:
+    if arg_type in [int, float, np.float32, np.float64, np.int32, np.int64, bool]:
```
duburcqa: `(int, float, np.float32, np.float64, np.int32, np.int64, bool)` ?

What about:
```python
import numbers
import numpy as np

if issubclass(arg_type, (numbers.Number, np.number)):
    [...]
```

Unless you want to exclude other numpy scalar types on purpose.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -57,18 +60,21 @@ def stringify_obj_type(obj: Any) -> str | None:
...
         return str(arg_type)
+    if arg_type is np.bool_:
+        # np is deprecating bool. Treat specially/carefully
+        return "np.bool_"
```
duburcqa, line 79: `np.bool_` is not deprecated by Numpy and will never be, so why are you deprecating it? `np.bool` is deprecated though.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -79,9 +85,9 @@ def hash_args(args: Sequence[Any]) -> str | None:
     g_num_calls += 1
     g_num_args += len(args)
     hash_l = []
-    for arg in args:
+    for arg_i, arg in enumerate(args):
```
duburcqa: I much prefer `i_arg`, because I usually use `arg_i` for the i-th argument.

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: yes, but let's say we have an NDArray. If you do `str(type(my_ndarray))` you'll get something liek `<class 'gstaichi.lang._ndarray.ScalarNdarray'>"`. This stringify is not required to return this. Nor does it. There is no requiremetn on the string at all, except that it somehow uniquely identifies the type, relative to other types stringified by this fucntion. So "nd" is valid, so is "abc", etc.

hughperkins: well, what I actually menat was a pythong typing string, like what we see in annotaionts, like `ti.types.NDArray[ti.i32, 2]`

But either way, there is no constarint on the stsring returend by this function, other than it uniquely idenitifes the type.

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: replaced `just something something` with `just a string that is `

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: It doesnt say we are deprecating bool_. It says that np is deprecating bool.

What makes you feel that we are deprecating `bool_`?

hughperkins:
@duburcqa Addressed or responoded to all review commetns AFAIK.

python/gstaichi/lang/_fast_caching/args_hasher.py
duburcqa: I was confused by the condition vs the comment. Sorry about that.

Note that you could replace `np.number` by `np.generic` to catch any Numpy scalar type, including `np.bool_`, but I think it makes more sense to use `np.number`, because you do not want to include char types and alike.

python/gstaichi/lang/_fast_caching/args_hasher.py
duburcqa: Yes yes I get it. I was confused when you said "python type string". I thought you were referring to `str` Python string type x)

hughperkins:
Thanks!

PR #139, opened 2025-08-19
[build] Run c++ tests on gpu ci runner

hughperkins:
Issue: #

### Brief Summary

- we merged a PR recenlty to run c++ tests
- however, we arent yet running them on a cuda box, so cuda tests won't run (eg the ptx cache test in pr review, https://github.com/Genesis-Embodied-AI/gstaichi/pull/130/files )
- this PR intends to run c++ tests also on a cuda box

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_fast_caching/args_hasher.py
+++ b/python/gstaichi/lang/_fast_caching/args_hasher.py
@@ -1,5 +1,6 @@
 import dataclasses
 import enum
+import numbers
 import time
 from typing import Any, Sequence
 
@@ -21,11 +22,11 @@ g_num_ignored_calls = 0
 FIELD_METADATA_CACHE_VALUE = "add_value_to_cache_key"
 
 
-def dataclass_to_repr(arg: Any) -> str:
+def dataclass_to_repr(path: tuple[str, ...], arg: Any) -> str:
     repr_l = []
     for field in dataclasses.fields(arg):
         child_value = getattr(arg, field.name)
-        _repr = stringify_obj_type(child_value)
+        _repr = stringify_obj_type(path + (field.name,), child_value)
         full_repr = f"{field.name}: ({_repr})"
         if field.metadata.get(FIELD_METADATA_CACHE_VALUE, False):
             full_repr += f" = {child_value}"
@@ -33,14 +34,18 @@ def dataclass_to_repr(arg: Any) -> str:
     return "[" + ",".join(repr_l) + "]"
 
 
-def stringify_obj_type(obj: Any) -> str | None:
+def stringify_obj_type(path: tuple[str, ...], obj: Any) -> str | None:
     """
-    stringify the type of obj
+    Convert an object into a string representation that only depends on its type.
 
     String should somehow represent the type of obj. Doesnt have to be hashed, nor does it have
-    to be the actual python type string, just something representative of the type, and won't collide
+    to be the actual python type string, just a string that is representative of the type, and won't collide
     with different (allowed) types.
+
+    `path` is used during debugging.
     """
+    # TODO: We should have a way of printing this without having to hack the code really. Using logger perhaps?
+    # (I have another PR that addreses this https://github.com/Genesis-Embodied-AI/gstaichi/pull/144/files)
     arg_type = type(obj)
     if isinstance(obj, ScalarNdarray):
         return f"[nd-{obj.dtype}-{len(obj.shape)}]"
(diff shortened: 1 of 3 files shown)


hughperkins:
oh, it's not running ptx cache tests because we didnt merge ptx cache pr yet 😅

hughperkins:
Ok, let's review/merge this PR as-is; and I'll check more about getting/ensuring ptx-cache test is running once both this and that PR are merged.

.github/workflows/manylinux_wheel.yml
```diff
@@ -45,6 +45,10 @@ jobs:
         with:
           name: manylinux_wheel_${{ matrix.PYTHON_CP_VERSION }}
           path: wheelhouse/*.whl
+      - uses: actions/upload-artifact@v4
+        with:
+          name: manylinux_cpptests_${{ matrix.PYTHON_CP_VERSION }}
+          path: build/gstaichi_cpp_tests
```
duburcqa, line 51: Why are you doing this? 

First, what do you want to do with these artefacts. Second, why using `manylinux_wheel` workflow for building cpp tests, which are completely unrelated?

duburcqa: What do not have the same compatibility requirements for the python wheels and executable that are not part of the wheel. In theory, the latter could be compiled with arch native flag enabled, which would make them theoretically incompatible with any other platform than the one that was used to compile them. Making sure that you can compile them on one linux machine and run them on some other linux machine is beyond expectations.

Ideally, you should be able to distribute wheels that contains C++ headers and precompiled binaries of taichi, so that you can build any executable on top of it later on, without having to recompile the entire project. This is exactly what I'm doing for Genesis: https://github.com/duburcqa/jiminy/blob/accdf3629d9d8c08bb97f2905f010f1616d2168a/.github/workflows/manylinux.yml#L199

I'm fine with considering this as out of scope, but it would be nice to create a dedicated issue to keep track of this design flaw.

gstaichi/platform/cuda/detect_cuda.cpp
```diff
@@ -10,12 +10,12 @@ bool is_cuda_api_available() {
...
-    if (count <= 0) {
-      std::cerr << "No CUDA devices found." << std::endl;
-      return false;
-    }
+    // int count;
+    // instance.device_get_count(&count);
+    // if (count <= 0) {
+    //   std::cerr << "No CUDA devices found." << std::endl;
+    //   return false;
+    // }
```
duburcqa: Can you explain why you can comment this code? Because it is useless I guess? If so, I would rather get rid of it completely.

From my understanding was doing its job (though I would much prefer `count == 0` compared to `count <= 0`).

.github/workflows/manylinux_wheel.yml
hughperkins: Discussed with Aleksey offline. We won't migrate to the c++ headers and precompiled binaries of taichi for now. But I'm aware of this option, if we hit related issues later on.

gstaichi/platform/cuda/detect_cuda.cpp
hughperkins: There are two questions here:

> why did you comment the code, and not delete it?

Because I forgot to delete it.

> Why did you get rid of this code?

I added this code, in a previous PR, that was merged a day or two
- the goal of this code was to allow C_API tests to not seg fault
- it did succeed in that goal
- however we hvae removed C API now anyway
- and this code causes CUDA not to load correctly, when running from python, unfortunately.

hughperkins: (deleted the code now)

hughperkins:
@duburcqa  Addressed, or responed to, all comments, AFAIK?

.github/workflows/manylinux_wheel.yml
duburcqa: > "Aleksey" 

LOL

gstaichi/platform/cuda/detect_cuda.cpp
duburcqa: I think it would be good to mention what you just said in comment. It sounds like something very tricky that we should remember because it is not obvious that just checking the number of devices would mess up with CUDA. If you don't like lengthy comment, you could just say:
> Note that one must be careful when detecting if CUDA is available. See `https://github.com/Genesis-Embodied-AI/gstaichi/pull/139#discussion_r2291124463`.

.github/workflows/manylinux_wheel.yml
hughperkins: oops 😅

hughperkins:
Thanks!

PR #140, opened 2025-08-19
[test] Fix broken merge of test ndarray max num args skip

hughperkins:
Issue: #

### Brief Summary

Somehow two forms of the if in test_ndarray_max_num_args are present, presumably through a broken merge. One of the i'fs correctly calls pytest.skip. The other simplye returns.

This PR removes the simple return, leaving the skip.

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_ndarray.py
```diff
@@ -1173,17 +1173,14 @@ def foo(
 @test_utils.test(exclude=[ti.metal])
 def test_ndarray_max_num_args() -> None:
     if platform.system() == "Darwin" and ti.lang.impl.current_cfg().arch == ti.vulkan:
-        # Mac doesn't support so many arguments, on Vulkan
-        return
+        pytest.skip(reason="Mac doesn't support so many arguments, on Vulkan")
     num_args = 512
```
duburcqa: I would suggest skipping one additional line for clarity.

duburcqa (approved):

tests/python/test_ndarray.py
```diff
@@ -1173,17 +1173,14 @@ def foo(
 @test_utils.test(exclude=[ti.metal])
 def test_ndarray_max_num_args() -> None:
     if platform.system() == "Darwin" and ti.lang.impl.current_cfg().arch == ti.vulkan:
```
duburcqa, line 1175: Could you elaborate on `.current_cfg().arch` instead of `.get_runtime().prog.config().arch`?

Because we are doing this on the CI for Genesis and it works just fine AFAIK:
```python
# Make sure that gstaichi arch is matching requirement
ti_runtime = ti.lang.impl.get_runtime()
ti_arch = ti_runtime.prog.config().arch
if backend != gs.cpu and ti_arch in (ti._lib.core.Arch.arm64, ti._lib.core.Arch.x64):
    device, device_name, total_mem, backend = get_device(gs.cpu)
```

hughperkins: They both appear to work:
```
import gstaichi as ti

ti.init(arch=ti.metal)

print(ti.lang.impl.current_cfg().arch)
print(ti.lang.impl.get_runtime().prog.config().arch)
```
output:
```
(taichi) system ~/git/taichi (hp/fast-cache-integration|✔) $ python /tmp/test.py
[GsTaichi] version 1.0.2, llvm 15.0.7, commit fe3dc006, osx, python 3.10.16
[GsTaichi] Starting on arch=metal
Arch.metal
Arch.metal
```

hughperkins: (I feel the first one is shorter though?)

duburcqa: Ok so basically it is just a proxy! Thank you :)

PR #142, opened 2025-08-20
[type] Clean typing on misc.py

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


duburcqa (approved):

python/gstaichi/lang/impl.py
```diff
@@ -349,6 +349,8 @@ def __init__(self, kernels=None):
         self.kernels: list[Kernel] = kernels or []
         self._signal_handler_registry = None
         self.unfinalized_fields_builder = {}
+        self.short_circuit_operators: bool = False
+        self.unrolling_limit: int = 0
```
yun-long, line 353: Question: When is type hint needed and when it can/should be ignored? Why is the newly added variables (`short_circuit_operators`) have a specific type but the older one doesn't  (`unfinalized_fields_builder`/`_signal_handler_registry`)?

yun-long (approved):
LGTM

python/gstaichi/lang/impl.py
duburcqa: The idea is that all new code must be up to standard. Updating old code will be done progressively.

hughperkins: ☝️

hughperkins: Also, I probably should have put in the description, but one of the main 'conceptual changes' in this PR is removing the `# type: ignore` in misc.py, and fixing the errors which then revealed themselves. (feel free to ask me to fill in a description , if I didnt; or to answer certain questions in the description which I didnt answer, etc)

hughperkins:
Thanks!

PR #143, opened 2025-08-20
[Misc] Add src_ll_cache flag to ti.init to disable src-ll-cache

hughperkins:
Issue: #

### Brief Summary

Add src_ll_cache flag to ti.init to disable src-ll-cache
- tested via new unit test

copilot:summary

### Walkthrough

copilot:walkthrough


python/gstaichi/lang/kernel_impl.py
```diff
@@ -693,7 +693,7 @@ def materialize(self, key: CompiledKernelKeyType | None, args: tuple[Any, ...],
         if key in self.materialized_kernels:
             return
 
-        if self.gstaichi_callable and self.gstaichi_callable.is_pure:
+        if self.runtime.src_ll_cache and self.gstaichi_callable and self.gstaichi_callable.is_pure:
```
duburcqa, line 699: Maybe you should disable printing of not pure function if src-ll cache is disabled?

python/gstaichi/lang/misc.py
```diff
@@ -307,6 +307,7 @@ def init(
     _test_mode=False,
     enable_fallback=True,
     require_version=None,
+    src_ll_cache: bool = True,
```
duburcqa, line 310: You should update the docstring documentation. It is not good enough to just declare new options. The documentation is what some users will be looking for to find out what are all the gstaichi's options one can play with. 

They are already some other options that are undocumented. This should be fixed.

tests/python/gstaichi/lang/fast_caching/test_src_ll_cache.py
```diff
@@ -37,3 +39,34 @@ def has_pure() -> None:
...
+
+@test_utils.test()
+@pytest.mark.parametrize(
+    "flag_value",
+    [
+        None,
+        False,
+        True,
+    ],
+)
```
duburcqa, line 52: One line should be fine.

python/gstaichi/lang/kernel_impl.py
hughperkins: that's not sufficien.t we have plenty of unpure kernels, even built into taichi. I mean, no-one is going to enable printing inpure functions except for debugging. Far too spammy.

python/gstaichi/lang/misc.py
hughperkins: I made a thread/question about this yesterady, https://genesis-ai-company.slack.com/archives/C08UKDR6KGR/p1755703601657789

hughperkins: (or, at least, relating to this)

duburcqa: Cross-posting my reply here: I don't need autocompletion but I want documentation.

hughperkins: Ok. Speaking for myself, I can never remember what the options are, and I'd like to be able to choose from available options, inside my IDE, without hunting for doc, but ok.

tests/python/gstaichi/lang/fast_caching/test_src_ll_cache.py
hughperkins: For parametrize, personally, I find it easier to have one test per line.

python/gstaichi/lang/kernel_impl.py
duburcqa: > that's not sufficien.t we have plenty of unpure kernels, even built into taichi. I mean, no-one is going to enable printing inpure functions except for debugging. Far too spammy.

What do you mean? I'm not saying it is sufficient, I'm saying you should ALSO check that src-ll cache is enabled, not just `self.gstaichi_callable and not self.gstaichi_callable.is_pure`. See your other PR https://github.com/Genesis-Embodied-AI/gstaichi/pull/145/files#r2296562927 (yes I'm doing cross-PR reviews :p Much wow)

image

hughperkins: After discussing with @duburcqa Alexis's main concern I feel is that the tag includes `[FASTCACHE]`, which suggests that we should only be printing this tag when fast cache is enabled. If we remove `[FASTCACHE]` from tht tag, then Alexis would have no major concerns with not including a check to see if fast cache is enabled. So, I'm going to remove `[FASTCACHE]` from the tag.

hughperkins:
@duburcqa  Addressed or responded to all comments AFAIK?

hughperkins:
thanks!

PR #144, opened 2025-08-20
[misc] Add logging for invalid params to pure kernels

hughperkins:
Issue: #

### Brief Summary

Add logging for invalid params to pure kernels; and unit tests for these.

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
Hmmm, test passes with `-s`, but fails without `-s`. which makes sense I suppose.

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -19,6 +21,12 @@ def create_cache_key(kernel_source_info: FunctionSourceInfo, args: Sequence[Any]
     """
     args_hash = args_hasher.hash_args(args)
     if args_hash is None:
+        # the bit in caps at start should not be modified without modifying corresponding text
+        # freetext bit can be freely modified
+        _logging.warn(
+            f"FASTCACHE_INVALID_FUNC The pure function {kernel_source_info.function_name} could not be "
+            "fast cached, because one or more parameter types were invalid"
+        )
```
duburcqa: See my comment https://github.com/Genesis-Embodied-AI/gstaichi/pull/145/files#r2291242086.

I would suggest using tags: `[FASTCACHE][INVALID_FUNC] ...`

tests/python/gstaichi/lang/fast_caching/test_src_ll_cache.py
```diff
@@ -37,3 +39,37 @@ def has_pure() -> None:
     assert has_pure._primal.src_ll_cache_observations.cache_key_generated
     assert has_pure._primal.src_ll_cache_observations.cache_validated
     assert has_pure._primal.src_ll_cache_observations.cache_loaded
+
+
+# Should be enough to run these on cpu I think, and anything involving
+# stdout/stderr capture is fairly flaky on other arch
+@test_utils.test(arch=ti.cpu)
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="Windows stderr not working with capfd")
```
duburcqa, line 47: Did you experience issues yourself when running this unit test on Windows. Because `capfd` is officially supported on Windows, although it may be work reliably on Windows according to some users feedback on Github issues.

hughperkins: Yeah, see my commit history of various attempts. All those failed commits are attempts to work around this issue.

duburcqa: Ok. Better to skip on Windows then. But it would be nice to clarify that it is "broken", not just not working.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -79,6 +77,10 @@ def stringify_obj_type(path: tuple[str, ...], obj: Any) -> str | None:
         return "np.bool_"
     if isinstance(obj, enum.Enum):
         return f"enum-{obj.name}-{obj.value}"
+    # The bit in caps should not be modified without updating corresponding test
+    # The rest of free text can be freely modified
+    # (will probably formalize this in more general doc / contributor guidelines at some point)
+    _logging.warn(f"FASTCACHE_PARAM_INVALID Parameter with path {path} and type {arg_type} not allowed by fast cache.")
```
duburcqa: `[FASTCACHE][PARAM_INVALID]`

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -79,6 +77,12 @@ def stringify_obj_type(path: tuple[str, ...], obj: Any) -> str | None:
         return "np.bool_"
     if isinstance(obj, enum.Enum):
         return f"enum-{obj.name}-{obj.value}"
+    # The bit in caps should not be modified without updating corresponding test
+    # The rest of free text can be freely modified
+    # (will probably formalize this in more general doc / contributor guidelines at some point)
+    _logging.warn(
+        f"[FASTCACHE][PARAM_INVALID] Parameter with path {path} and type {arg_type} not allowed by fast cache."
+    )
```
yun-long, line 85: probably we should type the `obj`  instead of `Any`?

e.g.

```python
def stringify_obj_type(path: tuple[str, ...], obj: AllowedObj) -> str | None:
```

yun-long (approved):

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: Possibly. Though I'm not sure if there is a type available for @ti.data_oriented. @duburcqa do you know if there is a type avialable for data oriented?

duburcqa (approved):

python/gstaichi/lang/_fast_caching/args_hasher.py
duburcqa: To do thinks properly, you should probably implement some typing protocol: https://typing.python.org/en/latest/spec/protocol.html#protocol-members

Or move to `pydantic`. It should work in this context.

Still, the most correct weak annotation in this context is `object`. It would already be better than `Any`, which basically means "I don't know and I don't care, leave me alone".

hughperkins: this method actually receives the objects, not the types. It's stringifying the type of the objects, but the input parameter is the object itself.

hughperkins: (for example, for fields, this means we can access the snode id)

hughperkins: oh, for some reason I thought you wrote 'type', but now is ays 'object' 🤔

hughperkins: updated to `object`

duburcqa: > oh, for some reason I thought you wrote 'type', but now is ays 'object'

Yes, I updated my answer while you were looking somewhere else!

hughperkins: Note: had to add type ignore for torch, but that's probably good: shows typign is turned on using object.

PR #146, opened 2025-08-20
[test] Use temporary_module to avoid leaving modules behind in namespace

hughperkins:
Issue: #

### Brief Summary

- when testing function hasher, we need to import various modules, temporarily. Some of the tests were leaving behind these modules in the namespace, after the test
- this caused spurious test failures
- we migrate to use a fixture that handles removing the modules from the namespace automatically after the test

copilot:summary

### Walkthrough

copilot:walkthrough


YilingQiao (approved):

duburcqa (approved):

hughperkins:
Thanks!

PR #159, opened 2025-08-26
[Perf] Improve testing of pure functions to call function repeatedly after cache load; and fix failure

hughperkins:
Issue: #

### Brief Summary

This addresses a bug where once a kernel has been loaded from src-ll-cache, the first time it is iexecuted, it works, but afterwards, nothing happens.

There are two parts to this PR:
- we add a test that does:
     - create a new empty cache, in a temp folder
     - run a kernel a few times, to populate hte cache
     - reinitialize gstaichi
     - run the kernel a few more times, and check that it ran each time, by checking that a passed-in tensor was modified correctly each time
 - we fix the bug that meant that subsequent calls after cache load did not work
      - we add a second dictionary, that stores the compiled_kernel_data, indexed by a python-side key (not the fast_cache_key, we should probably give it a better name than 'key'...)
      - we use this compiled_kernel_data for subsequent calls, which avoids inadvertently asking the c++ side to do some combination of reocmpiling, or loading incorrectly from cache

copilot:summary

### Walkthrough

copilot:walkthrough


python/gstaichi/lang/kernel_impl.py
```diff
@@ -781,7 +782,7 @@ def ast_to_dict(node: ast.AST | list | primitive_types._python_primitive_types):
                     )
                 struct_locals = _kernel_impl_dataclass.extract_struct_locals_from_context(ctx)
                 tree = _kernel_impl_dataclass.unpack_ast_struct_expressions(tree, struct_locals=struct_locals)
-                ctx.only_parse_function_def = self.compiled_kernel_data is not None
+                ctx.only_parse_function_def = self.compiled_kernel_data_by_key.get(key, None) is not None
```
duburcqa: `self.compiled_kernel_data_by_key.get(key) is not None`

By the way, the key can exist but map to None value nonetheless? If it is impossible, then you should rather use `key in self.compiled_kernel_data_by_key`

hughperkins: I'm not sure. I'd prefer to err on the side of caution.  I'm 95% sure it *can* be None, because: see lines 708 or so:

```
                self.compiled_kernel_data_by_key[key] = prog.load_fast_cache(
                    self.fast_checksum,
                    self.func.__name__,
                    prog.config(),
                    prog.get_device_caps(),
                )
                if self.compiled_kernel_data_by_key[key]:
                    self.src_ll_cache_observations.cache_loaded = True
```

probalby more like 100% sure it can be None actually.

duburcqa: I see! Still `get(key)` is more idiomatic than `get(key, None)`

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1041,19 +1042,19 @@ def recursive_set_args(needed_arg_type: Type, provided_arg_type: Type, v: Any, i
 
         try:
             prog = impl.get_runtime().prog
-            if not self.compiled_kernel_data:
-                self.compiled_kernel_data = prog.compile_kernel(prog.config(), prog.get_device_caps(), t_kernel)
+            if not compiled_kernel_data:
```
duburcqa, line 1045: `if compiled_kernel_data is None:` would be more explicit.

PR #161, opened 2025-08-28
[test] Instrument fe-ll-cache so we can test when we get a cache hit

hughperkins:
…d a unit test of this

Issue: #

### Brief Summary

Instrument fe-ll-cache so we can test when we get a cache hit, and add a unit test for this.

## Background

In the context of Genesis, we want to be able to unit test that when using ndarray, we can change ndarray dimensions (eg num-env), without triggering a recompile. A pre-requisite for that is to be able to detect in a convenient way when we triggered a recompile, or approximately equivaently, when we got an fe-ll-cache cache hit.

This PR adds an instrumentation to gstaichi to allow the convenient detection of fe-ll-cache cache hits.

Implementation:
- on the C++ side, when we compile a kernel, instead of returning a CompiledKernelData &, we return a CompileResult struct, which contains the CompiledKernelData &, and a bool called `cache_hit`.
- on the Python side, we add an FeLlCacheObservations dataclass to the Kernel object, and set cache_hit to True, when the compile result we receive from prog.compile_kernel is True

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_fast_caching/args_hasher.py
+++ b/python/gstaichi/lang/_fast_caching/args_hasher.py
@@ -43,6 +43,8 @@ def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
     String should somehow represent the type of obj. Doesnt have to be hashed, nor does it have
     to be the actual python type string, just a string that is representative of the type, and won't collide
     with different (allowed) types.
+
+    Note that fields are not included in fast cache.
     """
     arg_type = type(obj)
     if isinstance(obj, ScalarNdarray):
@@ -50,7 +52,10 @@ def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
     if isinstance(obj, VectorNdarray):
         return f"[ndv-{obj.n}-{obj.dtype}-{len(obj.shape)}]"
     if isinstance(obj, ScalarField):
-        return f"[f-{obj.snode._id}-{obj.dtype}-{obj.shape}]"
+        # disabled for now, because we need to think about how to handle field offset
+        # etc
+        # TODO: think about whether there is a way to include fields
+        return None
     if isinstance(obj, MatrixNdarray):
         return f"[ndm-{obj.m}-{obj.n}-{obj.dtype}-{len(obj.shape)}]"
     if "torch.Tensor" in str(arg_type):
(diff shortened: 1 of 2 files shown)


gstaichi/program/snode_rw_accessors_bank.cpp
```diff
@@ -41,17 +41,19 @@ void SNodeRwAccessorsBank::Accessors::write_float(const std::vector<int> &I,
   set_kernel_args(I, snode_->num_active_indices, &launch_ctx);
   launch_ctx.set_arg_float({snode_->num_active_indices}, val);
   prog_->synchronize();
-  const auto &compiled_kernel_data = prog_->compile_kernel(
+  CompileResult compile_result = prog_->compile_kernel(
       prog_->compile_config(), prog_->get_device_caps(), *writer_);
+  auto &compiled_kernel_data = compile_result.compiled_kernel_data;
```
bo-rc, line 46: I remember we are using C++17 and above? then the following is preferred (structured binding):

```cpp
auto [compiled_kernel_data, /*cache_hit*/] = prog_->compile_kernel(...);
```

There are multiple places in this PR this idiom should apply.

hughperkins: Thanks! Looks much nicer 🙌

duburcqa: > then the following is preferred (structured binding):

Structured bindings requires all output variables to be defined (if changes in C++23 or 26 if I remember correctly), so it did not play well with "warn about unused variables" plus "warning as an error".

setup.py
```diff
@@ -56,7 +56,7 @@ def get_version():
 
         version = scm_get_version()
         # Parse version string (e.g., "1.2.3" or "1.2.3.dev0+g1234567")
-        version_parts = version.split("+")[0].split(".dev")[0].split("rc")[0].split(".")
+        version_parts = version.split("+")[0].split(".dev")[0].split("rc")[0].split("b")[0].split(".")
```
CharlesMasson, line 59: What is this about?

hughperkins: Dealing with the various pre-release names. eg 1.1.1b0, 1.1.1rc1, etc.

gstaichi/program/snode_rw_accessors_bank.cpp
hughperkins: Hi, @bo-rc thoughts on what @duburcqa is saying about structured bindings?

duburcqa (approved):

PR #163, opened 2025-08-31
[perf] Disable fast cache for fields

hughperkins:
Issue: #

### Brief Summary

Disable fast cache for fields, since we don't yet take into account offset of fields, so kernels which should be recompiled are not being recompiled.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_fast_caching/args_hasher.py
+++ b/python/gstaichi/lang/_fast_caching/args_hasher.py
@@ -50,7 +50,9 @@ def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
     if isinstance(obj, VectorNdarray):
         return f"[ndv-{obj.n}-{obj.dtype}-{len(obj.shape)}]"
     if isinstance(obj, ScalarField):
-        return f"[f-{obj.snode._id}-{obj.dtype}-{obj.shape}]"
+        # disabled for now, because we need to think about how to handle field offset
+        # etc
+        return None
     if isinstance(obj, MatrixNdarray):
         return f"[ndm-{obj.m}-{obj.n}-{obj.dtype}-{len(obj.shape)}]"
     if "torch.Tensor" in str(arg_type):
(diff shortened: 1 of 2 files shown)


python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -50,15 +50,19 @@ def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
     if isinstance(obj, VectorNdarray):
         return f"[ndv-{obj.n}-{obj.dtype}-{len(obj.shape)}]"
     if isinstance(obj, ScalarField):
-        return f"[f-{obj.snode._id}-{obj.dtype}-{obj.shape}]"
+        # disabled for now, because we need to think about how to handle field offset
```
duburcqa, line 55: I would recommend added some tag that you could grep easily. I remember you already picked one in the past, that was neither FIXME nor TODO…

tests/python/gstaichi/lang/fast_caching/test_args_hasher.py
```diff
@@ -105,44 +105,14 @@ def _ti_init_same_arch() -> None:
...
-    deciding whether a kernel needs to be recompiled. A kernel is bound to the snode id of each kernel
-    used in that kernel. No matter whether the field was accessed as a global variable, or passed in as
-    a parameter. In addition the kernel is bound to the dtype and shape of the field, as well as pretty
-    much everything else about the field other than the actual values/data stored in it.
-
-    We need to check therefore that:
-    - if we reinitialize ti, and recreate the same field, that the hashes DO match
-    - if we do NOT reinitialize ti, and we recreate the same field, that the hashes do NOT match (because:
-      different snode id)
+    Disabling fields for now. Check they are disabled.
```
duburcqa: Are you sure that the entire comment should be deleted? Maybe you should add details to explain why your current approach does not work, or at least at a reference to the Genesis issue and this PR.

duburcqa (approved):

PR #164, opened 2025-09-02
[perf] ensure consistent module name when using src-ll cache

hughperkins:
Issue: #

### Brief Summary

ensure consistent module name when using src-ll cache
- the impact is that this means that the ptx kernel will only be cached once not twice
- added a test for consistency of the dumped module string

copilot:summary

### Walkthrough

copilot:walkthrough


yun-long:
didn't check the c++ implementation. on the python side, looks good to me.
maybe not for this PR, would recommend avoid using `print` but built-in `logger` in future

BTW: thank you for walking me through the code.

yun-long (approved):

hughperkins:
Thanks!

PR #167, opened 2025-09-04
[misc] Better error reporting

hughperkins:
Issue: #

### Brief Summary

Better error reporting for bugs related to ndarray migration.
- many diagnostics added when hit an error condition
- shouldn't affect behavior/output when in non-error condition

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/ast/ast_transformers/call_transformer.py
+++ b/python/gstaichi/lang/ast/ast_transformers/call_transformer.py
@@ -177,7 +177,11 @@ class CallTransformer:
             if dataclasses.is_dataclass(val):
                 dataclass_type = val
                 for field in dataclasses.fields(dataclass_type):
-                    child_name = create_flat_name(arg.id, field.name)
+                    try:
+                        child_name = create_flat_name(arg.id, field.name)
+                    except Exception as e:
+                        print("Exception whilst processing", field.name, "in", type(dataclass_type))
+                        raise e
                     load_ctx = ast.Load()
                     arg_node = ast.Name(
                         id=child_name,
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -336,6 +336,12 @@ def _process_args(self: "Func | Kernel", is_func: bool, args: tuple[Any, ...], k
         arg_str = ", ".join(map(str, args))
         expected_str = ", ".join(f"{arg.name} : {arg.annotation}" for arg in self.arg_metas)
         msg = f"Too many arguments. Expected ({expected_str}), got ({arg_str})."
+        for i in range(len_args):
+            if i < len(self.arg_metas):
+                print(i, self.arg_metas[i].name, type(args[i]))
+            else:
+                print(i, "<out of arg metas>", type(args[i]))
+        print("in function", self.func)
         raise GsTaichiSyntaxError(msg)
 
     for i, arg in enumerate(args):
@@ -351,14 +357,27 @@ def _process_args(self: "Func | Kernel", is_func: bool, args: tuple[Any, ...], k
         else:
             raise GsTaichiSyntaxError(f"Unexpected argument '{key}'.")
 
+    missing_annotations = []
     for i, arg in enumerate(fused_args):
         if arg is inspect.Parameter.empty:
             if self.arg_metas[i].annotation is inspect._empty:
-                raise GsTaichiSyntaxError(f"Parameter `{self.arg_metas[i].name}` missing.")
+                # raise GsTaichiSyntaxError(f"Parameter `{self.arg_metas[i].name}` missing.")
+                missing_annotations.append(f"Parameter `{self.arg_metas[i].name}` missing.")
             else:
-                raise GsTaichiSyntaxError(
+                missing_annotations.append(
                     f"Parameter `{self.arg_metas[i].name} : {self.arg_metas[i].annotation}` missing."
                 )
+                # raise GsTaichiSyntaxError(
+                #     f"Parameter `{self.arg_metas[i].name} : {self.arg_metas[i].annotation}` missing."
+                # )
+    if len(missing_annotations) > 0:
+        print("fused args:")
+        for i, arg in enumerate(fused_args):
+            print("  ", i, arg)
+        print("arg metas:")
+        for i, arg in enumerate(self.arg_metas):
+            print("  ", i, arg)
+        raise GsTaichiSyntaxError("\n".join(missing_annotations))
 
     return tuple(fused_args)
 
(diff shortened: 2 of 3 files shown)


python/gstaichi/lang/kernel_impl.py
```diff
@@ -351,14 +357,27 @@ def _process_args(self: "Func | Kernel", is_func: bool, args: tuple[Any, ...], k
         else:
             raise GsTaichiSyntaxError(f"Unexpected argument '{key}'.")
 
+    missing_annotations = []
```
SonSang: It's not an important stuff, but wouldn't it be better to name it `missing_parameters` or something like that? `missing_annotations` sounds like it is an error related to absence of type annotations, which are necessary for kernels.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -997,6 +1016,8 @@ def recursive_set_args(needed_arg_type: Type, provided_arg_type: Type, v: Any, i
                 set_arg_sparse_matrix_builder(indices, v)
                 return 1
             if dataclasses.is_dataclass(needed_arg_type):
+                if provided_arg_type != needed_arg_type:
+                    print("needed", needed_arg_type, "!= provided", provided_arg_type)
                 assert provided_arg_type == needed_arg_type
```
SonSang, line 1024: How about raising `GsTaichiRuntimeTypeError` in the above if statement and remove this line?

hughperkins: addressed in 01deb4ddb

python/gstaichi/lang/kernel_impl.py
hughperkins: addressed in 01deb4ddb

SonSang (approved):

python/gstaichi/lang/kernel_impl.py
```diff
@@ -336,6 +336,12 @@ def _process_args(self: "Func | Kernel", is_func: bool, args: tuple[Any, ...], k
         arg_str = ", ".join(map(str, args))
         expected_str = ", ".join(f"{arg.name} : {arg.annotation}" for arg in self.arg_metas)
         msg = f"Too many arguments. Expected ({expected_str}), got ({arg_str})."
+        for i in range(len_args):
+            if i < len(self.arg_metas):
+                print(i, self.arg_metas[i].name, type(args[i]))
+            else:
+                print(i, "<out of arg metas>", type(args[i]))
+        print("in function", self.func)
         raise GsTaichiSyntaxError(msg)
```
duburcqa: You are mixing prints are exception. I don't think it is right. I would suggest formatting a single clean error message containing all the information, possibly with multiple lines and bullet points.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -351,14 +357,23 @@ def _process_args(self: "Func | Kernel", is_func: bool, args: tuple[Any, ...], k
...
                     f"Parameter `{self.arg_metas[i].name} : {self.arg_metas[i].annotation}` missing."
                 )
+    if len(missing_parameters) > 0:
+        print("fused args:")
+        for i, arg in enumerate(fused_args):
+            print("  ", i, arg)
+        print("arg metas:")
+        for i, arg in enumerate(self.arg_metas):
+            print("  ", i, arg)
+        raise GsTaichiSyntaxError("\n".join(missing_parameters))
```
duburcqa: Same remark. You did half of the job already in this case.

python/gstaichi/lang/kernel_impl.py
hughperkins: Yeah, I kind of having been thinking the same thing tbh. Will look into this.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1109,7 +1126,11 @@ def construct_kernel_ret(self, launch_ctx: KernelLaunchContext, ret_type: Any, i
         raise GsTaichiRuntimeTypeError(f"Invalid return type on index={index}")
 
     def ensure_compiled(self, *args: tuple[Any, ...]) -> tuple[Callable, int, AutodiffMode]:
-        instance_id, arg_features = self.mapper.lookup(args)
+        try:
+            instance_id, arg_features = self.mapper.lookup(args)
+        except Exception as e:
```
duburcqa, line 1139: Catching all possible exception is usually not recommended but in get why you are doing this.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1109,7 +1126,11 @@ def construct_kernel_ret(self, launch_ctx: KernelLaunchContext, ret_type: Any, i
...
+            print("exception while trying to ensure compiled", self.func)
+            raise e
```
duburcqa: Same issue. I would suggest doing this instead:
```python
raise RuntimeError(f"Exception while trying to ensure compiled '{self.func}'.") from e
```

hughperkins: Neat. Thanks!

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -177,7 +177,11 @@ def _expand_Call_dataclass_args(args: tuple[ast.stmt]) -> tuple[ast.stmt]:
             if dataclasses.is_dataclass(val):
                 dataclass_type = val
                 for field in dataclasses.fields(dataclass_type):
-                    child_name = create_flat_name(arg.id, field.name)
+                    try:
+                        child_name = create_flat_name(arg.id, field.name)
+                    except Exception as e:
+                        print("Exception whilst processing", field.name, "in", type(dataclass_type))
+                        raise e
```
duburcqa: Same. Using `raise ... from e` pattern.

duburcqa (approved):

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -177,7 +177,10 @@ def _expand_Call_dataclass_args(args: tuple[ast.stmt]) -> tuple[ast.stmt]:
             if dataclasses.is_dataclass(val):
                 dataclass_type = val
                 for field in dataclasses.fields(dataclass_type):
-                    child_name = create_flat_name(arg.id, field.name)
+                    try:
+                        child_name = create_flat_name(arg.id, field.name)
+                    except Exception as e:
+                        raise RuntimeError("Exception whilst processing", field.name, "in", type(dataclass_type)) from e
```
duburcqa: This does not work as expected

duburcqa: `RuntimeError(f"Exception whilst processing {field.name} in {type(dataclass_type)}")`

hughperkins: good point

hughperkins:
Thanks!

duburcqa (approved):

python/gstaichi/lang/kernel_impl.py
hughperkins: so, it turns out that this breaks lots of tests:
Screenshot 2025-09-08 at 11 00 18 AM

Thoughts?

duburcqa: Two options: 
* change the unit test accordingly
* do `raise type(e)(...)` instead of `RuntimeError` systematically.

duburcqa (approved):

PR #168, opened 2025-09-06
[LLVM upgrade][PART 2] compiler.py

johnnynunez:


.github/workflows/scripts/ti_build/compiler.py
```diff
@@ -33,7 +33,7 @@ def setup_clang(as_compiler=True) -> None:
     """
     u = platform.uname()
     if u.system == "Linux":
-        for v in ("-15", "-14", "-13", "-12", "-11", "-10", ""):
+        for v in ("-21", "-20", "-19", "-18", "-17", "-16", "-15", "-14", "-13", "-12", "-11", "-10", ""):
```
hughperkins, line 36: This looks scary. What are your thought on how we can test this?

johnnynunez: It is llvm versions, taichi has the option to compile from llvm host

hughperkins: Yes. So, a general rule of thumb is: anything not tested tends not to work. So, ideally I'd like that anything we support is tested. It's not a strict rule, but would be good to test what we can I feel. I don't have a strong opinion on this point, but I kind of feel would be good to:
- remove llvm versions that aren't really necessary
- test many of the remaining versions, using CI

I wonder if we can remove all versions prior to 15 for example 🤔  I'm not sure really. @duburcqa do you have any opinions on this point?

hughperkins: @duburcqa Thoughts on this? I'm tempted to accept the change, but request that -14, -13, -12, -11, -10 are removed. Thoughts?

PR #169, opened 2025-09-06
[Part 2] vulkansdk

johnnynunez:
vulkansdk arm

### DIFF
--- a/.github/workflows/scripts/ti_build/entry.py
+++ b/.github/workflows/scripts/ti_build/entry.py
@@ -53,11 +53,20 @@ def build_wheel(python: Command, pip: Command) -> None:
     elif wheel_tag:
         proj_tags.extend(["egg_info", f"--tag-build={wheel_tag}"])
 
-    if platform.system() == "Linux":
+    u = platform.uname()
+    if (u.system, u.machine) == ("Linux", "x86_64"):
         if is_manylinux2014():
             extra.extend(["-p", "manylinux2014_x86_64"])
         else:
             extra.extend(["-p", "manylinux_2_27_x86_64"])
+    elif (u.system, u.machine) in (("Linux", "arm64"), ("Linux", "aarch64")):
+        if is_manylinux2014():
+            extra.extend(["-p", "manylinux2014_aarch64"])
+        else:
+            extra.extend(["-p", "manylinux_2_27_aarch64"])
+    else:
+        extra.extend(["-p", "manylinux2014_x86_64"])
+
     if platform.system() == "Darwin":
         extra.extend(["-p", "macosx-11.0-arm64"])
 
--- a/.github/workflows/scripts/ti_build/llvm.py
+++ b/.github/workflows/scripts/ti_build/llvm.py
@@ -27,6 +27,10 @@ def setup_llvm() -> None:
             out = get_cache_home() / "llvm15"
             url = "https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/taichi-llvm-15-linux.zip"
         download_dep(url, out, strip=1)
+    elif (u.system, u.machine) in (("Linux", "arm64"), ("Linux", "aarch64")):
+        out = get_cache_home() / "llvm15-manylinux2014"
+        # FIXME: ARM LLVM!
+        pass
     elif (u.system, u.machine) == ("Darwin", "arm64"):
         out = get_cache_home() / "llvm15-m1-nozstd"
         url = "https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/taichi-llvm-15-m1-nozstd.zip"
@@ -44,4 +48,8 @@ def setup_llvm() -> None:
 
     # We should use LLVM toolchains shipped with OS.
     # path_prepend('PATH', out / 'bin')
-    os.environ["LLVM_DIR"] = str(out)
+    if (u.system, u.machine) not in (("Linux", "arm64"), ("Linux", "aarch64")):
+        os.environ["LLVM_DIR"] = str(out)
+        # os.environ["LLVM_DIR"] = "/usr/lib/llvm-20/cmake"
+        # os.environ["CUDA_HOME"] = "/usr/local/cuda"
+        # os.environ["CPATH"] = "/usr/local/cuda/include"
--- a/.github/workflows/scripts/ti_build/vulkan.py
+++ b/.github/workflows/scripts/ti_build/vulkan.py
@@ -10,30 +10,34 @@ from .dep import download_dep
 from .misc import banner, get_cache_home, path_prepend
 from .python import path_prepend
 
-VULKAN_VERSION = "1.4.304.1"
-
 
 # -- code --
-@banner(f"Setup Vulkan {VULKAN_VERSION}")
+@banner("Setup Vulkan 1.4.313.0")
 def setup_vulkan():
     u = platform.uname()
-    if u.system == "Linux":
-        url = (
-            f"https://sdk.lunarg.com/sdk/download/{VULKAN_VERSION}/linux/vulkansdk-linux-x86_64-{VULKAN_VERSION}.tar.xz"
-        )
-        prefix = get_cache_home() / f"vulkan-{VULKAN_VERSION}"
-
+    if (u.system, u.machine) == ("Linux", "x86_64"):
+        url = "https://sdk.lunarg.com/sdk/download/1.4.313.0/linux/vulkansdk-linux-x86_64-1.4.313.0.tar.xz"
+        prefix = get_cache_home() / "vulkan-1.4.313.0"
         download_dep(url, prefix, strip=1)
         sdk = prefix / "x86_64"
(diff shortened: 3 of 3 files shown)


.github/workflows/scripts/ti_build/vulkan.py
```diff
@@ -10,30 +10,34 @@
 from .misc import banner, get_cache_home, path_prepend
 from .python import path_prepend
 
-VULKAN_VERSION = "1.4.304.1"
```
hughperkins, line 13: Let's continue to use a version string please, rather than hard-coding in each download.

.github/workflows/scripts/ti_build/entry.py
```diff
@@ -53,11 +53,20 @@ def build_wheel(python: Command, pip: Command) -> None:
...
 
-    if platform.system() == "Linux":
+    u = platform.uname()
+    if (u.system, u.machine) == ("Linux", "x86_64"):
         if is_manylinux2014():
             extra.extend(["-p", "manylinux2014_x86_64"])
         else:
             extra.extend(["-p", "manylinux_2_27_x86_64"])
+    elif (u.system, u.machine) in (("Linux", "arm64"), ("Linux", "aarch64")):
+        if is_manylinux2014():
```
hughperkins, line 63: Note to self: we should probalby get rid of manylinux2014, since we no longer use it. But fine with leaving it in for now.

.github/workflows/scripts/ti_build/entry.py
```diff
@@ -53,11 +53,20 @@ def build_wheel(python: Command, pip: Command) -> None:
...
+            extra.extend(["-p", "manylinux2014_aarch64"])
+        else:
+            extra.extend(["-p", "manylinux_2_27_aarch64"])
+    else:
```
hughperkins, line 67: what is this else here for? Perhaps we can sstructure like this:
```
if u.system == "Linux":
    if u.machine == "x86_64":
        ....
    elif u.machine in ["arm64", "aarch64"]:
       ....
    else:
         raise Exception("unrecognized machine {u.machine}"
```
?

.github/workflows/scripts/ti_build/llvm.py
```diff
@@ -27,6 +27,10 @@ def setup_llvm() -> None:
             out = get_cache_home() / "llvm15"
             url = "https://github.com/taichi-dev/taichi_assets/releases/download/llvm15/taichi-llvm-15-linux.zip"
```
hughperkins, line 28: Dont we need an arm build of llvm?

.github/workflows/scripts/ti_build/llvm.py
```diff
@@ -27,6 +27,10 @@ def setup_llvm() -> None:
...
         download_dep(url, out, strip=1)
+    elif (u.system, u.machine) in (("Linux", "arm64"), ("Linux", "aarch64")):
+        out = get_cache_home() / "llvm15-manylinux2014"
+        # FIXME: ARM LLVM!
```
hughperkins, line 32: Yes!

.github/workflows/scripts/ti_build/llvm.py
```diff
@@ -44,4 +48,8 @@ def setup_llvm() -> None:
 
     # We should use LLVM toolchains shipped with OS.
     # path_prepend('PATH', out / 'bin')
-    os.environ["LLVM_DIR"] = str(out)
+    if (u.system, u.machine) not in (("Linux", "arm64"), ("Linux", "aarch64")):
```
hughperkins, line 51: Please could you explain
1. what the original intent of this line is?
2. why does the new behavior vary for arm vs x86?

.github/workflows/scripts/ti_build/llvm.py
```diff
@@ -44,4 +48,8 @@ def setup_llvm() -> None:
...
+        os.environ["LLVM_DIR"] = str(out)
+        # os.environ["LLVM_DIR"] = "/usr/lib/llvm-20/cmake"
+        # os.environ["CUDA_HOME"] = "/usr/local/cuda"
```
hughperkins, line 54: Let's remove dead code please.

.github/workflows/scripts/ti_build/vulkan.py
```diff
@@ -10,30 +10,34 @@
...
+    if (u.system, u.machine) == ("Linux", "x86_64"):
+        url = "https://sdk.lunarg.com/sdk/download/1.4.313.0/linux/vulkansdk-linux-x86_64-1.4.313.0.tar.xz"
+        prefix = get_cache_home() / "vulkan-1.4.313.0"
         download_dep(url, prefix, strip=1)
         sdk = prefix / "x86_64"
         os.environ["VULKAN_SDK"] = str(sdk)
         path_prepend("PATH", sdk / "bin")
         path_prepend("LD_LIBRARY_PATH", sdk / "lib")
-        os.environ["VK_LAYER_PATH"] = str(sdk / "share" / "vulkan" / "explicit_layer.d")
+        os.environ["VK_LAYER_PATH"] = str(sdk / "etc" / "vulkan" / "explicit_layer.d")
```
hughperkins, line 26: does this mean this is broken currently? Any way we can test this in CI?

.github/workflows/scripts/ti_build/llvm.py
johnnynunez: Because not exists assets right now on that links for arm

.github/workflows/scripts/ti_build/vulkan.py
```diff
@@ -57,4 +61,4 @@ def setup_vulkan():
         os.environ["VK_LAYER_PATH"] = str(prefix / "Bin")
         path_prepend("PATH", prefix / "Bin")
     else:
-        return
+        return
```
hughperkins, line 64: Have you run `pre-commit` by the way? Run something like 
```
pre-commit run -a
```
I feel like it might have an opinion about linefeeds at the end of python files.

.github/workflows/scripts/ti_build/llvm.py
hughperkins: Yeah, we need an arm build of llvm right? Shall I create an empty repo for an llvm build, and you can PR an llvm build script onto that repo?

johnnynunez:
> https://releases.llvm.org/download.html

yes, we can build it. I have the scripts.
https://github.com/llvm/llvm-project/releases/tag/llvmorg-21.1.0 last version for example

hughperkins:
> > > > https://releases.llvm.org/download.html
> > > 
> > > 
> > > yes, we can build it. I have the scripts. https://github.com/llvm/llvm-project/releases/tag/llvmorg-21.1.0 last version for example
> > 
> > 
> > Ok. Yes, going directly for llvm 21.1.0 sounds interesting to me, given that it sounds like it is already working for you, in your personal development branches? But we will see: if it involves changing a lot of performance-related sections of the code, it might be tempting to break the upgrade into smaller bite-sized stages. Let's start with having a build for llvm-21.1.0, and then seeing how PRs to upgrade to llvm-21.1.0 look.
> 
> i didn't test it llvm21, it released 2 days ago... las version that i used was llvm20.1

Oh ok. Perhaps lets start with llvm20.1 then?

.github/workflows/scripts/ti_build/vulkan.py
johnnynunez: yes, i miss there, i will change it

.github/workflows/scripts/ti_build/entry.py
johnnynunez: i know, manylinux_2014 is deprecated

.github/workflows/scripts/ti_build/entry.py
johnnynunez: possible

.github/workflows/scripts/ti_build/llvm.py
johnnynunez: not.. we have to do it for 4(windows. windows-arm, ubuntu, ubuntu-arm)

.github/workflows/scripts/ti_build/llvm.py
johnnynunez: i prefer not to remove it, this is possible to build taichi with local llvm(installed in your SO host)

.github/workflows/scripts/ti_build/vulkan.py
johnnynunez: i don't know it, i used it, maybe your gstaichi was change there

.github/workflows/scripts/ti_build/llvm.py
hughperkins: I feel like this should be parametrized via CMAKE somehow, rather than by using commented out code in the python scripts. Whilst it might be obvious to you that one should navigate to llvm.py, and uncomment out these lines, I feel like this would not not be obvious to other people.

hughperkins:
Let's keep this open for now :) Since, as you say, arm isnt done yet.

PR #171, opened 2025-09-06
[LLVM upgrade][PART 4] codegen cpu/cuda

johnnynunez:
llvm20

### DIFF
--- a/gstaichi/codegen/llvm/struct_llvm.cpp
+++ b/gstaichi/codegen/llvm/struct_llvm.cpp
@@ -105,14 +105,14 @@ void StructCompilerLLVM::generate_types(SNode &snode) {
     // mutex
     aux_type = llvm::ArrayType::get(llvm::PointerType::getInt64Ty(*ctx),
                                     snode.max_num_elements());
-    body_type = llvm::ArrayType::get(llvm::PointerType::getInt8PtrTy(*ctx),
+    body_type = llvm::ArrayType::get(llvm::PointerType::get(*ctx, 0),
                                      snode.max_num_elements());
   } else if (type == SNodeType::dynamic) {
     // mutex and n (number of elements)
     aux_type =
         llvm::StructType::get(*ctx, {llvm::PointerType::getInt32Ty(*ctx),
                                      llvm::PointerType::getInt32Ty(*ctx)});
-    body_type = llvm::PointerType::getInt8PtrTy(*ctx);
+    body_type = llvm::PointerType::get(*ctx, 0);
   } else {
     TI_P(snode.type_name());
     TI_NOT_IMPLEMENTED;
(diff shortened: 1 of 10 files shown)


gstaichi/codegen/cpu/codegen_cpu.cpp
```diff
@@ -53,7 +58,7 @@ class TaskCodeGenCPU : public TaskCodeGenLLVM {
     {
       auto guard = get_function_creation_guard(
           {llvm::PointerType::get(get_runtime_type("RuntimeContext"), 0),
-           llvm::Type::getInt8PtrTy(*llvm_context),
+           llvm::PointerType::get(*llvm_context, 0),
```
hughperkins, line 61: Do you know if this is compatible with older vesrions of LLVM? Do you know what is the lowest version of LLVM this works for?

gstaichi/codegen/cpu/codegen_cpu.cpp
```diff
@@ -262,76 +267,85 @@ void KernelCodeGenCPU::optimize_module(llvm::Module *module) {
     options.NoInfsFPMath = 0;
     options.NoNaNsFPMath = 0;
   }
+
   options.HonorSignDependentRoundingFPMathOption = false;
   options.NoZerosInBSS = false;
   options.GuaranteedTailCallOpt = false;
 
-  llvm::legacy::FunctionPassManager function_pass_manager(module);
```
hughperkins, line 269: What are the implications of removing these? Not very keen on removing things without a fairly exhaustive descirption please of:
- what they do
- possible consequences of removing them

gstaichi/codegen/cpu/codegen_cpu.cpp
```diff
@@ -262,76 +267,85 @@ void KernelCodeGenCPU::optimize_module(llvm::Module *module) {
...
+#if LLVM_VERSION_MAJOR >= 18
+  const auto opt_level = llvm::CodeGenOptLevel::Aggressive;
+#else
+  const auto opt_level = llvm::CodeGenOpt::Aggressive;
+#endif
   llvm::StringRef mcpu = llvm::sys::getHostCPUName();
   std::unique_ptr<llvm::TargetMachine> target_machine(
       target->createTargetMachine(triple.str(), mcpu.str(), "", options,
                                   llvm::Reloc::PIC_, llvm::CodeModel::Small,
-                                  llvm::CodeGenOpt::Aggressive));
```
hughperkins, line 276: ditto for adding things really. If we are going to change things that affect performance, I think we need to run regression tests for before/after the changes.

On the whole, I'd prefer to change as little as possible please. The minimum necessary to get llvm20 working.

We can have a separate PR for performance tweaks, and run various performance benchmarks as part of that PR potentially.

gstaichi/codegen/cpu/codegen_cpu.cpp
```diff
@@ -262,76 +267,85 @@ void KernelCodeGenCPU::optimize_module(llvm::Module *module) {
...
-
+                                  opt_level));
   TI_ERROR_UNLESS(target_machine.get(), "Could not allocate target machine!");
 
+  module->setTargetTriple(triple.str());
   module->setDataLayout(target_machine->createDataLayout());
 
-  module_pass_manager.add(llvm::createTargetTransformInfoWrapperPass(
```
hughperkins, line 282: same observations as above hold for the rest of this file please.

gstaichi/codegen/llvm/llvm_codegen_utils.cpp
```diff
@@ -29,11 +29,17 @@ bool is_same_type(llvm::Type *a, llvm::Type *b) {
     return false;
   }
   if (a->isPointerTy()) {
+#if 0
     return is_same_type(a->getPointerElementType(), b->getPointerElementType());
+#else
```
hughperkins, line 34: wait, you're no longer checking the types? 🤔

johnnynunez:
> Is there anything in this PR that is still needed now we upgraded to LLVM 18?

I don't remember but... we can close it, if we need it we have the pr

PR #172, opened 2025-09-06
[bug] Fix ndarray crash after ti reset

hughperkins:
Issue: #

### Brief Summary

Summary: Fix intermittent ndarray crash after ti reset

Details:
- when an ndarray object is created:
    - a c++ object is created
    - a python object is created, which points to the c++ object
- when the python object is gc'd
    - the c++ object is deleted too
- when we call ti.init():
     - python objects still hang around for a while, until they are gc'd
     - the c++ objects are deleted
     - new ndarray c++ objects are often allocated with the exact same address as the old one
 - later, when the old python objects are gc'd, the c++ object they point to - which has since been replaced with a new c++ object - gets deleted
      - now, when new python ndarray objects try to access their c++ object, we get a crash, since those no longer exist

this pr adds a registration system for python-side ndarrays, and when we call ti.reset, the python side ndarrays are stripped of any links with the c++ object. This should address the issue

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_reset_ndarrays.py
+++ b/tests/python/test_reset_ndarrays.py
@@ -9,7 +9,7 @@ from tests import test_utils
 
 
 @test_utils.test(arch=[ti.cpu])
-def test_ndarray_reset() -> None:
+def test_ndarray_doesnt_crash_on_gc() -> None:
     if sys.platform != "darwin":
         pytest.skip("Only need to check Mac CPU")
     arch = getattr(ti, ti.cfg.arch.name)
@@ -44,3 +44,55 @@ def test_ndarray_reset() -> None:
             assert len(v.arr.shape) > 0, f"{v} it {n}"
         gc.collect()
         k1(z_param)
+
+
+@test_utils.test()
+def test_ndarray_reset() -> None:
+    arch = getattr(ti, ti.cfg.arch.name)
+    ti.reset()
+    ti.init(arch=arch)
+    rt = ti.lang.impl.get_runtime()
+    assert len(rt.ndarrays) == 0
+    a = ti.ndarray(ti.i32, shape=(55,))
+    assert len(rt.ndarrays) == 1
+    b = ti.ndarray(ti.i32, shape=(57,))
+    assert len(rt.ndarrays) == 2
+    c = ti.ndarray(ti.i32, shape=(211,))
+    assert len(rt.ndarrays) == 3
+    z_param = ti.ndarray(ti.i32, shape=(223,))
+    assert len(rt.ndarrays) == 4
+    bar_param = ti.ndarray(ti.i32, shape=(227,))
+    assert len(rt.ndarrays) == 5
+
+    @ti.kernel
+    def k1(z_param2: ti.types.NDArray[ti.i32, 1]) -> None:
+        z_param2[33] += 2
+
+    k1(z_param)
+    assert len(rt.ndarrays) == 5
+
+    ti.reset()
+    rt = ti.lang.impl.get_runtime()
+    assert len(rt.ndarrays) == 0
+
+    assert a.arr is None
+    assert b.arr is None
+    assert c.arr is None
+    assert z_param.arr is None
+    assert bar_param.arr is None
+
+    assert a.shape is None
+
+    gc.collect()
+    gc.collect()
+
+    rt = ti.lang.impl.get_runtime()
+    assert len(rt.ndarrays) == 0
+
+    assert a.arr is None
+    assert b.arr is None
+    assert c.arr is None
+    assert z_param.arr is None
+    assert bar_param.arr is None
+
+    assert a.shape is None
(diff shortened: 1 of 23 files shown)


hughperkins:
Note that I just realized that this will prevent ndarrays being garbage collected, in between calls to reset. Should probably make it a weak refeence.

duburcqa (approved):

tests/python/test_reset_ndarrays.py
```diff
@@ -0,0 +1,46 @@
...
+
+import pytest
+
+import gstaichi as ti
+
+from tests import test_utils
+
+
+@test_utils.test(arch=[ti.cpu])
+def test_ndarray_reset() -> None:
```
duburcqa: It may be a good idea to assert the number of ndarray object being registered?

hughperkins: Sure. Can you elaborate a little on what kind of error you are imagining might happen without this? You mean, like, inadvertently registering twice? and/or not registering at all?

duburcqa: We want to make sure we can release memory properly, to avoid OOM errors, and just because it should be part of our specifications that unused memory should be reclaimed. In practice, this is important if a user creates a scene, then delete and create another one, repeatedly. This is an actual use case. Not just a thought experiment.

hughperkins: Sure. But how do you see "releasing memory properly" relate to "asserting the number of ndarray objects being registered"?

hughperkins: Oh, perhaps this was a response to "Could you update the unit test to change that your patch does not prevent garbage collection in between two reset calls?"?

hughperkins: Addressed in 54356f1fc

tests/python/test_reset_ndarrays.py
```diff
@@ -0,0 +1,98 @@
...
+    a = ti.ndarray(ti.i32, shape=(55,))
+    assert len(rt.ndarrays) == 1
+    b = ti.ndarray(ti.i32, shape=(57,))
+    assert len(rt.ndarrays) == 2
+    c = ti.ndarray(ti.i32, shape=(211,))
+    assert len(rt.ndarrays) == 3
+    z_param = ti.ndarray(ti.i32, shape=(223,))
+    assert len(rt.ndarrays) == 4
+    bar_param = ti.ndarray(ti.i32, shape=(227,))
+    assert len(rt.ndarrays) == 5
```
duburcqa, line 65: I would add:
```
tmp = ti.ndarray(ti.i32, shape=(42,))
assert len(rt.ndarrays) == 6
del tmp
gc.collect()
assert len(rt.ndarrays) == 5
```

tests/python/test_reset_ndarrays.py
```diff
@@ -0,0 +1,98 @@
...
+    rt = ti.lang.impl.get_runtime()
+    assert len(rt.ndarrays) == 0
+
+    assert a.arr is None
+    assert b.arr is None
+    assert c.arr is None
+    assert z_param.arr is None
+    assert bar_param.arr is None
+
+    assert a.shape is None
```
duburcqa: I don't see the point of this section.

hughperkins: Tihs is the thing we are actually changing in this PR. The `.arr` is the c++ obejct that was being deleted, and causing the crash.

tests/python/test_reset_ndarrays.py
hughperkins: addressed in 8cd8153a2

tests/python/test_reset_ndarrays.py
duburcqa: But you are already checking it is none right better no?

hughperkins: I'm not sure I follow. What are you saying I am already checking is None?

duburcqa: You are checking twice the same thing without doing anything in between:

```python
    rt = ti.lang.impl.get_runtime()
    assert len(rt.ndarrays) == 0
    
    assert a.arr is None
    assert b.arr is None
    assert c.arr is None
    assert z_param.arr is None
    assert bar_param.arr is None

    assert a.shape is None

    gc.collect()
    gc.collect()

    rt = ti.lang.impl.get_runtime()
    assert len(rt.ndarrays) == 0

    assert a.arr is None
    assert b.arr is None
    assert c.arr is None
    assert z_param.arr is None
    assert bar_param.arr is None
```

hughperkins: Oh I see 🤔

hughperkins: pondering...

hughperkins: Fair enough .addressed in 0eed7d8a7

duburcqa (approved):

PR #176, opened 2025-09-09
[Lang] Allow static if expression as top level in for loop iterator

hughperkins:
Issue: #

### Brief Summary

Allow static if expression as top level in for loop iterator expression

Concretely, this allows structures such as:
```
for i in range(foo) if ti.static(some_flag) else range(ti.static(bar)):
```

this appears to generalize to:
- being an inner loop, e.g.
```
for b in range(B):
    for i in range(foo) if ti.static(some_flag) else range(ti.static(bar)):
```
- either side of the for iterator can be static or not, as long as the `if` expression itself is static

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/ast/ast_transformer.py
+++ b/python/gstaichi/lang/ast/ast_transformer.py
@@ -1108,9 +1108,10 @@ class ASTTransformer(Builder):
             ):
                 return ASTTransformer.build_range_for(ctx, node)
             elif isinstance(node.iter, ast.IfExp):
-                # lets just "hard-code" this for now...
-                # here we are handling cases like
-                # for i in range(foo) if ti.static(some_flag) else ti.static(range(bar))
+                # handle inline if expression as the top level iterator expression, e.g.:
+                #
+                #   for i in range(foo) if ti.static(some_flag) else ti.static(range(bar))
+                #
                 # this appears to generalize to:
                 # - being an inner loop
                 # - either side can be static or not, as long as the if expression itself is static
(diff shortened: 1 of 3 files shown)


hughperkins:
note: Need to add unit tests before taking out of 'draft'

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -1107,6 +1107,31 @@ def build_For(ctx: ASTTransformerContext, node: ast.For) -> None:
...
+                # here we are handling cases like
+                # for i in range(foo) if ti.static(some_flag) else ti.static(range(bar))
+                # this appears to generalize to:
+                # - being an inner loop
+                # - either side can be static or not, as long as the if expression itself is static
+                _iter = node.iter
+                build_stmt(ctx, node.iter)
+                is_static_if = get_decorator(ctx, node.iter.test) == "static"
+                if not is_static_if:
+                    raise GsTaichiSyntaxError("using non static if with for not currently supported")
```
duburcqa: "Using non static inlined if statement as for-loop iterable is not currently supported."

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -1107,6 +1107,31 @@ def build_For(ctx: ASTTransformerContext, node: ast.For) -> None:
                 and node.iter.func.id == "range"
             ):
                 return ASTTransformer.build_range_for(ctx, node)
+            elif isinstance(node.iter, ast.IfExp):
+                # lets just "hard-code" this for now...
```
duburcqa: Either is should have some kind of TODO tag, or `for now` should be removed, because it is clearly here to stay if there is no nothing to remain you about this special case handling.

YilingQiao (approved):

tests/python/test_for.py
```diff
@@ -0,0 +1,80 @@
...
+
+    V = ti.field if use_field else ti.ndarray
+    V_ANNOT = ti.Template if use_field else ti.types.NDArray[ti.i32, 2]
+
+    if is_inner:
+
+        @ti.kernel
+        def k1(a: V_ANNOT, n_left: ti.i32) -> None:
+            for b in range(B):
+                for i in range(n_left) if ti.static(static_value) else ti.static(range(N_right)):
```
YilingQiao, line 32: nice

duburcqa (approved):

PR #177, opened 2025-09-10
[Perf] Cache template values for fast cache

hughperkins:
Issue: #

### Brief Summary

When we call a kernel, and that kernel has primitive parmaeters annotated as ti.Template(), the kernel is compiled whenever the values passed in to those parmaeters change. However, the fast cache was not taking this into acccount. This PR:
- adds tests for this scenario
- fixes the fast cache cache key algo to take into account hte values of primitive kernel parameters annotated as ti.template

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_fast_caching/args_hasher.py
+++ b/python/gstaichi/lang/_fast_caching/args_hasher.py
@@ -30,7 +30,7 @@ def dataclass_to_repr(path: tuple[str, ...], arg: Any) -> str:
     repr_l = []
     for field in dataclasses.fields(arg):
         child_value = getattr(arg, field.name)
-        _repr = stringify_obj_type(path + (field.name,), child_value)
+        _repr = stringify_obj_type(path + (field.name,), child_value, None)
         full_repr = f"{field.name}: ({_repr})"
         if field.metadata.get(FIELD_METADATA_CACHE_VALUE, False):
             full_repr += f" = {child_value}"
@@ -112,7 +112,7 @@ def stringify_obj_type(path: tuple[str, ...], obj: object, arg_meta: ArgMetadata
     return None
 
 
-def hash_args(args: Sequence[Any], arg_metas: Sequence[ArgMetadata]) -> str | None:
+def hash_args(args: Sequence[Any], arg_metas: Sequence[ArgMetadata | None]) -> str | None:
     global g_num_calls, g_num_args, g_hashing_time, g_repr_time, g_num_ignored_calls
     g_num_calls += 1
     g_num_args += len(args)
--- a/python/gstaichi/lang/_fast_caching/args_hasher.py
+++ b/python/gstaichi/lang/_fast_caching/args_hasher.py
@@ -4,18 +4,17 @@ import numbers
 import time
 from typing import Any, Sequence
 
-from gstaichi.types.annotations import Template
 import numpy as np
 
 from gstaichi import _logging
+from gstaichi.types.annotations import Template
 
 from .._ndarray import ScalarNdarray
 from ..field import ScalarField
+from ..kernel_arguments import ArgMetadata
 from ..matrix import MatrixField, MatrixNdarray, VectorNdarray
 from ..util import is_data_oriented
 from .hash_utils import hash_iterable_strings
-from ..kernel_arguments import ArgMetadata
-
 
 g_num_calls = 0
 g_num_args = 0
@@ -119,7 +118,9 @@ def hash_args(args: Sequence[Any], arg_metas: Sequence[ArgMetadata]) -> str | No
     g_num_args += len(args)
     hash_l = []
     if len(args) != len(arg_metas):
-        raise RuntimeError(f"Number of args passed in {len(args)} doesnt match number of declared args {len(arg_metas)}")
+        raise RuntimeError(
+            f"Number of args passed in {len(args)} doesnt match number of declared args {len(arg_metas)}"
+        )
     for i_arg, arg in enumerate(args):
         start = time.time()
         arg_meta = arg_metas[i_arg]
--- a/python/gstaichi/lang/_fast_caching/args_hasher.py
+++ b/python/gstaichi/lang/_fast_caching/args_hasher.py
@@ -4,6 +4,7 @@ import numbers
 import time
 from typing import Any, Sequence
 
+from gstaichi.types.annotations import Template
 import numpy as np
 
 from gstaichi import _logging
@@ -13,6 +14,8 @@ from ..field import ScalarField
 from ..matrix import MatrixField, MatrixNdarray, VectorNdarray
 from ..util import is_data_oriented
 from .hash_utils import hash_iterable_strings
+from ..kernel_arguments import ArgMetadata
+
 
 g_num_calls = 0
 g_num_args = 0
@@ -36,7 +39,14 @@ def dataclass_to_repr(path: tuple[str, ...], arg: Any) -> str:
     return "[" + ",".join(repr_l) + "]"
 
 
-def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
+def _is_template(arg_meta: ArgMetadata | None) -> bool:
+    if arg_meta is None:
+        return False
+    annot = arg_meta.annotation
+    return annot is Template or isinstance(annot, Template)
+
+
+def stringify_obj_type(path: tuple[str, ...], obj: object, arg_meta: ArgMetadata | None) -> str | None:
     """
     Convert an object into a string representation that only depends on its type.
 
@@ -45,6 +55,9 @@ def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
     with different (allowed) types.
 
     Note that fields are not included in fast cache.
+
+    arg_meta should only be non-None for the top level arguments. e.g. counter-examples:
+    children of data_oriented, and children of dataclasses should have arg_meta set to None.
     """
     arg_type = type(obj)
     if isinstance(obj, ScalarNdarray):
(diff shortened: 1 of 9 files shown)


python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -28,15 +30,22 @@ def dataclass_to_repr(path: tuple[str, ...], arg: Any) -> str:
     repr_l = []
     for field in dataclasses.fields(arg):
         child_value = getattr(arg, field.name)
-        _repr = stringify_obj_type(path + (field.name,), child_value)
+        _repr = stringify_obj_type(path + (field.name,), child_value, None)
```
duburcqa: I suggest always using keyword arguments for "inlined value" (not stored in some temporary variable with a proper name).

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -94,14 +112,19 @@ def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
...
     hash_l = []
+    if len(args) != len(arg_metas):
+        raise RuntimeError(
+            f"Number of args passed in {len(args)} doesnt match number of declared args {len(arg_metas)}"
+        )
     for i_arg, arg in enumerate(args):
         start = time.time()
-        _hash = stringify_obj_type((str(i_arg),), arg)
+        arg_meta = arg_metas[i_arg]
+        _hash = stringify_obj_type((str(i_arg),), arg, arg_meta)
```
duburcqa: `_hash = stringify_obj_type((str(i_arg),), arg, arg_metas[i_arg])` is good enough I think. But as you want.

duburcqa (approved):

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: what do you mean by 'inlined value' in this context? What change are you proposing, concretely?

duburcqa: > what do you mean by 'inlined value' in this context
Exactly what is explained in parentheses: "[arguments whose values are] not stored in some temporary variable with a proper name"

> What change are you proposing, concretely?
```python
_repr = stringify_obj_type(path + (field.name,), child_value, arg_meta=None)
```

duburcqa: Another option is to male `arg_meta=None` the default value, so that you could just do:
```python
_repr = stringify_obj_type(path + (field.name,), child_value)
```

hughperkins: ` arg_meta=None)` => I see, thanks! Makes sense :)

> Another option is to male arg_meta=None

Yeah, I'm not a fan of default None, because when I forget to pass a parameter in, the IDE doesn't warn me.

tests/python/gstaichi/lang/fast_caching/test_src_ll_cache.py
```diff
@@ -139,3 +144,69 @@ def k1() -> None:
...
+        assert TEST_RAN in subprocess.check_output(
+            [sys.executable, __file__, src_ll_cache_template_params_child.__name__, create_args(a)]
+        ).decode("utf-8")
+
+
+# The following lines are critical for the tests to work. If they are missing, the test will
+# incorrectly pass, without doing anything.
+if __name__ == "__main__":
+    print("__main__")
+    globals()[sys.argv[1]](sys.argv[2:])
```
duburcqa: I would recommend to rather use a specific exit code in case of success. I would suggest 42:
https://unix.stackexchange.com/a/604262/406277

duburcqa: To do so, just use `sys.exit(42)`

hughperkins: I'd prefer to use text output on the whole, because it's more human friendly I feel.

I think the main advantage of using a return code would be performance? But for these tests, the performance gain here would be minimal.

Or ... what do you see as advantages to using a magic return code?

duburcqa: I like more magic code because it is basically designed for this: you have a program, you want to execute it, and then check that it exited with the correct status code. 

Using a string approach is more indirect and "hacky" in my view. I don't like relying on the standard output / error to check if something is valid. It just feels fragile and it scares me a lot. I had too much trouble in the past with that. Not to mention all the weird python and terminal modes that completely messes up the standard output / error via so environment variables and the like.

duburcqa: But if it works in your context, then I'm fine. You are already using stdout at many places in your unit tests to check if it ran successfully, usually because you have no simple alternative options when dealing with kernels. So here I think it is consistent with what you are already doing, and I'm fine with it. Anyway, I do not take the implementation of unit tests as the actual code base that will be running on user side. You can choose your preferred option.

duburcqa (approved):

tests/python/gstaichi/lang/fast_caching/test_src_ll_cache.py
hughperkins: Added return code check

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -28,15 +30,22 @@ def dataclass_to_repr(path: tuple[str, ...], arg: Any) -> str:
...
         if field.metadata.get(FIELD_METADATA_CACHE_VALUE, False):
             full_repr += f" = {child_value}"
         repr_l.append(full_repr)
     return "[" + ",".join(repr_l) + "]"
 
 
-def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
+def _is_template(arg_meta: ArgMetadata | None) -> bool:
+    if arg_meta is None:
+        return False
```
yun-long, line 43: if `arg_meta` is None, why calling this function in the first place? is it suppose to be a feature

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -45,6 +54,9 @@ def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
     with different (allowed) types.
 
     Note that fields are not included in fast cache.
+
+    arg_meta should only be non-None for the top level arguments. e.g. counter-examples:
+    children of data_oriented, and children of dataclasses should have arg_meta set to None.
```
yun-long, line 59: consider following one specific docstring formatter/style?
e.g., Google style

```python3
def foo(bar: int, baz: str = "hi") -> bool:
    """
    Do something with bar and baz.

    Args:
        bar (int): The main integer input.
        baz (str, optional): An optional string. Defaults to "hi".

    Returns:
        bool: True if successful, False otherwise.

    Raises:
        ValueError: If bar is negative.
    """
```

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -94,14 +112,18 @@ def stringify_obj_type(path: tuple[str, ...], obj: object) -> str | None:
     return None
 
 
-def hash_args(args: Sequence[Any]) -> str | None:
+def hash_args(args: Sequence[Any], arg_metas: Sequence[ArgMetadata | None]) -> str | None:
```
yun-long, line 115: what is the difference between `args` and `arg_metas`. They look the same to me. Maybe add some descriptions

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: need an `if` somewhere. Maybe it's nice to factorize the `if` into this function, so we don't have to copy and paste it every time we call this function?

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: `args` is the actual python objects passed in as parameters, when calling the functions.

Originally, there was only the one parameter, so `args` made sense. Happy to rename it to e.g. `py_args`?

hughperkins: (though I think that `arg_metas` is itself fairly descriptive of what arg_metas is? And args is the actual arguments?)

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: That would force the comments to be fairly verbose. In general, the types on the arguments - combined with the name - are already fairly descriptive I feel.

For example, in the above `baz (str, optional): An optional string. Defaults to "hi".` => this adds no new information that isn't already given by `baz: str = "hi"`

hughperkins: well, actually, that type in the example is incorrect. It should be:

`baz: str | None = "hi"`

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: ("meta" means "about" by the way; "metadata" is "data about data").

yun-long: normally, `args` already means `metadata`, or no?

python/gstaichi/lang/_fast_caching/args_hasher.py
yun-long: I have OCD. so, I prefer everything looks organized. but you call it.

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: No, args almost always measn the passed in args. eg.

```
def foo(some_int, *args, **kwargs):
    print(some_int, args, kwargs)

foo(123, 4, 5, 6, hello="world", paris="expensive")
```

```
$ python /tmp/test.py
123 (4, 5, 6) {'hello': 'world', 'paris': 'expensive'}
```

hughperkins: metadata, for args, is things like "the first argument is an integer, the second is an optional string".

hughperkins: (like, from the typing annotatinos basically).

PR #178, opened 2025-09-10
[build] Ruff now checks for unused imports

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/scripts/ti_build/misc.py
+++ b/.github/workflows/scripts/ti_build/misc.py
@@ -9,7 +9,7 @@ import sys
 from pathlib import Path
 from typing import Any, Callable
 
-from .bootstrap import get_cache_home  # noqa This is a re-export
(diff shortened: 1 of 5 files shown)


misc/ci_setup.py
```diff
@@ -160,7 +160,7 @@ def run(self):
 
         check_command_existence("wget")
         try:
-            import pip
+            import pip  # noqa
```
CharlesMasson:
```suggestion
            import pip  # noqa: F401
```

You can use this to specify the rule that's ignored. It may be useful to still apply the other rules.

CharlesMasson (approved):

duburcqa (approved):

misc/ci_setup.py
duburcqa: I would definitely recommend doing this

PR #179, opened 2025-09-10
[build] Migrate build metadata to pyproject.toml

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ license = "Apache-2.0"
 keywords = ["graphics", "simulation"]
 dynamic = ["version"]
 readme = "README.md"
-requires-python = ">=3.10,<=3.13"
+requires-python = ">=3.10,<3.14"
 dependencies = [
     "numpy>=1.26.0",
     "colorama",


hughperkins:
Oh, the version builds incorrectly. setting to draft

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+    "setuptools>=68.0.0",  # Required for Python 3.12+ compatibility
+    "cffi>=1.16.0",
+]
+
+[tool.setuptools_scm]
+write_to = "python/gstaichi/_version.py"
+write_to_template = "__version__ = '{version}'\n"
+
+[project.optional-dependencies]
+docs = [
```
CharlesMasson, line 37: A [dependency group](https://peps.python.org/pep-0735) may be more appropriate, given that I don't think we should publish this as an extra.

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+keywords = ["graphics", "simulation"]
+dynamic = ["version"]
+readme = "README.md"
+requires-python = ">=3.10,<4.0"
+dependencies = [
+    "numpy",
+    "colorama",
+    "dill",
+    "pydantic",
+    "rich",
```
CharlesMasson: I'd recommend setting lower bound for all of them. Otherwise, there is no guarantee that the code is actually compatible with the dependency versions that are installed.

hughperkins: Thoughts on what minimum versions to put?

I think our philosophy in general, is to not put a lower bound, and then if we specifically encounter an actual concrete issue with a lower version, then put in a bounds at that point.

(In an internal repo, we would simply lock to whatever we have installed right now, as the lower, but since this is an external repo, we want more flexibility I think?)

@duburcqa your thoughts on this point?

duburcqa: I have the same opinion than @hughperkins . Personally, I would do a small audit to determine whether some "old" version is not supported for some obvious reason (typically numpy 1.x vs numpy 2.x), but requiring support of Python>=3.8 is already restricting the oldest version that may be installed anyway.

duburcqa: In practice I have lower bound for half of my dependencies in Jiminy:
https://github.com/duburcqa/jiminy/blob/dev/python/jiminy_py/setup.py

hughperkins: Cool. Do either of you see any obvious old versions I should guard against? (I've updated numpy to be `>=2.0.0`, just now)

hughperkins: maybe pydantic >= 2.0.0?
Screenshot 2025-09-11 at 8 14 26 AM

hughperkins: >= 0.4.3 for colorama?

hughperkins: Screenshot 2025-09-11 at 8 15 22 AM

hughperkins: Added lowerobudns for all, along with comments justifying the choice

pyproject.toml
hughperkins: Addressed in a zillion commits I just pushed.

hughperkins: I guess https://github.com/Genesis-Embodied-AI/gstaichi/pull/179/commits/e7920ca98ad6056e18da3f07022a412b6351e9c1#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711 primarily

pyproject.toml
```diff
@@ -1,3 +1,105 @@
...
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+]
+license = "Apache-2.0"
+keywords = ["graphics", "simulation"]
+dynamic = ["version"]
+readme = "README.md"
+requires-python = ">=3.10,<4.0"
+dependencies = [
+    "numpy>=2.0.0",
```
duburcqa: I don't think numpy >= 2.0 should be a requirement for gstaichi. It is not even a requirement for genesis. 1.26 works just fine.

pyproject.toml
```diff
@@ -1,3 +1,105 @@
...
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+]
+license = "Apache-2.0"
+keywords = ["graphics", "simulation"]
+dynamic = ["version"]
+readme = "README.md"
+requires-python = ">=3.10,<4.0"
```
duburcqa: I don't think you properly support Python 3.14 that introduces free-threaded (aka no-gil) mode, among other major changes such as official JIT compilation support.

pyproject.toml
```diff
@@ -1,3 +1,105 @@
...
+dependencies = [
+    "numpy>=2.0.0",
+    "colorama>=0.4.3",  # 2019
+    "dill>=0.3.1.1",  # 2019
```
duburcqa: These requirements seem artificial. Not a great fan.

pyproject.toml
hughperkins: Well, I got that from:
Screenshot 2025-09-12 at 8 30 07 AM

where I inferred you said it's important to clarify whether we are using numpy 1 or numpy 2?

still, it is a requirement in requirements_test.txt for us though https://github.com/Genesis-Embodied-AI/gstaichi/blob/da95c6f9ed48b0d91dc54fe4522ebc0082db9ce9/requirements_test.txt

I can try removing it though, from `dependencies`.

hughperkins: UPdated to 1.26.0

pyproject.toml
hughperkins: no-gil mode is optional I think? It's a separate executable.

in general, I think that moving to future python versions appears to be generally fairly safe?

what advantage do you see in preventing people from installing with later versions?

pyproject.toml
hughperkins: I don't have an opinoin on these requirements. What are your recommendations on the lower bound for these?

pyproject.toml
duburcqa: It was just an example, because many libraries are still enforcing 'numpy<2.0' (notably Genesis in March), but usually libraries that support 'numpy>=2.0' are backward compatible

hughperkins: ok. what are your thoughts on what I put as lowerobund on numpy? (if any)

pyproject.toml
duburcqa: Fair enough. But it won't be the case anymore for Python 3.15.

> what advantage do you see in preventing people from installing with later versions?

What you allow is what you support. If you explicitly state that Python 3.15 will be supported, then you need to run CI tests and make sure it is, and support all the features it provides or guard against their use.

If you pretend that Python 3.14 is supported, you should run the CI on it. It is already supported by `actions/setup-python@v4` as `3.14-dev`

hughperkins: I feel like this contradicts your approach to backwrds compatbility, where you assert taht we should support older versions we cannot test.

Personally, my own opinion leans more towwards "if you state that [some version] is supproted then you need to run CI tests and make sure it is". This would mean for example, supporting mac os 14 and above, and not earlier mac os versions, for example.

What is your reason for feeling relaxed for backwards compatibility, but stricter towards forwards compatibilty?

duburcqa: I do not consider Python itself as a Python module dependency and therefore I don’t have the same requirement for it. Being too restrictive for module dependencies makes the life much harder for people using your library, while most often there is no need to be so restrictive. For Python, I tend to feel my experience that the probabilities are inverted: not supporting fresh new versions that are 2 weeks old is not a big deal since away your deps do no support it either, while the risk that it may break something is higher, especially nowadays with Python moving quite fast with the introduction for some major features.

pyproject.toml
duburcqa: Not having any in this case.

pyproject.toml
duburcqa: 1.26 is fine, anyway it is not possible to install anything older on Python 3.8+ if I remember correctly.

duburcqa: You need to enforce numpy 2.x in dev group to allow building wheels that support both numpy 1.x and numpy 2.x

hughperkins:
```
You need to enforce numpy 2.x in dev group to allow building wheels that support both numpy 1.x and numpy 2.x
```

Not following this. The current `requirements_dev.txt` makes no such enforcement.

On the whole, would prefer to not start changing the behavior of the build in a "migration" PR. Migration is already challenging enough I feel, and has high risk of regressions.

pyproject.toml
hughperkins: addressed in 73d7ba03d

pyproject.toml
hughperkins: addressed in 73d7ba03d

pyproject.toml
duburcqa: Then your requirements are broken. If you don’t want to follow this, you should enforce ‘numpy<2.0’ for the entire project. Otherwise I cannot approve something this wrong. But you can ask yiling approval if you prefer!

hughperkins: We test only with numpy >= 2.0.0.

So clealry it does run with numpy >= 2.0.0

hughperkins: For example my local env contains 2.x.x:

Screenshot 2025-09-12 at 9 30 50 AM

hughperkins: Why would enforcing >=2.0.0 in dev be required in order to allow running on numpy < 2.0.0?

hughperkins: numpy 1.26 is deprecated in september: https://scientific-python.org/specs/spec-0000/

hughperkins: I suppose I'm contradicting myself. Since we are already building using numpy>=2.0.0, I suppose there should be no issue in specifyin that in `dev`. I disagree however with "you should enforce ‘numpy<2.0’ for the entire project", since we are testing with "numpy>=2.0.0".

hughperkins: addressed in e79d35feb

PR #181, opened 2025-09-11
[Perf] Add ti.wipe_offline_cache()

hughperkins:
Issue: #

### Brief Summary

Add ti.wipe_offline_cache(), that wipes all offline cache files, using PTX, and LLVM IR files.

Use like:

```
ti.wipe_offline_cache()
```

Added unit test for this.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/__init__.py
+++ b/python/gstaichi/__init__.py
@@ -22,7 +22,7 @@ from gstaichi.types.primitive_types import *  # type: ignore
 
 def __getattr__(attr):
     if attr == "cfg":
-        return None if lang.impl.get_runtime()._prog is None else lang.impl.current_cfg()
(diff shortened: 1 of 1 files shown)


python/gstaichi/__init__.py
```diff
@@ -43,6 +42,7 @@ def __getattr__(attr):
     "ad",
     "algorithms",
     "experimental",
+    "lang",
```
hughperkins: Hmmm, I wonder if I should be exporting `lang`? But If I don't export `lang`, how to access the runtime object (PyGsTaichi) ? 🤔

hughperkins: Ok, I moved it to `ti.wipe_offline_cache()`, and added it to `test_api.py`

PR #190, opened 2025-09-13
[lang] Add pure parameter to ti.kernel

hughperkins:
Issue: #

### Brief Summary

- Minimal changes to get pure kernel parameter working for now
- Note that changes to def kernel nerfed slightly to avoid fixing pyright typing when calling kernels, since this would open a can of works about fixing ndarray typing
- adds tests for both data oriented kernels, and non data oriented (the logic, and stack depth, are slightly different)
- adds sanity test for pyright on kernel calls to ndarray, since pyright isnt actually run against any other unit tests currentlyy
- args hash for empty data oriented object is now valid
     - this mostly comes into play for unit tests which use an empty data oriented object, and this is then passed in as `self` into the kernels

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_fast_caching/args_hasher.py
+++ b/python/gstaichi/lang/_fast_caching/args_hasher.py
@@ -51,7 +51,7 @@ def stringify_obj_type(path: tuple[str, ...], obj: object, arg_meta: ArgMetadata
 
     String should somehow represent the type of obj. Doesnt have to be hashed, nor does it have
     to be the actual python type string, just a string that is representative of the type, and won't collide
-    with different (allowed) types.
+    with different (allowed) types. String should be non-empty.
 
     Note that fields are not included in fast cache.
 


python/gstaichi/lang/kernel_impl.py
```diff
@@ -1273,14 +1291,6 @@ def kernel(fn: Callable):
 
     Kernel's gradient kernel would be generated automatically by the AutoDiff system.
 
-    See also https://docs.taichi-lang.org/docs/syntax#kernel.
```
hughperkins, line 1276: remoed this, since not our doc page.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1273,14 +1291,6 @@ def kernel(fn: Callable):
...
-
-    Args:
-        fn (Callable): the Python function to be decorated
-
-    Returns:
```
hughperkins, line 1281: removed this since not true any more, and the truth is not really user-friendly

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -125,7 +125,7 @@ def hash_args(args: Sequence[Any], arg_metas: Sequence[ArgMetadata | None]) -> s
         start = time.time()
         _hash = stringify_obj_type((str(i_arg),), arg, arg_metas[i_arg])
         g_repr_time += time.time() - start
-        if not _hash:
+        if _hash is None:
```
hughperkins: `not _hash` also includes emptyt string ,which was causing empty data oriented to appear as invalid arg.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -82,7 +82,7 @@ def stringify_obj_type(path: tuple[str, ...], obj: object, arg_meta: ArgMetadata
     if dataclasses.is_dataclass(obj):
         return dataclass_to_repr(path, obj)
     if is_data_oriented(obj):
-        child_repr_l = []
+        child_repr_l = ["da"]
```
hughperkins, line 85: aoid empty data oriented producing empty string as hash

SonSang (approved):

tests/python/gstaichi/lang/test_kernel_impl.py
```diff
@@ -45,3 +45,69 @@ def my_cool_test_function(a: ti.types.NDArray[ti.types.vector(2, ti.i32), 2]):
...
+@test_utils.test()
+def test_pure_kernel_parameter() -> None:
+    arch = ti.lang.impl.current_cfg().arch
+    ti.init(arch=arch, offline_cache=False, src_ll_cache=True)
+
+    @ti.pure
+    @ti.kernel
+    def k1() -> None: ...
+
+    @ti.kernel(pure=True)
```
YilingQiao, line 60: Do we need tests with a pure kernel and actually do something insdie?

hughperkins: Addressed in f7d26946f

hughperkins: Thank you for checking the tests by the way :)

YilingQiao (approved):

tests/python/gstaichi/lang/test_kernel_impl.py
YilingQiao: thank you!

python/gstaichi/lang/_fast_caching/args_hasher.py
duburcqa: I think such magic should be documented in the code. A comment in a PR is not good enough. By the way, could you clarify if this small design flaw could be fixed at some point or it is too intricate?

python/gstaichi/lang/_fast_caching/args_hasher.py
duburcqa: I think stringify should raise an exception if mapping to empty hash, or empty is really valid?

if this is would be worth adding a comment stating whether empty hash is considered valid or not in the doc.

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: It's fixed. See below.

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: Empty hash is valid. This fixes the 'small design flaw' you talk about at https://github.com/Genesis-Embodied-AI/gstaichi/pull/190#discussion_r2348025518

hughperkins: (Like, in https://github.com/Genesis-Embodied-AI/gstaichi/pull/190#discussion_r2348025518 you assert that requiring the hash to be non-empty is a "small design flaw")

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: Though I'm not sure making the hash non-empty is a 'small design flaw'? I think intuitively we do expect hashes to be non-empty, as you allude to by saying "I think stringify should raise an exception if mapping to empty hash, or empty is really valid?" in https://github.com/Genesis-Embodied-AI/gstaichi/pull/190#discussion_r2348027559

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: Updated to no longer allow empty hash as valid hash.

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: See above, made a decision that hashes cannot be empty strings, and therefore it is standard that the hash is not empty string. Not sure we need a comment about this? The other hashes have no such comment. eg:
```
    if isinstance(obj, np.ndarray):
        return f"[np-{obj.dtype}-{obj.ndim}]"
```
We are not adding a comment about adding `np-` prefix here. `da` prefix is same idea I think?

duburcqa: You are right. I'm fine.

python/gstaichi/lang/_fast_caching/args_hasher.py
```diff
@@ -125,7 +125,7 @@ def hash_args(args: Sequence[Any], arg_metas: Sequence[ArgMetadata | None]) -> s
         start = time.time()
         _hash = stringify_obj_type((str(i_arg),), arg, arg_metas[i_arg])
         g_repr_time += time.time() - start
-        if not _hash:
+        if _hash:
```
duburcqa: Are you sure about this change?

hughperkins: haha, whoops. thanks!

python/gstaichi/lang/_fast_caching/args_hasher.py
duburcqa: Could you document in the method itself that the method is never supposed to return an empty string?

python/gstaichi/lang/_fast_caching/args_hasher.py
hughperkins: updated

duburcqa (approved):

PR #191, opened 2025-09-13
[build] Split windows build into build job and test job, so reruns are faster

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


.github/workflows/win.yml
```diff
@@ -43,16 +43,47 @@ jobs:
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.PYTHON_VERSION}}
+      - name: Set Python CP version
+        id: set_cp_version
+        run: |
+          pip install packaging
+          $cp = python -c "from packaging import tags; print(next(tags.sys_tags()).interpreter)"
+          echo "PYTHON_CP_VERSION=$cp"
+          echo "PYTHON_CP_VERSION=$cp" >> $env:GITHUB_OUTPUT
```
CharlesMasson, line 69: Can't we just use the matrix input value `${{ ... }}`?

duburcqa: It is not the same. Here it returns `cp312` or similar, which is a different "naming convention" than `actions/setup-python`.

CharlesMasson (approved):

PR #192, opened 2025-09-13
[build] Upgrade Eigen

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
Looks like this is just used by sparse_matrix.cpp and sparse_solver.cpp. Not sure if we are using either of those? Maybe worth upgrading since:
- won't break anything we are using ourselves
- presumably there are tests that sparse_matrix etc continue to work ok
- maybe makes the library more appealing to other people

?

duburcqa:
Why are you using main branch instead of latest release? Do you need to be this cutting edge? Usually I do not recommend using main branch as dependency because it is less stable / tested.

hughperkins:
> Why are you using main branch instead of latest release? Do you need to be this cutting edge? Usually I do not recommend using main branch as dependency because it is less stable / tested.

Not a conscious decision. Was wondering about that actually. Will look into using latest release instead.

duburcqa:
Eigen has been making a lot of progress since then, but the high-level API has been very stable. You should not expect any changes. Nevertheless, the performance has greatly improved, especially for CPU architectures that were still niche at that time, incl. ARM64, and many convenience helpers leveraging C++17 capabilities have been added.

duburcqa:
Note that one of the very few main dev of Eigen (they are 2) is one of the main dev of Tensorflow. Do whatever you want from this information :p

hughperkins:
> Note that one of the very few main dev of Eigen (they are 2) is also one of the main dev of Tensorflow. Do whatever you want from this information :p

😅

hughperkins:
Ok, upgrading to same commit as tensorflow, https://github.com/tensorflow/tensorflow/blob/master/tensorflow/lite/tools/cmake/modules/eigen.cmake , which at this time is 70d8d99d0df9fd967b135efd8d12ed20fc48d007

duburcqa (approved):

johnnynunez:
@hughperkins fyi: I tried last version some months ago and it was working for me(last commit)

PR #206, opened 2025-09-16
[misc] Add capabilities to facilitate fastcache debugging

hughperkins:
Issue: #

### Brief Summary

Add capabilities to facilitate fastcache debugging
- kernels have new method `to_string()`, available on python side, taht returns the ChiIr of the kernel, as a text string
- CompileResult have new attribute 'cache_key', available on python side, that returns the fe-ll cache cache key, as a string
- new env var option `TI_DUMP_KERNEL_CHECKSUMS=1`, which will:
     - for each kernel that has a fastcache key
          - the first time the kernel is compiled:
                - append both the fastcache key, and fe-ll cache key, to a csv file /tmp/cachekeys.csv, as a new row, along with the kernel name
                - write the ChiIr of the kernel to /tmp/kernels, with the filename set to the fe-ll cache key
     - this allows us to detect kernels where multiple fe-ll cache keys maps to a singel fastcache key
          - and compare the ChiIr for such kernels
          - to track down why the fastcache cache key is not correctly distinguishing between the two compiled variatns of this kernel

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/impl.py
+++ b/python/gstaichi/lang/impl.py
@@ -1,4 +1,5 @@
 import numbers
+import pathlib
 import weakref
 from types import FunctionType, MethodType
 from typing import TYPE_CHECKING, Any, Iterable, Sequence
@@ -363,6 +364,7 @@ class PyGsTaichi:
         self.short_circuit_operators: bool = False
         self.unrolling_limit: int = 0
         self.src_ll_cache: bool = True
+        self.debug_dump_path: pathlib.Path = pathlib.Path("/tmp")
 
     @property
     def compiling_callable(self) -> KernelCxx | Kernel | Function:
--- a/python/gstaichi/lang/impl.py
+++ b/python/gstaichi/lang/impl.py
@@ -1,5 +1,4 @@
 import numbers
-import pathlib
 import weakref
 from types import FunctionType, MethodType
 from typing import TYPE_CHECKING, Any, Iterable, Sequence
@@ -364,7 +363,6 @@ class PyGsTaichi:
         self.short_circuit_operators: bool = False
         self.unrolling_limit: int = 0
         self.src_ll_cache: bool = True
-        self.debug_dump_path: pathlib.Path = pathlib.Path("/tmp")
 
     @property
     def compiling_callable(self) -> KernelCxx | Kernel | Function:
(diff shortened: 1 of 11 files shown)


python/gstaichi/lang/kernel_impl.py
```diff
@@ -1082,6 +1083,22 @@ def recursive_set_args(needed_arg_type: Type, provided_arg_type: Type, v: Any, i
             prog = impl.get_runtime().prog
             if not compiled_kernel_data:
                 compile_result: CompileResult = prog.compile_kernel(prog.config(), prog.get_device_caps(), t_kernel)
+                if os.environ.get("TI_DUMP_KERNEL_CHECKSUMS", "0") == "1":
+                    checksums_file_path = pathlib.Path("/tmp/checksums.csv")
+                    kernels_dump_dir = pathlib.Path("/tmp/kernels")
```
duburcqa: At this point we should be able to specify a custom location where to store all these dump via some environment variables. We have been relying on hard-coded path too often already for all these `TS_DUMP_*`.

hughperkins: Sure. Was surprised no-one has called me out on that tbh 😂  What are your thoughts on how/where we specify such paths?

duburcqa: In the same way we specify the offline cache path.

hughperkins: addressed in a977655c0

YilingQiao (approved):

python/gstaichi/lang/impl.py
```diff
@@ -363,6 +364,7 @@ def __init__(self, kernels=None):
         self.short_circuit_operators: bool = False
         self.unrolling_limit: int = 0
         self.src_ll_cache: bool = True
+        self.debug_dump_path: pathlib.Path = pathlib.Path("/tmp")
```
duburcqa: I think this should be coming Taichi core c++ instead of being native Python, because we are also dumping some debug stuff on the C++ side, and it should be possible to decide where to store them in the exact same way. Moreover, it should be controlled using some environment variable, similar to cache location.

hughperkins: cache location is not controlled by env var AFAIK: that's a Genesis thing I think?

hughperkins: Screenshot 2025-09-24 at 8 42 03 AM

hughperkins: moved debug_dump_path to c++ side in 5c1e507d7 , and added unit test

hughperkins:
Thanks!

PR #209, opened 2025-09-17
upgrade glibc according to runners for better stability

johnnynunez:


.github/workflows/scripts/ti_build/entry.py
```diff
@@ -33,11 +33,11 @@ def build_wheel(python: Command, pip: Command) -> None:
...
     u = platform.uname()
     if (u.system, u.machine) == ("Linux", "x86_64"):
-        extra.extend(["-p", "manylinux_2_27_x86_64"])
+        extra.extend(["-p", "manylinux_2_34_x86_64"])
     elif (u.system, u.machine) in (("Linux", "arm64"), ("Linux", "aarch64")):
-        extra.extend(["-p", "manylinux_2_27_aarch64"])
+        extra.extend(["-p", "manylinux_2_34_aarch64"])
     if platform.system() == "Darwin":
-        extra.extend(["-p", "macosx-11.0-arm64"])
+        extra.extend(["-p", "macosx-14.0-arm64"])
```
hughperkins, line 40: we don't want to restrict artifically to mac 14+. If we find that gstaichi wont run on certain versions, we'll add a restriction. AFAIK, it's possible that it works ok on 11+.

.github/workflows/scripts/ti_build/entry.py
```diff
@@ -33,11 +33,11 @@ def build_wheel(python: Command, pip: Command) -> None:
     cmake_args.writeback()
     u = platform.uname()
     if (u.system, u.machine) == ("Linux", "x86_64"):
-        extra.extend(["-p", "manylinux_2_27_x86_64"])
+        extra.extend(["-p", "manylinux_2_34_x86_64"])
```
hughperkins, line 36: We're aiming for manylinux 2 27 currently, for x86.

.github/workflows/scripts/ti_build/entry.py
```diff
@@ -33,11 +33,11 @@ def build_wheel(python: Command, pip: Command) -> None:
...
     elif (u.system, u.machine) in (("Linux", "arm64"), ("Linux", "aarch64")):
-        extra.extend(["-p", "manylinux_2_27_aarch64"])
+        extra.extend(["-p", "manylinux_2_34_aarch64"])
```
hughperkins, line 38: We might do this. Note that this doesnt affect the build process. And at test time, this *reduces* the possible installation targets. ie if this is set to 2 27, we can install on: 2 27, 2 28, 2 34. If we set it to 2 34, it can only be installed on: 2 34.

hughperkins: (this basically affects the filename fo the wheel, which is used by pip to decide if it's ok to try to install the wheel)

PR #210, opened 2025-09-17
[misc] Move taskgen class declaration to header

hughperkins:
Issue: #

### Brief Summary

Move taskgen class declaration to header, so it's more intuitively obvious that this class exists, and so it's easy to see the methods and properties on it.

copilot:summary

### Walkthrough

copilot:walkthrough


gstaichi/codegen/spirv/spirv_codegen.cpp
duburcqa, line 1: Could you avoid this reformatting? I would suggest reformatting the entire codebase in a dedicated PR instead of doing it here and there when files get modified. It just make the life of reviewers unnecessarily harder :)

gstaichi/ir/ir.cpp
```diff
@@ -69,9 +69,10 @@ class StatementTypeNameVisitor : public IRVisitor {
   StatementTypeNameVisitor() {
   }
 
-#define PER_STATEMENT(x)         \
-  void visit(x *stmt) override { \
-    type_name = #x;              \
+#define PER_STATEMENT(x)                      \
+  void visit(x *stmt) override {              \
+    std::cout << "visiting " #x << std::endl; \
```
duburcqa: Are you sure about this print? It looks inappropriate.

gstaichi/codegen/spirv/spirv_codegen.cpp
hughperkins: It's not under my control.`pre-commit run -a` does it. I'm unclear why it does that only when I modify the file. It seems like the kind of logic I was trying to add at the start, but for pyright. I guess I can dig what logic is changing this.

gstaichi/ir/ir.cpp
hughperkins: looks like my debug stuff yeah. Thanks! Good spot :)

gstaichi/codegen/spirv/spirv_codegen.cpp
hughperkins: Oh I see what you mean. Oh, so it's already in a dedicated PR at https://github.com/Genesis-Embodied-AI/gstaichi/pull/210  Just waiting on your approval :)

hughperkins: oh its the same Pr

duburcqa (approved):

PR #225, opened 2025-09-22
Fix support of Autodiff for torch non-CPU non-CUDA tensor.

duburcqa:


hughperkins:
Thanks! Note: I'll wait until there is a unit test before looking at this.

tests/python/test_ad_ndarray.py
```diff
@@ -9,11 +11,11 @@
 if has_pytorch():
     import torch
 
-archs_support_ndarray_ad = [ti.cpu, ti.cuda]
-
+archs_support_ndarray_ad = [ti.cpu, ti.cuda, ti.metal]
+default_fp = ti.f64 if sys.platform != "darwin" else ti.f32
```
hughperkins: `default_fp` is misleading I feel. Because my brain interpreted it as meaning `ti.f32`. Maybe `target_fp`?

hughperkins: I wonder why we only test f64? I wonder if we should parametrize over f64, and f32? I mean, I understand that's pretty much out of the scope of your changes, which are localized to adding the new metal tests. But just kind of pondering out loud.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -927,23 +927,27 @@ def call_back():
                                 "Non contiguous gradient tensors are not supported, please call tensor.grad.contiguous() before passing it into gstaichi kernel."
                             )
 
-                    tmp = v
-                    if (str(v.device) != "cpu") and not (
-                        str(v.device).startswith("cuda") and gstaichi_arch == _ti_core.Arch.cuda
+                    v_taichi = v
+                    grad_taichi = v.grad
+                    if (v.device.type != "cpu") and not (
+                        v.device.type == "cuda" and gstaichi_arch == _ti_core.Arch.cuda
```
hughperkins, line 933: You're using both `v` and `v_taichi`, which I find confusing. Maybe don't create an alias of `v`?

python/gstaichi/lang/kernel_impl.py
```diff
@@ -927,23 +927,27 @@ def call_back():
...
-                        # We just replace it with a CPU tensor and by the end of kernel execution we'll use the
-                        # callback to copy the values back to the original CUDA tensor.
-                        host_v = v.to(device="cpu", copy=True)
-                        tmp = host_v
-                        callbacks.append(get_call_back(v, host_v))
+                        # For a torch tensor to be passed as as input argument (in and/or out) of a taichi kernel, its
+                        # memory must be hosted either on CPU, or on CUDA if and only if GsTaichi is using CUDA backend.
+                        # We just replace it with a CPU tensor and by the end of kernel execution we'll use the callback
+                        # to copy the values back to the original tensor.
+                        v_taichi = v.to(device="cpu")
```
hughperkins: Maybe at this point, you can call this something like `v_cpu`, and remove the earlier aliasing of v to `v_taichi`, so `v_cpu` always contains the same thing (kind of 'single source assigement'), ie it contains v convertted to cpu.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -927,23 +927,27 @@ def call_back():
...
+                        callbacks.append(get_call_back(v, v_taichi))
+                        if v.grad is not None:
+                            grad_taichi = v.grad.to(device="cpu")
```
hughperkins: again, you aliased `v.grad` to `grad_taichi`, but now you're still using `v.grad`. Maybe don't create the initial grad_taichi, (or simply assign it as `None`), and create it only here? Also, maybe this would be more intuitive if we named it something like `v_cpu_grad`?

python/gstaichi/lang/kernel_impl.py
duburcqa: I did this to align with what was originally done. And anyway, creating an alias will be necessary for the gradient.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -927,23 +927,27 @@ def call_back():
                                 "Non contiguous gradient tensors are not supported, please call tensor.grad.contiguous() before passing it into gstaichi kernel."
                             )
 
-                    tmp = v
-                    if (str(v.device) != "cpu") and not (
-                        str(v.device).startswith("cuda") and gstaichi_arch == _ti_core.Arch.cuda
+                    v_taichi = v
```
hughperkins: (I defintiely prefer not using the name `tmp` by the way 🙌)

python/gstaichi/lang/kernel_impl.py
duburcqa: I don't understand why you want me to do.

duburcqa: Ok I will try something :)

python/gstaichi/lang/kernel_impl.py
```diff
@@ -927,23 +927,28 @@ def call_back():
...
+                    if (v.device.type != "cpu") and not (
+                        v.device.type == "cuda" and gstaichi_arch == _ti_core.Arch.cuda
                     ):
-                        # Getting a torch CUDA tensor on GsTaichi non-cuda arch:
-                        # We just replace it with a CPU tensor and by the end of kernel execution we'll use the
-                        # callback to copy the values back to the original CUDA tensor.
-                        host_v = v.to(device="cpu", copy=True)
-                        tmp = host_v
-                        callbacks.append(get_call_back(v, host_v))
+                        # For a torch tensor to be passed as as input argument (in and/or out) of a taichi kernel, its
```
hughperkins, line 935: I needed a lot of context when we were discussing earlier, in order to understand what was happening here. Maybe we should add some of that context? eg something liek:

```
# Handle things like:
# - Mac MPS Torch tensor, which cannot be passed into taichi direclty
# - CUDA torch tensor, when using gstaichi with arch == cpu
# In such cases, we make a CPU copy of the incoming tensor, and pass
# this copy to the kernel. After kernel execution,  we copy the data back 
# into the original tensor.
```

hughperkins:
@erizmr could you review/approve this PR too please? (since I believe you worked on a lot of the taichi AD implementation?)

erizmr (approved):
LGTM.

hughperkins:
for the failing unit tests, there's still a few failing tests remaining, but I guess you can see the approach I'm taking?

basically:
- extension data64 => only runs on machines that support f64
- adstack => adstack allows for loops with dynamic ranges to be used with AD. By requiring this, we filter out any machine that do not support adstack, and therefore filter out any machines where dynamic ranges are not allowed with AD, for backprop.

hughperkins:
Actually, it occurs to me, when default_fp is fp32, perahps we should be converting the torch tensor to fp32? (and not adding the filtering on data64).

hughperkins:
related aside: I find it odd that we are testing *only* fp64 on platforms that support fp64.  I feel we should test:
- fp32 everywhere
- fp64 *as well* on platforms taht suport it.

hughperkins:
oh this is why the torch is fp64:
```
    input = torch.rand(4, dtype=torch.double, device=device, requires_grad=True)
```
`torch.double`. We should paremterie that too I feel.

hughperkins:
ok, this test poc passes:

```
import sys

import pytest

import gstaichi as ti
from gstaichi.lang.exception import GsTaichiRuntimeError
from gstaichi.lang.util import has_pytorch

from tests import test_utils

if has_pytorch():
    import torch

archs_support_ndarray_ad = [ti.cpu, ti.cuda, ti.metal]
# target_fp_l = [(ti.f32, "float"), (ti.f64, "duoble")] if sys.platform != "darwin" else [(ti.f32, "float")]
target_fp_l = [(ti.f32, "float"), (ti.f64, "double")]


@pytest.mark.skipif(not has_pytorch(), reason="Pytorch not installed.")
@pytest.mark.parametrize("default_fp,torch_type_name", target_fp_l)
@test_utils.test(arch=archs_support_ndarray_ad)
def test_simple_demo(default_fp, torch_type_name: str) -> None:
    assert ti.lang is not None
    arch = ti.lang.impl.current_cfg().arch
    torch_type = getattr(torch, torch_type_name)
    ti.init(arch=arch, default_fp=default_fp, offline_cache=False)
    @test_utils.torch_op(output_shapes=[(1,)])
    @ti.kernel
    def test(x: ti.types.ndarray(), y: ti.types.ndarray()):
        for i in x:
            a = 2.0
            for j in range(1):
                a += x[i] / 3
            y[0] += a

    device = "cuda" if ti.lang.impl.current_cfg().arch == ti.cuda else "cpu"
    input = torch.rand(4, dtype=torch_type, device=device, requires_grad=True)
    eps = 0.0001 if default_fp == ti.f64 else 0.1
    torch.autograd.gradcheck(test, input, eps=eps)
```

hughperkins:
(admittedly using `--arch cpu` 😅 )

hughperkins:
Passes without needing `--arch cpu`:

```
import sys

import pytest

import gstaichi as ti
from gstaichi.lang.exception import GsTaichiRuntimeError
from gstaichi.lang.util import has_pytorch

from tests import test_utils

if has_pytorch():
    import torch

archs_support_ndarray_ad = [ti.cpu, ti.cuda, ti.metal]
target_fp_l = [(ti.f32, "float"), (ti.f64, "double")]


@pytest.mark.skipif(not has_pytorch(), reason="Pytorch not installed.")
@pytest.mark.parametrize("default_fp,torch_type_name", target_fp_l)
@test_utils.test(arch=archs_support_ndarray_ad)
def test_simple_demo(default_fp, torch_type_name: str) -> None:
    assert ti.lang is not None
    arch = ti.lang.impl.current_cfg().arch
    if sys.platform == "darwin" and arch in [ti.metal, ti.vulkan] and default_fp == ti.f64:
        pytest.skip("fp64 not supported on mac gpu")
    torch_type = getattr(torch, torch_type_name)
    ti.init(arch=arch, default_fp=default_fp, offline_cache=False)
    @test_utils.torch_op(output_shapes=[(1,)])
    @ti.kernel
    def test(x: ti.types.ndarray(), y: ti.types.ndarray()):
        for i in x:
            a = 2.0
            for j in range(1):
                a += x[i] / 3
            y[0] += a

    device = "cuda" if ti.lang.impl.current_cfg().arch == ti.cuda else "cpu"
    input = torch.rand(4, dtype=torch_type, device=device, requires_grad=True)
    eps = 0.0001 if default_fp == ti.f64 else 0.1
    torch.autograd.gradcheck(test, input, eps=eps)
```

Screenshot 2025-10-06 at 18 02 21

hughperkins:
yeah, so, the existing tests pretty much all require adstack (inner for loops have dynamic range). So we'll probably need to write some new tests

hughperkins:
Ok, so after fighting with this for a few hours 😅 :
1. test_ad_ndarray won't currently test your change because it relies on adstack - because dynamic for loops combined with backprop, which needs adstack - and adstack isnt supported on mac metal
2. I tried copying and pasting test_ad_ndarray to test_ad_ndarray_no_adstatic, and migrating from dynamic for loops to static for loops. However the tests still fail on mac metal, reasons unclear. I feel fixing that strays beyond the scope of your current changes
3. Therefore, I feel to test your chcanges we should create new dedicated tests, dedicated speciiclly at testing your changes
     - might be just a handful of simple tests, might not be anything very complicated

I've saved my changes to a different branch, so feel free to wipe my changes from this branch.

PR #226, opened 2025-09-23
[NOT MERGE] NOW COMPILES AARCH64

johnnynunez:
looking for opinion, it is using GCC


.github/workflows/scripts/ti_build/entry.py
```diff
@@ -101,6 +101,14 @@ def action_wheel():
     else:
         sccache("--start-server")
 
+    import os
+    inc_base = "/opt/rh/gcc-toolset-14/root/usr/include/c++/14"
```
hughperkins, line 105: this seems system specific. should not be part of entry.py I feel. should be part of https://github.com/Genesis-Embodied-AI/gstaichi/blob/main/.github/workflows/scripts_new/manylinux_wheel/2_build.sh or thereabouts

.github/workflows/scripts/ti_build/entry.py
```diff
@@ -101,6 +101,14 @@ def action_wheel():
...
+    extra = ":".join([inc_base, f"{inc_base}/aarch64-redhat-linux", f"{inc_base}/backward"])
+    os.environ["CPLUS_INCLUDE_PATH"] = os.environ.get("CPLUS_INCLUDE_PATH", "")
```
hughperkins, line 107: ditto for all of these bits I feel.

cmake/GsTaichiCXXFlags.cmake
```diff
@@ -138,6 +138,11 @@ else()
 endif()
 set(HOST_ARCH ${ARCH} CACHE INTERNAL "Host arch")
 
+if (UNIX AND (CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64"))
```
hughperkins, line 141: This looks reasonable to me. Nit: should this be `LINUX` rather than `UNIX`? I searched for `LINUX` and found loads of matches. for `UNIX`, only found one. I think we should standardize on `LINUX` on the whole.

.github/workflows/manylinux_wheel.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     strategy:
       matrix:
         PYTHON_CP_VERSION: ['cp310', 'cp311', 'cp312', 'cp313']
-        os: ['ubuntu-22.04', 'ubuntu-22.04-arm']
+        os: ['ubuntu-22.04', 'self-hosted-runner-gh200']
```
hughperkins, line 26: obviously these should not be changd :)

PR #216, opened 2025-09-21
[Build] Build using freshly built LLVM 15.0.7

hughperkins:
Issue: #

### Brief Summary

- use llvm built from gstaichi-sdk-builds
- switch to building linux using gcc

Genesis tests ran to completion https://github.com/Genesis-Embodied-AI/Genesis/pull/1762

(Not sure if any effect on performance. @duburcqa  ?)

copilot:summary

### Walkthrough

copilot:walkthrough

PR #227, opened 2025-09-24
some changes to build for arm with gcc

johnnynunez:


.github/workflows/scripts_new/manylinux_wheel/2_build.sh
```diff
@@ -2,7 +2,31 @@
...
+# Add Taichi LLVM toolchain to PATH
+export PATH="$PWD/taichi-llvm-15.0.7-linux-${PLATFORM}/bin:$PATH"
+
+# Taichi build options
 export GSTAICHI_CMAKE_ARGS="-DTI_WITH_VULKAN:BOOL=ON -DTI_WITH_CUDA:BOOL=ON -DTI_BUILD_TESTS:BOOL=ON"
-./build.py wheel
+
+# GCC toolset include paths
+inc_base="/opt/rh/gcc-toolset-14/root/usr/include/c++/14"
+extra="$inc_base:$inc_base/$inc_subdir:$inc_base/backward"
```
hughperkins: what is `inc_subdir`? I dont see where it is being defined?

.github/workflows/scripts_new/manylinux_wheel/2_build.sh
```diff
@@ -2,7 +2,31 @@
...
+
+export CPLUS_INCLUDE_PATH="${CPLUS_INCLUDE_PATH:+$CPLUS_INCLUDE_PATH:}$extra"
+export CPATH="${CPATH:+$CPATH:}/opt/rh/gcc-toolset-14/root/usr/include"
```
hughperkins: is CPATH defined before this? If not, then the reference to CPATH on the right hand side is not being used I think? And can be removed? I feel like if CPATH was defined then the right hand side would not work, because there is no `:` in front of `/opt/rh/gcc-toolset-14/root/usr/include`?

hughperkins:
To fix the linter errors, you can use `pre-commit run -a`

.github/workflows/scripts_new/manylinux_wheel/2_build.sh
johnnynunez: you have on the top, it is from you. aarch detection basically

hughperkins: oh right :)

.github/workflows/scripts_new/manylinux_wheel/2_build.sh
```diff
@@ -2,7 +2,31 @@
 
 set -ex
 
-export PATH=$PWD/clang+llvm-15.0.4-x86_64-linux-gnu-rhel-8.4/bin:$PATH
+PLATFORM=$(uname -m)
+if [ "$PLATFORM" = "x86_64" ]; then
+    PLATFORM="x86_64"
+    inc_subdir="x86_64-redhat-linux"
```
hughperkins: we can move this out of the loop I feel
```
inc_subdir="${PLATFORM}-redhat-linux"
```
But maybe we don't need to create this additional variable, since it is only used in one place I think?

.github/workflows/scripts_new/manylinux_wheel/2_build.sh
johnnynunez: checking

hughperkins:
Interesting that arm build works better with 2 28 than 2 34 👀

hughperkins:
(changing to draft for now: will set back to 'ready to review' once https://github.com/Genesis-Embodied-AI/gstaichi/pull/216 has been merged. This makes it easier for reviewers to know which PRs to review)

hughperkins:
Note to reviewers and to @johnnynunez : we can ignore the failures for `upload` and `api_docs`. They are a permissions issue, and do not indicate any issue with this PR.

johnnynunez:
@duburcqa @hughperkins for arm is working, now next work upgrade to LLVM20/21

hughperkins:
> > (changing to draft for now: will set back to 'ready to review' once #216 has been merged. This makes it easier for reviewers to know which PRs to review)
> 
> if the tests passed, i hope to see a release with llvm15, then we can upgrade it to llvm20/21

Yeah, I have no major concerns with passing directly to llvm20/21, as long as code changes are not too great, and are easy-ish to review.
- would need to carefully document our changes to pass manager I feel, forming an alignment between the code for the previous pass manager, and the code with the new pass manager, and ensuring we are neither adding nor removing new passes, within the constraints of the architectural change
- and there will need to be a performance regression test (you don't need to handle this yourself, but you'll need to wait for this to be done)

gstaichi/runtime/cpu/jit_cpu.cpp
```diff
@@ -131,10 +131,21 @@ class JITSessionCPU : public JITSession {
         mangle_(es_, this->dl_),
         module_counter_(0),
         memory_manager_(nullptr) {
+#if !(defined(__APPLE__) && defined(__aarch64__))
+    // On ELF-based targets LLVM's ORC JIT expects the object layer to claim
+    // responsibility for every symbol emitted by the module. Without this the
+    // resolver may encounter symbols that were not pre-registered in the
+    // materialization responsibility set and abort with
+    // "Resolving symbol outside this responsibility set" (observed on
+    // manylinux aarch64).
```
duburcqa: Why include Mac if no issue was observed on Mac ?

gstaichi/runtime/llvm/runtime_module/runtime.cpp
```diff
@@ -54,6 +54,69 @@ __asm__(".symver powf,powf@GLIBC_2.2.5");
...
+    !defined(__APPLE__)
+// JIT session error: Symbols not found: [ __aarch64_ldadd4_acq_rel, ... ]
+// This is an issue with newer clang versions (>13) on aarch64, where
+// atomics are outlined by default. The JIT environment does not
+// automatically link libatomic/libgcc.
+//
+// We provide the implementations here using inline assembly. The
+// implementations use a load-exclusive/store-exclusive loop to be compatible
+// with all ARMv8-A processors, including those that do not support the LSE
+// (Large System Extensions) instructions.
```
duburcqa: This seems very hacky. I don't like it much.

gstaichi/runtime/cpu/jit_cpu.cpp
hughperkins: Not my PR, but, what do you mean by 'include Mac'? The conditional has `!(defined(__APPLE__)`, suggesting not on Mac?

duburcqa: I mean: why include Mac in this new exclusion rule

hughperkins: I was assuming this is only needed for ARM on linux. Mac already works. No need to modify things for non-linux arm?

duburcqa: Yes, this is what I'm suggesting.

duburcqa: wait, the condition is not clear to me. you are adding a new code block that is triggered on anything but Apple ARM ??

hughperkins: "(not apple) and (arm)"

hughperkins: oh wait

hughperkins: "not(apple and arm)"

hughperkins: yes, looks like it. so really it should better be "linux and arm"?

duburcqa: this is my feeling yes.

.github/workflows/scripts/ti_build/llvm.py
```diff
@@ -36,6 +36,8 @@ def setup_llvm() -> None:
             out = get_cache_home() / f"llvm-{llvm_version}-x86-{build_version}"
             url = release_url_template.format(platform="linux-x86_64")
         download_dep(url, out, strip=1)
+    elif (u.system, u.machine) in (("Linux", "arm64"), ("Linux", "aarch64")):
```
hughperkins: dont you need to modify the initial if u.system == "linux" above?

PR #228, opened 2025-09-25
[Bug] Fix memory OOM in Genesis push_differentiable example

hughperkins:
Issue: #

### Brief Summary

- calling build_stmts twice, on a Call node args, adds an extra 8GB of memory usage when calling push_differentiable.py, https://github.com/Genesis-Embodied-AI/Genesis/blob/main/examples/differentiable_push.py , from [Genesis](https://github.com/Genesis-Embodied-AI/Genesis/tree/main)
- we address this by only calling build_stmt on new args, which haven't yet been built, i.e. which don't yet have a `ptr` attribute
- note that the additional calls to build_stmts were originally added in https://github.com/Genesis-Embodied-AI/gstaichi/pull/73

Before Genesis migrated to gstaichi, in https://github.com/Genesis-Embodied-AI/Genesis/pull/1550, push_differentiable used to use ~22GB of memory, which we can observe by running `while true; do { nvidia-smi | grep 575W; sleep 1; } done` in parallel, in a second terminal session:

Screenshot 2025-09-25 at 9 20 59 AM

After migrating to gstaichi, this test used ~32GB of memory:

Screenshot 2025-09-25 at 8 18 56 AM

This PR fixes this regression, so this test uses ~22GB of memory again.

The extra memory was being used because we were parsing incoming kernel parameters multiple times. This PR modifies the code so that each kernel parameter is only processed once.

copilot:summary

### Walkthrough

copilot:walkthrough


python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -257,11 +257,17 @@ def build_Call(ctx: ASTTransformerContext, node: ast.Call, build_stmt, build_stm
...
+
             # create variables for the now-expanded dataclass members
-            build_stmts(ctx, node.args)
-            build_stmts(ctx, node.keywords)
+            for arg in node.args:
+                if not hasattr(arg, "ptr"):
+                    build_stmt(ctx, arg)
+            for arg in node.keywords:
+                if not hasattr(arg, "ptr"):
+                    build_stmt(ctx, arg)
```
yun-long: Minor naming suggestion: `for keyword in node.keywords:` in stead of  `for arg in node.keywords`.

Second, the solution seem a bit hacky, as now we are calling similar functions multiple times. 

```
build_stmts(ctx, node.args)
build_stmts(ctx, node.keywords)
build_stmt(ctx, arg)
build_stmt(ctx, keyword)
```

not sure, but how about?
```
new_args, new_keywords = expand_call_dataclass(node.args, node.keywords)

build_stmts(ctx, new_args)
build_stmts(ctx, new_keywords)

node.args = new_args
node.keywords = new_keywords
```

yun-long (approved):

hughperkins:
Actually, it's not clear to me that non-static ranges should fail. So let's merge this anyway. (albeit likely with @yun-long 's suggestion incorporated to make this a bit more principled).

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: addressed in [bd51400](https://github.com/Genesis-Embodied-AI/gstaichi/pull/228/commits/bd5140092894b8f08ba37da3859c37e5f4477566)

yun-long:
LGTM

hughperkins:
https://linear.app/genesis-ai-company/issue/CMP-154/fix-bug-with-extend-in-httpsgithubcomgenesis-embodied

hughperkins:
https://github.com/Genesis-Embodied-AI/gstaichi/pull/235

PR #229, opened 2025-09-27
[Bug] Fail on non-static range in backwards

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/ast/ast_transformer_utils.py
+++ b/python/gstaichi/lang/ast/ast_transformer_utils.py
@@ -7,6 +7,7 @@ from enum import Enum
 from textwrap import TextWrapper
 from typing import TYPE_CHECKING, Any, List
 
+from gstaichi._lib import core as _ti_core
 from gstaichi._lib.core.gstaichi_python import ASTBuilder
 from gstaichi.lang import impl
 from gstaichi.lang._ndrange import ndrange
@@ -17,6 +18,9 @@ from gstaichi.lang.exception import (
     GsTaichiSyntaxError,
     handle_exception_from_cpp,
 )
+
+AutodiffMode = _ti_core.AutodiffMode
+
 
 if TYPE_CHECKING:
     from gstaichi.lang.kernel_impl import (
@@ -168,6 +172,7 @@ class ASTTransformerContext:
         start_lineno: int,
         ast_builder: ASTBuilder | None,
         is_real_function: bool,
+        autodiff_mode: AutodiffMode,
     ):
         self.func = func
         self.local_scopes: list[dict[str, Any]] = []
@@ -199,6 +204,7 @@ class ASTTransformerContext:
         self.is_real_function = is_real_function
         self.kernel_args: list = []
         self.only_parse_function_def: bool = False
+        self.autodiff_mode = autodiff_mode
 
     # e.g.: FunctionDef, Module, Global
     def variable_scope_guard(self):
--- a/python/gstaichi/lang/ast/ast_transformer.py
+++ b/python/gstaichi/lang/ast/ast_transformer.py
@@ -40,6 +40,8 @@ from gstaichi.lang.struct import Struct, StructType
 from gstaichi.types import primitive_types
 from gstaichi.types.utils import is_integral
 
+AutodiffMode = _ti_core.AutodiffMode
+
 
 def reshape_list(flat_list: list[Any], target_shape: Sequence[int]) -> list[Any]:
     if len(target_shape) < 2:
@@ -1106,6 +1108,8 @@ class ASTTransformer(Builder):
                 and isinstance(node.iter.func, ast.Name)
                 and node.iter.func.id == "range"
             ):
+                if ctx.autodiff_mode == AutodiffMode.REVERSE:
+                    raise Exception("Cannot use non static range in Backwards mode")
                 return ASTTransformer.build_range_for(ctx, node)
             elif isinstance(node.iter, ast.IfExp):
                 # Handle inline if expression as the top level iterator expression, e.g.:
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -310,6 +310,7 @@ def _get_tree_and_ctx(
     assert current_kernel is not None
     current_kernel.visited_functions.add(function_source_info)
 
+    autodiff_mode = current_kernel.autodiff_mode
     return tree, ASTTransformerContext(
         excluded_parameters=excluded_parameters,
         is_kernel=is_kernel,
@@ -323,6 +324,7 @@ def _get_tree_and_ctx(
         file=function_source_info.filepath,
         ast_builder=ast_builder,
         is_real_function=is_real_function,
+        autodiff_mode=autodiff_mode,
     )
 
 


SonSang:
AD does not work properly in the backward pass when non-static `inner` loops are used. However, the results are quite inconsistent and unpredictable, so I summarize here some cases that I have faced so far.

* Compiles and runs silently, but memory usage is largely increased
* Compiles and runs silently, but gives `nan` value
* Compiles and runs silently, but gradient is not computed correctly
* Does not compile at all

Here is the example where it does not compile at all:
```
import gstaichi as ti

ti.init(arch=ti.cpu, debug=True)

B = 10
x = ti.field(ti.f32, shape=(3, B), needs_grad=True)
y = ti.field(ti.f32, shape=(3, B), needs_grad=True)
loss = ti.field(ti.f32, shape=(), needs_grad=True)

x[0, 0] = 1.0
y[0, 0] = 2.0

@ti.kernel
def k1(use_static: ti.template()):
    for i_b in ti.ndrange(B):
        z = ti.Vector.zero(ti.f32, 3)
        
        # Non-static inner loop is not supported in backward
        for i_3 in ti.static(range(3)) if ti.static(use_static) else range(3):
            z[i_3] += x[i_3, i_b] * y[i_3, i_b]

        loss[None] += z.x + z.y + z.z

for i in range(2):
    try:
        use_static = True if i == 0 else False

        loss.fill(0.0)
        k1(use_static) 
        
        loss.grad[None] = 1.0
        x.grad.fill(0.0)
        y.grad.fill(0.0)
        k1.grad(use_static)

        assert x.grad[0,0] == 2.0
        assert y.grad[0,0] == 1.0

        print("Passed for use_static =", use_static)
        input("Press Enter to continue...")
    except Exception as e:
        print("Failed for use_static =", use_static)
        if i == 1:
            print("This is expected failure, because non-static inner loop is not supported")
            print("Error:", e)
```

hughperkins:
> AD does not work properly in the backward pass when non-static `inner` loops are used. However, the results are quite inconsistent and unpredictable, so I summarize here some cases that I have faced so far.
> 
> * Compiles and runs silently, but memory usage is largely increased
> * Compiles and runs silently, but gives `nan` value
> * Compiles and runs silently, but gradient is not computed correctly
> * Does not compile at all
> 
> Here is the example where it does not compile at all:
> 
> ```
> import gstaichi as ti
> 
> ti.init(arch=ti.cpu, debug=True)
> 
> B = 10
> x = ti.field(ti.f32, shape=(3, B), needs_grad=True)
> y = ti.field(ti.f32, shape=(3, B), needs_grad=True)
> loss = ti.field(ti.f32, shape=(), needs_grad=True)
> 
> x[0, 0] = 1.0
> y[0, 0] = 2.0
> 
> @ti.kernel
> def k1(use_static: ti.template()):
>     for i_b in ti.ndrange(B):
>         z = ti.Vector.zero(ti.f32, 3)
>         
>         # Non-static inner loop is not supported in backward
>         for i_3 in ti.static(range(3)) if ti.static(use_static) else range(3):
>             z[i_3] += x[i_3, i_b] * y[i_3, i_b]
> 
>         loss[None] += z.x + z.y + z.z
> 
> for i in range(2):
>     try:
>         use_static = True if i == 0 else False
> 
>         loss.fill(0.0)
>         k1(use_static) 
>         
>         loss.grad[None] = 1.0
>         x.grad.fill(0.0)
>         y.grad.fill(0.0)
>         k1.grad(use_static)
> 
>         assert x.grad[0,0] == 2.0
>         assert y.grad[0,0] == 1.0
> 
>         print("Passed for use_static =", use_static)
>         input("Press Enter to continue...")
>     except Exception as e:
>         print("Failed for use_static =", use_static)
>         if i == 1:
>             print("This is expected failure, because non-static inner loop is not supported")
>             print("Error:", e)
> ```

This PR is aiming to address the examples where it does compile. Since if it does not compile, then behavior is correct already, and does not need to be modified.

Could you provide an example please for each of:
* Compiles and runs silently, but memory usage is largely increased
* Compiles and runs silently, but gives `nan` value
* Compiles and runs silently, but gradient is not computed correctly

(or at least one of these, ideally)

python/gstaichi/lang/ast/ast_transformer_utils.py
```diff
@@ -199,6 +204,17 @@ def __init__(
         self.is_real_function = is_real_function
         self.kernel_args: list = []
         self.only_parse_function_def: bool = False
+        self.autodiff_mode = autodiff_mode
+        self.loop_depth: int = 0
+        from gstaichi import extension  # pylint: disable=import-outside-toplevel
```
duburcqa: I prefer keeping local import on top of the corresponding scope / code block (in this case, on top of the function definition). I find it much more readable. But do as you please if you don't share my opinion.

hughperkins: It broke the linter, since it said that torch might not be bound.

Open to alternative approaches to fix the linter.

duburcqa: I don't understand. I'm only suggesting a minor change that should not have any effect on the linter:
```python
        is_real_function: bool,
        autodiff_mode: AutodiffMode,
    ):
        from gstaichi import extension

        self.func = func
        self.local_scopes: list[dict[str, Any]] = []
        [...]
```

tests/python/test_ad_for.py
```diff
@@ -3,10 +3,13 @@
 from tests import test_utils
 
 
-@test_utils.test()
+@test_utils.test(require=ti.extension.adstack)
 def test_ad_nested_for():
     N = 5
 
+    assert ti.lang is not None
+    ti.lang.impl.current_cfg().ad_stack_experimental_enabled = True
```
duburcqa: `@test_utils.test` should do `ti.lang.impl.current_cfg().ad_stack_experimental_enabled = True` automatically whenever `ti.extension.adstack` is required.

The point of pytest vs unittest is just to expose facilities to hide all the boilerplate, for devs and readers to focus on what really matters. As a rule of thumb, if you need to copy-paste the same code in multiple test, you probably need a fixture to do this for you behind the scene without you paying attention to it.

tests/python/test_ad_ndarray.py
```diff
@@ -32,6 +31,8 @@ def test(x: ti.types.ndarray(), y: ti.types.ndarray()):
 @pytest.mark.skipif(not has_pytorch(), reason="Pytorch not installed.")
 @test_utils.test(arch=archs_support_ndarray_ad, default_fp=ti.f64)
 def test_ad_reduce():
+    import torch
```
duburcqa: What is this?

hughperkins: otherwise there are linter errors about 'torch might be unbound'

tests/python/test_ad_ndarray.py
```diff
@@ -6,15 +6,14 @@
 
 from tests import test_utils
 
-if has_pytorch():
-    import torch
-
```
duburcqa, line 11: What about moving all the tests that requires torch in a dedicated test file and do:
```python
try:
    import torch
except ImportError:
    pytest.skip("PyTorch not installed. Skipping...", allow_module_level=True)
```

hughperkins: addressed in 546998697

PR #230, opened 2025-09-27
[Perf] Detect pure violation

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/ast/ast_transformer.py
+++ b/python/gstaichi/lang/ast/ast_transformer.py
@@ -11,7 +11,7 @@ from typing import Any, Sequence, Type
 import numpy as np
 
 from gstaichi._lib import core as _ti_core
-from gstaichi.lang import expr, impl, matrix, mesh
+from gstaichi.lang import expr, impl, matrix, mesh, exception
 from gstaichi.lang import ops as ti_ops
 from gstaichi.lang._ndrange import _Ndrange
 from gstaichi.lang.ast.ast_transformer_utils import (
@@ -19,6 +19,7 @@ from gstaichi.lang.ast.ast_transformer_utils import (
     Builder,
     LoopStatus,
     ReturnStatus,
+    PtrSource,
     get_decorator,
 )
 from gstaichi.lang.ast.ast_transformers.call_transformer import CallTransformer
@@ -71,10 +72,18 @@ def boundary_type_cast_warning(expression: Expr) -> None:
 class ASTTransformer(Builder):
     @staticmethod
     def build_Name(ctx: ASTTransformerContext, node: ast.Name):
-        node.ptr = ctx.get_var_by_name(node.id)
+        print("build_Name", ast.dump(node))
+        ptr_source, node.ptr = ctx.get_var_by_name(node.id)
         if isinstance(node, (ast.stmt, ast.expr)) and isinstance(node.ptr, Expr):
             node.ptr.dbg_info = _ti_core.DebugInfo(ctx.get_pos_info(node))
             node.ptr.ptr.set_dbg_info(node.ptr.dbg_info)
+        node.ptr_source = ptr_source
+        print('is_pure', ctx.is_pure, 'ptr_source', ptr_source)
+        if ctx.is_pure and ptr_source == PtrSource.GLOBAL and not ctx.static_scope_status.is_in_static_scope:
+            print("ctx.static_scope_status", ctx.static_scope_status.is_in_static_scope)
+            if isinstance(node.ptr, (float, int, Field)):
+                raise exception.GsTaichiCompilationError("Accessing global variable", node.id, type(node.ptr))
+            print("node.ptr", node.ptr, type(node.ptr))
         return node.ptr
 
     @staticmethod
@@ -242,6 +251,7 @@ class ASTTransformer(Builder):
         if not ASTTransformer.is_tuple(node.slice):
             node.slice.ptr = [node.slice.ptr]
         node.ptr = impl.subscript(ctx.ast_builder, node.value.ptr, *node.slice.ptr)
+        node.ptr_source = node.value.ptr_source
         return node.ptr
 
     @staticmethod
@@ -274,7 +284,12 @@ class ASTTransformer(Builder):
 
     @staticmethod
     def build_List(ctx: ASTTransformerContext, node: ast.List):
+        node.ptr_source = PtrSource.LOCAL
         build_stmts(ctx, node.elts)
+        for elt in node.elts:
+            if elt.ptr_source == PtrSource.GLOBAL:
+                node.ptr_source = PtrSource.GLOBAL
+            # print("elt", ast.dump(elt), getattr(elt, "ptr_source", None))
         node.ptr = [elt.ptr for elt in node.elts]
         return node.ptr
 
@@ -301,17 +316,26 @@ class ASTTransformer(Builder):
 
     @staticmethod
     def process_generators(ctx: ASTTransformerContext, node: ast.GeneratorExp, now_comp, func, result):
+        node.ptr_source = PtrSource.LOCAL
         if now_comp >= len(node.generators):
+            print("now_comp >= len node.generators")
             return func(ctx, node, result)
         with ctx.static_scope_guard():
+            print("using static scope guard, building stmts")
             _iter = build_stmt(ctx, node.generators[now_comp].iter)
 
         if isinstance(_iter, impl.Expr) and _iter.ptr.is_tensor():
+            print("iter is expr, and iter.ptr is tensor")
             shape = _iter.ptr.get_shape()
             flattened = [Expr(x) for x in ctx.ast_builder.expand_exprs([_iter.ptr])]
             _iter = reshape_list(flattened, shape)
 
         for value in _iter:
+            # these are actual python objects, not ast nodes
+            # print("value", value, type(value))
+            # if value.ptr_source == PtrSource.GLOBAL:
+            #     node.ptr_source = PtrSource.GLOBAL
+            # print('iterate over _iter')
             with ctx.variable_scope_guard():
                 ASTTransformer.build_assign_unpack(ctx, node.generators[now_comp].target, value, True)
                 with ctx.static_scope_guard():
@@ -351,6 +375,7 @@ class ASTTransformer(Builder):
     @staticmethod
     def build_Constant(ctx: ASTTransformerContext, node: ast.Constant):
         node.ptr = node.value
+        node.ptr_source = PtrSource.LOCAL
         return node.ptr
 
     @staticmethod
@@ -582,6 +607,7 @@ class ASTTransformer(Builder):
 
     @staticmethod
     def build_Attribute(ctx: ASTTransformerContext, node: ast.Attribute):
+        print("build_Attribute", ast.dump(node))
         # There are two valid cases for the methods of Dynamic SNode:
         #
         # 1. x[i, j].append (where the dimension of the field (3 in this case) is equal to one plus the number of the
@@ -613,9 +639,11 @@ class ASTTransformer(Builder):
             raise e
 
         if ASTTransformer.build_attribute_if_is_dynamic_snode_method(ctx, node):
+            print("is dynamic snode")
             return node.ptr
 
         if isinstance(node.value.ptr, Expr) and not hasattr(node.value.ptr, node.attr):
+            print("is expr")
             if node.attr in Matrix._swizzle_to_keygroup:
                 keygroup = Matrix._swizzle_to_keygroup[node.attr]
                 Matrix._keygroup_to_checker[keygroup](node.value.ptr, node.attr)
@@ -647,9 +675,18 @@ class ASTTransformer(Builder):
                 node.ptr = getattr(tensor_ops, node.attr)
                 setattr(node, "caller", node.value.ptr)
         elif dataclasses.is_dataclass(node.value.ptr):
+            print("is dataclass")
             node.ptr = next(field.type for field in dataclasses.fields(node.value.ptr))
         else:
+            print("following .")
+            # print("parent is", type(node.value.ptr))
             node.ptr = getattr(node.value.ptr, node.attr)
+            node.ptr_source = node.value.ptr_source
+            if ctx.is_pure and node.ptr_source == PtrSource.GLOBAL and not ctx.static_scope_status.is_in_static_scope:
+                print("node.ptr_source is global")
+                if isinstance(node.ptr, (int, float, Field)):
+                    raise exception.GsTaichiCompilationError(f"Accessing global var {node.attr} from outside function scope within pure kernel")
+            print('node.ptr', node.ptr, type(node.ptr))
         return node.ptr
 
     @staticmethod
--- a/python/gstaichi/lang/ast/ast_transformer.py
+++ b/python/gstaichi/lang/ast/ast_transformer.py
@@ -11,15 +11,15 @@ from typing import Any, Sequence, Type
 import numpy as np
 
 from gstaichi._lib import core as _ti_core
-from gstaichi.lang import expr, impl, matrix, mesh, exception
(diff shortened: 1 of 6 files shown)


python/gstaichi/lang/ast/ast_transformer_utils.py
```diff
@@ -25,6 +26,13 @@
     )
 
 
+class PtrSource(Enum):
+    GLOBAL = 1
+    TEMPLATE = 2
+    BUILTIN = 3
+    LOCAL = 4
```
duburcqa: Could you add documentation to explain what are all these pointer source types?

python/gstaichi/lang/ast/ast_transformer_utils.py
```diff
@@ -35,7 +43,10 @@ def __call__(self, ctx: "ASTTransformerContext", node: ast.AST):
                 raise GsTaichiSyntaxError(error_msg)
             info = ctx.get_pos_info(node) if isinstance(node, (ast.stmt, ast.expr)) else ""
             with impl.get_runtime().src_info_guard(info):
-                return method(ctx, node)
+                res = method(ctx, node)
+                if not hasattr(node, "ptr_source"):
```
duburcqa: Having to rely on `hasattr` is usually a code smell.

python/gstaichi/lang/ast/ast_transformer_utils.py
```diff
@@ -265,22 +280,35 @@ def check_loop_var(self, loop_var: str) -> None:
...
+    def get_var_by_name(self, name: str) -> tuple[PtrSource, Any]:
         for s in reversed(self.local_scopes):
             if name in s:
-                return s[name]
-        if name in self.global_vars:
+                val = s[name]
+                ptr_source = PtrSource.LOCAL
+                if self.is_pure and is_data_oriented(val):
+                    # a data oriented is as good as global, from pov of pure constraints
+                    ptr_source = PtrSource.GLOBAL
```
duburcqa: This sounds a bit hacky no? And the error message in the end may be confusing?

python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -44,7 +44,9 @@ def _decl_and_create_variable(
             ctx.kernel_args.append(name)
         if annotation == annotations.template or isinstance(annotation, annotations.template):
             assert ctx.global_vars is not None
-            return True, ctx.global_vars[name]
+            if name in ctx.template_vars:
+                return True, ctx.template_vars[name]
+            return True, ctx.global_vars.get(name, ctx.template_vars.get(name))
```
duburcqa: You already know that `name not in ctx.template_vars`, which means that `ctx.template_vars.get(name)` evaluate to `None` systematically?

python/gstaichi/lang/ast/ast_transformer_utils.py
hughperkins: In hindsight, after writing the code, I was thinking, perhaps I should have a single boolean, like `uses_global`. All the conditional logic I've added only ever toggles from 'not GLOBAL' to 'GLOBAL'

python/gstaichi/lang/ast/ast_transformer_utils.py
hughperkins: nodes are created by `ast.parse`. We can't control the classes being returned.

python/gstaichi/lang/ast/ast_transformer_utils.py
hughperkins: I'm open to renaming `global` to something else if you wish. `Non-pure` perhaps?

hughperkins: Perhaps we call it `is_pure_compliant`, and default to `True`.  (Or we could have `violates_pure`, and defaults to `False`.)

python/gstaichi/lang/ast/ast_transformer_utils.py
duburcqa: I see, but `ptr_source` is something added systematically added "manually" by gstaichi? Or it may or may not exist depending on some external conditional that is hard to foresee?

python/gstaichi/lang/ast/ast_transformer_utils.py
duburcqa: I like `is_pure_compliant`, because it not only makes very explicit which property you are tracking, but also why you are tracking this additional property in the first place.

python/gstaichi/lang/ast/ast_transformer_utils.py
duburcqa: Yes, I'm in favour of keeping this as simple as possible. If an enum is not required for now, I would remove it for now. Adding it back should not be too difficult since the code path would already be in place.

python/gstaichi/lang/ast/ast_transformer_utils.py
hughperkins: Cool. One think I noticed is that when we initialize this, often we don't know whether the node is pure compliant or not. If we initialize with `is_pure_compliant=True`, then it makes it sound like we are sure that it is compliant, which might not be the case. So I've flipped it around to`violates_pure=False`. And then, if we discovert hat it does violate pure, we can flip this to `True`, for a node

python/gstaichi/lang/ast/ast_transformer_utils.py
hughperkins: It never exists after ast.parse. We add it when we are walking the tree. ast_transformer_utils is a generic location to add it, during walk, I feel. But sometimes, nodes are walked multiple times.

duburcqa: In that case, you should initialise `ptr_source`to `None` right after call `ast.parse`. And the replace `hasattr` by `is None`.

Still, on the long run, the correct design would be to stop hijacking nodes returned by `ast.parse` and create our own `AstNode` class that is initialised from a built-in ast node. This class would extend the original ast node using composition, and expose additional information by defining its own custom attributes, `ptr_source` being part of them.

python/gstaichi/lang/ast/ast_transformer_utils.py
duburcqa: Not that it may be better to initialise to `None` to distinguish, True, False and undetermined.

python/gstaichi/lang/ast/ast_transformer_utils.py
hughperkins: Well, how to replace? Wouldnt taht mean walking the tree?

python/gstaichi/lang/ast/ast_transformer_utils.py
hughperkins: The thing is, we never know when its determined. It basically defaults to False, until we find a violation.

Think of it like a flag 'is_criminal' on a human. If someone is convicted of a crime, then, assuming they were correctly convicted, we know we can set this flag to True. But otherwise, we don't know if it's true or false.

I suppose what we could do is have only two possible states:
- `None`
- `True`

hughperkins: (or another analogy:when running pytest, a flag like `exist_failing_tests`. Initially False, later might become True).

hughperkins: I could rename it to `pure_violation_detected` perhaps?

python/gstaichi/lang/ast/ast_transformer_utils.py
duburcqa: I don't know well enough this part of the code, it would take me 1 hour to run and analyse the code to better understand how and where nodes are constructed.

hughperkins: The nodes are constructed by `ast.parse`, which is built into Python.

hughperkins: (I did ponder creating our own nodes, when I was trying to improve typing, but I decided that's a can of worms / rabbit hole that I don't feel like opening / wandering down for now)

duburcqa: Ok know I get it, `ast.parse` does not return a single node but a hierarchy I guess... so yes you would need to walk through the entire tree...

hughperkins: Yes exactly. And that's what we are doing here. This function is the most generic function during the walk. It's called for *every* node, during the walk.

python/gstaichi/lang/ast/ast_transformer_utils.py
duburcqa: Nono, it was just a side note. `violates_pure` is just fine I think, but no strong opinion. I'm already happy.

python/gstaichi/lang/ast/ast_transformer_utils.py
duburcqa: Ok let that slip in, but it is not ideal.

hughperkins: Yeah, messes with typing for sure. I'm fairly OCD about typing, so I might poke at this at some point.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -277,6 +277,21 @@ def build_Call(ctx: ASTTransformerContext, node: ast.Call, build_stmt, build_stm
                 assert not hasattr(arg, "ptr")
                 build_stmt(ctx, arg)
 
+        violates_pure = False
+        # if any arg violates pure, then node also violates global
```
duburcqa: This comment does not seem to be related to anything happening here?

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -71,10 +73,19 @@ def boundary_type_cast_warning(expression: Expr) -> None:
 class ASTTransformer(Builder):
     @staticmethod
     def build_Name(ctx: ASTTransformerContext, node: ast.Name):
-        node.ptr = ctx.get_var_by_name(node.id)
+        violates_pure, node.ptr, violates_pure_reason = ctx.get_var_by_name(node.id)
         if isinstance(node, (ast.stmt, ast.expr)) and isinstance(node.ptr, Expr):
             node.ptr.dbg_info = _ti_core.DebugInfo(ctx.get_pos_info(node))
             node.ptr.ptr.set_dbg_info(node.ptr.dbg_info)
+        node.violates_pure = violates_pure
+        node.violates_pure_reason = violates_pure_reason
```
duburcqa: Why not to assign this directly? (as you are doing for `ptr`)
```python
node.violates_pure, node.ptr, node.violates_pure_reason = ctx.get_var_by_name(node.id)
```

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -274,8 +288,15 @@ def build_Tuple(ctx: ASTTransformerContext, node: ast.Tuple):
...
     def build_List(ctx: ASTTransformerContext, node: ast.List):
+        node.violates_pure = False
         build_stmts(ctx, node.elts)
+        reason = []
+        for elt in node.elts:
+            if elt.violates_pure:
+                node.violates_pure = True
+                reason.append("list member violates pure " + str(elt))
         node.ptr = [elt.ptr for elt in node.elts]
+        node.violates_pure_reason = "\n".join(reason) if len(reason) > 0 else None
```
duburcqa: `"\n".join(reason) if reason else None` ?

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -301,6 +322,7 @@ def process_dictcomp(ctx: ASTTransformerContext, node, result) -> None:
 
     @staticmethod
     def process_generators(ctx: ASTTransformerContext, node: ast.GeneratorExp, now_comp, func, result):
+        node.violates_pure = False
```
duburcqa: I would recommend doing `node.violates_pure_reason = None` to make sure all instances (weakly) define the same attributes.

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -351,6 +373,7 @@ def build_Index(ctx: ASTTransformerContext, node: ast.Index):
     @staticmethod
     def build_Constant(ctx: ASTTransformerContext, node: ast.Constant):
         node.ptr = node.value
+        node.violates_pure = False
```
duburcqa: Same. I would add `node.violates_pure_reason`

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -650,6 +675,23 @@ def build_Attribute(ctx: ASTTransformerContext, node: ast.Attribute):
...
+                    violation = True
+                    if violation and isinstance(node.ptr, enum.Enum):
+                        violation = False
+                    if violation and node.value.ptr in [ti_math, math, np]:
+                        # ignore this built-in module
+                        violation = False
+                    if violation:
+                        message = f"WARNING: Accessing global var {node.attr} from outside function scope within pure kernel {node.value.violates_pure_reason}"
+                        if node.attr.upper() == node.attr:
+                            print(message)
```
duburcqa: You should use some proper logging mechanism. Either `warnings` or `logging`. Otherwise it can be silenced / redirected / post-processed easily.

python/gstaichi/lang/ast/ast_transformer_utils.py
```diff
@@ -46,7 +52,13 @@ def __call__(self, ctx: "ASTTransformerContext", node: ast.AST):
             if not isinstance(e, GsTaichiCompilationError):
                 msg = ctx.get_pos_info(node) + traceback.format_exc()
                 raise GsTaichiCompilationError(msg) from None
-            msg = ctx.get_pos_info(node) + str(e)
+            msg = (
+                "gstaichi stack trace:\n====\n"
+                + stack_trace
+                + "\n====\n\nYour code:\n"
+                + ctx.get_pos_info(node)
+                + str(e)
```
duburcqa: I would recommend using f-string but as you wish.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -830,6 +839,8 @@ def ast_to_dict(node: ast.AST | list | primitive_types._python_primitive_types):
                 self.runtime._compiling_callable = None
 
         gstaichi_kernel = impl.get_runtime().prog.create_kernel(gstaichi_ast_generator, kernel_name, self.autodiff_mode)
+        for violation in ctx.pure_violations:
+            print("WARNING: violates pure", violation)
```
duburcqa: Same.  Please avoid plain old print in favour of more flexible warnings mechanism.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: typo. `global` => `pure`. will fix.

hughperkins: addressed in 0d9810fa7

python/gstaichi/lang/ast/ast_transformer.py
hughperkins: reason is initially empty list.

hughperkins: oh I see empty list evaluates to False 🤔

hughperkins: interesting

python/gstaichi/lang/ast/ast_transformer.py
hughperkins: removed entirely, since we are defaulting to False anyway.

python/gstaichi/lang/kernel_impl.py
hughperkins: removed since not using this list currently

PR #234, opened 2025-10-07
[Test] Disable flaky OOB test on windows

hughperkins:
Issue: #

### Brief Summary

Looks like out of bounds detection in debug mode is flaky on windows. This test fails often. pytest.mark.fail'ing it for now.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_ndarray.py
+++ b/tests/python/test_ndarray.py
@@ -1,5 +1,6 @@
 import copy
 import platform
+import sys
 
 import numpy as np
 import pytest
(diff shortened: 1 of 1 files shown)


tests/python/test_ndarray.py
```diff
@@ -833,6 +834,9 @@ def access_arr(input: ti.types.NDArray, x: ti.i32) -> ti.f32:
     gdb_trigger=False,
 )
 def test_matrix_ndarray_oob():
+    if sys.platform == "win32":
+        pytest.xfail("out of bound access detection flaky on windows")
```
duburcqa: If would restrict "dynamic" xfail (i.e. `pytest.xfail` as opposed to `pytest.mark.xfail`) call only for conditions that cannot be determined statically, which is very VERY uncommon but do exist: https://github.com/Genesis-Embodied-AI/Genesis/blob/09a1c5f763879b75711b04e5c065217cfdd5f17e/tests/test_render.py#L1056

Moreover, I would specify the type of error that may occur, to only accept the failures that you are expecting and catch the rest. See example in genesis: https://github.com/Genesis-Embodied-AI/Genesis/blob/09a1c5f763879b75711b04e5c065217cfdd5f17e/tests/test_render.py#L65
```python
@pytest.mark.xfail(sys.platform == "darwin", raises=AssertionError, reason="Flaky on MacOS with CPU-based OpenGL")
```

PR #239, opened 2025-10-17
[Perf] Avoid fast cache corruption and recover from errors

hughperkins:
Issue: #

### Brief Summary

- sometimes the fast cache became corrupted, for example through
    - running tests in parallel
    - aggressively terminating a test in progress
- once the cache was corrupted:
    - any attempt to run the related kernel crashed
    - the cache could not recover
 
This PR:
- swallows any exceptions with failing to read a cache file
- will automatically attempt to write a new versoin of the cache file
- tests both these things

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/_test_tools/__init__.py
+++ b/python/gstaichi/_test_tools/__init__.py
@@ -1,6 +1,6 @@
 import gstaichi as ti
 
-from . import textwrap2
+from . import textwrap2, warnings_helper
 
 
 def ti_init_same_arch(**options) -> None:
@@ -15,4 +15,4 @@ def ti_init_same_arch(**options) -> None:
     ti.init(**options)
 
 
-__all__ = ["textwrap2"]
+__all__ = ["textwrap2", "warnings_helper"]
--- a/python/gstaichi/_test_tools/warnings_helper.py
+++ b/python/gstaichi/_test_tools/warnings_helper.py
@@ -1,0 +1,11 @@
+import warnings
+
+_seen: set[str] = set()
+
+
+def warn_once(message: str) -> None:
+    global _seen
+    if message in _seen:
+        return
+    _seen.add(message)
+    warnings.warn(message)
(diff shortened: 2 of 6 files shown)


python/gstaichi/lang/_fast_caching/python_side_cache.py
```diff
@@ -40,13 +42,32 @@ def _touch(self, filepath):
...
             return None
-        self._touch(filepath)
-        with open(filepath) as f:
-            return f.read()
+        try:
+            with open(filepath) as f:
+                res = f.read()
+            self._touch(filepath)
+            return res
+        except Exception as e:
```
duburcqa: I would recommend catching the actual exception that you are expected. Otherwise you may "silently" catch other issues that should rather be raising an exception.

In this case, `pydantic.ValidationError, json.JSONDecodeError, UnicodeDecodeError`.

python/gstaichi/lang/_fast_caching/python_side_cache.py
```diff
@@ -40,13 +42,32 @@ def _touch(self, filepath):
...
+        tmp_path = None
+        try:
+            target_dir = os.path.dirname(filepath)
+            fd, tmp_path = tempfile.mkstemp(dir=target_dir, prefix=f"{key}.", suffix=".tmp")
+            with os.fdopen(fd, "w") as f:
+                f.write(value)
+                f.flush()
+                os.fsync(f.fileno())
+            os.replace(tmp_path, filepath)
+        except Exception as e:
```
duburcqa: I would recommend catching the actual exception that you are expected. Otherwise you may "silently" catch other issues that should rather be raising an exception.

In this particular context, I don't think raising any kind of exception can be considered "expected". I would just remove it entirely.

hughperkins: You're saying, if we fail to write to cache, we make the program fail to run?

duburcqa (approved):

python/gstaichi/lang/_fast_caching/python_side_cache.py
duburcqa: Yes, I think it makes sense. I don't remember observing anything like this, even in case of race conditions. Reading is another story.

PR #243, opened 2025-10-20
[Perf] Add fastcache= to ti.kernel and mark other approaches as deprecated

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/manylinux_wheel.yml
+++ b/.github/workflows/manylinux_wheel.yml
@@ -28,6 +28,18 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: Free Disk Space
+        run: |
+          df -h
+          pushd /opt/python
+          rm -Rf cp39-cp39 cp314-cp314 cp314-cp314t cp38-cp38 cp313-cp313t pp310-pypy310_pp73 pp311-pypy311_pp73
+          for cpversion in cp310 cp311 cp312 cp313; do {
+            if [[ "${cpversion}" != "${{ matrix.PYTHON_CP_VERSION }}" ]]; then
+              rm -Rf ${cpversion}-${cpversion}
+            fi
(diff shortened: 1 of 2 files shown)


python/gstaichi/lang/kernel_impl.py
```diff
@@ -1349,7 +1351,12 @@ def decorator(fn: F, has_kernel_params: bool = True) -> F:
             level = 4
 
         wrapped = _kernel_impl(fn, level_of_class_stackframe=level)
-        wrapped.is_pure = pure
+        wrapped.is_pure = pure or fastcache
+        if pure:
+            warnings_helper.warn_once(
+                "@ti.kernel parameter `pure` is deprecated. Please use parameter `fastcache`. "
+                "`pure` parameter is intended to be removed in 4.0.0"
+            )
```
duburcqa, line 1359: I would rather suggest to update the default value `pure: bool | None = None`. It would allow to properly detect unused vs disabled. For instance in genesis we actively set it to false.

hughperkins: It only has two possible values from gstaichi pov: false or true
- if true, then the kernel is eligible for fastcache
- if false it is not
I can update the type at line 160 to be explicilty bool.

What advantage do you see to it having 3 possible values?

duburcqa: > It only has two possible values from gstaichi pov: false or true

Yes, this is exactly my point!

> What advantage do you see to it having 3 possible values?

As I said, detecting use of deprecated argument(s).

hughperkins: 🤔

hughperkins: oh I see.

PR #244, opened 2025-10-20
[Build] Free up CI runner disk space

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/scripts_new/manylinux_wheel/1_prerequisites.sh
+++ b/.github/workflows/scripts_new/manylinux_wheel/1_prerequisites.sh
@@ -8,17 +8,9 @@ yum install -y git wget
 git config --global --add safe.directory /__w/gstaichi/gstaichi
 git submodule update --init --jobs 2
 
-# wget -q https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.4/clang+llvm-15.0.4-x86_64-linux-gnu-rhel-8.4.tar.xz
-# tar -xf clang+llvm-15.0.4-x86_64-linux-gnu-rhel-8.4.tar.xz
-
-wget -q https://github.com/Genesis-Embodied-AI/gstaichi-sdk-builds/releases/download/llvm-15.0.7-202510071403/taichi-llvm-15.0.7-linux-x86_64.zip
-unzip taichi-llvm-15.0.7-linux-x86_64.zip
-ls
-mkdir -p ~/.cache/ti-build-cache
-# yes, this is super hacky...
-mv taichi-llvm-15.0.7-linux-x86_64 ~/.cache/ti-build-cache/llvm-15.0.7-202510071403
-ls ~/.cache/ti-build-cache/
-df -h
+wget -q https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.4/clang+llvm-15.0.4-x86_64-linux-gnu-rhel-8.4.tar.xz
+tar -xf clang+llvm-15.0.4-x86_64-linux-gnu-rhel-8.4.tar.xz
+rm clang+llvm-15.0.4-x86_64-linux-gnu-rhel-8.4.tar.xz
 
 # clang++ searches for libstd++.so, not libstdc++.so.6
 # without this, then the compiler checks will fail
--- a/.github/workflows/scripts_new/manylinux_wheel/2_build.sh
+++ b/.github/workflows/scripts_new/manylinux_wheel/2_build.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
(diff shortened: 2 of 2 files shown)


.github/workflows/scripts_new/manylinux_wheel/1_prerequisites.sh
```diff
@@ -8,8 +8,17 @@ yum install -y git wget
...
+# tar -xf clang+llvm-15.0.4-x86_64-linux-gnu-rhel-8.4.tar.xz
+
+wget -q https://github.com/Genesis-Embodied-AI/gstaichi-sdk-builds/releases/download/llvm-15.0.7-202510071403/taichi-llvm-15.0.7-linux-x86_64.zip
+unzip taichi-llvm-15.0.7-linux-x86_64.zip
+ls
+mkdir -p ~/.cache/ti-build-cache
+# yes, this is super hacky...
+mv taichi-llvm-15.0.7-linux-x86_64 ~/.cache/ti-build-cache/llvm-15.0.7-202510071403
+ls ~/.cache/ti-build-cache/
+df -h
```
duburcqa: I don't understand what this does. Could you elaborate?

By the way, you could free space by just remove the archive after download no? I don' think it is done automatically by `tar -xf` but I may be wrong.

hughperkins: yes, thats a good point. i do that for the main download, but not for this one. Perhaps I could just revert this change for now, and remove the archive as you say. I'll try that. Would be a lot less fragile, though not perfectly efficient in disk space.

As to your question....

Yes, so ....

- clang: We use clang to compile the runtime module, in gstaichi/runtime/llvm/runtime_module/CMakeLists.txt
- llvm: We use llvm library, within the gstaichi c++ code, to manipulate the incoming python AST, and gradually transform it into our desired output format
- (then, the clang-built runtime module is linked with the things we've compiled using llvm)

We get llvm from the pre-built llvm sdk.

As far as clang, we have a couple of choices:
- can use an operating system supplied version
- can provide/download our own

In the case of Mac, the operating system provided one is too new to work with llvm 15: the runtime module built with clang 16 or 17 won't load into llvm 15, as far as I can tell.

So, I switched from using the operating system clang to a clang we provide/download ourselves.

For linux, I can't remember exactly why I changed to using the downloaded version. There seem to be a couple of possibilites in my mind:
- complete uniformity across all linux versions: always using the exact same version of clang
- don't need to figure out how to install the operating system version, and hack around with which operating system provided version will work ok-ish

Anyway, the clang and llvm are configured in two separate places:
- clang is configured in /CMakeLists.txt, and used in /gstaichi/runtime/llvm/runtime_module/CMakeLists.txt
    - this clang is the one you see being downloaded here
- llvm is configured and downloaded in /.github/workflows/scripts/ti_build/llvm.py

Till now, these were downloaded and unzipped entirely separately (its' 3.5GB after unzip by the way...).

Then, I figured, lets just download once.

BUT, since they're configured completely independently, what I have done very hackily here is:
- take the clang download, and place it exactly where llvm.py is going to cache it
    - llvm.py sees the already downloaded llvm, and thinks "woo, it's already there! nothing to do" and just uses that
    - however, llvm.py, well it used to look for it in a generic place (~/.cache/ti_build_cache/llvm), but I changed it to be very specific now (with the build version and so on)
    - soooo, in order for this to work, this clang download has to position the llvm download exactly in this folder with a very specific name
    - as soon as someone changes the build version in llvm.py, or antying at all, then bam! Back to two separate downloads, and out of disk space again 😅 , and in a very uninuitive way.


Soo.... I will try reverting this change, and removing the archive, and see if thats sufficient ot avoid the disk space issue.

hughperkins: By the way, when one builds 'llvm', one has the choice:
- build just llvm (definitley possible)
- build clang + llvm (waht I'm doing currenlty)
- build just clang (not sure if this is possible, but seems plausible

I kind of like building both, because then we can make sure we use the same version of clang as llvm. But it might not be strictly necessary, and might not stand up to close scrutiny, or tight disk space pressure. But I kind of want to go in this direction, ideally, as long as I don't encounter any major obstacles.

hughperkins: appears that builds :)

hughperkins: Alright, maybe I merge this, and then I can faff around with the cleaner version https://github.com/Genesis-Embodied-AI/gstaichi/pull/245 , in parallel with building release for the cache corruption fix.

PR #250, opened 2025-10-26
[Perf] Optimize kernel launch overhead

duburcqa:
This brings major simulation speed up for Genesis, using both field and ndarray. It was benchmarked by running the following command on CoreWeave compute node (Nvidia H100):
```bash
py-spy record -r 85 -o benchmark.svg -- pytest -m "benchmarks" -n 0 \
    "tests/test_rigid_benchmarks.py::test_speed[False-30000-CG-batched_franka]"
```

| data type    |  BEFORE | AFTER | RATIO |
| -------- | ------- | ------- | ------- | 
| field  |  18.4M    | 22.0M | +20% |
| ndarray  | 1.86M     | 3.3M | +77% |

#### BEFORE

image

#### AFTER

image

#### BEFORE

image

#### AFTER

image


python/gstaichi/lang/util.py
```diff
@@ -239,17 +239,32 @@ def to_gstaichi_type(dt):
...
 
+class DataTypeCxxWrapper(_ti_core.DataTypeCxx):
+    __slots__ = ('_hash',)
+
+    def __init__(self, dtype: _ti_core.Type):
+        super().__init__(dtype)
+        self._hash = super().__hash__()
+
+    def __hash__(self):
+        return self._hash
```
duburcqa, line 261: This wrapper is the only part that is "dirty". It is used to speed up hashing (for now), and could be extended to shape and element_type.

python/gstaichi/lang/_template_mapper.py
```diff
@@ -86,121 +73,152 @@ def extract_arg(raise_on_templated_floats: bool, arg: Any, annotation: Annotatio
...
+        # [Primitive arguments] Return the value
+        if raise_on_templated_floats and arg_type is float:
+            raise ValueError("Floats not allowed as templated types.")
+        return arg
+    if annotation_type is ndarray_type.NdarrayType:
+        if isinstance(arg, Ndarray):
+            # FIXME: Disabling this check because it is very expensive
+            # if __debug__:
+            #     annotation.check_matched(arg.get_type(), arg_name)
+            #     assert arg.shape is not None
```
duburcqa: I commented this myself. It is not ideal. But it is damn slow...

duburcqa: Re-enabled with a runtime evaluation of the debug flag to allow overwrite by the user (i.e. me in genesis).

python/gstaichi/lang/_template_mapper.py
```diff
@@ -39,6 +35,149 @@
 ]
 
 
+_ExprCxx = _ti_core.ExprCxx
+_composite_types = {list, dict, set}
```
hughperkins: should these types be factorized into some more generic file? (similar to primitive_types.py)

python/gstaichi/lang/_template_mapper.py
```diff
@@ -39,6 +35,149 @@
...
+            return arg.ptr
+        if arg_type is Expr:
+            return arg.ptr.get_underlying_ptr_address()
+        if arg_type is _ExprCxx:
+            return arg.get_underlying_ptr_address()
+        if arg_type is tuple:
+            return tuple([_extract_arg(raise_on_templated_floats, item, annotation, arg_name) for item in arg])
+        if issubclass(arg_type, Ndarray):
+            raise GsTaichiRuntimeTypeError(
+                "Ndarray shouldn't be passed in via `ti.template()`, please annotate your kernel using `ti.types.ndarray(...)` instead"
```
hughperkins: `ti.types.NDArray[...]` is my preferred usage.  Also, `ti.Template` rather than `ti.template()`.

python/gstaichi/lang/_template_mapper.py
```diff
@@ -39,6 +35,149 @@
...
+            )
+        if arg_type in _composite_types or getattr(arg_type, "_data_oriented", False):
```
hughperkins: we have a utility function for this. You can search for `is_data_oriented`

python/gstaichi/lang/_template_mapper.py
```diff
@@ -39,6 +35,149 @@
...
+                annotation.check_matched(ty, arg_name)
+            return ty.element_type, len(arg.shape), ty.needs_grad, annotation.boundary
+        # external arrays
+        shape = getattr(arg, "shape", None)
+        if shape is None:
+            raise GsTaichiRuntimeTypeError(f"Invalid type for argument {arg_name}, got {arg}")
+        shape = tuple(shape)
+        element_shape: tuple[int, ...] = ()
+        dtype = to_gstaichi_type(arg.dtype)
+        if isinstance(annotation.dtype, MatrixType):
```
hughperkins: out of curiosity, why preserve this isinstance?

python/gstaichi/lang/_template_mapper.py
duburcqa: I don't think so. In my view it is very specific to this file. It is only defined as a global variable to avoid allocating this variable systematically every time this method is called.

Same for `_ExprCxx = _ti_core.ExprCxx`. This is only to spare on lookup and therefore speed up the check.

python/gstaichi/lang/_template_mapper.py
duburcqa: What do you mean exactly? This is a strict copy paste of the original code :)

python/gstaichi/lang/_template_mapper.py
duburcqa: Yes, but calling a utility add one extra function call interaction, which is very costly in Python. So it must be inlined.

python/gstaichi/lang/_template_mapper.py
duburcqa: Because the type is derived in practice, so you specifically need the extra work that isinstance is doing. Same for `issubclass(arg_type, Ndarray)` right above.

python/gstaichi/lang/_template_mapper.py
```diff
@@ -1,31 +1,27 @@
-import dataclasses
 import weakref
+from dataclasses import _FIELD, _FIELDS
```
hughperkins: These are private. Should we be importing these?

python/gstaichi/lang/_template_mapper.py
```diff
@@ -39,6 +35,149 @@
...
+                    f"Invalid value for argument {arg_name} - required array has ndim={annotation.ndim}, "
+                    f"array with {len(shape)} dimensions is provided"
+                )
+        needs_grad = getattr(arg, "requires_grad", False) if annotation.needs_grad is None else annotation.needs_grad
+        if element_shape:
+            element_type = _ti_core.get_type_factory_instance().get_tensor_type(element_shape, dtype)
+        else:
+            element_type = arg.dtype
+        return element_type, len(shape) - len(element_shape), needs_grad, annotation.boundary
+    annotation_fields = getattr(annotation, _FIELDS, None)
```
hughperkins: should we be using `.fields()` public api, rather than `_FIELDS` private attribute and `getattr`?

python/gstaichi/lang/_template_mapper.py
duburcqa: No we shouldn't. But this makes a very sizeable performance difference by avoiding several function interaction and runtime checks. I would rather keep it like this and see if it breaks with future python versions but I doubt so.

python/gstaichi/lang/_template_mapper.py
duburcqa: As mentioned before. We should in theory, but it is much slower. we can keep is_dataclass and fields calls everywhere where it is not performance critical, but here it makes a big difference.

python/gstaichi/lang/_template_mapper.py
hughperkins: You have been pretty strict on not allowing `getattr`, in your code-reviews of my code. Specifically for this particular usage. Which is why I made the helper function. How to reconcile these two situations?

python/gstaichi/lang/_template_mapper.py
hughperkins: oh, fair enough.

python/gstaichi/lang/_template_mapper.py
duburcqa: In this particular case we are (literally) chasing nanoseconds, not even microseconds. We have to get our hands dirty, and the expense of code that would be harder to maintain. Or to give up optimization. But it is the game in Python. Luckily, only 2 files in gstaichi require this level of optimization, _template_mapper and kernel_impl.

python/gstaichi/lang/_template_mapper.py
duburcqa: I can updated it if you have something in mind. So far I just did translation from inefficient to efficient idempotent code. So I don't necessarily know what is going on here.

python/gstaichi/lang/_template_mapper_hotpath.py
```diff
@@ -0,0 +1,181 @@
+# pyright: reportAttributeAccessIssue=false
```
hughperkins, line 1: could we add a docstring explaining what is special about this file please?

also, on the whole, I'd prefer the name to be very clear, like `_template_mapper_performance_critical_impl.py`, or semantically similar.

Currently, this file, from the name, just sounds like it's trying to hide some implmentation-specific stuff from clients, which isn't really its primary raison d'etre I feel?

What we want is so that in the future, people don't add random stuff into this file, not realizing what its very specific purpose is. And reviewsers will approve that, also not knowing whta its very specific purpose is.

PR #251, opened 2025-10-28
[Perf] Cleanup launch kernel logic

duburcqa:
This PR does not aim at improving performance. Just removing unnecessary code complexity and boilerplate. Still the simulation is marginally faster for ndarray (3.31M to 3.35M FPS).

### DIFF
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -18,6 +18,8 @@ from dataclasses import (
     dataclass,
     is_dataclass,
 )
+
+# Must import 'partial' directly instead of the entire module to avoid attribute lookup overhead.
 from functools import partial, update_wrapper, wraps
 from typing import Any, Callable, Type, TypeVar, cast, overload
 
@@ -680,6 +682,7 @@ def _recursive_set_args(
         for field in needed_arg_fields.values():
             if field._field_type is not _FIELD:
                 continue
+            # Storing attribute in a temporary to avoid repeated attribute lookup (~20ns penalty)
             field_type = field.type
             assert not isinstance(field_type, str)
             field_value = getattr(v, field.name)
@@ -720,12 +723,15 @@ def _recursive_set_args(
             element_dim = needed_arg_dtype.ndim
             array_shape = v.shape[element_dim:] if is_soa else v.shape[:-element_dim]
         if isinstance(v, np.ndarray):
-            if v.flags.f_contiguous:
+            # Check ndarray flags is expensive (~250ns), so it is important to order branches according to hit stats
+            if v.flags.c_contiguous:
+                pass
+            elif v.flags.f_contiguous:
                 # TODO: A better way that avoids copying is saving strides info.
                 v_contiguous = np.ascontiguousarray(v)
                 v, v_orig_np = v_contiguous, v
                 callbacks.append(partial(np.copyto, v_orig_np, v))
-            elif not v.flags.c_contiguous:
+            else:
                 raise ValueError(
                     "Non contiguous numpy arrays are not supported, please call np.ascontiguousarray(arr) "
                     "before passing it into gstaichi kernel."


python/gstaichi/lang/kernel_impl.py
```diff
@@ -19,7 +18,8 @@
     dataclass,
     is_dataclass,
 )
-from typing import Any, Callable, Iterable, Type, TypeVar, cast, overload
+from functools import partial, update_wrapper, wraps
```
hughperkins, line 23: could you add a comment explaining why you are importing the functions directly, rahter than the module? (I'm assuming it's because it improve perf?)

python/gstaichi/lang/kernel_impl.py
```diff
@@ -632,25 +635,23 @@ def cast_int(x: int | np.integer) -> int:
...
-    actual_argument_slot: int,
-    exceed_max_arg_num: bool,
-    set_later_list: list[tuple[Callable, Iterable[Any]]],
     needed_arg_type: Type,
     provided_arg_type: Type,
     v: Any,
     indices: tuple[int, ...],
-    callbacks: list[Callable[[], None]],
-    buffers: list[np.ndarray],
+    actual_argument_slot: int,
```
hughperkins, line 644: is there a reason why actual_argument_slot is moving within the parameter list? (not asking it not to be, just curious)

python/gstaichi/lang/kernel_impl.py
```diff
@@ -679,19 +680,17 @@ def _recursive_set_args(
         for field in needed_arg_fields.values():
             if field._field_type is not _FIELD:
                 continue
-            assert not isinstance(field.type, str)
+            field_type = field.type
```
hughperkins, line 686: You know, I originally had things like field_type, but you asked me not to do that, because you felt it wastes a variable 😛

Could you explain why you are adding it back now, and what are your criterion for when to create a new variable field_type, and when to use field.type instead?

python/gstaichi/lang/kernel_impl.py
```diff
@@ -721,25 +720,17 @@ def _recursive_set_args(
             element_dim = needed_arg_dtype.ndim
             array_shape = v.shape[element_dim:] if is_soa else v.shape[:-element_dim]
         if isinstance(v, np.ndarray):
-            # numpy
-            if v.flags.c_contiguous:
-                launch_ctx.set_arg_external_array_with_shape(indices, int(v.ctypes.data), v.nbytes, array_shape, 0)
-            elif v.flags.f_contiguous:
+            if v.flags.f_contiguous:
```
hughperkins: can you add a comment explaining why you are sequencing the conditional the way that you are (so that someone editing the code knows this)

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1166,34 +1140,29 @@ def launch_kernel(self, t_kernel: KernelCxx, compiled_kernel_data: CompiledKerne
...
-        if has_ret or self.has_print:
-            runtime_ops.sync()
-
-        if has_ret:
-            ret = []
-            for i, ret_type in enumerate(ret_dt):
-                ret.append(self.construct_kernel_ret(launch_ctx, ret_type, (i,)))
-            if len(ret_dt) == 1:
-                ret = ret[0]
         for c in callbacks:
```
hughperkins, line 1149: could you explain why you reordered for c in callbacks to be before the return type bit?

python/gstaichi/lang/kernel_impl.py
duburcqa: Importing 'partial' is important for performance, the other one for consistency. Are you ok with it? We can mix both if you want.

python/gstaichi/lang/kernel_impl.py
duburcqa: Because logically its place below indices I feel: it is closely related to indices but of less importance.

python/gstaichi/lang/kernel_impl.py
hughperkins: my question is: is there a performance reason for not doing `import functools`? Or is it simply personal preference?

python/gstaichi/lang/kernel_impl.py
duburcqa: > You know, I originally had things like field_type, but you asked me not to do that, because you felt it wastes a variable

Yeah, once again, we were not really aware at that time that we were chasing nanoseconds.

> Could you explain why you are adding it back now, and what are your criterion for when to create a new variable field_type, and when to use field.type instead?

I will had a comment. Performance concern to avoid repeated attribute lookup.

python/gstaichi/lang/kernel_impl.py
duburcqa: You are right. This is bad.

python/gstaichi/lang/kernel_impl.py
duburcqa: Because there is no reason to do this afterward. It is just confusing. The callbacks are about updating input arguments that are supposed to be read-write. For me it is part of "executing" the kernel. Fetching the return value comes after.

python/gstaichi/lang/kernel_impl.py
duburcqa: And my answer is:

Importing 'partial' directly is important for performance. Importing 'update_wrapper', 'wraps' directly is for consistency.

hughperkins: Ok. Could you add a comment stating that we are importing partial directly for performacne reasons please? (otherwise someone might come along later, and import just functools)

hughperkins: Please don't reoslve the convesrsations yourself. I belive the convention is taht the reviwer resolves conversations. (There si a thread I can dig up about this, for gs-core, but which I think we can apply to gstaichi).

hughperkins: (unresolving it for now)

python/gstaichi/lang/kernel_impl.py
```diff
@@ -19,7 +18,10 @@
     dataclass,
     is_dataclass,
 )
-from typing import Any, Callable, Iterable, Type, TypeVar, cast, overload
+
+# Most import 'partial' directly instead of the entire module to avoid attribute lookup overhead.
```
hughperkins: "Must"

hughperkins: also, comma before and after "instead of the entire module"

duburcqa: Oh right.

python/gstaichi/lang/kernel_impl.py
duburcqa: > Please don't reoslve the convesrsations yourself. I belive the convention is taht the reviwer resolves conversations.

As you wish!

duburcqa: > (There si a thread I can dig up about this, for gs-core, but which I think we can apply to gstaichi).

I'm not used to it because it does not apply to Genesis.

PR #253, opened 2025-10-28
[Perf] Optimize of kernel launch overhead on C++ side

duburcqa:
| BEFORE | AFTER |
| -- | -- |
| 4.25M | 4.51M |

### DIFF
--- a/gstaichi/program/launch_context_builder.cpp
+++ b/gstaichi/program/launch_context_builder.cpp
@@ -57,6 +57,15 @@ void LaunchContextBuilder::set_arg_float(const std::vector<int> &arg_id,
   }
 }
 
+void LaunchContextBuilder::set_args_float(const std::vector<int> &args_id,
+                                          const std::vector<float64> &vec) {
+  const size_t num_arrs = args_id.size();
+  TI_ASSERT(num_arrs == vec.size());
+  for (int i = 0; i < num_arrs; i++) {
+    set_arg_float({args_id[i]}, vec[i]);
+  }
+}
+
 template <typename T>
 void LaunchContextBuilder::set_struct_arg(std::vector<int> arg_indices, T d) {
   auto dt = kernel_->args_type->get_element_type(arg_indices);
@@ -136,9 +145,27 @@ void LaunchContextBuilder::set_arg_int(const std::vector<int> &arg_id,
   }
 }
 
+void LaunchContextBuilder::set_args_int(const std::vector<int> &args_id,
+                                        const std::vector<int64> &vec) {
+  const size_t num_arrs = args_id.size();
+  TI_ASSERT(num_arrs == vec.size());
+  for (int i = 0; i < num_arrs; i++) {
+    set_arg_int({args_id[i]}, vec[i]);
+  }
+}
+
 void LaunchContextBuilder::set_arg_uint(const std::vector<int> &arg_id,
                                         uint64 d) {
   set_arg_int(arg_id, d);
+}
+
+void LaunchContextBuilder::set_args_uint(const std::vector<int> &args_id,
+                                         const std::vector<uint64> &vec) {
+  const size_t num_arrs = args_id.size();
+  TI_ASSERT(num_arrs == vec.size());
+  for (int i = 0; i < num_arrs; i++) {
+    set_arg_uint({args_id[i]}, vec[i]);
+  }
 }
 
 template <>
@@ -230,7 +257,7 @@ void LaunchContextBuilder::set_arg_external_array_with_shape(
   }
   set_array_runtime_size(arg_id, size);
   set_array_device_allocation_type(arg_id, DevAllocType::kNone);
-  for (uint64 i = 0; i < shape.size(); ++i) {
+  for (uint64 i = 0; i < shape.size(); i++) {
     set_struct_arg(concatenate_vector<int>(arg_id, {0, (int32)i}),
                    (int32)shape[i]);
   }
@@ -245,17 +272,17 @@ void LaunchContextBuilder::set_arg_ndarray(const std::vector<int> &arg_id,
 }
 
 void LaunchContextBuilder::set_args_ndarray(
-    const std::vector<int> &arg_ids,
+    const std::vector<int> &args_id,
     const std::vector<Ndarray *> &arrs) {
-  const size_t num_arrs = arg_ids.size();
+  const size_t num_arrs = args_id.size();
 
   array_ptrs.reserve(array_ptrs.size() + num_arrs);
   array_runtime_sizes.reserve(array_runtime_sizes.size() + num_arrs);
   device_allocation_type.reserve(device_allocation_type.size() + num_arrs);
 
-  TI_ASSERT(num_arrs == arg_ids.size());
-  for (int i = 0; i < num_arrs; ++i) {
-    const int arg_id = arg_ids[i];
+  TI_ASSERT(num_arrs == args_id.size());
+  for (int i = 0; i < num_arrs; i++) {
+    const int arg_id = args_id[i];
     const Ndarray &arr = *arrs[i];
 
     intptr_t ptr = arr.get_device_allocation_ptr_as_int();
@@ -280,6 +307,18 @@ void LaunchContextBuilder::set_arg_ndarray_with_grad(
   TI_ASSERT_INFO(arr.shape.size() <= gstaichi_max_num_indices,
                  "External array cannot have > {max_num_indices} indices");
   set_arg_ndarray_impl(arg_id, ptr, arr.shape, ptr_grad);
+}
+
+void LaunchContextBuilder::set_args_ndarray_with_grad(
+    const std::vector<int> &args_id,
+    const std::vector<Ndarray *> &arrs,
+    const std::vector<Ndarray *> &arrs_grad) {
+  const size_t num_arrs = args_id.size();
+  TI_ASSERT(num_arrs == arrs.size());
+  TI_ASSERT(num_arrs == arrs_grad.size());
+  for (int i = 0; i < num_arrs; i++) {
+    set_arg_ndarray_with_grad({args_id[i]}, *arrs[i], *arrs_grad[i]);
+  }
 }
 
 void LaunchContextBuilder::set_arg_texture(const std::vector<int> &arg_id,
@@ -311,7 +350,7 @@ void LaunchContextBuilder::set_arg_rw_texture_impl(
   array_ptrs[arg_id] = (void *)alloc_ptr;
   set_array_device_allocation_type(arg_id, DevAllocType::kRWTexture);
   TI_ASSERT(shape.size() <= gstaichi_max_num_indices);
-  for (int i = 0; i < shape.size(); ++i) {
+  for (int i = 0; i < shape.size(); i++) {
     set_struct_arg(concatenate_vector<int>(arg_id, {0, i}), shape[i]);
   }
 }
--- a/gstaichi/program/launch_context_builder.h
+++ b/gstaichi/program/launch_context_builder.h
@@ -26,10 +26,25 @@ class LaunchContextBuilder {
   LaunchContextBuilder &operator=(const LaunchContextBuilder &) = delete;
 
   void set_arg_float(const std::vector<int> &arg_id, float64 d);
+  // Bulk processing of multiple scalar float arguments at the same time.
+  // This is mainly useful to mitigate pybind11 function call overhead.
+  // In this context, 'args_id' is a vector gathering the position of each
+  // of these scalar arguments in the corresponding kernel. As a result, the
+  // length 'args_id' and 'vec' must be equal.
+  void set_args_float(const std::vector<int> &args_id,
+                      const std::vector<float64> &vec);
 
   // Created signed and unsigned version for argument range check of pybind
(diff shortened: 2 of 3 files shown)


gstaichi/program/launch_context_builder.cpp
```diff
@@ -242,6 +244,25 @@ void LaunchContextBuilder::set_arg_ndarray(const std::vector<int> &arg_id,
   set_arg_ndarray_impl(arg_id, ptr, arr.shape);
 }
 
+void LaunchContextBuilder::set_arg_ndarray(const std::vector<int> &arg_ids,
```
hughperkins: can we make the name plural? `set_arg_ndarrays`?

Do we still need the old method by the way? If not ,could we remove it, to simplify the code?

gstaichi/program/launch_context_builder.cpp
```diff
@@ -294,7 +315,7 @@ void LaunchContextBuilder::set_arg_ndarray_impl(const std::vector<int> &arg_id,
   // Set array ptr
   array_ptrs[concatenate_vector<int>(
       arg_id, {TypeFactory::DATA_PTR_POS_IN_NDARRAY})] = (void *)devalloc_ptr;
-  if (devalloc_ptr != 0) {
+  if (devalloc_ptr_grad != 0) {
```
hughperkins, line 365: why is this changing? Was the original version a bug?

gstaichi/program/launch_context_builder.cpp
```diff
@@ -224,8 +224,10 @@ void LaunchContextBuilder::set_arg_external_array_with_shape(
                  "External array cannot have > {max_num_indices} indices");
   array_ptrs[concatenate_vector<int>(
       arg_id, {TypeFactory::DATA_PTR_POS_IN_NDARRAY})] = (void *)ptr;
-  array_ptrs[concatenate_vector<int>(
-      arg_id, {TypeFactory::GRAD_PTR_POS_IN_NDARRAY})] = (void *)grad_ptr;
+  if (grad_ptr != 0) {
```
hughperkins, line 254: why is this conditional being added? Was the absence of the conditional a bug before?

gstaichi/program/launch_context_builder.cpp
duburcqa: Yes :)

gstaichi/program/launch_context_builder.cpp
duburcqa: No, just an optimisation. Not calling dealloc on nullptr.

gstaichi/program/launch_context_builder.cpp
duburcqa: done.

gstaichi/program/launch_context_builder.cpp
duburcqa: But a soft one. Just doing useless deallocation.

gstaichi/program/launch_context_builder.cpp
```diff
@@ -242,6 +244,33 @@ void LaunchContextBuilder::set_arg_ndarray(const std::vector<int> &arg_id,
   set_arg_ndarray_impl(arg_id, ptr, arr.shape);
 }
 
+void LaunchContextBuilder::set_args_ndarray(
+    const std::vector<int> &arg_ids,
+    const std::vector<Ndarray *> &arrs) {
```
hughperkins, line 276: could we add some doc string one what the arg_ids represents please? I know there wasn't any before, but not having any also makes the code quite hard to read I feel?

gstaichi/program/launch_context_builder.cpp
```diff
@@ -242,6 +244,33 @@ void LaunchContextBuilder::set_arg_ndarray(const std::vector<int> &arg_id,
...
+  const size_t num_arrs = arg_ids.size();
+
+  array_ptrs.reserve(array_ptrs.size() + num_arrs);
+  array_runtime_sizes.reserve(array_runtime_sizes.size() + num_arrs);
+  device_allocation_type.reserve(device_allocation_type.size() + num_arrs);
```
hughperkins, line 281: why are we adding these?

gstaichi/program/launch_context_builder.cpp
```diff
@@ -242,6 +244,33 @@ void LaunchContextBuilder::set_arg_ndarray(const std::vector<int> &arg_id,
   set_arg_ndarray_impl(arg_id, ptr, arr.shape);
 }
 
+void LaunchContextBuilder::set_args_ndarray(
+    const std::vector<int> &arg_ids,
+    const std::vector<Ndarray *> &arrs) {
+  const size_t num_arrs = arg_ids.size();
+
+  array_ptrs.reserve(array_ptrs.size() + num_arrs);
```
hughperkins, line 279: same question: why are these being added?

gstaichi/program/launch_context_builder.cpp
```diff
@@ -242,6 +244,33 @@ void LaunchContextBuilder::set_arg_ndarray(const std::vector<int> &arg_id,
...
+  array_runtime_sizes.reserve(array_runtime_sizes.size() + num_arrs);
+  device_allocation_type.reserve(device_allocation_type.size() + num_arrs);
+
+  TI_ASSERT(num_arrs == arg_ids.size());
+  for (int i = 0; i < num_arrs; ++i) {
```
hughperkins: I feel like `i++` is more conventional? Is there any reason why we need `++i` here?

gstaichi/program/launch_context_builder.cpp
```diff
@@ -242,6 +244,33 @@ void LaunchContextBuilder::set_arg_ndarray(const std::vector<int> &arg_id,
...
+    const int arg_id = arg_ids[i];
+    const Ndarray &arr = *arrs[i];
+
+    intptr_t ptr = arr.get_device_allocation_ptr_as_int();
+    array_ptrs[{arg_id, TypeFactory::DATA_PTR_POS_IN_NDARRAY}] = (void *)ptr;
+    set_array_device_allocation_type({arg_id}, DevAllocType::kNdarray);
+    const std::vector<int> &shape = arr.shape;
+    size_t total_size = 1;
```
hughperkins, line 292: I wonder if converting the shape into totla size should be factorized out into a separate function?

hughperkins: oh, there's also the set_struct_arg function being called though 🤔

gstaichi/program/launch_context_builder.cpp
```diff
@@ -242,6 +244,33 @@ void LaunchContextBuilder::set_arg_ndarray(const std::vector<int> &arg_id,
...
+    for (int32 i = 0; i < shape.size(); i++) {
+      set_struct_arg({arg_id, 0, i}, (int32)shape[i]);
```
hughperkins, line 294: could we add some doc about what {arg_id, 0, i} represents please? I know it wasnt added by your PR, but it's fairly confusing without hte doc string I think?

gstaichi/program/launch_context_builder.h
```diff
@@ -74,6 +74,8 @@ class LaunchContextBuilder {
                             const std::vector<int> &shape,
                             intptr_t devalloc_ptr_grad = 0);
   void set_arg_ndarray(const std::vector<int> &arg_id, const Ndarray &arr);
+  void set_args_ndarray(const std::vector<int> &arg_ids,
```
hughperkins: do we still need `set_arg_ndarray` for anything? Can we remove it?

Same question with all the other duplicate set_arg functions.

gstaichi/python/export_lang.cpp
```diff
@@ -635,21 +635,21 @@ void export_lang(py::module &m) {
 
   py::class_<LaunchContextBuilder>(m, "KernelLaunchContext")
       .def("set_arg_int", &LaunchContextBuilder::set_arg_int)
-      .def("set_arg_int",
+      .def("set_args_int",
```
hughperkins: question: is there any reason why we couldn't move the iteration code inside a function, similar to how you are doing for set_args_ndarray?

gstaichi/program/launch_context_builder.cpp
duburcqa: This is to improve efficiency. Memory is allocated in a bulk once, then all key assignment that are coming after are basically free. This improves performance because allocating larger chunks of memory at once is more efficient than element by element.

gstaichi/program/launch_context_builder.cpp
duburcqa: Same remark.

gstaichi/program/launch_context_builder.cpp
duburcqa: > I feel like i++ is more conventional? Is there any reason why we need ++i here?

Yes, it is a copy-paste of what was done below in the existing code. Besides, I do prefer ++i because on paper it is supposed to be faster, but in fact with modern compilers it changes nothing.

gstaichi/program/launch_context_builder.cpp
duburcqa: Honestly I have no idea. I'm just inlining and (weakly) batching the existing code, I have no idea what I'm doing.

gstaichi/python/export_lang.cpp
duburcqa: I just tried to be less invasive. It is necessary to do this because of the Python bindings overhead, so I thought it was not necessary to clutter the original C++ API.

I wanted to do the same for 'set_args_ndarray', but then I realised that it necessary to go one level deeper to unlock more performance (which is basically what this PR is doing in the first place), so I did because it also has value from a C++ perspective.

gstaichi/program/launch_context_builder.cpp
hughperkins: My uqesetion is: why add, not mutiply? what do the two terms represent?

gstaichi/program/launch_context_builder.cpp
hughperkins: further down it has `i++`. I think we should be consistent.

gstaichi/program/launch_context_builder.h
duburcqa: I think this function may be called on the C++ side, and I don't feel the urge to rush at removing the old one right away. This file would require way more optimisation anyway, it is poorly designed and the reason of 40% of the simulation overhead. But this is more involving, like 1 week of work. For I figured out that it may be nice to only do the minimal changes for now.

gstaichi/program/launch_context_builder.cpp
hughperkins: oh yeah, I know the experience, was hoping you had some secret insights that I could steal 😂

gstaichi/program/launch_context_builder.cpp
duburcqa: As for the rest, just a copy paste at this point.

gstaichi/python/export_lang.cpp
hughperkins: I'd prefer to keep the export_lang.cpp clean if we can. Since we are doing one level deeper for ndarray anyway, maybe we could go one level deeper for the others, for consistency? Do you see any peformance issues in going one level deeper for the others too?

gstaichi/program/launch_context_builder.cpp
duburcqa: Oh I see. You need to specify how much memory to expect to use at some point, so that it can be reserved  right away. Here, we are adding one item to each map per array being passed (`arg_ids.size()`), and we know how many items are currently stored in each of these maps (`array_ptrs.size()`). So the total memory we need is just the sum of both.

gstaichi/program/launch_context_builder.cpp
duburcqa: Well it is random, they chose something different for each for-loop in this file....

As you wish.

gstaichi/program/launch_context_builder.cpp
duburcqa: Rest assured soon I will, after I'm done refactoring this entire mess during my free time :p

gstaichi/python/export_lang.cpp
duburcqa: > Do you see any peformance issues in going one level deeper for the others too?

Not at all. Let's go!

gstaichi/program/launch_context_builder.h
hughperkins: I think that would show up in unit tests? Not keen on leaving un-used code around. Please could you try removing the potentially unused fucntions, and see what happens?

gstaichi/program/launch_context_builder.cpp
hughperkins: ah, got it. Makes sense. Thanks!

gstaichi/program/launch_context_builder.cpp
duburcqa: done.

hughperkins: I'm not seeing any docstring 🤔 Could you provide a link to the docstring you have added?

hughperkins: Oh you added int othe header perhaps?

YilingQiao (approved):

PR #254, opened 2025-10-29
[Perf] Diagnose launch context buffer cache-ability

duburcqa:


python/gstaichi/lang/kernel_impl.py
```diff
@@ -658,7 +658,7 @@ def _recursive_set_args(
     indices: tuple[int, ...],
     actual_argument_slot: int,
     callbacks: list[Callable[[], Any]],
-) -> int:
+) -> tuple[int, bool]:
     """
     Returns the number of kernel args set e.g. templates don't set kernel args, so returns 0
     a single ndarray is 1 kernel arg, so returns 1
```
YilingQiao: what does the additional bool variable mean? Can we add some documentation?

duburcqa: sure!

PR #255, opened 2025-10-29
[Perf] Add python caching of launch context buffer

duburcqa:
From 4.51M FPS to 8.47M FPS

image

### DIFF
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -653,7 +653,7 @@ class KernelBatchedArgType(IntEnum):
 _FLOAT, _INT, _UINT, _TI_ARRAY, _TI_ARRAY_WITH_GRAD = KernelBatchedArgType
 
 
-Hash: TypeAlias = int
+ArgsHash: TypeAlias = int
 
 
 def _destroy_callback(kernel_ref: ReferenceType["Kernel"], ref: ReferenceType):
@@ -928,8 +928,8 @@ class Kernel:
         # * '_prog_weakref'is used for bounding the lifetime of the entire cache to the Taichi programm managing all
         #   the launch context being stored in cache.
         # See 'launch_kernel' for details regarding the intended use of caching.
-        self._launch_ctx_cache: dict[Hash, KernelLaunchContext] = {}
-        self._launch_ctx_cache_tracker: dict[Hash, list[ReferenceType[Ndarray]]] = {}
+        self._launch_ctx_cache: dict[ArgsHash, KernelLaunchContext] = {}
+        self._launch_ctx_cache_tracker: dict[ArgsHash, list[ReferenceType[Ndarray]]] = {}
         self._prog_weakref: ReferenceType[Program] | None = None
 
     def ast_builder(self) -> ASTBuilder:


YilingQiao (approved):

gstaichi/program/launch_context_builder.h
```diff
@@ -25,6 +25,13 @@ class LaunchContextBuilder {
   LaunchContextBuilder(const LaunchContextBuilder &) = delete;
   LaunchContextBuilder &operator=(const LaunchContextBuilder &) = delete;
 
+  // Copy all the arguments already added to an existing launcher context.
```
hughperkins, line 28: Interesting

python/gstaichi/lang/kernel_impl.py
```diff
@@ -906,6 +916,10 @@ def __init__(self, _func: Callable, autodiff_mode: AutodiffMode, _classkernel=Fa
         self.src_ll_cache_observations: SrcLlCacheObservations = SrcLlCacheObservations()
         self.fe_ll_cache_observations: FeLlCacheObservations = FeLlCacheObservations()
 
+        self._launch_ctx_cache: dict[int, KernelLaunchContext] = {}
```
hughperkins: can we document what the int represents please?

python/gstaichi/lang/kernel_impl.py
```diff
@@ -906,6 +916,10 @@ def __init__(self, _func: Callable, autodiff_mode: AutodiffMode, _classkernel=Fa
...
+        self._launch_ctx_cache_tracker: dict[int, list[ReferenceType]] = {}
```
hughperkins: ditto. what does the int represent?

python/gstaichi/lang/kernel_impl.py
hughperkins: (if it's just a seuqence number, can we use a list instead?)

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1102,10 +1116,13 @@ def ast_to_dict(node: ast.AST | list | primitive_types._python_primitive_types):
     def launch_kernel(self, t_kernel: KernelCxx, compiled_kernel_data: CompiledKernelData | None, *args) -> Any:
         assert len(args) == len(self.arg_metas), f"{len(self.arg_metas)} arguments needed but {len(args)} provided"
 
-        actual_argument_slot = 0
-        launch_ctx = t_kernel.make_launch_context()
-        launch_ctx_buffer: DefaultDict[KernelBatchedArgType, list[tuple]] = defaultdict(list)
-        callbacks: list[Callable[[], None]] = []
+        # Keep track of taichi runtime to automatically clear cache if destroyed
+        if self._prog_weakref is None:
+            prog = impl.get_runtime().prog
```
hughperkins, line 1133: can you explain why we don't simply systematiclly call `prog = impl.get_runtime().prog` ?

python/gstaichi/lang/kernel_impl.py
duburcqa: It is the hash of the input arguments

hughperkins: please could we add that as a comment?

python/gstaichi/lang/kernel_impl.py
duburcqa: If we already have a weak ref available, it is much faster to use it rather than for through python bindings: 180ns -> 5ns.

python/gstaichi/lang/kernel_impl.py
duburcqa: Of course. Do you want a type alias for clarity?

python/gstaichi/lang/kernel_impl.py
duburcqa: Do you want a comment about this in the code?

python/gstaichi/lang/kernel_impl.py
hughperkins: No opinion either way; as long as it's clear what the int or type alias represents.

python/gstaichi/lang/kernel_impl.py
hughperkins: Yes please :)

python/gstaichi/lang/kernel_impl.py
hughperkins: can we state that the int for eac hof these is also the hash please? (or use a type alias, whichever you feel is easier)

hughperkins: (maybe a type alias is easier? 🤔 )

hughperkins: Also, what is the difference btween '_launch_ctx_cache' and 'launch_ctx_cache_tracker'? Can you add a comment on this, and on what the list[referencetype] continas please?

python/gstaichi/lang/kernel_impl.py
```diff
@@ -649,6 +652,17 @@ class KernelBatchedArgType(IntEnum):
 _FLOAT, _INT, _UINT, _TI_ARRAY, _TI_ARRAY_WITH_GRAD = KernelBatchedArgType
 
 
+Hash: TypeAlias = int
```
hughperkins: Thanks! Could we call this `ArgsHash` please, to make ti clear?

duburcqa: sure!

duburcqa: Good idea indeed.

python/gstaichi/lang/kernel_impl.py
duburcqa: done

PR #259, opened 2025-11-01
[Perf] Prune unused dataclass fields from being passed to kernels

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -669,11 +669,6 @@ class FeLlCacheObservations:
     cache_hit: bool = False
 
 
-@dataclass
-class LaunchStats:
-    kernel_args_count_by_type: dict[str, int]
-
-
 def cast_float(x: float | np.floating | np.integer | int) -> float:
     if not isinstance(x, (int, float, np.integer, np.floating)):
         raise ValueError(f"Invalid argument type '{type(x)}")
@@ -686,7 +681,7 @@ def cast_int(x: int | np.integer) -> int:
     return int(x)
 
 
-class KernelBatchedArgType(IntEnum):
+class _KernelBatchedArgType(IntEnum):
     FLOAT = 0
     INT = 1
     UINT = 2
@@ -695,10 +690,15 @@ class KernelBatchedArgType(IntEnum):
 
 
 # Define proxies for fast lookup
-_FLOAT, _INT, _UINT, _TI_ARRAY, _TI_ARRAY_WITH_GRAD = KernelBatchedArgType
+_FLOAT, _INT, _UINT, _TI_ARRAY, _TI_ARRAY_WITH_GRAD = _KernelBatchedArgType
 
 
 ArgsHash: TypeAlias = int
+
+
+@dataclass
+class LaunchStats:
+    kernel_args_count_by_type: dict[_KernelBatchedArgType, int]
 
 
 def _destroy_callback(kernel_ref: ReferenceType["Kernel"], ref: ReferenceType):
@@ -713,7 +713,7 @@ def _recursive_set_args(
     used_py_dataclass_parameters: set[tuple[str, ...]],
     py_dataclass_basename: tuple[str, ...],
     launch_ctx: KernelLaunchContext,
-    launch_ctx_buffer: DefaultDict[KernelBatchedArgType, list[tuple]],
+    launch_ctx_buffer: DefaultDict[_KernelBatchedArgType, list[tuple]],
     needed_arg_type: Type,
     provided_arg_type: Type,
     v: Any,
@@ -1266,7 +1266,7 @@ class Kernel:
         except (TypeError, KeyError):
             pass
         if not launch_ctx_cache_tracker:  # Empty or none
-            launch_ctx_buffer: DefaultDict[KernelBatchedArgType, list[tuple]] = defaultdict(list)
+            launch_ctx_buffer: DefaultDict[_KernelBatchedArgType, list[tuple]] = defaultdict(list)
             actual_argument_slot = 0
             is_launch_ctx_cacheable = True
             template_num = 0
@@ -1296,33 +1296,32 @@ class Kernel:
                 i_out += num_args_
                 is_launch_ctx_cacheable &= is_launch_ctx_cacheable_
 
+            kernel_args_count_by_type = defaultdict(int)
+            kernel_args_count_by_type.update(
+                {key: len(launch_ctx_args) for key, launch_ctx_args in launch_ctx_buffer.items()}
+            )
+            self.launch_stats = LaunchStats(kernel_args_count_by_type=kernel_args_count_by_type)
+
             # All arguments to context in batches to mitigate overhead of calling Python bindings repeatedly.
             # This is essential because calling any pybind11 function is adding ~180ns penalty no matter what.
             # Note that we are allowed to do this because GsTaichi Launch Kernel context is storing the input
             # arguments in an unordered list. The actual runtime (gfx, llvm...) will later query this context
             # in correct order.
-            kernel_args_count_by_type = defaultdict(int)
-            self.launch_stats = LaunchStats(kernel_args_count_by_type=kernel_args_count_by_type)
             if launch_ctx_args := launch_ctx_buffer.get(_FLOAT):
                 indices, vec = zip(*launch_ctx_args)
-                kernel_args_count_by_type["float"] = len(indices)
                 launch_ctx.set_args_float([index for index, in indices], vec)  # type: ignore
             if launch_ctx_args := launch_ctx_buffer.get(_INT):
                 indices, vec = zip(*launch_ctx_args)
                 launch_ctx.set_args_int([index for index, in indices], vec)  # type: ignore
-                kernel_args_count_by_type["int"] = len(indices)
             if launch_ctx_args := launch_ctx_buffer.get(_UINT):
                 indices, vec = zip(*launch_ctx_args)
                 launch_ctx.set_args_uint([index for index, in indices], vec)  # type: ignore
-                kernel_args_count_by_type["uint"] = len(indices)
             if launch_ctx_args := launch_ctx_buffer.get(_TI_ARRAY):
                 indices, arrs = zip(*launch_ctx_args)
                 launch_ctx.set_args_ndarray([index for index, in indices], arrs)  # type: ignore
-                kernel_args_count_by_type["ndarray"] = len(indices)
             if launch_ctx_args := launch_ctx_buffer.get(_TI_ARRAY_WITH_GRAD):
                 indices, arrs, arrs_grad = zip(*launch_ctx_args)
                 launch_ctx.set_args_ndarray_with_grad([index for index, in indices], arrs, arrs_grad)  # type: ignore
-                kernel_args_count_by_type["ndarray_with_grad"] = len(indices)
 
             if is_launch_ctx_cacheable and args_hash is not None:
                 # TODO: It some rare occurrences, arguments can be cached yet not hashable. Ignoring for now...
@@ -1726,4 +1725,4 @@ def data_oriented(cls):
     return cls
 
 
-__all__ = ["data_oriented", "func", "kernel", "pyfunc", "real_func"]
+__all__ = ["data_oriented", "func", "kernel", "pyfunc", "real_func", "_KernelBatchedArgType"]
--- a/tests/python/test_py_dataclass.py
+++ b/tests/python/test_py_dataclass.py
@@ -7,6 +7,7 @@ from typing import Any
 import pytest
 
 import gstaichi as ti
+from gstaichi.lang import _KernelBatchedArgType
 
 from tests import test_utils
 
@@ -958,8 +959,8 @@ def test_print_used_leaves():
     assert u1[1] == 333
     assert u1[2] == 0
     kernel_args_count_by_type = k1._primal.launch_stats.kernel_args_count_by_type
-    assert kernel_args_count_by_type["ndarray"] == 3
-    assert kernel_args_count_by_type["int"] == 1
+    assert kernel_args_count_by_type[_KernelBatchedArgType.TI_ARRAY] == 3
+    assert kernel_args_count_by_type[_KernelBatchedArgType.INT] == 1
 
     u1[0] = 0
     u1[1] = 0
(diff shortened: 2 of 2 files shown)


python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -57,37 +60,45 @@ def store(cache_key: str, function_source_infos: Iterable[FunctionSourceInfo]) -
...
-def _try_load(cache_key: str) -> Sequence[HashedFunctionSourceInfo] | None:
+def _try_load(cache_key: str) -> tuple[Sequence[HashedFunctionSourceInfo], set[str]] | tuple[None, None]:
     cache = PythonSideCache()
     maybe_cache_value_json = cache.try_load(cache_key)
     if maybe_cache_value_json is None:
-        return None
+        return None, None
     try:
-        cache_value_obj = CacheValue.parse_raw(maybe_cache_value_json)
+        cache_value_obj = CacheValue.model_validate_json(maybe_cache_value_json)
```
hughperkins, line 84: note: finally got rid of those pydantic warnings! (maybe 🙏 )

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -57,37 +60,45 @@ def store(cache_key: str, function_source_infos: Iterable[FunctionSourceInfo]) -
...
     """
     loads function source infos from cache, if available
     checks the hashes against the current source code
     """
-    maybe_hashed_function_source_infos = _try_load(cache_key)
-    if not maybe_hashed_function_source_infos:
-        return False
-    return function_hasher.validate_hashed_function_infos(maybe_hashed_function_source_infos)
+    maybe_hashed_function_source_infos, used_py_dataclass_leaves = _try_load(cache_key)
+    if maybe_hashed_function_source_infos is None or used_py_dataclass_leaves is None:
```
duburcqa, line 97: Either your typing is wrong or the second clause of this condition is redundant.

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -87,6 +89,8 @@ def build_Name(ctx: ASTTransformerContext, node: ast.Name):
                     warnings.warn(message)
                 else:
                     raise exception.GsTaichiCompilationError(message)
+        if isinstance(node.ptr, Generator):
+            raise ValueError("Cannot use generator inside kernels or functions")
```
duburcqa: This is misleading. It is not that generators cannot be used inside kernels or functions, but rather cannot be stored in variables, which is very different and much less problematic.

python/gstaichi/lang/ast/ast_transformer_utils.py
```diff
@@ -195,6 +195,10 @@ def __init__(
         is_real_function: bool,
         autodiff_mode: AutodiffMode,
         raise_on_templated_floats: bool,
+        # during 1st pass, we collect the names of used parameters
+        used_py_dataclass_parameters_collecting: set[str],
+        # during 2nd pass, we only handle these names
+        used_py_dataclass_parameters_enforcing: set[str] | None,
```
duburcqa, line 201: Could you elaborate why distinguishing parameters during collection and enforcing phases is never ? The same buffer cannot be used for both? since I have the feeling these 2 phases come one after the other and never simultaneously?

python/gstaichi/lang/kernel_impl.py
```diff
@@ -353,24 +355,41 @@ def _get_tree_and_ctx(
...
+        used_py_dataclass_parameters_enforcing=used_py_dataclass_parameters_enforcing,
     )
+    return tree, ctx
+
+
+def _process_args(
+    self: "Func | Kernel", is_pyfunc: bool, is_func: bool, args: tuple[Any, ...], kwargs
+) -> tuple[Any, ...]:
+    if is_func and not is_pyfunc:
+        assert isinstance(self, Func)
```
duburcqa: Is this performance critical? If so, you should use `type(self) is Func` instead.

python/gstaichi/lang/_fast_caching/src_hasher.py
hughperkins: The current typing doesn't couple the two Nones together, though perhaps that is possible 🤔  Currently I have something like `tuple[Blah | None, Foo | None]`. But I suppose you are saying I could have osmthing like `tuple[Blah, Foo] | tuple[None, None]` ?

python/gstaichi/lang/kernel_impl.py
```diff
@@ -721,11 +765,17 @@ def _recursive_set_args(
         for field in needed_arg_fields.values():
             if field._field_type is not _FIELD:
                 continue
+            field_name = field.name
+            field_full_name = py_dataclass_basename + (field_name,)
```
duburcqa, line 772: I'm wondering whether this is faster than `field_full_name = (*py_dataclass_basename, field_name)`. I would expect the later to be more efficient because one less tuple must be allocated and freed.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1198,21 +1298,28 @@ def launch_kernel(self, t_kernel: KernelCxx, compiled_kernel_data: CompiledKerne
...
                 launch_ctx.set_args_uint([index for index, in indices], vec)  # type: ignore
+                kernel_args_count_by_type["uint"] = len(indices)
             if launch_ctx_args := launch_ctx_buffer.get(_TI_ARRAY):
                 indices, arrs = zip(*launch_ctx_args)
                 launch_ctx.set_args_ndarray([index for index, in indices], arrs)  # type: ignore
+                kernel_args_count_by_type["ndarray"] = len(indices)
             if launch_ctx_args := launch_ctx_buffer.get(_TI_ARRAY_WITH_GRAD):
                 indices, arrs, arrs_grad = zip(*launch_ctx_args)
                 launch_ctx.set_args_ndarray_with_grad([index for index, in indices], arrs, arrs_grad)  # type: ignore
+                kernel_args_count_by_type["ndarray_with_grad"] = len(indices)
```
duburcqa: What about using our `KernelBatchedArgType` as key and just do:
```python
self.launch_stats = LaunchStats(kernel_args_count_by_type={
    key: len(launch_ctx_args) for key, launch_ctx_args in launch_ctx_buffer.items()
})
```

duburcqa: Moreover, keep in mind that `launch_ctx_buffer` does not track all possible types of input arguments. Only the one that are considered "cache-friendly", i.e. what we are using in simulation hot-path.

python/gstaichi/lang/_fast_caching/src_hasher.py
duburcqa: It is just that you typed the output of `_try_load` as `tuple[Blah, Foo] | tuple[None, None]` instead of `tuple[Blah | None, Foo | None]`.

python/gstaichi/lang/kernel_impl.py
hughperkins: > Moreover, keep in mind that `launch_ctx_buffer` does not track all possible types of input arguments. Only the one that are considered "cache-friendly", i.e. what we are using in simulation hot-path.

Yeah. it's good enough for unit testing though, which is the purpose of these stats.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -917,6 +968,12 @@ def __init__(self, _func: Callable, autodiff_mode: AutodiffMode, _classkernel=Fa
         self.kernel_function_info: FunctionSourceInfo | None = None
         self.compiled_kernel_data_by_key: dict[CompiledKernelKeyType, CompiledKernelData] = {}
         self._last_compiled_kernel_data: CompiledKernelData | None = None  # for dev/debug
+        # for collecting, we'll grab an empty set if it doesnt exist
+        self.used_py_dataclass_leaves_by_key_collecting: dict[CompiledKernelKeyType, set[str]] = defaultdict(set)
+        # however, for enforcing, we want None if it doesn't exist (we'll use .get() instead of [] )
+        self.used_py_dataclass_leaves_by_key_enforcing: dict[CompiledKernelKeyType, set[str]] = {}
+        self.used_py_dataclass_leaves_by_key_enforcing_dotted: dict[CompiledKernelKeyType, set[tuple[str, ...]]] = {}
```
duburcqa, line 978: I'm not sure `_by_key` add any value and the name is already very long. As you want.

duburcqa, line 978: Why is it necessary to track both `_enforcing` and `_enforcing_dotted` separately? It seems redundant.

python/gstaichi/lang/_fast_caching/src_hasher.py
hughperkins: No matter how I type the return type, checking the first type doesnt appear to change the inferred type of the second type?

Screenshot 2025-11-10 at 09 52 47
Screenshot 2025-11-10 at 09 52 51
Screenshot 2025-11-10 at 09 52 55
Screenshot 2025-11-10 at 09 52 59

hughperkins: (though perhaps this is a bug in vscode pylance, and doesnt generalize to pyright itself? 🤔 )

hughperkins: Well, pyright seems to complain smilarly for both? :
Screenshot 2025-11-10 at 09 55 12
Screenshot 2025-11-10 at 09 55 14

python/gstaichi/lang/ast/ast_transformer_utils.py
hughperkins: The high level reason is that trying to debug with a single buffer is really hard.

But more concretely:
- used_py_dataclass_parameters_collecting is conceptually mutable, but never None
- used_py_dataclass_parameters_enforcing is never mutable, but can be None

hughperkins: (basically, there were tons of bugs where I was mutating the enforcing buffer by accident, or using the collecting buffer for enforcement etc, when I had only a single buffer; was just super confusing I felt).

python/gstaichi/lang/kernel_impl.py
hughperkins: I'm fairly sure I tested, and this was faster. (I initially tried (*py_dataclass_basename, field_name) ). NOt sure if I documented hti though 🤔

hughperkins: I switched in 538ca10332cca4763 , but I cant find any doc/measurements to back that up. I guess I could remeasure.

python/gstaichi/lang/kernel_impl.py
hughperkins: it added value to me, because otherwise it confuses with the non by_key version.

hughperkins: for performance.

Originally I just had one, but then tons of time went into running the transforms from one to the other. So we just do it once, and it's much faster.

hughperkins: I made this optimization in c77d587773b3

python/gstaichi/lang/ast/ast_transformer.py
hughperkins: Good point 👍

python/gstaichi/lang/kernel_impl.py
hughperkins: its just for typing to be honest

hughperkins: updated to use `if typing.TYPE_CHECKING:`

python/gstaichi/lang/kernel_impl.py
hughperkins: So... I tried remeasuring, and there was zero difference. However if I try timeit:

```
import timeit

print(timeit.timeit("a + (6,)", setup="a = tuple(list(range(10)))"))
print(timeit.timeit("(*a, 6)", setup="a = tuple(list(range(10)))"))
```
(gstaichi) ~/git/gstaichi (hp/prune-unused-dataclass-fields-adding-fastcache|…3△2) $ python /tmp/test5.py
0.026368291990365833
0.043114166997838765
(gstaichi) ~/git/gstaichi (hp/prune-unused-dataclass-fields-adding-fastcache|…3△2) $ python /tmp/test5.py
0.023982166952919215
0.0423166659893468
(gstaichi) ~/git/gstaichi (hp/prune-unused-dataclass-fields-adding-fastcache|…3△2) $ python /tmp/test5.py
0.026562208018731326
0.0428738750051707

python/gstaichi/lang/kernel_impl.py
hughperkins: Updated to populate the dict direclty from launch_ctx_buffer.items()

python/gstaichi/lang/_fast_caching/src_hasher.py
duburcqa: If you type it `tuple[Blah, Foo] | tuple[None, None]`, checking the first return value should be sufficient. If it is not, then it is a bug of pylance. But not a big deal.

hughperkins: well, pyright has the same issue, see above?

duburcqa: Try replacing `Blah` and `Foo` by `int` in your example

python/gstaichi/lang/ast/ast_transformer_utils.py
duburcqa: Ok I'm fine with this.

python/gstaichi/lang/kernel_impl.py
duburcqa: Ok, use whether to prefer then :)

python/gstaichi/lang/kernel_impl.py
duburcqa: ok.

duburcqa: I mean, it is not possible to compute `self.used_py_dataclass_leaves_by_key_enforcing_dotted` straight away, you need to compute `self.used_py_dataclass_leaves_by_key_enforcing` first ? Or you may even use both at runtime depending on the situation?

duburcqa (approved):

python/gstaichi/lang/_fast_caching/src_hasher.py
hughperkins: with ints and pyright:

Screenshot 2025-11-10 at 13 10 48
Screenshot 2025-11-10 at 13 10 51

python/gstaichi/lang/kernel_impl.py
hughperkins: they are both used
- in some cases we compare with `__ti_some_param__ti_some_attribute...`, in some cases we compare with `(some_param, some_attribute, ...)`

python/gstaichi/lang/_fast_caching/src_hasher.py
duburcqa: Weird. OK!

python/gstaichi/lang/kernel_impl.py
duburcqa: Ok I see!

PR #260, opened 2025-11-02
[Perf] Store template mapper key at instance level

duburcqa:
This is necessary because a single class may be associated with instances using different memory layout (although this is fairly unusual).

### DIFF
--- a/tests/python/test_py_dataclass.py
+++ b/tests/python/test_py_dataclass.py
@@ -866,8 +866,9 @@ def test_ndarray_struct_multiple_child_structs_field():
     assert f[0] == 77
 
 
+@pytest.mark.parametrize("use_slots", [False, True])
 @test_utils.test()
-def test_template_mapper_cache(monkeypatch):
+def test_template_mapper_cache(use_slots, monkeypatch):
     # Mock '_extract_arg' to track the number of (recursive) calls
     counter = 0
     _extract_arg_orig = ti.lang._template_mapper_hotpath._extract_arg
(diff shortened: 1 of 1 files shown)


hughperkins:
Please could you add a test, taht fails without your prposed fix.

tests/python/test_py_dataclass.py
```diff
@@ -864,3 +864,32 @@ def k1(p1: P1) -> None:
...
+    value = ti.ndarray(ti.i32, shape=(1,))
+    value.fill(0)
+    my_struct_1d = MyStruct(value=value)
+    value = ti.ndarray(ti.f32, shape=(1, 2))
+    value.fill(0.0)
+    my_struct_2d = MyStruct(value=value)
+
+    my_kernel(my_struct_1d, my_struct_2d)
+    assert my_struct_1d.value[0] == 1
+    assert abs(my_struct_2d.value[0, 0] - 1.0) < 1e-9
```
hughperkins: Can we define EPS, rather than writing 1e-9 each time please?

Also, looks like python provides `maths.isclose`, https://docs.python.org/3/library/math.html#math.isclose Thoughts on our using that?

tests/python/test_py_dataclass.py
```diff
@@ -864,3 +864,32 @@ def k1(p1: P1) -> None:
     assert d[0] == 55
     assert e[0] == 66
     assert f[0] == 77
+
+
+@test_utils.test()
+def test_template_mapper_cache():
+    gc.collect()
```
hughperkins: why are we running gc.collect() before the test? I dont think this should be ncessary any more?

tests/python/test_py_dataclass.py
hughperkins: actually, is there any reason wy it wouldnt be perfectly equal? We arent doing any calculations, and 1 as a float should be exact I think?

python/gstaichi/lang/_template_mapper_hotpath.py
```diff
@@ -193,7 +197,11 @@ def _extract_arg(raise_on_templated_floats: bool, arg: Any, annotation: Annotati
             ]
         )
         if is_frozen:
-            annotation._key = key
+            try:
+                object.__setattr__(arg, "_key", key)
+            except AttributeError:
```
hughperkins, line 202: so what should happen in this case? will this cause an error? if so, should we be swallowing the error like this? If not, can we add a test for this please?

hughperkins: oh i see, this is exactly the case that the new test tests for?

hughperkins: Hmmm, I'm not sure. I think we are testing with `frozen=True`, but not with `slots=True`, is this a fair impression?

tests/python/test_py_dataclass.py
duburcqa: No tolerance is needed indeed. I will remove it if you think it would be better.

tests/python/test_py_dataclass.py
duburcqa: I don't know, I took inspiration from the other tests. I will remove it.

python/gstaichi/lang/_template_mapper_hotpath.py
duburcqa: There is a comment to explain what will happen in the except close. To put it simply: caching is disabled for such dataclass.

I will add a unit test for it.

tests/python/test_py_dataclass.py
hughperkins: Yes, they predate our adding ndarray registration that means gc is no longer needed.

tests/python/test_py_dataclass.py
```diff
@@ -864,3 +864,50 @@ def k1(p1: P1) -> None:
...
+        for i in ti.ndrange(my_struct.value.shape[0]):
+            my_struct.value[i] += 1
+
+    value = ti.ndarray(ti.i32, shape=(1,))
+    value.fill(0)
+    my_struct = MyStruct(value=value)
+
+    my_kernel(my_struct)
+    assert my_struct.value[0] == 1
+    my_kernel(my_struct)
```
hughperkins: Don't we need to test what happens if we create a brand new instance of MyStruct, as we did in the first test? What will happen? Should that throw an exception?

hughperkins: (I think I need to add Sonarqube, to check coverage, at some point :) )

duburcqa: > Don't we need to test what happens if we create a brand new instance of MyStruct, as we did in the first test? What will happen? Should that throw an exception?

No because there is simply no caching for dataclasses that uses slots. But I can add it for completeness if you want? You just check more explicitly that caching is ignored?

hughperkins: not necessarily asking to check explicitly that caching is ignored, just for correctness.

duburcqa: I'm now explicitly checking that caching is ignored.

hughperkins: why is it in the first test, you create two different instances of MyStruct, with different value types, but in this test you are only creating a single instance of MyStruct?

duburcqa: Because it is checking something completely different. I'm checking that dataclasses that enable slots are ignored by "caching" mechanism. Without caching, we are back to the "usual" dataclass, which is already thoroughly tested.

hughperkins: Could you simply add a check about what actually happens with two such instances please? Involves fewer deductive steps, so safer/easier to understand.

duburcqa: Sure !

duburcqa: I unified tests for slots=True and False.

YilingQiao (approved):

PR #261, opened 2025-11-04
[Build] Add build for ARM

hughperkins:
Issue: #

### Brief Summary

- based on @johnnynunez's branch/PR at https://github.com/Genesis-Embodied-AI/gstaichi/pull/227
- merged from main
- turned `assert`, inside kernels, into a `nop` for linux arm (otherwise it crashes)
- skipped any tests involving testing out of bounds in debug mode, which fail on linux arm
(basically, both of these things (assert => nop, oob testing not working in debug) are 'nice to have' for development, but aren't critically necessary in order to run things on linux arm).

copilot:summary

### Walkthrough

copilot:walkthrough


.github/workflows/scripts/ti_build/entry.py
```diff
@@ -26,8 +26,11 @@ def build_wheel(python: Command, pip: Command) -> None:
...
 
     cmake_args.writeback()
-    if platform.system() == "Linux":
+    u = platform.uname()
+    if (u.system, u.machine) == ("Linux", "x86_64"):
         extra.extend(["-p", "manylinux_2_27_x86_64"])
+    elif (u.system, u.machine) in (("Linux", "arm64"), ("Linux", "aarch64")):
+        extra.extend(["-p", "manylinux_2_27_aarch64"])
     if platform.system() == "Darwin":
         extra.extend(["-p", "macosx-11.0-arm64"])
```
duburcqa: Sounds like the perfect spot to use `match: case` pattern if you wanted:

```python
u = platform.uname()
match (u.system, u.machine):
    case ("Linux", "x86_64"):
        extra.extend(["-p", "manylinux_2_27_x86_64"])
    case ("Linux", "arm64") | ("Linux", "aarch64"):
        extra.extend(["-p", "manylinux_2_27_aarch64"])
    case ("Darwin", _):
        extra.extend(["-p", "macosx-11.0-arm64"])
```

.github/workflows/scripts/ti_build/llvm.py
```diff
@@ -28,14 +28,18 @@ def setup_llvm() -> str:
         platform="{platform}",
     )
 
-    if u.system == "Linux":
+    if (u.system, u.machine) == ("Linux", "x86_64"):
```
duburcqa: Same. Could you match case.

.github/workflows/scripts/ti_build/vulkan.py
```diff
@@ -17,12 +17,22 @@
 @banner(f"Setup Vulkan {VULKAN_VERSION}")
 def setup_vulkan():
     u = platform.uname()
-    if u.system == "Linux":
+    if (u.system, u.machine) == ("Linux", "x86_64"):
```
duburcqa: Same.

.github/workflows/scripts_new/linux/1_prerequisites.sh
```diff
@@ -16,6 +16,9 @@ chmod +x ${LLVM_DIR}/bin/*
 clang --version
 which clang
 
+python -c 'import sys; print("sys.platform", sys.platform)'
+python -c 'import os; print("os.uname()[4]", os.uname()[4])'
+
```
duburcqa: I'm not convinced python is the best tool to print system information. You are in bash, you have more appropriate tools for this. But I'm fine with it. Still, I would at least replace `[4]` by its corresponding name.

.github/workflows/scripts_new/manylinux_wheel/2_build.sh
```diff
@@ -2,11 +2,33 @@
...
 
+PLATFORM=$(uname -m)
+if [ "$PLATFORM" = "x86_64" ]; then
+    PLATFORM="x86_64"
+elif [ "$PLATFORM" = "aarch64" ]; then
+    PLATFORM="aarch64"
+else
+    echo "Unsupported architecture: $PLATFORM"
+    exit 1
+fi
```
duburcqa, line 9:
```bash
PLATFORM=$(uname -m)
if [ "$PLATFORM" != "x86_64" ] && [ "$PLATFORM" != "aarch64" ]; then
    echo "Unsupported architecture: $PLATFORM"
    exit 1
fi
```

.github/workflows/linux.yml
duburcqa, line 1: I would remove all modifications in this file if they are not needed.

Otherwise, you should finish the work and modify the job name to include the current architecture being tested, by at least adding `${{ matrix.OS }}` in name. I think it is important to keep track of this information at first glance.

cmake/GsTaichiCXXFlags.cmake
```diff
@@ -138,6 +138,11 @@ else()
 endif()
 set(HOST_ARCH ${ARCH} CACHE INTERNAL "Host arch")
 
+if (LINUX AND (CMAKE_SYSTEM_PROCESSOR STREQUAL "aarch64" OR CMAKE_SYSTEM_PROCESSOR STREQUAL "arm64"))
```
duburcqa: I'm not convinced that you need to enable atomic for `"arm64"` (i.e. macos). It was not the case before, why would it be the case now.

gstaichi/runtime/cpu/jit_cpu.cpp
```diff
@@ -131,10 +131,21 @@ class JITSessionCPU : public JITSession {
...
+#if !(defined(__APPLE__) && defined(__aarch64__))
+    // On ELF-based targets LLVM's ORC JIT expects the object layer to claim
+    // responsibility for every symbol emitted by the module. Without this the
+    // resolver may encounter symbols that were not pre-registered in the
+    // materialization responsibility set and abort with
+    // "Resolving symbol outside this responsibility set" (observed on
+    // manylinux aarch64).
+    object_layer_.setOverrideObjectFlagsWithResponsibilityFlags(true);
+    object_layer_.setAutoClaimResponsibilityForObjectSymbols(true);
+#else
```
duburcqa: I still don't understand this part. I would avoid altering the current behaviour as much as possible. So the condition should be at best `#if defined(__aarch64__) && !defined(__APPLE__)`. Moreover, Apple using `__arm64__`, not `__aarch64__`, so in the end it should just be: `#if defined(__aarch64__):`

Actually, the condition `defined(__APPLE__) && defined(__aarch64__)` may be possible on manylinux ARM container running on MacOS host, eventually, but I did not checked that. I don't really think it makes any sense still.

gstaichi/runtime/llvm/runtime_module/runtime.cpp
```diff
@@ -54,6 +54,69 @@ __asm__(".symver powf,powf@GLIBC_2.2.5");
 __asm__(".symver expf,expf@GLIBC_2.2.5");
 #endif
 
+#if (defined(__aarch64__) || defined(__arm64__)) && defined(__clang__) && \
+    !defined(__APPLE__)
```
duburcqa: This condition seems overcomplicated. If it is not compatible with Apple, why check `__arm64__` in the first place?

```cpp
#if defined(__aarch64__) && defined(__clang__)
```

python/gstaichi/lang/ast/ast_transformer.py
```diff
@@ -1332,6 +1334,10 @@ def ti_format_list_to_assert_msg(raw) -> tuple[str, list]:
 
     @staticmethod
     def build_Assert(ctx: ASTTransformerContext, node: ast.Assert) -> None:
+        if sys.platform == "linux" and os.uname()[4] in ["arm64", "aarch64"]:
```
duburcqa: I would recommend using `platform.machine()` or at least:
```python
    u = platform.uname()
    if u.system == "linux" and u.machine in ("arm64", "aarch64"):
        [...]
```

Besides, ideally use immutable sequences when applicable: `("arm64", "aarch64")`

tests/python/test_assert.py
```diff
@@ -1,10 +1,16 @@
...
+import sys
+
 import pytest
 
 import gstaichi as ti
 from gstaichi.lang.misc import get_host_arch_list
 
 from tests import test_utils
 
+if sys.platform == "linux" and os.uname()[4] in ["arm64", "aarch64"]:
```
duburcqa: same.

tests/python/test_assert_skip.py
```diff
@@ -0,0 +1,37 @@
+import os
+import sys
+
+import pytest
+
+import gstaichi as ti
+
+from tests import test_utils
+
+if sys.platform != "linux" or os.uname()[4] not in ["arm64", "aarch64"]:
```
duburcqa: same.

tests/python/test_ast_refactor.py
```diff
@@ -794,6 +797,9 @@ def foo(x: tc.template()) -> tc.i32:
 
 @test_utils.test(require=ti.extension.assertion, debug=True, gdb_trigger=False)
 def test_assert_message():
+    if sys.platform == "linux" and os.uname()[4] in ["arm64", "aarch64"]:
+        pytest.skip("assert not currently supported on linux arm64 or aarch64")
```
duburcqa: I would suggest adding a helper to detect whether the current arch is ARM on linux (local to unit test).

Moreover, you should prefer static `pytest.mark.skip` over runtime `pytest.skip` everytime the condition can be resolved statically, which is almost always the case, here included.

.github/workflows/scripts_new/manylinux_wheel/2_build.sh
hughperkins: already updated thi sI think?

.github/workflows/scripts_new/linux/1_prerequisites.sh
hughperkins: I'm using this to diagnose these exact same statements that are used in test conditionals 45 minutes later.

cmake/GsTaichiCXXFlags.cmake
hughperkins: this is "if linux and (aarch64 or arm)"? Am I misreading?

gstaichi/runtime/cpu/jit_cpu.cpp
hughperkins: should probably be, conceptually "if linux and (aarch64 or arm64)" I think?

gstaichi/runtime/llvm/runtime_module/runtime.cpp
hughperkins: yes, should proablby check for linux, rather than not apple, conceptaully?

hughperkins: actually, we arent using clang any more 🤔

hughperkins: might try deleting 🤔

tests/python/test_ast_refactor.py
hughperkins: Creating a helper just to compare platform.machine() with arm64 and aarch64 seems like over-engineerin I feel?

hughperkins: migrated to pytest.mark.skipif

hughperkins:
everything failed. going to revert the changes to cpp conditionals

build.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
```
hughperkins, line 1: this change is because, when running in a manylinux container, `python3` picks up some random python 3.9 installation, rather than the intended python 3.10 or so python.

gstaichi/runtime/llvm/runtime_module/runtime.cpp
hughperkins: oh this is the runtime module. We are probably using clang for this. almost certainly.

.github/workflows/scripts_new/manylinux_wheel/2_build.sh
duburcqa: Yeah, you were faster than me.

hughperkins: :)

cmake/GsTaichiCXXFlags.cmake
duburcqa: You are right! I was confused because `LINUX AND arm64` is impossible. Any ARM architecture on Linux is reported as aarch64, even if it is an Apple M processor.

image

gstaichi/runtime/cpu/jit_cpu.cpp
duburcqa: Still I think it should just be `if linux and aarch64`, arm64 is MacOS specific, not Apple ARM-based CPU arch.

gstaichi/runtime/llvm/runtime_module/runtime.cpp
duburcqa: Still `__arm64__` is impossible on `__linux__`

build.py
duburcqa: Anyway, we should move on a stop using python3 instead of Python. Python 2 is long dead. This habit is why 
Guido van Rossum do not want to release Python 4 ever x)

gstaichi/runtime/cpu/jit_cpu.cpp
hughperkins: So I googled around, and whilst it's true that aarch64 is what is used on linux, it looks like linux is the exception, rather than mac is the eceptoin? https://www.reddit.com/r/asm/comments/10chhth/comment/lreujfs/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

Screenshot 2025-11-05 at 17 08 47

(Eg windows uses 'arm64' name?)

build.py
hughperkins: :)

.github/workflows/scripts_new/linux/1_prerequisites.sh
duburcqa: Sure, I get it, but then, what about:
```bash
python -c 'import platform; print("\n".join(f"{k.capitalize():<9}: {v}" for k, v in platform.uname()._asdict().items()))'
```

It gives:
```
System   : Darwin
Node     : Mac.lan
Release  : 25.0.0
Version  : Darwin Kernel Version 25.0.0: Wed Sep 17 21:41:26 PDT 2025; root:xnu-12377.1.9~141/RELEASE_ARM64_T6041
Machine  : arm64
Processor: arm
```

hughperkins: I didn't need that level of detail.  I just needed to confirm uname()[4] and platform. Giving more detail meant there was *more* room for error, since I still had to count to 4 and stuff....

hughperkins: (I'm fine with removing it now, if you prefer?)

duburcqa: I just figured that it was easier to print all of them than filter out what you don't care about. The point is ideally you should use the same source of truth everywhere, which is now `uname` + `.system`, `.machine` from what I see. But you can keep them as 2 separated python calls if you prefer.

gstaichi/runtime/cpu/jit_cpu.cpp
hughperkins: (I think it's already "if linux and aarch64" now?)

.github/workflows/scripts_new/linux/1_prerequisites.sh
hughperkins: Migrated to print .system and .machine.

YilingQiao (approved):

PR #262, opened 2025-11-09
[Pref] Various performance improvements.

duburcqa:
* Add template mapper arg caching.
* Avoid dynamic cast.
* Replace dynamic-size std::vector by fixed-size std::array.


cmake/GsTaichiCXXFlags.cmake
duburcqa, line 1: `haswell` is a more sensible baseline (2013). I have been using it for years without any issue. It corresponds more or less to `x86-64-v3`.

PR #263, opened 2025-11-10
[Perf] Avoid dynamic cast if possible.

duburcqa:
```bash
```

| PR (branch) | runtime FPS |
| -------- | ------- |
| **+ #263 (duburcqa/avoid_dyn_cast)** | **1200FPS**  |
| + #264 (duburcqa/template_mapper_arg_cache) | 2470FPS  |
| + #265 (duburcqa/update_x86_arch) | 2520FPS  |

### DIFF
--- a/gstaichi/ir/type.h
+++ b/gstaichi/ir/type.h
@@ -38,9 +38,24 @@ class TI_DLL_EXPORT Type {
 
   template <typename T>
   const T *cast() const {
+    // Note that `Type` is a concrete type, not to be confused with placeholder
+    // `x##Type` that we be replaced by the preprocessor. More specifically,
+    // `Type` is the base type of all the derived types that are listed in
+    // "gstaichi/inc/type_kind.inc.h", e.g. `PrimitiveType`, `PointerType`,
+    // `StructType` or `TensorType`. All these derived types can be implicitly
+    // upcasted in their base type `Type`, so calling this method is
+    // superfluous. Still, this method is supposed to mimic the behavior of
+    // `dynamic_cast`, so this special case must be handled separately.
     if constexpr (std::is_same_v<typename std::remove_cv<T>::type, Type>) {
       return this;
     }
+
+    // This switch-case mechanism determines at runtime the correct branch based
+    // on the true derived type of the current instance. On its side, the inner
+    // branch is statically resolved at compile-time to either return downcasted
+    // `this` if the branch corresponds to requested type specified by the user,
+    // nullptr otherwise. This allows for mimicking the behavior of
+    // `dynamic_cast` without paying the cost and lack of portability of RTTI.
     switch (static_cast<int>(type_kind)) {
 #define PER_TYPE_KIND(x)                                                       \
   case static_cast<int>(TypeKind::x):                                          \


gstaichi/runtime/cuda/kernel_launcher.cpp
```diff
@@ -32,17 +32,17 @@ void KernelLauncher::launch_llvm_kernel(Handle handle,
...
       transfers;
 
   // |device_ptrs| stores pointers on device for all arrays args, including
   // external arrays and ndarrays, no matter whether the data is originally on
   // device or host.
   // This is the source of truth for us to look for device pointers used in CUDA
   // kernels.
-  std::unordered_map<std::vector<int>, void *,
-                     hashing::Hasher<std::vector<int>>>
+  std::unordered_map<std::pair<int, int>, void *,
```
hughperkins: could you add a comment stating what is the key of this map please?

gstaichi/runtime/cuda/kernel_launcher.cpp
```diff
@@ -32,17 +32,17 @@ void KernelLauncher::launch_llvm_kernel(Handle handle,
   // for data_ptr and TypeFactory::GRAD_PTR_POS_IN_NDARRAY for grad_ptr. Value
   // is [host_ptr, temporary_device_alloc]. Invariant: temp_devallocs.size() !=
   // 0 <==> transfer happened.
-  std::unordered_map<std::vector<int>, std::pair<void *, DeviceAllocation>,
-                     hashing::Hasher<std::vector<int>>>
+  std::unordered_map<std::pair<int, int>, std::pair<void *, DeviceAllocation>,
```
hughperkins: could you add a comment stating what is the key of this map please?

gstaichi/runtime/cuda/kernel_launcher.cpp
duburcqa: This was not supposed to be part of this PR. Fixed.

gstaichi/runtime/cuda/kernel_launcher.cpp
duburcqa: This was not supposed to be part of this PR. Fixed.

gstaichi/ir/type.cpp
```diff
@@ -157,16 +157,19 @@ std::string StructType::to_string() const {
...
 
 size_t StructType::get_element_offset(const std::vector<int> &indices) const {
-  const Type *type_now = this;
-  size_t offset = 0;
-  for (auto ind : indices) {
-    if (auto tensor_type = type_now->cast<TensorType>()) {
-      TI_ASSERT(ind < tensor_type->get_num_elements())
+  int ind = indices[0];
+  size_t offset = elements_[ind].offset;
+  const Type *type_now = elements_[ind].type;
```
hughperkins, line 162: Is this function basically doing the same as before, but switching the order of testing of StructType vs TensorType?

duburcqa: Yes, because it is more often a `StructType` than a `TensorType` in practice. Minor improvement.

gstaichi/ir/type.h
```diff
@@ -34,25 +38,55 @@ class TI_DLL_EXPORT Type {
 
   template <typename T>
   const T *cast() const {
-    return dynamic_cast<const T *>(this);
+    if constexpr (std::is_same_v<typename std::remove_cv<T>::type, Type>) {
+      return this;
+    }
+    switch (static_cast<int>(type_kind)) {
+#define PER_TYPE_KIND(x)                                                       \
```
hughperkins, line 60: coudl you give an example of when the first if would be true; and an example of where the first if would be false, and the if inside the switch would evaluate to true?

duburcqa: > coudl you give an example of when the first if would be true; and an example of where the first if would be false, and the if inside the switch would evaluate to true?

I don't understand. You mean an example in comment or in the codebase ?

hughperkins: I'm not sure. Maybe we could start in this PR thread.

duburcqa:
```cpp
if constexpr (std::is_same_v<typename std::remove_cv<T>::type, Type>) {
    return this;
}
```

This condition is true when casting to common base type `Type`. In this case, there is nothing to do since implicit upcasting is allowed.

-------------------------

```cpp
switch (static_cast<int>(type_kind)) {
    case static_cast<int>(TypeKind::x):
}
```

This case is selected if the type kind of the derived type is `TypeKind::x`, where `x` is the argument of the macro. The type kind is stored in `type_kind`. Note that casting to int is necessary to apply switch-case on c-style enums.

-------------------------

```cpp
if constexpr (std::is_same_v<typename std::remove_cv<T>::type, x##Type>) { 
    return static_cast<const T *>(this);                                     
} else {                                                                   
    return nullptr;                                                          
}
```

This conditions is true if the desired derived type `T` corresponds to `x##Type`. In this case, we can safety case to `T` without additional check because we are already in the switch-case corresponding to `TypeKind::x`, which means that the derived type is `x##Type`. This condition is false if the derived type  `T` does not match the true derived type, in which case `nullptr` is returned, in the same way as `dynamic_cast`.

hughperkins: Could you give a concrete example of a pair of types from gstaichi in common usage please, for each of the conditionals?

duburcqa: The most common types are probably `PrimitiveType`, `StructType`, and `TensorType`. Maybe I can unroll it for you:
```cpp
    switch (type_kind) {
      case TypeKind::Primitive:                                          
        if (std::is_same_v<T, PrimitiveType>) { 
          return static_cast<PrimitiveType *>(this);                                     
        } else {                                                                   
          return nullptr;                                                          
        }
      case TypeKind::Struct:                                          
        if (std::is_same_v<T, StructType>) { 
          return static_cast<StructType *>(this);                                     
        } else {                                                                   
          return nullptr;                                                          
        }
      case TypeKind::Tensor:                                          
        if (std::is_same_v<T, TensorType>) { 
          return static_cast<TensorType *>(this);                                     
        } else {                                                                   
          return nullptr;                                                          
        }
    }
```

YilingQiao (approved):

gstaichi/ir/type.h
hughperkins: Thats the switch. What about the first conditional?

duburcqa: Let's assume the user requested cast to `PrimitiveType`, then it gives:
```cpp
    switch (type_kind) {
      case TypeKind::Primitive:                                          
        if (std::is_same_v< PrimitiveType, PrimitiveType>) { 
          return static_cast<PrimitiveType *>(this);                                     
        } else {                                                                   
          return nullptr;                                                          
        }
      case TypeKind::Struct:                                          
        if (std::is_same_v< PrimitiveType, StructType>) { 
          return static_cast<StructType *>(this);                                     
        } else {                                                                   
          return nullptr;                                                          
        }
      case TypeKind::Tensor:                                          
        if (std::is_same_v< PrimitiveType, TensorType>) { 
          return static_cast<TensorType *>(this);                                     
        } else {                                                                   
          return nullptr;                                                          
        }
    }
```

Which reduces (at compile time) to this:
```cpp
    switch (type_kind) {
      case TypeKind::Primitive:                                          
        return static_cast<PrimitiveType *>(this);         
      case TypeKind::Struct:                                          
        return nullptr;        
      case TypeKind::Tensor:                                          
        return nullptr;             
    }
```

So basically, it returns the downcasted type if the actual type of the object is matching, nullptr otherwise, without resorting on safe but slow dynamic cast but rather by tracking the underlying type manually (much like std::variant).

hughperkins: Sure. But that's just the switch. What is the use-case of the conditional above that?
```
    if constexpr (std::is_same_v<typename std::remove_cv<T>::type, Type>) {
      return this;
    }
```

duburcqa: I already explained this one on top of my first reply.

duburcqa: “This condition is true when casting to common base type Type. In this case, there is nothing to do since implicit upcasting is allowed.”

It is true when to user explicitly request cast to `Type`.

```cpp
if (std::is_same_v<T, Type>) {
    return this;
}
```

hughperkins: After discussing with @duburcqa offline, realized/discvoered that `Type` in ths first conditioanl means the actual base calss called `Type`: it is not a generic Type variable. So it's already the concrete type name.

@duburcqa has agreed to add a comment to the first conditional to clarify this point.

duburcqa: done :)

PR #264, opened 2025-11-10
[Perf] Add template mapper key caching.

duburcqa:
```bash
```

| PR (branch) | runtime FPS |
| -------- | ------- |
| + #263 (duburcqa/avoid_dyn_cast) | 1200FPS  |
| **+ #264 (duburcqa/template_mapper_arg_cache)** | **2470FPS**  |
| + #265 (duburcqa/update_x86_arch) | 2520FPS  |

### DIFF
--- a/python/gstaichi/lang/_template_mapper.py
+++ b/python/gstaichi/lang/_template_mapper.py
@@ -65,7 +65,7 @@ class TemplateMapper:
         assert prog is not None
 
         mapping_cache_tracker: list[ReferenceType] | None = None
-        args_hash: ArgsHash = tuple(map(id, args))
+        args_hash: ArgsHash = tuple([id(arg) for arg in args])
         try:
             mapping_cache_tracker = self._mapping_cache_tracker[args_hash]
         except KeyError:


python/gstaichi/lang/_template_mapper.py
```diff
@@ -21,24 +36,56 @@ def __init__(self, arguments: list[ArgMetadata], template_slot_locations: list[i
...
         except KeyError:
-            count = len(self.mapping)
-            self.mapping[key] = count
-            return count, key
+            count = self.mapping[key] = len(self.mapping)
+
+        mapping_cache_tracker_: list[ReferenceType] = []
+        clear_callback = lambda ref: mapping_cache_tracker_.clear()
+        try:
+            mapping_cache_tracker_ += [ReferenceType(arg, clear_callback) for arg in args]
```
hughperkins, line 85: Some queetions:
1. under what circumstances will we have multiple sets of args with the same hash? Is this in case of hash collisions?
2. what is the advantage of storing all the collided args in one large list, rahter than in one tuple per args set?

duburcqa: > under what circumstances will we have multiple sets of args with the same hash? Is this in case of hash collisions?

Only if all the arguments have the exact same memory address. This cannot happen because we invalid a cache entry as soon as one of the arguments get destroyed.

duburcqa: > what is the advantage of storing all the collided args in one large list, rather than in one tuple per args set?

We are using one independent list per args set. It is a list instead of a tuple to be able to clear it in-place. It simplify the implementation, but it would be possible to use a tuple if you think it is important.

hughperkins: I don't. I trust your decisions on this; just trying to understand them :)

duburcqa:
> You mentioned that your PR runs faster? Is that because you are using _destroy_callback, and you trust that if any reference has been deleted, then this callback will have been called, and so you can directly use tuple(map(id, args)) as the hash, without any additional checks?

I think doing this is faster, but I'm thinking in may just be because other sources of overhead where dominating...

hughperkins (approved):
Thank you! 🙌

hughperkins (approved):
Thanks!

PR #265, opened 2025-11-10
[cpu] Move from 'nehalem' to 'x86-64-v3' as x86 micro-architecture baseline.

duburcqa:
`x86-64-v3` is a more sensible baseline than `nehalem`. The first micro-architecture supporting is `haswell`, which was introduced in 2013. Now, many linux-based operating systems are building linux and packages for this instruction set, including Ubuntu 25.04, so it is wildly accepted as the new standard in 2025.

Note that it does bring some minor performance improvement in practice, but the motivation is mainly moving away from `nehalem`, which is very old (2008) and a non-standard choice as a baseline.

------------------------

```bash
```

| PR (branch) | runtime FPS |
| -------- | ------- |
| + #263 (duburcqa/avoid_dyn_cast) | 1200FPS  |
| + #264 (duburcqa/template_mapper_arg_cache) | 2470FPS  |
| **+ #265 (duburcqa/update_x86_arch)** | **2520FPS** |

### DIFF
--- a/python/gstaichi/lang/_fast_caching/src_hasher.py
+++ b/python/gstaichi/lang/_fast_caching/src_hasher.py
@@ -39,15 +39,18 @@ def create_cache_key(
         return None
     kernel_hash = function_hasher.hash_kernel(kernel_source_info)
     config_hash = config_hasher.hash_compile_config()
-    cache_key = hash_iterable_strings((kernel_hash, args_hash, config_hash))
+    cache_key = hash_iterable_strings((kernel_hash, args_hash, config_hash, "pruned"))
     return cache_key
 
 
 class CacheValue(BaseModel):
     hashed_function_source_infos: list[HashedFunctionSourceInfo]
+    used_py_dataclass_leaves: set[str]
 
 
-def store(cache_key: str, function_source_infos: Iterable[FunctionSourceInfo]) -> None:
+def store(
+    cache_key: str, function_source_infos: Iterable[FunctionSourceInfo], used_py_dataclass_leaves: set[str]
+) -> None:
     """
     Note that unlike other caches, this cache is not going to store the actual value we want.
     This cache is only used for verification that our cache key is valid. Big picture:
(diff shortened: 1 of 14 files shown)


cmake/GsTaichiCXXFlags.cmake
```diff
@@ -118,12 +118,10 @@ if ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64" OR "${CMAKE_SYSTEM_PROCESSOR}"
...
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DTI_ARCH_x64")
         if ("arm64" IN_LIST CMAKE_OSX_ARCHITECTURES)
-            # TODO: (penguinliong) Will probably need this in a future version
-            # of Clang. Clang11 doesn't recognize this.
-            #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mcpu=apple-m1")
+            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mcpu=apple-m1")
         else()
-            message("Setting -march=nehalem for x86_64 processors")
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=nehalem")
+            message("Setting -march=haswell for x86_64 processors")
```
hughperkins: Question: how did you decide on haswell, cf somthing newer or older?

duburcqa: As mentioned in the description of this PR, 'haswell' is a very common baseline nowadays that corresponds to 'x86-64-v3' instruction set. Haswell architecture was introduced in 2013, while 'nehalem' was introduced in 2008.
See for reference: https://en.wikipedia.org/wiki/List_of_Intel_CPU_microarchitectures

image

hughperkins: for example, what is your reason for not choosing eg Skylake instead?

duburcqa: I updated the PR description. 'x86-64-v3' (which is more or less 'haswell') is a fairly standard choice nowadays that many linux distributions and packages are enforcing. Picking any more recent architecture may unlock additional performance boost but will prevent some machines from using GsTaichi. We may consider this acceptable, but I don't think the performance boost to expect is sufficient to justify this.

PR #268, opened 2025-11-10
[Misc] Remove textures

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/gstaichi/ir/type.cpp
+++ b/gstaichi/ir/type.cpp
@@ -157,16 +157,19 @@ std::string StructType::to_string() const {
 }
 
 size_t StructType::get_element_offset(const std::vector<int> &indices) const {
-  const Type *type_now = this;
-  size_t offset = 0;
-  for (auto ind : indices) {
-    if (auto tensor_type = type_now->cast<TensorType>()) {
-      TI_ASSERT(ind < tensor_type->get_num_elements())
+  int ind = indices[0];
+  size_t offset = elements_[ind].offset;
+  const Type *type_now = elements_[ind].type;
+  for (int i = 1; i < indices.size(); i++) {
+    ind = indices[i];
+    if (auto struct_type = type_now->cast<StructType>()) {
+      offset += struct_type->elements_[ind].offset;
+      type_now = struct_type->elements_[ind].type;
+    } else {
+      auto tensor_type = type_now->as<TensorType>();
+      TI_ASSERT(ind < tensor_type->get_num_elements());
       offset += tensor_type->get_element_offset(ind);
       type_now = tensor_type->get_element_type();
-    } else {
-      offset += type_now->as<StructType>()->elements_[ind].offset;
-      type_now = type_now->as<StructType>()->elements_[ind].type;
     }
   }
   return offset;
--- a/gstaichi/ir/type.h
+++ b/gstaichi/ir/type.h
@@ -20,6 +20,10 @@ enum class TypeKind : int {
 #undef PER_TYPE_KIND
 };
 
+#define PER_TYPE_KIND(x) class x##Type;
+#include "gstaichi/inc/type_kind.inc.h"
+#undef PER_TYPE_KIND
+
 class TI_DLL_EXPORT Type {
  public:
   explicit Type(TypeKind type_kind) : type_kind(type_kind) {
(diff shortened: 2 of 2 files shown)


python/gstaichi/lang/impl.py
```diff
@@ -140,7 +139,7 @@ def begin_frontend_struct_for(ast_builder, group, loop_range):
             'use "for I in ti.grouped(x)" to group all indices into a single vector I?'
         )
     dbg_info = _ti_core.DebugInfo(get_runtime().get_current_src_info())
-    if isinstance(loop_range, (AnyArray, RWTextureAccessor)):
+    if isinstance(loop_range, (AnyArray)):
```
duburcqa: Remove parentheses.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1049,7 +1041,7 @@ def extract_arguments(self) -> None:
             else:
                 if isinstance(
                     annotation,
-                    (template, ndarray_type.NdarrayType, texture_type.TextureType, texture_type.RWTextureType),
+                    (template, ndarray_type.NdarrayType),
```
duburcqa: Keep everything on one line.

python/gstaichi/_snode/fields_builder.py
duburcqa, line 1: What is the modification for doing updating this file? It seems completely unrelated.

hughperkins: circular imports.

hughperkins: after removing the import of texture stuff from impl.py, for some reason this file became ground zero for a zillion circular imports, mostly related to importing impl.

hughperkins: (If you've got any suggestions on how to fix it more gracefully, open to them. I guess the texture imports were somehow importing some stuff that fields builder needs too, so that when the time came to import fields builder, fields builder didnt choke up the import from impl?)

duburcqa: I will try to help you on this.

duburcqa:
```diff
diff --git a/python/gstaichi/types/__init__.py b/python/gstaichi/types/__init__.py
index 935e54da5..f8412739d 100644
--- a/python/gstaichi/types/__init__.py
+++ b/python/gstaichi/types/__init__.py
@@ -15,4 +15,6 @@ from gstaichi.types.ndarray_type import *  # type: ignore
 from gstaichi.types.primitive_types import *  # type: ignore
 from gstaichi.types.utils import *  # type: ignore

+import gstaichi.lang
+
 __all__ = ["quant"]
```

duburcqa: If you apply this patch, you can revert all our modifications in this file.

hughperkins: Nice! Dont suppose... please could you explain how this works please, so I can use this to solve future circular imports?

duburcqa: Sure. I will write a note. But first, I would argue this patch is cleaner because it is applied on top-level `__init__.py`, so easier to spot / more centralised.
```diff
diff --git a/python/gstaichi/__init__.py b/python/gstaichi/__init__.py
index 9a907d9a2..3d0713d42 100644
--- a/python/gstaichi/__init__.py
+++ b/python/gstaichi/__init__.py
@@ -1,5 +1,6 @@
 # type: ignore

+import gstaichi.lang
 from gstaichi import (
     ad,
     algorithms,
```

duburcqa: First, circular imports do not trigger circular import error systematically, but only if you need to TRULY import circularly, which means that none of the two modules has already been imported successfully before. It may sound like it is the same, since you cannot import one without importing the other anyway, but in practice it is a bit more subtle than this.

Let's start from the concrete circular import error you are facing here. Here is the trace:
```python
ImportError                               Traceback (most recent call last)
Cell In[1], line 1
----> 1 import gstaichi as ti
      4 def test_kernel_print():
      5     N = 10000

File ~/workspace/src/gstaichi/python/gstaichi/__init__.py:3
----> 3 from gstaichi import (
      4     ad,
      5     algorithms,
      6     experimental,
      7     linalg,
      8     math,
      9     sparse,
     10     tools,
     11     types,
     12 )
     13 from gstaichi._funcs import *
     14 from gstaichi._lib import core as _ti_core

File ~/workspace/src/gstaichi/python/gstaichi/ad/__init__.py:3
----> 3 from gstaichi.ad._ad import *

File ~/workspace/src/gstaichi/python/gstaichi/ad/_ad.py:15
     12 import numpy as np
     14 import gstaichi.types.primitive_types as types
---> 15 from gstaichi import _snode
     16 from gstaichi.lang import impl
     17 from gstaichi.lang._ndarray import Ndarray

File ~/workspace/src/gstaichi/python/gstaichi/_snode/__init__.py:3
----> 3 from gstaichi._snode.fields_builder import FieldsBuilder
      5 __all__ = ["FieldsBuilder"]

File ~/workspace/src/gstaichi/python/gstaichi/_snode/fields_builder.py:7
      5 from gstaichi._lib import core as _ti_core
      6 from gstaichi._lib.core.gstaichi_python import SNodeCxx
----> 7 from gstaichi._snode.snode_tree import SNodeTree
      8 from gstaichi.lang import impl, snode
      9 from gstaichi.lang.exception import GsTaichiRuntimeError

File ~/workspace/src/gstaichi/python/gstaichi/_snode/snode_tree.py:8
----> 8 from gstaichi.lang import impl
      9 from gstaichi.lang.exception import GsTaichiRuntimeError
     12 class SNodeTree:

File ~/workspace/src/gstaichi/python/gstaichi/lang/__init__.py:3
----> 3 from gstaichi.lang import impl, simt  # noqa: F401
      4 from gstaichi.lang._fast_caching.function_hasher import pure  # noqa: F401
      5 from gstaichi.lang._ndarray import *

File ~/workspace/src/gstaichi/python/gstaichi/lang/impl.py:15
      8 from gstaichi._lib import core as _ti_core
      9 from gstaichi._lib.core.gstaichi_python import (
     10     DataTypeCxx,
     11     Function,
     12     KernelCxx,
     13     Program,
     14 )
---> 15 from gstaichi._snode.fields_builder import FieldsBuilder
     16 from gstaichi.lang._ndarray import ScalarNdarray
     17 from gstaichi.lang._ndrange import GroupedNDRange, _Ndrange
 ```
This tells you that when you import `gstaichi`, you will start by importing a bunch of sub-modules, including `ad`. Then, `ad` requires importing `gstaichi._snode.snode_tree`, which in turns requires importing `from gstaichi.lang import impl`. At this point, you get an exception, because `lang` has never been imported so far, so it will have to be imported right now, including `lang.impl`. This is a problem, because importing the latter while trigger `from gstaichi._snode.fields_builder import FieldsBuilder`, which is impossible, because you are already importing `gstaichi._snode.fields_builder` AND `FieldsBuilder` is not already in `sys.module` (otherwise it would work), which leads to circular import error.

So now, all you need to do is making sure that `gstaichi.lang.impl` has been imported before `gstaichi._snode.snode_tree`! You can do this in different ways, the most obvious one being the second patch I suggested, the first one being more convoluted: importing `gstaichi.lang.impl` right away will trigger import of `gstaichi._snode.fields_builder` (and gstaichi._snode.snode_tree), which is just fine because they only import `gstaichi.lang.impl` rather than a specific object. What is saving you here is that a module is added to `sys.modules` registry at the very beginning of its import rather than the end, which means that `gstaichi.lang.impl` itself is already considered imported and there you escape circular import error by breaking the circle.

hughperkins: "a module is added to sys.modules registry at the very beginning of its import rather than the end" => oh, interestnig. This is new information to me. I always assumed the other way around 🤔 

Doing some more experimentation, it looks like circular *import*s are actually fine, because of this. What breaks things is when one tries to access attributes in another module, before that module has defined them yet?

duburcqa: > I always assumed the other way around 🤔

I know right!

> What breaks things is when one tries to access attributes in another module, before that module has defined them yet?

Exactly!

hughperkins: Based on the knowledge you just imparted to me, I tried a different approach, to practice :)

PR #269, opened 2025-11-11
[Perf] Minor launch kernel python overhead optimisation.

duburcqa:
Slightly faster:
* Early-return in `_check_gradient_field_not_placed`
* Define proxies for Pybind11 enum values.

From 5000FPS to 5300FPS on MacOS for non-batched cpu-based single franka using ndarray.

### DIFF
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -82,6 +82,13 @@ from .._test_tools import warnings_helper
 MAX_ARG_NUM = 512
 
 
+# Define proxies for fast lookup
+_NONE, _VALIDATION, _FORWARD, _REVERSE = (
+    AutodiffMode.NONE,
+    AutodiffMode.VALIDATION,
+    AutodiffMode.FORWARD,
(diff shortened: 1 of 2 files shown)


python/gstaichi/lang/kernel_impl.py
```diff
@@ -84,6 +84,13 @@
 MAX_ARG_NUM = 512
 
 
+# Define proxies for fast lookup
+_NONE, _VALIDATION, _FORWARD, _REVERSE = (
```
hughperkins, line 86: could you add a comment about why we are adding the enum values into varaible please, so we dont inadvertently revert the change in the future?

duburcqa: Wait, I already added a comment no ? `# Define proxies for fast lookup`

hughperkins: Yes, fair.

hughperkins (approved):
Thanks! 🙌

YilingQiao (approved):

PR #270, opened 2025-11-11
[Type] Add to_dlpack to ndarray tensors

hughperkins:
Issue: #

### Brief Summary

Use like:

```
my_dlpack = my_ndarray.to_dlpack()
my_torch_tensor = torch.utils.dlpack.from_dlpack(my_dlpack)
```

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/gstaichi/python/dlpack_funcs.cpp
+++ b/gstaichi/python/dlpack_funcs.cpp
@@ -63,7 +63,7 @@ pybind11::capsule ndarray_to_dlpack(Program *program,
   DataType ndarray_data_type = ndarray->get_element_data_type();
   uint8_t data_type_code = kDLInt;
 
-  uint8_t element_bits = -1;
+  uint8_t element_bits = 0;
   PrimitiveTypeID type_id = ndarray_data_type->as<PrimitiveType>()->type;
   switch (type_id) {
     case PrimitiveTypeID::i32: {


gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,130 @@
...
+#endif  // TI_WITH_CUDA
+#include "gstaichi/rhi/cpu/cpu_device.h"
+
+namespace gstaichi {
+namespace lang {
+pybind11::capsule ndarray_to_dlpack(Program *program,
+                                    pybind11::object owner,
+                                    Ndarray *ndarray) {
+  auto *owner_holder = new pybind11::object(owner);
+  new pybind11::object(owner);
```
duburcqa: What is this ??

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+  DeviceAllocation devalloc = ndarray->get_device_allocation();
+
+  void *raw_ptr = nullptr;
+  DLDeviceType device_type = DLDeviceType::kDLCPU;
+
+  cpu::CpuDevice *cpu_device = dynamic_cast<cpu::CpuDevice *>(devalloc.device);
```
duburcqa: Avoid using dynamic cast if possible, because it is very slow. I have the feeling it should be easy to detect if it can be casted to CPUDevice no?

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,131 @@
...
+  if (cpu_device != nullptr) {
+    cpu::CpuDevice::AllocInfo alloc_info = cpu_device->get_alloc_info(devalloc);
+    raw_ptr = alloc_info.ptr;
+  }
+#if TI_WITH_CUDA
+  cuda::CudaDevice *cuda_device =
+      dynamic_cast<cuda::CudaDevice *>(devalloc.device);
```
duburcqa: Same, it should at least be an `else if` or inside a `else` clause to avoid trying to cast something that has no chance to work.

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  std::vector<int> ndarray_shape = ndarray->total_shape();
+  int ndim = ndarray_shape.size();
+
+  int64_t *shape = nullptr;
+  if (ndim > 0) {
+    shape = new int64_t[ndim];
+    for (int i = 0; i < ndim; i++) {
+      shape[i] = ndarray_shape[i];
+    }
```
duburcqa: Note that you could do: `std::copy(ndarray_shape.begin(), ndarray_shape.end(), shape);`

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,131 @@
...
+    strides[ndim - 1] = 1;
+    for (int i = ndim - 2; i >= 0; i--) {
+      strides[i] = strides[i + 1] * shape[i + 1];
+    }
+  }
+
+  DataType ndarray_data_type = ndarray->get_element_data_type();
+  uint8_t data_type_code = kDLInt;
+
+  uint8_t element_bits = 32;
```
duburcqa: Strange to init to 32.

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,131 @@
...
+  dl_tensor.data = raw_ptr;
+  dl_tensor.device.device_type = device_type;
+  dl_tensor.device.device_id = 0;
+  dl_tensor.ndim = ndim;
+  dl_tensor.dtype = DLDataType{data_type_code, element_bits, 1};
+  dl_tensor.shape = shape;
+  dl_tensor.strides = strides;
+  dl_tensor.byte_offset = 0;
+
+  managed_tensor->manager_ctx = static_cast<void *>(owner_holder);
```
duburcqa: Any raw pointer can be implicitly casted to void * without `static_cast`

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,131 @@
...
+  managed_tensor->deleter = [](DLManagedTensor *self) {
+    auto *owner = reinterpret_cast<pybind11::object *>(self->manager_ctx);
+    pybind11::gil_scoped_acquire gil;
+    delete owner;  // DECREFs the Python object
+    if (self->dl_tensor.shape != nullptr) {
```
duburcqa, line 116: `if (self->dl_tensor.shape) {` should be good enough no ?

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,131 @@
...
+      delete[] self->dl_tensor.shape;
+      delete[] self->dl_tensor.strides;
+    }
+    delete self;
+  };
+  auto deleter = [](PyObject *capsule) {};
```
duburcqa: I would suggest to call this `capsule_deleter` or just inline its definition:
```cpp
  pybind11::capsule capsule = pybind11::capsule(
      static_cast<void *>(managed_tensor), "dltensor", [](PyObject *capsule) {});
```

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+  pybind11::capsule capsule = pybind11::capsule(
+      static_cast<void *>(managed_tensor), "dltensor", deleter);
```
duburcqa: Same, casting to `void *` is pedantic.

gstaichi/python/dlpack_funcs.cpp
hughperkins: This is how we incref

gstaichi/python/dlpack_funcs.cpp
hughperkins: I have to init it to something. It's pretty arbitrary. 32 seems a pretty standard size for most of what we are using.

gstaichi/python/dlpack_funcs.cpp
hughperkins: I'm not a fan of using 'magic' comparisons, since Iv'e been burned a few times when I thought I was comparing to None, in python, but eg set() was also evaluating to false. Not sure if the same situation is possible in c++? ie if I remove `!= nullptr`, is it *guaranteed* that this will only evaluate to true if shape is not `nullptr`?

gstaichi/python/dlpack_funcs.cpp
duburcqa: You do it twice? This is worth a comment at least.

gstaichi/python/dlpack_funcs.cpp
duburcqa: That is the issue, it gives the feeling that this value is valid and therefore may be used in some cases. I would rather set it to 0.

gstaichi/python/dlpack_funcs.cpp
duburcqa: Here `self->dl_tensor.shape` is a pointer. Pointers are guaranteed to be evaluated as False if and only if the pointer is nullptr, True otherwise. But I get your point. If you are more comfortable with explicit comparison, just do it.

gstaichi/python/dlpack_funcs.cpp
hughperkins: Ah no, thats a bug 😅  Thanks! (basically, I cant compile with 'unused variables', so when I was playing around, I commented out the frist line, and used the second line instead).

gstaichi/python/dlpack_funcs.cpp
hughperkins: Not entirely convinced, since I quite like variables to hold valid values, where possible, but ok.

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -0,0 +1,129 @@
...
+    strides[ndim - 1] = 1;
+    for (int i = ndim - 2; i >= 0; i--) {
+      strides[i] = strides[i + 1] * shape[i + 1];
+    }
+  }
+
+  DataType ndarray_data_type = ndarray->get_element_data_type();
+  uint8_t data_type_code = kDLInt;
+
+  uint8_t element_bits = -1;
```
duburcqa: `-1` for `uint8_t` is uncanny

hughperkins: ah, whoops 😅

YilingQiao (approved):

PR #278, opened 2025-11-17
[Bug] Fix ndarray memory leak

hughperkins:
Issue: #

### Brief Summary

See https://github.com/taichi-dev/taichi/pull/8764 by @BernardoCovas for more details.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -177,7 +177,6 @@ class GsTaichiCallable:
         self._primal: Kernel | None = None
         self._adjoint: Kernel | None = None
         self.grad: Kernel | None = None
-        self._is_staticmethod: bool = False
         self.is_pure: bool = False
         update_wrapper(self, fn)
 
@@ -191,7 +190,7 @@ class GsTaichiCallable:
 
 
 class BoundGsTaichiCallable:
-    def __init__(self, instance: Any, gstaichi_callable: "GsTaichiCallable"):
+    def __init__(self, instance: Any, gstaichi_callable: GsTaichiCallable):
         self.wrapper = gstaichi_callable.wrapper
         self.instance = instance
         self.gstaichi_callable = gstaichi_callable
@@ -205,10 +204,14 @@ class BoundGsTaichiCallable:
 
     def __setattr__(self, k: str, v: Any) -> None:
         # Note: these have to match the name of any attributes on this class.
-        if k in ("wrapper", "instance", "gstaichi_callable"):
+        if k in {"wrapper", "instance", "gstaichi_callable"}:
             object.__setattr__(self, k, v)
         else:
             setattr(self.gstaichi_callable, k, v)
+
+    def grad(self, *args, **kwargs) -> "Kernel":
+        assert self.gstaichi_callable._adjoint is not None
+        return self.gstaichi_callable._adjoint(self.instance, *args, **kwargs)
 
 
 def func(fn: Callable, is_real_function: bool = False) -> GsTaichiCallable:
@@ -962,6 +965,7 @@ class Kernel:
         # A materialized kernel is a KernelCxx object which may or may not have
         # been compiled. It generally has been converted at least as far as AST
         # and front-end IR, but not necessarily any further.
+        self._dump_kernel_checksums = os.environ.get("TI_DUMP_KERNEL_CHECKSUMS", "0") == "1"
         self.materialized_kernels: dict[CompiledKernelKeyType, KernelCxx] = {}
         self.has_print = False
         self.gstaichi_callable: GsTaichiCallable | None = None
@@ -1344,7 +1348,7 @@ class Kernel:
                 prog_device_cap = prog.get_device_caps()
 
                 compile_result: CompileResult = prog.compile_kernel(prog_config, prog_device_cap, t_kernel)
-                if os.environ.get("TI_DUMP_KERNEL_CHECKSUMS", "0") == "1":
+                if self._dump_kernel_checksums:
                     debug_dump_path = pathlib.Path(impl.current_cfg().debug_dump_path)
                     checksums_file_path = debug_dump_path / "checksums.csv"
                     kernels_dump_dir = debug_dump_path / "kernels"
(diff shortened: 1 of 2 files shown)


tests/python/test_ndarray.py
```diff
@@ -1199,3 +1199,16 @@ def my_kernel({args}) -> None:
...
+@test_utils.test()
+def test_ndarray_del(dtype) -> None:
+    def foo():
+        nd = ti.ndarray(ti.i32, (1000,))
+        assert nd.arr.shape != []
+        return nd.arr
+
+    arr = foo()
+    # after deletion, the shape is now empty, so this shows the arr was deleted
+    assert arr.shape == []
```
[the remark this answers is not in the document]
hughperkins: this isnt true. shape is a list. changing to tuple causes test to fail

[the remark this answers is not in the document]
hughperkins: see below

tests/python/test_ndarray.py
```diff
@@ -1199,3 +1199,16 @@ def my_kernel({args}) -> None:
         my_kernel(*arg_objs_l)
     for i in range(num_args):
         assert arg_objs_l[i][0] == i + 1
+
+
+@pytest.mark.parametrize("dtype", [ti.i32, ti.types.vector(3, ti.f32), ti.types.matrix(2, 2, ti.f32)])
+@test_utils.test()
+def test_ndarray_del(dtype) -> None:
+    def foo():
+        nd = ti.ndarray(ti.i32, (1000,))
```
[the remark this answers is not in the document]
hughperkins: addressed 😅

tests/python/test_ndarray.py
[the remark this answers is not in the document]
duburcqa: Seriously ? I would consider this as a bug, though a minor one

python/gstaichi/lang/_ndarray.py
```diff
@@ -38,6 +38,12 @@ def __init__(self):
         # we register with runtime, in order to enable reset to work later
         impl.get_runtime().ndarrays.add(self)
 
+    def __del__(self):
+        if impl is not None and impl.get_runtime is not None and impl.get_runtime() is not None:
+            prog = impl.get_runtime()._prog
+            if prog is not None:
```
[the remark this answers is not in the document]
duburcqa: I would even advice to use getattr. Better make sure nothing wrong can happen in destructors.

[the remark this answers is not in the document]
hughperkins: `if prog is not None and getattr(self, 'arr') is not None:` ?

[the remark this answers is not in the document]
duburcqa:
```python
arr = getattr(self, 'arr')
if arr is not None:
    prog = [...]
    if prog is not None:
        [...]
```

python/gstaichi/lang/_ndarray.py
```diff
@@ -38,6 +38,14 @@ def __init__(self):
         # we register with runtime, in order to enable reset to work later
         impl.get_runtime().ndarrays.add(self)
 
+    def __del__(self):
+        if impl is not None and impl.get_runtime is not None and impl.get_runtime() is not None:
+            arr = getattr(self, "arr")
+            if arr is not None:
+                prog = impl.get_runtime()._prog
+                if prog is not None:
+                    prog.delete_ndarray(self.arr)
```
duburcqa: `arr`

gstaichi/python/export_lang.cpp
```diff
@@ -370,6 +370,8 @@ void export_lang(py::module &m) {
            [](Program *program, SNode *snode, int element_ndim, int n, int m) {
              return field_to_dlpack(program, snode, element_ndim, n, m);
            })
+      .def("_unsupported_get_num_ndarrays",
+           &Program::unsupported_get_num_ndarrays)
```
duburcqa: Why is this considered unsupported?

By the way, I would recommend using readonly property instead of function for functions that are basically getters. (and read/write properties if there is any setter).

hughperkins: > Why is this considered unsupported?

because it's only for tests, and I don't want to add new supported features, which then have to be maintained, reduce agility etc. No-one is asking for this functionality. There is no reason to make it part of our standard supported API I feel.

hughperkins: (coudl also call it `debug_` or something, but I think `unsupported_` or maybe `non_api_` or something simlar is clearer I feel?)

hughperkins: (`internal_` perhaps?)

hughperkins: (I tried to use `_` prefix, but clang tidy wouldnt allow me to)

hughperkins: (I kind of like `internal_` tbh. Thoughts?)

duburcqa: Usually `_` prefix should be sufficient to indicate it is not part of public API. But you can add internal if you think it is really really necessary.

hughperkins: as stated, clang-tidy wouldnt allow `_` prefix, since it's a pubblic method.

hughperkins: oh, you mean, on the python side? I guess I'd like the name to be ~somewhat consistent with the c++ side, on the whole.

duburcqa: As you want, but this is not a common practice at all for python-first package. It makes the API just plain horrible to use. The good thing is that taichi never exposes the python bindings to the user, so we are good I guess.

hughperkins: You're saying:
- on the C++ side, we aren't providing a supported API anyway, and therefore we don't need `unsupported_` prefix?
- and on the python side, we should simply put `_` prefix, and that is sufficient?

duburcqa: Yes.

But what I'm truly saying is that the C++ core must be at the service of the Python API, because gstaichi targets Python users and nothing else. So if you have to choose, good user experience in Python comes before good user experience in C++. Obviously, this should not compromise maintainability, which may involve having a difference private API in C++ vs Public API in Python.

In this case, it is not a big deal, but I think it is very important for the users to feel right at home even when it comes to Python bindings.

duburcqa (approved):

PR #279, opened 2025-11-19
[Pref] Avoid runtime overhead due to custom data_oriented attribute getter.

duburcqa:
Huge speedup of Genesis zero-copy getters (https://github.com/Genesis-Embodied-AI/Genesis/pull/2011) thanks to this improvement:

```bash
GS_ENABLE_NDARRAY=1 GS_ENABLE_ZEROCOPY=1 ipython
>>> %run examples/rigid/single_franka.py --cpu
>>> scene = gs._scene_registry[0]()
>>> %timeit scene.entities[0].get_verts()
```

* Before: `16.1 μs ± 93.4 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)`
* After: `10.6 μs ± 42.5 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)`

### DIFF
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -177,7 +177,6 @@ class GsTaichiCallable:
         self._primal: Kernel | None = None
         self._adjoint: Kernel | None = None
         self.grad: Kernel | None = None
-        self._is_staticmethod: bool = False
         self.is_pure: bool = False
         update_wrapper(self, fn)
 
@@ -1621,16 +1620,12 @@ class _BoundedDifferentiableMethod:
         self._kernel_owner = kernel_owner
         self._primal = wrapped_kernel_func._primal
         self._adjoint = wrapped_kernel_func._adjoint
-        self._is_staticmethod = wrapped_kernel_func._is_staticmethod
         self.__name__: str | None = None
 
     def __call__(self, *args, **kwargs):
         try:
             assert self._primal is not None
-            if self._is_staticmethod:
-                return self._primal(*args, **kwargs)
             return self._primal(self._kernel_owner, *args, **kwargs)
-
         except (GsTaichiCompilationError, GsTaichiRuntimeError) as e:
             if impl.get_runtime().print_full_traceback:
                 raise e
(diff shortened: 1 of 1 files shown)


python/gstaichi/lang/kernel_impl.py
```diff
@@ -1507,35 +1513,30 @@ def _kernel_impl(_func: Callable, level_of_class_stackframe: int, verbose: bool
     # Having |primal| contains |grad| makes the tape work.
     primal.grad = adjoint
 
+    @wraps(_func)
+    def wrapped_func(*args, **kwargs):
```
hughperkins, line 1515: out of curiosity, what is the reason for moving this function?

duburcqa: Because now it is used inside `wrapped_classkernel`, which is no longer a `noreturn` but a real valid function that this doing something: checking that self is correctly passed because calling `wrapped_func`.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1670,41 +1671,34 @@ def data_oriented(cls):
...
-                    return ret()
-                return ret
-        if is_property:
-            return x(self)
-        return x
+            ret = property(ret)
+        return ret
 
-    cls.__getattribute__ = _getattr
+    for name, attr in cls.__dict__.items():
```
hughperkins, line 1688: Could you add a comment explaining what this is doing please?

hughperkins:
> > Do you hvae some objective benchmark demonstrating the effect of this PR please?
> 
> Are you happy with the one I provided or you want something more isolated?

Looks great! Makes it easier to understand the context if we look back and want to know why we made these changes 6-12 months from now.

python/gstaichi/lang/kernel_impl.py
duburcqa: done!

hughperkins: what do you mean by "throw _BoundedDifferentiableMethod systematically"? I don't see any throw in this for loop.

duburcqa: Through* bad english

PR #280, opened 2025-11-20
[Perf] Implement to_dlpack for ndarrays on Metal

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/manylinux_wheel.yml
+++ b/.github/workflows/manylinux_wheel.yml
@@ -17,7 +17,7 @@ jobs:
   build_wheel:
     name: Manylinux wheel Build
     runs-on: ${{ matrix.os }}
-    container: ${{ matrix.os == 'ubuntu-22.04' && 'quay.io/pypa/manylinux_2_28_x86_64:latest' || matrix.os == 'ubuntu-22.04-arm' && 'quay.io/pypa/manylinux_2_34_aarch64:latest' }}
+    container: ${{ matrix.os == 'ubuntu-22.04' && 'quay.io/pypa/manylinux_2_28_x86_64:latest' || matrix.os == 'ubuntu-22.04-arm' && 'quay.io/pypa/manylinux_2_34_aarch64:2025.11.11-1' }}
     concurrency:
       # group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}-${{ matrix.PYTHON_CP_VERSION }}-${{ matrix.PYTHON_CP_VERSION != 'cp310' && 'all' || github.sha }}-build
       group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}-${{ matrix.PYTHON_CP_VERSION }}-${{matrix.os}}-build


gstaichi/python/dlpack_funcs.cpp
```diff
@@ -41,14 +45,30 @@ std::pair<void *, DLDeviceType> get_raw_ptr(Arch arch,
...
+    metal::MetalDevice *metal_device =
+        static_cast<metal::MetalDevice *>(dev_alloc.device);
+    device_type = DLDeviceType::kDLMetal;
+    const metal::MetalMemory &memory =
+        metal_device->get_memory(dev_alloc.alloc_id);
+
+    RhiResult result = memory.mapped_ptr(&raw_ptr);
+    if (result != RhiResult::success || raw_ptr == nullptr) {
+      MTLBuffer_id mtl_buffer = memory.mtl_buffer();
+      raw_ptr = reinterpret_cast<void *>(mtl_buffer);
```
duburcqa: Are you sure explicit cast via `reinterpret_cast` is necessary in this context?

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -168,14 +193,24 @@ pybind11::capsule field_to_dlpack(Program *program,
...
   void *raw_ptr = nullptr;
   DLDeviceType device_type = DLDeviceType::kDLCPU;
-  std::tie(raw_ptr, device_type) = get_raw_ptr(arch, program, tree_device_ptr);
-  raw_ptr = (void *)((uint64_t)raw_ptr + field_in_tree_offset);
+  bool pointer_arithmetic_ok = true;
+  std::tie(raw_ptr, device_type, pointer_arithmetic_ok) =
+      get_raw_ptr(arch, program, tree_device_ptr);
+  if (field_in_tree_offset != 0) {
+    if (pointer_arithmetic_ok) {
+      raw_ptr = (void *)((uint64_t)raw_ptr + field_in_tree_offset);
```
duburcqa: Are you sure explicit cast is necessary in this context?

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -208,6 +243,7 @@ pybind11::capsule field_to_dlpack(Program *program,
   dl_tensor.dtype = DLDataType{data_type_code, element_bits, 1};
   dl_tensor.shape = shape;
   dl_tensor.strides = strides;
+  // note: torch doesn't yet support non-zero byte_offset
```
duburcqa: Could you replace this by a FIXME + link to issue/PR in torch?

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -43,8 +50,8 @@ def test_dlpack_types(tensor_type, dtype, shape: tuple[int], poses: list[tuple[i
     }[dtype]
     assert tt.dtype == expected_torch_type
     for i, pos in enumerate(poses):
-        assert tt[pos] == ti_tensor[pos]
-        assert tt[pos] != 0
+        assert tt[pos].item() == ti_tensor[pos]
+        assert tt[pos].item() != 0
```
duburcqa: Are you sure adding `.item()` is necessary?

gstaichi/python/dlpack_funcs.cpp
hughperkins: Screenshot 2025-11-20 at 09 48 15

gstaichi/python/dlpack_funcs.cpp
hughperkins: apparently not:

Screenshot 2025-11-20 at 09 50 25

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: removed

duburcqa: There is a special handling in torch that allows for 0D tensor to be implicitly interpreted as a native python scalar whatever the device.

gstaichi/python/dlpack_funcs.cpp
hughperkins: updated

gstaichi/python/dlpack_funcs.cpp
duburcqa: Yes you are right. Is `static_cast` enough? I don't think so. If not, then go for `reinterpret_cast` to use modern-style cast only?

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -27,10 +26,18 @@ def ti_to_torch(ti_tensor: ti.types.NDArray) -> torch.Tensor:
...
+@pytest.mark.parametrize("tensor_type", [ti.ndarray, ti.field])
 def test_dlpack_types(tensor_type, dtype, shape: tuple[int], poses: list[tuple[int, ...]]) -> None:
+    if ti.cfg.arch == ti.metal and dtype is ti.f64:
+        pytest.skip("Metal does not support f64")
     ti_tensor = tensor_type(dtype, shape)
     for i, pos in enumerate(poses):
         ti_tensor[pos] = i * 10 + 10
+    if ti.cfg.arch == ti.metal and dtype is ti.u1:
+        with pytest.raises(RuntimeError):
+            dlpack = ti_tensor.to_dlpack()
```
duburcqa, line 39: I would rather have a xfail than making sure it fails. This is weird, you are promoting a bug as a feature that should be maintained.

hughperkins: before I added TI_ERROR, this wsa segfaulting. Would xfail ensure this is the case too? (i'm not saying it wouldnt, just posing the question)

hughperkins: (ie would it check that it doesnt segfault, and cause a full test failure or similar if it does?)

duburcqa: You need to try it. I cannot tell for sure. It is was just me, I would just skip this case (statically, via `pytest.mark.skip` with parameter-specific marker. See https://docs.pytest.org/en/stable/example/parametrize.html#set-marks-or-test-id-for-individual-parametrized-test)

hughperkins: Yeah, I would prefer to check taht there is no segfault. Neither of us like segfaults. Happy to add a TODO or something if you want?

hughperkins: I can add an xfail *after* checking no segfault if ou want?

duburcqa: As you want. Want there should be at least some FIXME to track this bug.

YilingQiao (approved):

hughperkins:
I'll re-open a new one...

PR #282, opened 2025-11-21
[Perf] Speed up argument processing on Python-side.

duburcqa:
Benchmark for non-batched cpu-based single franka On Genesis zero-copy branch (https://github.com/Genesis-Embodied-AI/Genesis/pull/2011) (219fb52f):

### BEFORE

* field: 5900FPS
* ndarray: 5800FPS

image

### AFTER

* field: 6450FPS
* ndarray: 6250FPS

image

### DIFF
--- a/tests/python/test_argument.py
+++ b/tests/python/test_argument.py
@@ -73,10 +73,22 @@ def test_function_keyword_args():
         assert b == 2
         assert c == 4
 
+    @ti.func
+    def all_default(a=1, b=2, c=3):
+        assert a == 1
+        assert b == 2
+        assert c == 3
+
+    @ti.func
+    def do_nothing():
+        pass
+
     @ti.kernel
     def baz():
         foo(1, b=2)
         bar(b=2, a=1, c=4)
+        all_default()
+        do_nothing()
 
     baz()
 


python/gstaichi/lang/kernel_impl.py
```diff
@@ -390,57 +393,47 @@ def _process_args(
...
+                    if key == arg_meta.name:
+                        if i < num_args:
+                            raise GsTaichiSyntaxError(f"Multiple values for argument '{key}'.")
+                        break
+                else:
+                    raise GsTaichiSyntaxError(f"Unexpected argument '{key}'.")
+    elif num_missing_args:
+        for arg_meta in self.arg_metas_expanded[num_args:]:
+            if fused_args is _ARG_EMPTY:
+                raise GsTaichiSyntaxError(f"Missing argument '{arg_meta.name}'.")
```
hughperkins, line 438: I remember that I added all that debug info, otherwise it's really hard to debug stuff. And not just for me, but people who are using gstaichi. To be fair, you are the person who will have to deal with its absence, so ultimately I will defer to your decision on this, and you can always put it back later :) But: are you sure you want to remove this debug info?

python/gstaichi/lang/kernel_impl.py
```diff
@@ -390,57 +393,47 @@ def _process_args(
...
-
-    for key, value in kwargs.items():
-        for i, arg in enumerate(self.arg_metas_expanded):
-            if key == arg.name:
-                if i < len_args:
-                    raise GsTaichiSyntaxError(f"Multiple values for argument '{key}'.")
+    missing_arg_metas = self.arg_metas_expanded[num_args:]
+    num_missing_args = len(missing_arg_metas)
+    fused_args: list[Any] = [*args, *[arg_meta.default for arg_meta in missing_arg_metas]]
+    if kwargs:
```
hughperkins, line 414: can you remind me what test(s) we have for this section of code please? (ideally with urls, line numbers, etc).

python/gstaichi/lang/kernel_impl.py
duburcqa: Yes, I removed them on purpose, because ultimately I found it more confusing than helpful. This was too verbose and I was struggling understanding what was the issue. In this case, all I need to know is which variable is missing, passed multiple times, or unexpected. Nothing more, nothing less. Same as for any using Python function.

But I'm also ok to restore the original behaviour if you think it is important.

hughperkins: No. You are the client. When it's me debugging, I throw in whatever spam I want, all over the place :)

python/gstaichi/lang/kernel_impl.py
duburcqa: They are defined in `tests/python/test_argument.py`.

I double checked manually that it is going through all the branches!

duburcqa: Actually there is one single branch that is not checked in unit test yet. Let me add it.

duburcqa: Fixed! It was actually broken x)

hughperkins: could you provide a url pelase, cos I'm really lazy, and can then just click on it, rather than having to go to gstaichi website, and click my way through the hierarchy :P (You'll notice I usually provide you with direct URLs when referring to code I think?)

duburcqa: Here it is:
https://github.com/Genesis-Embodied-AI/gstaichi/blob/main/tests/python/test_argument.py

hughperkins: Thanks! Hmmm, I'm going to add a ticket to Linear to add SonarQube or smilar to the CI, which makes checking these things much easier.

hughperkins: In the absence of Sonarqube, I asked Cursor what it thinks. It says:

Screenshot 2025-11-24 at 10 18 07

Thoughts?

duburcqa: * 1.: Maybe, I don't know what this is, and did not modified this part of the code.
* 2.: Same as 1.
* 3.: Not true
* 4.: Not true
* 5.: Not true
* 6.: This is supported and not an edge case. Just the normal code flow. But sure there is no unit test for this. I can add one if you want.
* 7.: Not true.

The code Cursor is referring to does not exist.

hughperkins: Yeah, for 1 and 2, seems plausible that it doesnt relate to your change, though I think that 2 refers to python dataclasses, but I feel we've probably got ok-ish coverage for python dataclasses, so doesn't worry me unduly.

Point 3, what do you mean by 'not true'? What test do we have for this currnetly?
- I see a mixture https://github.com/Genesis-Embodied-AI/gstaichi/blob/64195c3dcc43ef82ba5bb7dcfbaaf43a593c4993/tests/python/test_argument.py#L15
- missing kwargs https://github.com/Genesis-Embodied-AI/gstaichi/blob/64195c3dcc43ef82ba5bb7dcfbaaf43a593c4993/tests/python/test_argument.py#L19
- more missing kwargs https://github.com/Genesis-Embodied-AI/gstaichi/blob/64195c3dcc43ef82ba5bb7dcfbaaf43a593c4993/tests/python/test_argument.py#L30
- this one doesnt even have any kernel args at all https://github.com/Genesis-Embodied-AI/gstaichi/blob/64195c3dcc43ef82ba5bb7dcfbaaf43a593c4993/tests/python/test_argument.py#L81
- etc ...

Could you also point to a test case for each for 4,5,6,7 please?

For 'does not exist', it exists in my cloned branch, see screenshot, but sesems to no longer exist in cloud 🙌 
Screenshot 2025-11-24 at 10 54 47

duburcqa: Honestly I don't care which test is checking what because I already made sure we are going through all the branches but OK.

* 3.: `test_function_keyword_args`
* 4.: `test_function_keyword_args_duplicate`
* 5.: `test_kernel_args_missing`, `test_kernel_keyword_args_missing`
* 6.: No test for this. I see no point. It is not an edge case at all.
* 7.: We have check for defaults but not all default (`test_function_keyword_args`). This is NOT an edge case either.

I will had a unit test for 6. and 7. even if I don't get the point.

duburcqa: Done.

* 6.: `test_function_keyword_args`
* 7.: `test_function_keyword_args`

hughperkins: thanks! Yes, like I say, Sonarqube would make this all much easier, more objective. We should set that up soon-ish I feel.

hughperkins: (or some equivalent. but I've only ever used Sonarqube, and I know that Sonarqube has a clear ui showing, for the ocde, which new lines are uncovered, etc)

hughperkins: sonarqube-covered-test-units

duburcqa: > Sonarqube would make this all much easier, more objective. 

I would say more observable. Because my check was objective, I added breakpoint() add the beginning of each branch and deleted them one by one after trigger any of them when running the test suite.

hughperkins: Yes, observable to the reviewer. And more repeatable. Icannot see that you have done that, nor can either of us easily re-run your checks after each new commit.

duburcqa: This would be indeed much much better, big fan of coverage report.

duburcqa: For the record:
https://github.com/Genesis-Embodied-AI/Genesis/pull/1223

hughperkins: We used to have sonarqube run asynchronously, once everything else had finished running. Since, as you allude to, it's very slow.

PR #283, opened 2025-11-23
[Perf] Redo fastcache kernel key to reuse existing front end cache key

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/manylinux_wheel.yml
+++ b/.github/workflows/manylinux_wheel.yml
@@ -17,7 +17,7 @@ jobs:
   build_wheel:
     name: Manylinux wheel Build
     runs-on: ${{ matrix.os }}
-    container: ${{ matrix.os == 'ubuntu-22.04' && 'quay.io/pypa/manylinux_2_28_x86_64:latest' || matrix.os == 'ubuntu-22.04-arm' && 'quay.io/pypa/manylinux_2_34_aarch64:latest' }}
+    container: ${{ matrix.os == 'ubuntu-22.04' && 'quay.io/pypa/manylinux_2_28_x86_64:latest' || matrix.os == 'ubuntu-22.04-arm' && 'quay.io/pypa/manylinux_2_34_aarch64:2025.11.11-1' }}
     concurrency:
       # group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}-${{ matrix.PYTHON_CP_VERSION }}-${{ matrix.PYTHON_CP_VERSION != 'cp310' && 'all' || github.sha }}-build
       group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}-${{ matrix.PYTHON_CP_VERSION }}-${{matrix.os}}-build
--- a/python/gstaichi/lang/misc.py
+++ b/python/gstaichi/lang/misc.py
@@ -756,6 +756,21 @@ def is_extension_enabled(ext: Extension) -> bool:
     return is_extension_supported(arch, ext)
 
 
+def dump_compile_config() -> None:
+    """
+    Dumps the compile config, which can be usful for example in diagnosing fastcache issues, since
+    the fastcache cache keys depend on the compile config.
+    """
+    config = impl.current_cfg()
+    config_l = []
+    for _k in sorted(dir(config)):
+        if _k.startswith("_"):
+            continue
+        v = getattr(config, _k)
+        config_l.append(f"{_k}={v}")
+    print("\n".join(config_l))
+
+
 __all__ = [
     "i",
     "ij",
@@ -787,6 +802,7 @@ __all__ = [
     "assume_in_range",
     "block_local",
     "cache_read_only",
+    "dump_compile_config",
     "init",
     "mesh_local",
     "no_activate",
--- a/tests/python/test_api.py
+++ b/tests/python/test_api.py
@@ -120,6 +120,7 @@ user_api[ti] = [
     "dataclass",
     "deactivate",
     "deactivate_all_snodes",
+    "dump_compile_config",
     "eig",
     "exp",
     "experimental",


gstaichi/compilation_manager/kernel_compilation_manager.cpp
```diff
@@ -122,12 +122,13 @@ void KernelCompilationManager::dump() {
...
     }
   }
+
   // Add new data
   for (auto &[kernel_key, kernel] : caching_kernels_) {
     if (kernel.metadata.cache_mode == CacheData::MemAndDiskCache) {
       auto [iter, ok] =
           dataWrapperByCacheKey.insert({kernel_key, std::move(kernel)});
-      TI_ASSERT(!ok || iter->second.metadata.size == 0);
+      TI_ASSERT(ok);
```
hughperkins, line 131: I have no idea why it was asserting "not ok" before 🤔

duburcqa: Why are you not checking size anymore ?
```cpp
TI_ASSERT(ok && iter->second.metadata.size != 0);
```

gstaichi/compilation_manager/kernel_compilation_manager.h
```diff
@@ -79,17 +79,18 @@ class KernelCompilationManager final {
...
                                             const std::string &kernel_name,
                                             const CompileConfig &compile_config,
                                             const DeviceCapabilityConfig &caps);
 
+  // ONLY public for testing
+  CompiledKernelData &cache_kernel(
+      const std::string &kernel_key,
+      const CompileConfig &compile_config,
+      std::unique_ptr<CompiledKernelData> compiled_kernel_data,
+      const Kernel &kernel_def);
```
duburcqa: This is usually discouraged. Two options: 
* make the class non-final and this method protected so that you can defined a derived testing class
* Forward-declare a testing class as friend as this one, so that you can excess whatever you want for testing without breaking the public API.

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -63,42 +67,48 @@ def store(
 
     Update! We are now going to store parameter pruning infomation, which is:
     - used_py_dataclass_leaves: set[str]
+
+    Update 2: we are going to store the cache key used by the c++ kernel cache, so that we can use that
+    to retrieve the immutable cached c++ kernel later, rather than, before, we were storing the c++
+    cached kernel using the fast cache key, leading to bugs, when cached kernel file then had to be mutable.
```
duburcqa, line 73: Note a big fan of update enumeration. This is a docstring not a change log. What matters is the current behaviour, not the history.

gstaichi/compilation_manager/kernel_compilation_manager.cpp
hughperkins: it was an "or" not an "and".

gstaichi/compilation_manager/kernel_compilation_manager.h
hughperkins: I tried the latter earlier, but means adding include gtest/gtest_prod all over the place, which became messy I felt. Will look into the first option.

hughperkins: maybe I'll take another stab at the second approcah though.

hughperkins: addressed in 0d0b63e16

duburcqa (approved):

gstaichi/compilation_manager/kernel_compilation_manager.cpp
duburcqa: Yes, not the entire condition was toggled I think.

duburcqa: It should be `TI_ASSERT(!(!ok || iter->second.metadata.size == 0));`

hughperkins: That seems complex. What are your intentions behind this expression?

Remidner that we have agreed that we want to prevent mutation. That means taht `ok` should always be true.

hughperkins: (related question: given the tests ae currently passing, what case(s) do you feel we are missing in our test that aer not causing the current expression to fail in the way that you might expect it to?)

duburcqa: > What are your intentions behind this expression?

I have zero intention, it is just a strict minimal fix of the original assert. I don't even know what it does.

duburcqa: I you want to change this, ideally you should convince me that the condition about `iter->second.metadata.size` can be safely removed. Is this related to being immutable now?

hughperkins: metadata size is set to 0 when we first add soemthing. It is populated during dump. There is no obvious reason to me why we should be checking for at this point in the code, where we are simply:
- chekcing if somethign is memory only, or should be dumped to file
- if the latter, then moving it into the list of things we will dump to file

If we want to check it, it would make more sense to check for it I feel either or both:
- in `cache_kernel`. function (which simply direclty sets it to 0 anywya https://github.com/Genesis-Embodied-AI/gstaichi/blob/6a2e5365115cddeeeec17a0a2dec7759114e5a58/gstaichi/compilation_manager/kernel_compilation_manager.cpp#L276)
- or just before populating it https://github.com/Genesis-Embodied-AI/gstaichi/blob/6a2e5365115cddeeeec17a0a2dec7759114e5a58/gstaichi/compilation_manager/kernel_compilation_manager.cpp#L145

duburcqa: Great! This argument is much better than "I have no idea why it was asserting "not ok" before" without further details x) Thank you

duburcqa: No opinion whether it should be checked or not at a more appropriate place. You know better!

PR #289, opened 2025-11-24
[Perf] Fastcache key contains gstaichi version

hughperkins:
Issue: #

### Brief Summary

- Fastcache key contains gstaichi version
- Unfortunately, I can't think of any obvious way to test that this does what we want
     - I constructed a test that used a second venv, into which I installed gstaichi 4.0.0
     - and then checked that the cache key was different
     - but it was
     - but that was before I had added the gstaichi vesrion to the cache key...
     - it was different simply as a result of the previous pr, for dupe kernels

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_fast_caching/src_hasher.py
+++ b/python/gstaichi/lang/_fast_caching/src_hasher.py
@@ -15,8 +15,6 @@ from .fast_caching_types import HashedFunctionSourceInfo
 from .hash_utils import hash_iterable_strings
 from .python_side_cache import PythonSideCache
 
-GSTAICHI_VERSION_STR: str = gstaichi.__version_str__
-
 
 def create_cache_key(
     raise_on_templated_floats: bool,
@@ -44,7 +42,7 @@ def create_cache_key(
     config_hash = config_hasher.hash_compile_config()
     cache_key = hash_iterable_strings(
         (
-            GSTAICHI_VERSION_STR,
+            gstaichi.__version_str__,
             kernel_hash,
             args_hash,
             config_hash,


python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -39,8 +40,10 @@ def create_cache_key(
         return None
     kernel_hash = function_hasher.hash_kernel(kernel_source_info)
     config_hash = config_hasher.hash_compile_config()
+    v = gstaichi.__version__
     cache_key = hash_iterable_strings(
         (
+            f"{v[0]}.{v[1]}.{v[2]}",
```
duburcqa: This seems to be called frequently? I would rather define it once and for all as a global variable in upper case on top of this model. By the way, why not just do `".".join(gstaichi.__version__)` ?

hughperkins: for the second question, I tried that initialy too, but the elements are ints.

hughperkins: for the former, addressed in b7a54f65a

duburcqa: I see. What about: `".".join(map(str, gstaichi.__version__))`. But I guess it may be magic for some.

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -14,6 +15,8 @@
 from .hash_utils import hash_iterable_strings
 from .python_side_cache import PythonSideCache
 
+GSTAICHI_VERSION_STR: str | None = None
```
duburcqa: Why not just initialise this here right away? It is just a basic string manipulation, no big deal.

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -28,6 +31,7 @@ def create_cache_key(
     - kernel function (but not sub functions)
     - compilation config (which includes arch, and debug)
     """
+    global GSTAICHI_VERSION_STR
```
duburcqa: I tend to avoid `global ...` when I'm dealing with module constant. But as you want.

hughperkins: i tried without, but it wouldnt allow the assignment

hughperkins: well, it wouldnt allow the 'if' I guess. said 'unbound'

python/gstaichi/lang/_fast_caching/src_hasher.py
hughperkins: I tried that, but gives circular reference 
```
python/gstaichi/_snode/snode_tree.py:8: in <module>
    from gstaichi.lang import impl
python/gstaichi/lang/__init__.py:3: in <module>
    from gstaichi.lang import impl, simt  # noqa: F401
python/gstaichi/lang/impl.py:28: in <module>
    from gstaichi.lang.kernel_impl import BoundGsTaichiCallable, GsTaichiCallable, Kernel
python/gstaichi/lang/kernel_impl.py:44: in <module>
    from gstaichi.lang._fast_caching import src_hasher
python/gstaichi/lang/_fast_caching/src_hasher.py:19: in <module>
    v = gstaichi.__version__
E   AttributeError: partially initialized module 'gstaichi' has no attribute '__version__' (most likely due to a circular import)
```

python/gstaichi/lang/_fast_caching/src_hasher.py
hughperkins: addressed in [2c96032](https://github.com/Genesis-Embodied-AI/gstaichi/pull/289/commits/2c96032bfc8d7b6f6d5e2a9925caacebbe102540)

python/gstaichi/lang/_fast_caching/src_hasher.py
duburcqa: I see. `__version__` attribute should be defined right away. In your case, it is the very last attribute being defined in the your init file. I would suggest to move it on top.

python/gstaichi/lang/_fast_caching/src_hasher.py
duburcqa: Yeas, I think it is because you are setting it inside this function. This issue should vanish once it would be done outside.

python/gstaichi/lang/_fast_caching/src_hasher.py
duburcqa: https://github.com/Genesis-Embodied-AI/gstaichi/blob/7db9a9bf70f5894219c22889067169efa54afde3/python/gstaichi/__init__.py#L31

hughperkins: addressed in 2d8e17ddf

python/gstaichi/__init__.py
```diff
@@ -28,12 +35,6 @@ def __getattr__(attr):
     raise AttributeError(f"module '{__name__}' has no attribute '{attr}'")
 
 
-__version__ = (
-    _ti_core.get_version_major(),
-    _ti_core.get_version_minor(),
-    _ti_core.get_version_patch(),
-)
-
 del _ti_core
```
duburcqa: What is this? I would recommend deleting this line. Never seen this pattern in my entire life.

python/gstaichi/lang/_fast_caching/src_hasher.py
```diff
@@ -14,6 +15,8 @@
 from .hash_utils import hash_iterable_strings
 from .python_side_cache import PythonSideCache
 
+GSTAICHI_VERSION_STR: str = gstaichi.__version_str__
```
duburcqa: Now that you have a precomputed string variant, you could avoid defining this global variable and just use `gstaichi.__version_str__` directly x)

python/gstaichi/__init__.py
hughperkins: deleted

python/gstaichi/lang/_fast_caching/src_hasher.py
hughperkins: done

PR #293, opened 2025-11-25
[Type] Fix U1 on Metal and Vulkan

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_types.py
+++ b/tests/python/test_types.py
@@ -1,3 +1,4 @@
+import numpy as np
 import pytest
 
 import gstaichi as ti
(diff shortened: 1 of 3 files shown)


tests/python/test_types.py
```diff
@@ -163,3 +163,23 @@ def run():
...
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_types_fields_and_dtypes(tensor_type, dtype) -> None:
+    """
+    Check u1 works. Used to be broken for metal fields.
+    """
+    assert ti.cfg is not None
+    arch = ti.cfg.arch.name
+    if tensor_type == ti.field and dtype == ti.u1 and arch == "vulkan":
+        pytest.xfail("u1 field vulkan doesn't work currently")
```
duburcqa: So that could be the reason why vulkan is broken for Genesis ?!

hughperkins: Do you mean, you feel better to look into fixing this?

hughperkins: cursor fixed this. but then noticed that to_numpy is broken. it's looking into that :)

hughperkins: Update: to numpy fixed for field. gettting cursor to check for ndarray (which is also broken)

(mgiht split out to numpy into a separate PR, we'll see)

duburcqa: > Do you mean, you feel better to look into fixing this?

At least I may have workaround of Vulkan failing in Genesis.

hughperkins: Note that to_numpy is also broken, for obth ndarray and field, curerntly.

duburcqa: You mean for u1 only? Because we never observed any issue so far and we are relying on to_numpy extensively.

hughperkins: yeah I think.

seems like cursor is really struggling with ndarray to numpy for some reason.  Imight have to. do some actual work myself 😂 

I think in the mantime, I might add the vulkan read element fix into this PR, and then I'll put to numpy in a separate pr.

hughperkins: Ok, added fix for vulkan get element.

gstaichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2095,10 +2098,16 @@ spirv::Value TaskCodegen::load_buffer(const Stmt *ptr, DataType dt) {
   if (ptr_val.stype.dt == PrimitiveType::u64) {
     ti_buffer_type = dt;
   } else if (dt->is_primitive(PrimitiveTypeID::u1)) {
+    // For u1, Metal stores values as i32 (4 bytes) in the buffer.
```
duburcqa: Metal and Vulkan

tests/python/test_types.py
```diff
@@ -163,3 +163,19 @@ def run():
...
     for f in fs:
         assert f == val
+
+
+@pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_types_fields_and_dtypes(tensor_type, dtype) -> None:
+    """
+    Check u1 works. Used to be broken for metal fields.
```
duburcqa, line 174: and vulkan

hughperkins: I'm not certain it used to be broken for vulkan. I never tested in the absence of the fix in this PR.

PR #296, opened 2025-11-26
[Test] Add tests for consistency between kernel accessors, external accessors, to/from numpy

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_tensor_consistency.py
+++ b/tests/python/test_tensor_consistency.py
@@ -21,7 +21,7 @@ _TI_DTYPE_TO_NP_DTYPE = {
 @pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
 @pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
 @test_utils.test()
-def test_types_fields_and_dtypes_kernel_write_to_numpy_consistency(tensor_type, dtype) -> None:
+def test_tensor_consistency_kernel_write_to_numpy_consistency(tensor_type, dtype) -> None:
     """
     write from kernel => to_numpy => check
     """
@@ -52,7 +52,7 @@ def test_types_fields_and_dtypes_kernel_write_to_numpy_consistency(tensor_type, 
 @pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
 @pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
 @test_utils.test()
-def test_types_fields_and_dtypes_kernel_from_numpy_to_numpy_consistency(tensor_type, dtype) -> None:
+def test_tensor_consistency_kernel_from_numpy_to_numpy_consistency(tensor_type, dtype) -> None:
     """
     write to numpy => from_numpy => to_numpy => check
     """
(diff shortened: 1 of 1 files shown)


tests/python/test_types.py
```diff
@@ -163,3 +164,195 @@ def run():
...
+    check consistency
+    """
+    assert ti.cfg is not None
+    arch = ti.cfg.arch
+    if dtype == ti.u1 and arch in [ti.vulkan, ti.metal]:
+        pytest.xfail("u1 doesnt work on vulkan or metal doesn't work currently, neither on field nor ndarray")
+
+    poses = [0, 2, 5, 11]
+
+    ti_dtype_to_np_dtype = {
```
erizmr: Just one nit: shall we move this mapping outside functions? It looks we have one duplicate in each function

hughperkins: addressed in 5d937ce5b

hughperkins: (good spot :) )

tests/python/test_tensor_consistency.py
```diff
@@ -0,0 +1,323 @@
...
+    @ti.kernel
+    def k1(a: TensorType) -> None:
+        for b_ in range(1):
+            for pos in ti.static(poses):
+                a[pos] = 1
+
+    k1(a)
+
+    a_np = a.to_numpy()
+
```
duburcqa, line 47: I would avoid spreading this in too many lines, but as you want.

tests/python/test_tensor_consistency.py
```diff
@@ -0,0 +1,323 @@
...
+    for pos in poses:
+        a_np[pos] = 1
+
+    a = tensor_type(dtype, (16,))
+    a.from_numpy(a_np)
+
+    b_np = a.to_numpy()
+
+    for i in range(16):
```
duburcqa, line 78: Same. Basically I don't like having to scroll too much for understanding what is happening in some code. But just a personal taste I guess.

tests/python/test_tensor_consistency.py
```diff
@@ -0,0 +1,323 @@
...
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_tensor_consistency_kernel_write_to_numpy_consistency(tensor_type, dtype) -> None:
+    """
+    write from kernel => to_numpy => check
+    """
+    assert ti.cfg is not None
+    arch = ti.cfg.arch
+    if dtype == ti.u1 and arch in [ti.vulkan, ti.metal]:
+        pytest.xfail("u1 doesnt work on vulkan or metal doesn't work currently, neither on field nor ndarray")
```
duburcqa, line 31: Maybe you could add some auto-fixture that could run this check for all tests in this file?

tests/python/test_tensor_consistency.py
hughperkins: yeah, I prefer to separate out the different concepts, ie
- kernel call
- copy to host
- check

tests/python/test_tensor_consistency.py
hughperkins: Was tempted to do this, but might be removing these shortly anyway... will see the results of the tests in https://github.com/Genesis-Embodied-AI/gstaichi/pull/299

hughperkins: Yeah, they are removed in https://github.com/Genesis-Embodied-AI/gstaichi/pull/302 now

tests/python/test_tensor_consistency.py
hughperkins: Same answer as above: I like to separate out the different concepts, ideally.

PR #298, opened 2025-11-27
[Bug] Fix dlpack zero-copy memory alignment

erizmr:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/gstaichi/python/dlpack_funcs.cpp
+++ b/gstaichi/python/dlpack_funcs.cpp
@@ -98,11 +98,6 @@ std::pair<uint8_t, uint8_t> get_type_info(Arch arch, DataType dt) {
       break;
     }
     case PrimitiveTypeID::u1: {
-#if TI_WITH_METAL
-      if (arch_is_metal(arch)) {
-        TI_ERROR("DLPack conversion for bool type is not supported on Metal");
-      }
-#endif
       data_type_code = static_cast<uint8_t>(kDLBool);
       element_bits = 8;
       break;
--- a/tests/python/gstaichi/lang/test_dlpack.py
+++ b/tests/python/gstaichi/lang/test_dlpack.py
@@ -42,11 +42,6 @@ def test_dlpack_types(tensor_type, dtype, shape: tuple[int], poses: list[tuple[i
     ti_tensor = tensor_type(dtype, shape)
     for i, pos in enumerate(poses):
         ti_tensor[pos] = i * 10 + 10
-    if ti.cfg.arch == ti.metal and dtype is ti.u1:
-        # TODO: fix u1 on Metal
-        with pytest.raises(RuntimeError):
-            dlpack = ti_tensor.to_dlpack()
-        pytest.xfail(reason="ti.u1 doesn't work on Metal currently")
     ti.sync()
     dlpack = ti_tensor.to_dlpack()
     tt = torch.utils.dlpack.from_dlpack(dlpack)
--- a/gstaichi/codegen/spirv/spirv_codegen.cpp
+++ b/gstaichi/codegen/spirv/spirv_codegen.cpp
@@ -727,11 +727,11 @@ void TaskCodegen::visit(ExternalPtrStmt *stmt) {
       linear_offset = ir_->mul(linear_offset, size_var);
       linear_offset = ir_->add(linear_offset, indices);
     }
+    size_t type_size =
+        ir_->get_primitive_type_size(stmt->ret_type.ptr_removed());
     linear_offset = ir_->make_value(
         spv::OpShiftLeftLogical, ir_->i32_type(), linear_offset,
-        ir_->int_immediate_number(ir_->i32_type(),
-                                  log2int(ir_->get_primitive_type_size(
-                                      stmt->ret_type.ptr_removed()))));
+        ir_->int_immediate_number(ir_->i32_type(), log2int(type_size)));
     if (caps_->get(DeviceCapability::spirv_has_no_integer_wrap_decoration)) {
       ir_->decorate(spv::OpDecorate, linear_offset,
                     spv::DecorationNoSignedWrap);
@@ -2095,7 +2095,7 @@ spirv::Value TaskCodegen::load_buffer(const Stmt *ptr, DataType dt) {
   if (ptr_val.stype.dt == PrimitiveType::u64) {
     ti_buffer_type = dt;
   } else if (dt->is_primitive(PrimitiveTypeID::u1)) {
-    ti_buffer_type = PrimitiveType::i32;
+    ti_buffer_type = PrimitiveType::u8;
   }
 
   auto buf_ptr = at_buffer(ptr, ti_buffer_type);
@@ -2117,9 +2117,7 @@ void TaskCodegen::store_buffer(const Stmt *ptr, spirv::Value val) {
   if (ptr_val.stype.dt == PrimitiveType::u64) {
     ti_buffer_type = val.stype.dt;
   } else if (val.stype.dt->is_primitive(PrimitiveTypeID::u1)) {
-    ti_buffer_type = PrimitiveType::i32;
-    val = ir_->make_value(spv::OpSelect, ir_->i32_type(), val,
-                          ir_->const_i32_one_, ir_->const_i32_zero_);
+    ti_buffer_type = PrimitiveType::i8;
   }
 
   auto buf_ptr = at_buffer(ptr, ti_buffer_type);
--- a/tests/python/test_tensor_consistency.py
+++ b/tests/python/test_tensor_consistency.py
@@ -1,0 +1,279 @@
+import numpy as np
+import pytest
+
+import gstaichi as ti
+
+from tests import test_utils
+
+_TI_DTYPE_TO_NP_DTYPE = {
+    ti.u1: np.bool_,
+    ti.u8: np.uint8,
+    ti.u16: np.uint16,
+    ti.u32: np.uint32,
+    ti.u64: np.uint64,
+    ti.i8: np.int8,
+    ti.i16: np.int16,
+    ti.i32: np.int32,
+    ti.i64: np.int64,
+}
+
+
+@pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_tensor_consistency_kernel_write_to_numpy_consistency(tensor_type, dtype) -> None:
+    """
+    write from kernel => to_numpy => check
+    """
+    poses = [0, 2, 5, 11]
+    a = tensor_type(dtype, (16,))
+
+    TensorType = ti.types.NDArray if tensor_type == ti.ndarray else ti.Template
+
+    @ti.kernel
+    def k1(a: TensorType) -> None:
+        for b_ in range(1):
+            for pos in ti.static(poses):
+                a[pos] = 1
+
+    k1(a)
+
+    a_np = a.to_numpy()
+
+    for i in range(16):
+        assert a_np[i] == (1 if i in poses else 0)
+
+
+@pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_tensor_consistency_kernel_from_numpy_to_numpy_consistency(tensor_type, dtype) -> None:
+    """
+    write to numpy => from_numpy => to_numpy => check
+    """
+    poses = [0, 2, 5, 11]
+
+    np_dtype = _TI_DTYPE_TO_NP_DTYPE[dtype]
+
+    a_np = np.zeros(dtype=np_dtype, shape=(16,))
+
+    for pos in poses:
+        a_np[pos] = 1
+
+    a = tensor_type(dtype, (16,))
+    a.from_numpy(a_np)
+
+    b_np = a.to_numpy()
+
+    for i in range(16):
+        assert b_np[i] == (1 if i in poses else 0)
+
+
+@pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_tensor_consistency_np_read_write_np_consistency(tensor_type, dtype) -> None:
+    """
+    write to numpy => read from kernel => write from kernel => numpy => check
+    check consistency
+    """
+    poses = [0, 2, 5, 11]
+
+    np_dtype = _TI_DTYPE_TO_NP_DTYPE[dtype]
+
+    a_np = np.zeros(dtype=np_dtype, shape=(16,))
+    a = tensor_type(dtype, (16,))
+    b = tensor_type(dtype, (16,))
+
+    for pos in poses:
+        a_np[pos] = 1
+    a.from_numpy(a_np)
+
+    TensorType = ti.types.NDArray if tensor_type == ti.ndarray else ti.Template
+
+    @ti.kernel
+    def k1(a: TensorType, b: TensorType) -> None:
+        for b_ in range(1):
+            for pos in ti.static(poses):
+                b[pos] = a[pos]
+
+    k1(a, b)
+
+    b_np = b.to_numpy()
+
+    for i in range(16):
+        assert b_np[i] == (1 if i in poses else 0)
+
+
+@pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_tensor_consistency_from_numpy_accessor_read_consistency(tensor_type, dtype) -> None:
+    """
+    write to numpy => from_numpy => accessor read => check
+    check consistency
+    """
+    poses = [0, 2, 5, 11]
+
+    np_dtype = _TI_DTYPE_TO_NP_DTYPE[dtype]
+    a_np = np.zeros(dtype=np_dtype, shape=(16,))
+    a = tensor_type(dtype, (16,))
+
+    for pos in poses:
+        a_np[pos] = 1
+    a.from_numpy(a_np)
+
+    for i in range(16):
+        assert a[i] == (1 if i in poses else 0)
+
+
+@pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_tensor_consistency_accessor_write_to_numpy_consistency(tensor_type, dtype) -> None:
+    """
+    accessor write => to_numpy => check
+    """
+    poses = [0, 2, 5, 11]
+
+    a = tensor_type(dtype, (16,))
+    for pos in poses:
+        a[pos] = 1
+
+    a_np = a.to_numpy()
+
+    for i in range(16):
+        assert a_np[i] == (1 if i in poses else 0)
+
+
+@pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@pytest.mark.parametrize("std_dtype", [ti.i8, ti.i32])
+@test_utils.test()
+def test_tensor_consistency_from_numpy_kern_read(tensor_type, dtype, std_dtype) -> None:
+    """
+    write numpy => from_numpy => kernel read => kernel write to standard type => to_numpy => check
+    """
+    poses = [0, 2, 5, 11]
+    N = 16
+
+    np_dtype = _TI_DTYPE_TO_NP_DTYPE[dtype]
+    a_np = np.zeros(dtype=np_dtype, shape=(N,))
+    a = tensor_type(dtype, (N,))
+    b = tensor_type(std_dtype, (N,))
+
+    for pos in poses:
+        a_np[pos] = 1
+    a.from_numpy(a_np)
+
+    TensorType = ti.types.NDArray if tensor_type == ti.ndarray else ti.Template
+
+    @ti.kernel
+    def k1(a: TensorType, b: TensorType) -> None:
+        for b_ in range(1):
+            for i in range(N):
+                b[i] = a[i]
+
+    k1(a, b)
+
+    b_np = b.to_numpy()
+
+    for i in range(N):
+        assert b_np[i] == (1 if i in poses else 0)
+
+
+@pytest.mark.parametrize("tensor_type", [ti.field, ti.ndarray])
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@pytest.mark.parametrize("std_dtype", [ti.i8, ti.i32])
+@test_utils.test()
+def test_tensor_consistency_kern_write_to_numpy(tensor_type, dtype, std_dtype) -> None:
+    """
+    write to std type numpy => from_numpy => std type kernel read => kernel write => to_numpy => check
+    """
+    poses = [0, 2, 5, 11]
+    N = 16
+
+    np_dtype = _TI_DTYPE_TO_NP_DTYPE[dtype]
+    a_np = np.zeros(dtype=np_dtype, shape=(N,))
+    a = tensor_type(std_dtype, (N,))
+    b = tensor_type(dtype, (N,))
+
+    for pos in poses:
+        a_np[pos] = 1
+    a.from_numpy(a_np)
+
+    TensorType = ti.types.NDArray if tensor_type == ti.ndarray else ti.Template
+
+    @ti.kernel
+    def k1(a: TensorType, b: TensorType) -> None:
+        for b_ in range(1):
+            for i in range(N):
+                b[i] = a[i]
+
+    k1(a, b)
+
+    b_np = b.to_numpy()
+
+    for i in range(N):
+        assert b_np[i] == (1 if i in poses else 0)
+
+
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_tensor_consistency_ext_to_kern(dtype) -> None:
+    """
+    write to numpy => pass directly to kernel => test in kern
+    """
+    poses_l = [0, 2, 5, 11]
+    N = 16
+
+    np_dtype = _TI_DTYPE_TO_NP_DTYPE[dtype]
+    a_np = np.zeros(dtype=np_dtype, shape=(N,))
+    anti_poses_l = list(range(N))  # i.e positions with zeros
+    for pos in poses_l:
+        a_np[pos] = 1
+        anti_poses_l.remove(pos)
+
+    result = ti.ndarray(ti.i32, ())
+    result[()] = 1
+
+    @ti.kernel
+    def k1(a: ti.types.NDArray, result: ti.types.NDArray) -> None:
+        for b_ in range(1):
+            for pos in ti.static(poses_l):
+                if a[pos] != 1:
+                    result[()] = 0
+            for pos in ti.static(anti_poses_l):
+                if a[pos] != 0:
+                    result[()] = 0
+
+    k1(a_np, result)
+    ti.sync()
+    assert result[()] == 1
+
+
+@pytest.mark.parametrize("dtype", [ti.u1, ti.u8, ti.u16, ti.u32, ti.u64, ti.i8, ti.i32, ti.i16, ti.i64])
+@test_utils.test()
+def test_tensor_consistency_kern_to_ext(dtype) -> None:
+    """
+    write directly to numpy array in kernel => check in numpy
+    """
+
+    poses_l = [0, 2, 5, 11]
+    N = 16
+
+    np_dtype = _TI_DTYPE_TO_NP_DTYPE[dtype]
+    a_np = np.zeros(dtype=np_dtype, shape=(N,))
+
+    @ti.kernel
+    def k1(a: ti.types.NDArray) -> None:
+        for b_ in range(1):
+            for pos in ti.static(poses_l):
+                a[pos] = 1
+
+    k1(a_np)
+
+    ti.sync()
+
+    for i in range(N):
+        assert a_np[i] == (1 if i in poses_l else 0)


gstaichi/python/dlpack_funcs.cpp
```diff
@@ -164,11 +164,19 @@ pybind11::capsule field_to_dlpack(Program *program,
...
 
   void *raw_ptr = nullptr;
   DLDeviceType device_type = DLDeviceType::kDLCPU;
   std::tie(raw_ptr, device_type) = get_raw_ptr(arch, program, tree_device_ptr);
+
+  void *raw_ptr_check = nullptr;
+  for (int i = 0; i < field_in_tree_offset * 3; i++) {
+    raw_ptr_check = (void *)((uint64_t)raw_ptr + i);
+    int64_t *as_f64_check = reinterpret_cast<int64_t *>(raw_ptr_check);
+    TI_INFO("  offset {} raw as f64: {} {} {}", i, as_f64_check[0], as_f64_check[1], as_f64_check[2]);
```
hughperkins: can we remove the TI_INFOs throughout the PR please?
- I think they make the PR harder to read
- they are only needed whilst working on this feature I think?

gstaichi/python/dlpack_funcs.cpp
```diff
@@ -164,11 +164,19 @@ pybind11::capsule field_to_dlpack(Program *program,
...
 
-  int field_in_tree_offset = program->get_field_in_tree_offset(tree_id, snode);
+  int field_in_tree_offset = program->get_field_in_tree_offset(tree_id, snode, /*is_memory_aligned=*/true);
 
   void *raw_ptr = nullptr;
   DLDeviceType device_type = DLDeviceType::kDLCPU;
   std::tie(raw_ptr, device_type) = get_raw_ptr(arch, program, tree_device_ptr);
+
+  void *raw_ptr_check = nullptr;
+  for (int i = 0; i < field_in_tree_offset * 3; i++) {
```
hughperkins: I think this whole loop is for dev/debug right? Could we remove it please?

(what I usually do is save this stuff to a separate branch, so I still have it for dev/debug, but keeps the PR clean).

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -182,7 +182,7 @@ class LlvmProgramImpl : public ProgramImpl {
     runtime_exec_->check_runtime_error(result_buffer);
   }
 
-  size_t get_field_in_tree_offset(int tree_id, const SNode *child) override {
+  size_t get_field_in_tree_offset(int tree_id, const SNode *child, bool is_memory_aligned=false) override {
```
hughperkins: could you describe the goal of `is_memory_aligned` parameter please
- when would we set it to true?
- when would we set it to false?
- why wouldnt we always set it to the same value? (and thus could remove it entirely)

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -182,7 +182,7 @@ class LlvmProgramImpl : public ProgramImpl {
...
     // FIXME: Compute the proper offset. Current method taken from GGUI code
```
hughperkins: is this comment still valid? can it be removed?

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -191,11 +191,29 @@ class LlvmProgramImpl : public ProgramImpl {
 
     int child_id = root->child_id(dense_parent);
 
+    // The maximum cell size bytes is used to handle memory alignment for dlpack usage
+    // When compute field offset, SNode doesn't take the alignment into account yet, we need to mannully handle it here
```
hughperkins: I think we can remove word "manully", since it's handled by code, right?

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -191,11 +191,29 @@ class LlvmProgramImpl : public ProgramImpl {
...
       SNode *child = root->ch[i].get();
-      offset += child->cell_size_bytes * child->num_cells_per_container;
+      size_t child_cell_size_bytes = child->cell_size_bytes;
+      if (is_memory_aligned) {
+        max_cell_size_bytes = child_cell_size_bytes > max_cell_size_bytes
+                                ? child_cell_size_bytes
+                                : max_cell_size_bytes;
+      }
+      offset += child_cell_size_bytes * child->num_cells_per_container;
+      TI_INFO("child Snode name {} child cell size bytes: {} max_cell_size_bytes {} child num cells per container {}", child->get_name(), child_cell_size_bytes, max_cell_size_bytes, child->num_cells_per_container);
```
hughperkins: remove TI_INFO

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -191,11 +191,29 @@ class LlvmProgramImpl : public ProgramImpl {
...
     }
 
+    if (is_memory_aligned) {
+      auto align_up = [](size_t x, size_t align) {
+        return (x + align - 1) / align * align;
+      };
+      offset = align_up(offset, max_cell_size_bytes);
+    }
+    TI_INFO("Root name {} Parent Snode name {} max_cell_size_bytes {} Computed aligned field offset: {}", root->get_name(), dense_parent->get_name(), max_cell_size_bytes, offset);
```
hughperkins: remove TI_INFO

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -191,11 +191,29 @@ class LlvmProgramImpl : public ProgramImpl {
...
+        max_cell_size_bytes = child_cell_size_bytes > max_cell_size_bytes
+                                ? child_cell_size_bytes
+                                : max_cell_size_bytes;
+      }
+      offset += child_cell_size_bytes * child->num_cells_per_container;
+      TI_INFO("child Snode name {} child cell size bytes: {} max_cell_size_bytes {} child num cells per container {}", child->get_name(), child_cell_size_bytes, max_cell_size_bytes, child->num_cells_per_container);
     }
 
+    if (is_memory_aligned) {
+      auto align_up = [](size_t x, size_t align) {
```
hughperkins: Gentle preference to factorize this anonymous function out to a separate function. @duburcqa what are your thoughts on this?

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -191,11 +191,29 @@ class LlvmProgramImpl : public ProgramImpl {
...
+      if (is_memory_aligned) {
+        max_cell_size_bytes = child_cell_size_bytes > max_cell_size_bytes
+                                ? child_cell_size_bytes
+                                : max_cell_size_bytes;
+      }
+      offset += child_cell_size_bytes * child->num_cells_per_container;
+      TI_INFO("child Snode name {} child cell size bytes: {} max_cell_size_bytes {} child num cells per container {}", child->get_name(), child_cell_size_bytes, max_cell_size_bytes, child->num_cells_per_container);
     }
 
+    if (is_memory_aligned) {
```
hughperkins: if it's already memory aligned, why would we align it? Perhaps rename the parameter to `align_memory`, in imperative?

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -191,11 +191,29 @@ class LlvmProgramImpl : public ProgramImpl {
 
     int child_id = root->child_id(dense_parent);
 
+    // The maximum cell size bytes is used to handle memory alignment for dlpack usage
+    // When compute field offset, SNode doesn't take the alignment into account yet, we need to mannully handle it here
+    size_t max_cell_size_bytes = root->ch[child_id].get()->cell_size_bytes;
     for (int i = 0; i < child_id; ++i) {
       SNode *child = root->ch[i].get();
-      offset += child->cell_size_bytes * child->num_cells_per_container;
+      size_t child_cell_size_bytes = child->cell_size_bytes;
```
hughperkins: I wonder if it's worth factorizing out the work at each iteration into a separate function? @duburcqa any opinions/thougths on this?

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -189,3 +189,80 @@ def test_dlpack_field_multiple_tree_nodes():
     assert c_t[0] == 333
     assert d_t[0] == 444
     assert e_t[0] == 555
+
+
+@test_utils.test(arch=dlpack_arch)
+@pytest.mark.parametrize("tensor_type", [ti.field])
+@pytest.mark.parametrize("dtype", [ti.i32, ti.i64, ti.f32, ti.f64, ti.u1, ti.i8])
+@pytest.mark.parametrize("shape", [3, 1, 4, 5, 7, 2])
+def test_dlpack_mixed_types_field(tensor_type, dtype, shape: tuple[int]) -> None:
```
hughperkins: what do you mean by 'mixed type'? It looks like this is only creating a single tensor, of type vector?

When I think "mixed type", I think like eg https://github.com/Genesis-Embodied-AI/gstaichi/blob/e43faa4a3605dd709ccad65aaffd7a87c03c7398/tests/python/test_types.py#L377-L378 , which is reading from one type, and writing to another, for example.

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -189,3 +189,80 @@ def test_dlpack_field_multiple_tree_nodes():
...
+        torch.utils.dlpack.from_dlpack(pos.to_dlpack()).cpu().numpy(), 
+        pos.to_numpy(),
+    )
+
+
+@test_utils.test(arch=dlpack_arch)
+@pytest.mark.parametrize("tensor_type", [ti.field])
+@pytest.mark.parametrize("dtypes", [(ti.i32, ti.i64, ti.f32, ti.f64, ti.u1, ti.i8)])
+@pytest.mark.parametrize("shapes", [(3, 1, 4, 5, 7, 2)])
+def test_dlpack_multiple_mixed_types_field(tensor_type, dtypes, shapes: tuple[int]) -> None:
```
hughperkins: Personally I feel this would be easier for me to read if the types and shapes were pre-'zipped' in the parametrize, eg 

```
@pytest.mark.parametrize("tensor_geometries", [
    ([(ti.i32, (3,)), (ti.i64, (1,)), (ti.f32, (4,)), (ti.f64, (5,)), (ti.u1, (7,)), (ti.i8, (2,)))
]
)
```
I may have understood how this is being used:
- it looks like there is just one tuple of dtypes, that is always used, rather than multiple tuples of dtypes, is this a fair impression?
- similarly, it looks like there is just one tuple of shapes, that is always being used, is this a fair impression?
    - it looks like all 'shapes' are 1-dimensional, and each number represents the size of the 0th dimension in a 1-dimensional shape, is this a fair impression?
    - I feel like this might be easier to see clearly if 'pre-zipped' as I suggest
@duburcqa not sure what your thoughts are on this?

hughperkins: (I actullly have a gentle preference to remove the parametrize's in this case altogether, since the parametrize is not actually being using to run the function multiple times right? I might have misunderstood, but it slooks like 'shapes' and 'dtypes' only contain a single item each, from parametrize point of view. is tis a fair impression? Not sure if @duburcqa you have a pov on use of paraemtrize with a single value?)

hughperkins: (Ditto for `tensor_type` I think? ie also only contains a single value, so could simply be moved directly into the function potentailly, depending on what @duburcqa 's thoughts are on this.)

duburcqa: A compromise could be:
```python
@pytest.mark.parametrize("dtypes, shapes", 
    [(ti.i32, (3,)), (ti.i64, (1,)), (ti.f32, (4,)), (ti.f64, (5,)), (ti.u1, (7,)), (ti.i8, (2,))]
)
```

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -221,3 +189,80 @@ def test_dlpack_field_multiple_tree_nodes():
...
+    np.testing.assert_allclose(
+        torch.utils.dlpack.from_dlpack(pos.to_dlpack()).cpu().numpy(), 
+        pos.to_numpy(),
+    )
+
+
+@test_utils.test(arch=dlpack_arch)
+def test_dlpack_joints_case() -> None:
+    import numpy as np
+
```
duburcqa: Why importing numpy inside the test itself? taichi already requires numpy AFAIK.

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -193,26 +193,31 @@ class LlvmProgramImpl : public ProgramImpl {
 
     // The maximum cell size bytes is used to handle memory alignment for dlpack usage
     // When compute field offset, SNode doesn't take the alignment into account yet, we need to mannully handle it here
-    size_t max_cell_size_bytes = root->ch[child_id].get()->cell_size_bytes;
+    
+    // The second ->ch is to get the actual data continer
+    DataType dt = root->ch[child_id].get()->ch[0]->dt;
+    size_t max_cell_size_bytes = data_type_size(dt.get_element_type());
+    TI_DEBUG("data type {} data type size {}", dt.get_element_type().to_string(), data_type_size(dt.get_element_type()));
```
hughperkins: Let's remove TI_DEBUG too pelase :) They are only useful whilts working on this PR I feel?

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -191,11 +191,34 @@ class LlvmProgramImpl : public ProgramImpl {
...
+
     for (int i = 0; i < child_id; ++i) {
       SNode *child = root->ch[i].get();
+      if (is_memory_aligned) {
+        size_t child_cell_size_bytes = data_type_size(child->ch[0]->dt.get_element_type());
+        max_cell_size_bytes = child_cell_size_bytes > max_cell_size_bytes
+                                ? child_cell_size_bytes
+                                : max_cell_size_bytes;
+        TI_DEBUG("child Snode name {} child cell size bytes: {} max_cell_size_bytes {} child num cells per container {}, data type {}, data type size {}", child->get_name(), child_cell_size_bytes, max_cell_size_bytes, child->num_cells_per_container, child->ch[0]->dt.get_element_type().to_string(), child_cell_size_bytes);
```
hughperkins: remove TI_DEBUG

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -191,11 +191,34 @@ class LlvmProgramImpl : public ProgramImpl {
...
+      }
       offset += child->cell_size_bytes * child->num_cells_per_container;
     }
 
+    if (is_memory_aligned) {
+      auto align_up = [](size_t x, size_t align) {
+        return (x + align - 1) / align * align;
+      };
+      offset = align_up(offset, max_cell_size_bytes);
+      TI_DEBUG("Root name {} Parent Snode name {} max_cell_size_bytes {} Computed aligned field offset: {}", root->get_name(), dense_parent->get_name(), max_cell_size_bytes, offset);
```
hughperkins: remove TI_DEBUG

gstaichi/python/dlpack_funcs.cpp
erizmr: yes, this is purely for debug use. It is removed for ready to review.

hughperkins: cool. could you move them to a separate branch please?

eg you can see that for my own prs I have:
- whatI PR https://github.com/Genesis-Embodied-AI/gstaichi/compare/main...hp/metal-u1
- what I use for dev/debug https://github.com/Genesis-Embodied-AI/gstaichi/compare/main...hp/metal-u1-instruementation

I keep the main PR branch clean, so:
- easy to review
- can be merged directly, after approval, with no further changes required

hughperkins: (you can use `cherry-pick` etc, to copy things from the dev branch to the PR branch)

hughperkins: (and/or Beyond Compare and simliar https://www.scootersoftware.com/ )

hughperkins: Oh, I see, this PR is in 'draft'. Should I ignore hte PR for now, and wait until 'ready to review'?

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: Here the motivation to use `is_memory_aligned` is to keep the current change behind this option, so it would not break other use cases. To my understanding, we only found this issue when using dlpack conversion right?

Regarding to the questions:
- Currently I only set it to True in the dlpack `field_to_dlpack`.
- For all other cases, I keep them all False by default.
- The current Snode tree offsets are not memory alignment awared. It would be a design choice/substantial work whether we need to make it memory alignment awared. I think we can enitrely remove it or make the memory alignment to True by default after we make the decision.

hughperkins: why would the memory alignment of fields vary across use cases? The memory alignment is fixed by the underlying field implementation itselef right? dlpack doesnt actually change the layout in any way I think? Just publishes what is already tehre I think?

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: I think this comment is related to the gfx snode tree offset implementation here: https://github.com/Genesis-Embodied-AI/gstaichi/blob/5428d9f5f8dfc2e8b5dd14bd5ac64bef15049624/gstaichi/runtime/gfx/snode_tree_manager.cpp#L31

This is a refactored version comparing to the llvm version, we haven't touch it yet. I think we can still keep the comment.

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: Moved it to debug in the ready to review version.

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: ditto

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: But doesnt your code now calculate the correct offset?

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: I think TI_DEBUG could be removed too please
- makes the PR harder to read
- is oly needed whilst developping this PR I think?

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: This `is_memory_aligned` indicates that LLVM did the alignment but the snode doesn't know it, i.e., it is still computing the offset without taking the padding into considration. Therefore we need to match it once it is memory aligned. But I am open to change it to `align_memory` if people think it is a better name

tests/python/gstaichi/lang/test_dlpack.py
erizmr: ah, sorry for the confusion made here, the mixed type here indicate on one SNode tree, we have different data types such as ti.f32 and ti.f64 which trigger the memory alignment and Snode offset mismatch. Maybe I can change the name to e.g., mix type for memory alignment?

gstaichi/runtime/program_impls/llvm/llvm_program.h
duburcqa: In this case, I think it is good enough to just do `offset_aligned = ((offset + align - 1) / align) * align;`

tests/python/gstaichi/lang/test_dlpack.py
erizmr: Moved to top level

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: I found there is a similar func in [type_utils.cpp](https://github.com/Genesis-Embodied-AI/gstaichi/blob/5428d9f5f8dfc2e8b5dd14bd5ac64bef15049624/gstaichi/ir/type_utils.h#L23), now turn to use the existing one.

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: haha yes it is correct offset now, then let me modify the comment

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: I agree the memory alignment of fields would not vary use cases. My original thought is to make a safe change only limited to dlpack use case and do not cause regressions in existing use cases.

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: Yes. Please change it to imperative.

hughperkins: (after all, everything is memory aligned, in general. A byte is 8 bit aligned. A 32bit int is 32 bit aligned. I would imagine 64 bit int is 64bit aligned. etc. There isn't really any such thing as 'not memory aligned', in general, I feel?)

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -182,20 +183,49 @@ class LlvmProgramImpl : public ProgramImpl {
...
     int child_id = root->child_id(dense_parent);
 
+    // The maximum cell size bytes is used to handle memory alignment for dlpack
+    // usage When compute field offset, SNode doesn't take the alignment into
+    // account yet, we need to handle it here
+
+    // The second ->ch is to get the actual data continer
+    DataType dt = root->ch[child_id].get()->ch[0]->dt;
+    size_t max_cell_size_bytes = data_type_size(dt.get_element_type());
+    TI_DEBUG("data type {} data type size {}",
```
hughperkins: remove TI_DEBUG please

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: Can you see any examples where it will break existing code?

Ideally, we can trust our unit tests, so maybe simply try making it default to making the alignment, and see if any tests break?

hughperkins: (historically, I tend to trust our unit tests. If the unit tests pass, and then we turn out to break something, we can add some new unit tests at that point, and fix the issue, I feel. Unless @duburcqa you have a strong objection to this approach, in this particular situation?)

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: Comment still seems to exist?

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: "manully" still seems to exist?

hughperkins: oh wait, I was looking at wrong version. great! 🙌

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: Great! 🙌

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: Oh, I see, curr_field has one dtype, and pos has another?

Perhaps you can add a docstring to point this out, so lazy people like me, who don't read the code carefully, can see this easily :)

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: Chatted with alexis. Looks like without factorizing the loop body into a separate function
=> marking conversaiton resolved.

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -189,3 +190,76 @@ def test_dlpack_field_multiple_tree_nodes():
     assert c_t[0] == 333
     assert d_t[0] == 444
     assert e_t[0] == 555
+
+
+@test_utils.test(arch=dlpack_arch)
+@pytest.mark.parametrize("tensor_type", [ti.field])
```
hughperkins: please remove this from parametrize, since it only takes a single value, ie ti.efield

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -221,3 +226,76 @@ def test_dlpack_field_multiple_tree_nodes():
...
+    ti.sync()
+
+    np.testing.assert_allclose(
+        torch.utils.dlpack.from_dlpack(pos.to_dlpack()).cpu().numpy(),
+        pos.to_numpy(),
+    )
+
+
+@test_utils.test(arch=dlpack_no_metal_arch)
+def test_dlpack_joints_case_memory_alignment() -> None:
```
hughperkins: please add _field postfix to the name, or run for ndarray too

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -6,6 +7,10 @@
 from tests import test_utils
 
 dlpack_arch = [ti.cpu, ti.cuda, ti.metal]
+dlpack_no_metal_arch = [
+    ti.cpu,
+    ti.cuda,
```
hughperkins: Pleas add back metal. dlpack works fine on metal ndarray

hughperkins (changes requested):

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: I made some modifications to the comments, the motivation is to keep the context the assumptions we made here when getting these field offsets

tests/python/gstaichi/lang/test_dlpack.py
erizmr: postfix added

tests/python/gstaichi/lang/test_dlpack.py
erizmr: used xfail for metal backend instead

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: New comment looks good. Thanks!

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -182,20 +183,46 @@ class LlvmProgramImpl : public ProgramImpl {
...
     size_t offset = 0;
 
     SNode *dense_parent = child->parent;
     SNode *root = dense_parent->parent;
 
     int child_id = root->child_id(dense_parent);
 
+    // The maximum cell size bytes is used to handle memory alignment for dlpack
+    // usage When compute field offset, SNode doesn't take the alignment into
+    // account yet, we need to handle it here
```
duburcqa: This comment is not very clear and does not follow English grammar rules with punctuation and uppercases.

Can you clarify why this is needed to accommodate dlpack limitations? What is assumed by dlpack that may be wrong, and why we cannot adapt our use of dlpack instead?

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -182,20 +183,46 @@ class LlvmProgramImpl : public ProgramImpl {
...
+
+    // The second ->ch is to get the actual data continer
```
duburcqa: `data container`

gstaichi/runtime/program_impls/llvm/llvm_program.h
duburcqa: I'm in favour of removing any option that is not used. If there is no use-case for nonaligned memory, then do not add this option and change the current behaviour if needed. Our objective is to remove as many features as possible, as long as it is not going to be used in a foreseeable future. Not to add more unused features.

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -221,3 +222,87 @@ def test_dlpack_field_multiple_tree_nodes():
...
+
+
+@test_utils.test(arch=dlpack_arch)
+@pytest.mark.parametrize("dtype", [ti.i32, ti.i64, ti.f32, ti.f64, ti.u1, ti.i8, ti.types.vector(3, ti.i32)])
+@pytest.mark.parametrize("shape", [3, 1, 4, 5, 7, 2])
+def test_dlpack_mixed_types_memory_alignment_field(dtype, shape: tuple[int]) -> None:
+    from gstaichi.lang import impl
+
+    if impl.current_cfg().arch == ti.metal:
+        pytest.xfail(reason="dlpack for field hasn't been supported on Metal backend yet.")
```
duburcqa: I would rather add this as a file-local auto-fixture to avoid duplicating this check in every single test

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: Let me do some tests to see if it breaks any tests

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: Sorry for the confusion made here, will polish the comments.

tests/python/gstaichi/lang/test_dlpack.py
erizmr: Have made it a file-local auto-fixture.

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -9,6 +10,15 @@
 dlpack_ineligible_arch = [ti.vulkan]
 
 
+@pytest.fixture(autouse=True)
+def selective_metal_xfail(request):
+    if "metal_xfail" in request.keywords:
+        from gstaichi.lang import impl
+
+        if impl.current_cfg().arch == ti.metal:
+            pytest.xfail(reason="dlpack for field hasn't been supported on Metal backend yet.")
```
duburcqa: The marker logic is overkill since all tests in this file are affected.

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: Did the tests, no broken exisintg test found by making the alignment default. I will go with removing this is_memory_aligned option.

tests/python/gstaichi/lang/test_dlpack.py
erizmr: Please checkout the updated version. The autouse is removed, and the metal_xfail is only applied to the new field specific test cases.

gstaichi/runtime/program_impls/llvm/llvm_program.h
erizmr: Please checkout the polished comments.

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -9,6 +10,14 @@
 dlpack_ineligible_arch = [ti.vulkan]
 
 
+@pytest.fixture
+def metal_xfail():
```
duburcqa, line 14: Sorry I thought all the uni tests were affected by this issue. Apparently it is not the case. As a result, using a fixture may not be a good idea in the first place. But it is not a big deal.

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -184,18 +185,51 @@ class LlvmProgramImpl : public ProgramImpl {
 
   size_t get_field_in_tree_offset(int tree_id, const SNode *child) override {
     // FIXME: Compute the proper offset. Current method taken from GGUI code
```
hughperkins: why is this comment still here? Are we not now computing the proper offset? Is the current method actually taken from GGUI code? I think you've modified the method extensively now right? I could accept "Method based on similar method in GGUI code" if you wish, but fairly open to just removing this comment altogether. ie removing "// FIXME: Compute the proper offset. Current method taken from GGUI code".

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -184,18 +185,51 @@ class LlvmProgramImpl : public ProgramImpl {
...
+    // SNodes record only the actual byte size (cell_size_bytes) without
+    // accounting for alignment padding. For example, if three ti.f32 fields and
+    // one ti.f64 field share the same SNode tree, the SNode’s logical layout
+    // is:
+    //   | ti.f32 | ti.f32 | ti.f32 | ti.f64 | = 20 bytes.
+    // However, in real memory, alignment requirements introduce padding,
+    // producing:
+    //   | ti.f32 | ti.f32 | ti.f32 | padding (4 bytes) | ti.f64 | = 24 bytes.
+    //
+    // This causes the computed offsets to differ from the true in-memory
```
hughperkins: Leaving aside dlpack, dont we need to know the correct padding and alignment in order to calculate the offset of each field within the allocated buffer? Where/how is that calculated? Why can't we use the same mechanism here, rather than having two parallel mechanisms?

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -221,3 +230,75 @@ def test_dlpack_field_multiple_tree_nodes():
     assert c_t[0] == 333
     assert d_t[0] == 444
     assert e_t[0] == 555
+
+
+@test_utils.test(arch=dlpack_arch)
+@pytest.mark.parametrize("dtype", [ti.i32, ti.i64, ti.f32, ti.f64, ti.u1, ti.i8, ti.types.vector(3, ti.i32)])
```
hughperkins, line 231: oh wait. I'm wrong. ignore.

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -221,3 +230,75 @@ def test_dlpack_field_multiple_tree_nodes():
...
+@pytest.mark.parametrize("shape", [3, 1, 4, 5, 7, 2])
+def test_dlpack_mixed_types_memory_alignment_field(dtype, shape: tuple[int], metal_xfail) -> None:
```
hughperkins, line 233: Honestly, I think I'd rather have the xfail clearly visible in the standard place. But fine, not going to ask you to flip backwards and forwards between two different things :)

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: Docstring looks good. Thanks!

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: since:
1. it's moved to top level
2. @duburcqa approved the entire review
... so I'm going to mark this conversation resolved, to make it easier for me to see what is left.

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: I'm still seeing a fixme comment about alignment not being computed yet?

Screenshot 2025-11-28 at 07 44 23

gstaichi/runtime/program_impls/llvm/llvm_program.h
hughperkins: (marking resolved, since alexis already approved the pr)

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: (marking resolved, since alexis already approved the pr)

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: (marking resolved, since alexis already approved the pr)

gstaichi/runtime/program_impls/llvm/llvm_program.h
```diff
@@ -183,19 +183,15 @@ class LlvmProgramImpl : public ProgramImpl {
...
-
-    int child_id = root->child_id(dense_parent);
-
-    for (int i = 0; i < child_id; ++i) {
-      SNode *child = root->ch[i].get();
-      offset += child->cell_size_bytes * child->num_cells_per_container;
+    for (const SNode *sn = child; sn != nullptr; sn = sn->parent) {
+      // Only add offsets for nodes that have a parent (root has no parent).
+      if (sn->parent)
+        offset += sn->offset_bytes_in_parent_cell;
```
hughperkins, line 193: Nice! 🔥

hughperkins:
Approved conceptually. Before I actually click the "confirm" button, @duburcqa since the last update was a fairly big change, pelase could you re-review/re-approve?

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -9,6 +10,14 @@
 dlpack_ineligible_arch = [ti.vulkan]
 
 
+@pytest.fixture
+def metal_xfail():
+    from gstaichi.lang import impl
+
+    if impl.current_cfg().arch == ti.metal:
+        pytest.xfail(reason="dlpack for field hasn't been supported on Metal backend yet.")
```
duburcqa, line 18: Metal is supported now?

hughperkins: Not for field it's not.

Pending pytorch pr, right?

duburcqa: I see. My bad. Forgot about it.

PR #304, opened 2025-11-28
[SPIRV] Add additional spir-v dump before optimization

hughperkins:
Issue: #

### Brief Summary

So now we have both before and after opt:

[
Screenshot 2025-11-28 at 08 10 04
Screenshot 2025-11-28 at 08 10 07
](url)

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/gstaichi/codegen/spirv/spirv_codegen.cpp
+++ b/gstaichi/codegen/spirv/spirv_codegen.cpp
@@ -727,11 +727,11 @@ void TaskCodegen::visit(ExternalPtrStmt *stmt) {
       linear_offset = ir_->mul(linear_offset, size_var);
       linear_offset = ir_->add(linear_offset, indices);
     }
+    size_t type_size =
+        ir_->get_primitive_type_size(stmt->ret_type.ptr_removed());
     linear_offset = ir_->make_value(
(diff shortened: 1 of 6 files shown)


gstaichi/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2437,6 +2437,16 @@ void KernelCodegen::run(GsTaichiKernelAttributes &kernel_attribs,
     TaskCodegen cgen(tp);
     auto task_res = cgen.run();
 
+    if (dump_ir) {
```
hughperkins, line 2438: this paragraph is the new bit in this pr.

The other changes are refactorization/tidying up.

hughperkins:
> Is this for checking the difference of spirv between before/after optimization?

It's because, when I'm debugging spirv_codegen.cpp, I want to see the spirv that is closest to what is actually being written out/ generated, not some spirv that went through an optimization step, and who-knows-what transformations been applied to it.

Concretely, yesterday, for fixing the u1 issue, I used the spirv before optimization output to do that.

duburcqa (approved):

PR #306, opened 2025-11-28
[AMDGPU] Enable AMDGPU build and AMDGPU test runner

hughperkins:
Issue: #

### Brief Summary

Fixes for AMD GPU build
- enables WITH_AMDGPU, in CI wheel builds, on manylinux and windows
- builds to completion with WITH_AMDGPU set, following upgrade of code to llvm 20
- tests pass for `--arch amdgpu`, following changes:
    - replaced exception in `print` with `nop`
    - added xfail for multiple shared array test, since we don't use shared arrays in genesis currently
- adds workflow to run amdgpu tests on new amdgpu self-hosted runner

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/gstaichi/runtime/amdgpu/kernel_launcher.cpp
+++ b/gstaichi/runtime/amdgpu/kernel_launcher.cpp
@@ -122,7 +122,6 @@ void KernelLauncher::launch_llvm_kernel(Handle handle,
   if (transfers.size()) {
     for (auto itr = transfers.begin(); itr != transfers.end(); itr++) {
       auto &idx = itr->first;
-      // Change from idx.pop_back() to idx.arg_id
       auto arg_id = idx.arg_id;
       AMDGPUDriver::get_instance().memcpy_device_to_host(
           itr->second.first, (void *)device_ptrs[idx],


hughperkins:
On first 25% of all tests, around ~10% fail, 90% pass:
Screenshot 2025-11-28 at 16 40 18

hughperkins:
Note: There are two sets of changes here:
- Cursor stuff, to get amd to build, following llvm20 migration
- Changes to get tests to pass

I feel like these should perhaps be reviewd separately? (cursor stuff to build as the first PR).

Let me know your thoughts.

In any case:
- everything up to and including [d4e080f](https://github.com/Genesis-Embodied-AI/gstaichi/pull/306/commits/d4e080f448a4562fcc1213f173dfb8b9193c801f) is "cursor stuff to get build working"
- everything after that is "changes to get tests to passA"

gstaichi/runtime/amdgpu/kernel_launcher.cpp
```diff
@@ -122,8 +122,8 @@ void KernelLauncher::launch_llvm_kernel(Handle handle,
   if (transfers.size()) {
     for (auto itr = transfers.begin(); itr != transfers.end(); itr++) {
       auto &idx = itr->first;
-      auto arg_id = idx;
-      arg_id.pop_back();
+      // Change from idx.pop_back() to idx.arg_id
```
duburcqa: Just remove this comment. It is indeed what must be done.

hughperkins: If there was ever any doubt tat I Cursored this, my cover is blown now 😂

duburcqa:
> I feel like these should perhaps be reviewd separately? (cursor stuff to build as the first PR).

I'm fine with doing both at once, because merging the fix alone first does not really make sense since you cannot validate that it works anyway, and the other way around is plain and simple impossible.

erizmr (approved):

PR #313, opened 2025-12-05
[Perf] dlpack with amdgpu

hughperkins:
Issue: #

### Brief Summary

- udpate to_dlpack to work with amdgpu
- ~move existing .bc files to 5x folder~
- add .bc files for amdgpu to recent ones from rocm 7.0, to new 70 folder (so works on MI300X for example, 942)
- migrate to use 70 folder
- ~select 5x folder for gfx10 and gfx11 (for v520, which doesnt work with the 70 bc files, empirically)~
     - ~otherwise use 70 folder~
- search for ld.lld automatically, so we don't have to add it to PATH
- use default context for amdgpu, instead of creating a new one, for pytorch compatibility (which is consistent with how the cuda version works)
- check for new env var TI_AMDGPU_V520=1 in test_dlpack.py and in test_torch_ad.py. When this is true:
    - move the torch tensors onto cpu, before testing their values, so we avoid running any torch kernels
    - skip two tests that won't run on v520, because uses unsupported torch kernels (eg test_torch_ad_gpu uses `.zeros_like()`)

Note that this branch runs on MI300X, without using the TI_AMDGPU_V520 env var, see later comment below.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/gstaichi/codegen/spirv/spirv_codegen.cpp
+++ b/gstaichi/codegen/spirv/spirv_codegen.cpp
@@ -1409,6 +1409,20 @@ void TaskCodegen::visit(InternalFuncStmt *stmt) {
         spv_op, stype,
         ir_->int_immediate_number(ir_->i32_type(), spv::ScopeSubgroup), arg0,
         arg1);
+  } else if (stmt->func_name == "spirv_clock_i64") {
+    // OpReadClockKHR returns a 64-bit unsigned integer
+    // Scope: Device (1) for device-wide clock
+    if (caps_->get(DeviceCapability::spirv_has_shader_clock) &&
+        caps_->get(DeviceCapability::spirv_has_int64)) {
+      spirv::Value clock_val = ir_->make_value(
+          spv::OpReadClockKHR, ir_->u64_type(),
+          ir_->uint_immediate_number(ir_->u32_type(), spv::ScopeDevice));
+      // Cast u64 to i64 as the return type is i64
+      val = ir_->make_value(spv::OpBitcast, ir_->i64_type(), clock_val);
+    } else {
+      // Return 0 if shader clock is not supported
+      val = ir_->int_immediate_number(ir_->i64_type(), 0);
+    }
   }
   ir_->register_value(stmt->raw_name(), val);
 }
--- a/gstaichi/codegen/spirv/spirv_ir_builder.cpp
+++ b/gstaichi/codegen/spirv/spirv_ir_builder.cpp
@@ -74,6 +74,11 @@ void IRBuilder::init_header() {
         .add(spv::CapabilityPhysicalStorageBufferAddresses)
         .commit(&header_);
   }
+  if (caps_->get(cap::spirv_has_shader_clock)) {
+    ib_.begin(spv::OpCapability)
+        .add(spv::CapabilityShaderClockKHR)
+        .commit(&header_);
+  }
 
   ib_.begin(spv::OpExtension)
       .add("SPV_KHR_storage_buffer_storage_class")
@@ -107,6 +112,10 @@ void IRBuilder::init_header() {
     ib_.begin(spv::OpExtension)
         .add("SPV_EXT_shader_atomic_float_min_max")
         .commit(&header_);
+  }
+
+  if (caps_->get(cap::spirv_has_shader_clock)) {
+    ib_.begin(spv::OpExtension).add("SPV_KHR_shader_clock").commit(&header_);
   }
 
   if (caps_->get(cap::spirv_has_physical_storage_buffer)) {
--- a/gstaichi/inc/internal_ops.inc.h
+++ b/gstaichi/inc/internal_ops.inc.h
@@ -18,7 +18,7 @@ PER_INTERNAL_OP(do_nothing)
 PER_INTERNAL_OP(refresh_counter)
 PER_INTERNAL_OP(test_internal_func_args)
 
-// Vulkan
+// SPIRV
 PER_INTERNAL_OP(workgroupBarrier)
 PER_INTERNAL_OP(workgroupMemoryBarrier)
 PER_INTERNAL_OP(localInvocationId)
@@ -43,8 +43,10 @@ PER_INTERNAL_OP(subgroupInclusiveMax)
 PER_INTERNAL_OP(subgroupInclusiveAnd)
 PER_INTERNAL_OP(subgroupInclusiveOr)
 PER_INTERNAL_OP(subgroupInclusiveXor)
+PER_INTERNAL_OP(spirv_clock_i64)
 
 // CUDA
+PER_INTERNAL_OP(cuda_clock_i64)
 PER_INTERNAL_OP(block_barrier)
 PER_INTERNAL_OP(block_barrier_and_i32)
 PER_INTERNAL_OP(block_barrier_or_i32)
@@ -65,3 +67,9 @@ PER_INTERNAL_OP(cuda_match_any_sync_i32)
 PER_INTERNAL_OP(cuda_match_all_sync_i32)
 PER_INTERNAL_OP(cuda_active_mask)
 PER_INTERNAL_OP(warp_barrier)
+
+// AMDGPU
+PER_INTERNAL_OP(amdgpu_clock_i64)
+
+// CPU
+PER_INTERNAL_OP(cpu_clock_i64)
--- a/gstaichi/inc/rhi_constants.inc.h
+++ b/gstaichi/inc/rhi_constants.inc.h
@@ -32,6 +32,7 @@ PER_DEVICE_CAPABILITY(spirv_has_subgroup_arithmetic)
 PER_DEVICE_CAPABILITY(spirv_has_subgroup_ballot)
 PER_DEVICE_CAPABILITY(spirv_has_non_semantic_info)
 PER_DEVICE_CAPABILITY(spirv_has_no_integer_wrap_decoration)
+PER_DEVICE_CAPABILITY(spirv_has_shader_clock)
 #endif
 
 #ifdef PER_BUFFER_FORMAT
--- a/gstaichi/ir/type_system.cpp
+++ b/gstaichi/ir/type_system.cpp
@@ -319,7 +319,7 @@ void Operations::init_internals() {
   // block_barrier, grid_memfence, cuda_all_sync, cuda_any_sync, cuda_uni_sync,
   // cuda_ballot, cuda_shfl_sync, cuda_shfl_up_sync, cuda_shfl_down_sync,
   // cuda_shfl_xor_sync, cuda_match_any_sync, cuda_match_all_sync,
-  // cuda_active_mask, warp_barrier
+  // cuda_active_mask, warp_barrier, cuda_clock_i64
 
 #define CUDA_VOTE_SYNC(name) \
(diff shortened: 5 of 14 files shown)


hughperkins:
> This does not work If I understood correctly?

Yeah, I have a ticket to fix it.

.github/workflows/test_gpu.yml
```diff
@@ -170,20 +170,20 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.10'
+      - name: install torch
+        # do this now, otherwise gstaichi install will install non-rocm version instead
```
duburcqa: I don't think gstaichi should install torch for the user. Same as we are doing for genesis. We just raise an exception with a link to the official get started instructions. There are too many options, it is impossible to guess for the user which version should be installed. And if gstaichi is silently installing the wrong one, it is way worse than throwing an error because the issue may just go unnoticed while dramatically impeding performance.

hughperkins: Oh, I think the comment is wrong. But for some reason, without `--upgrade`, tests were failing, and on inspecting the instance, the venv had the non-rocm version of torch.

My spidey senses are telling me the venv is being re-used somehow, but I haven't got around to digging into that for now, since `--upgrade` fixes it.

gstaichi/runtime/amdgpu/jit_amdgpu.cpp
```diff
@@ -182,9 +185,40 @@ std::string JITSessionAMDGPU::compile_module_to_hsaco(
...
+        "/opt/rocm-6.0.0/llvm/bin/ld.lld",
+    };
+    for (const auto &path : common_paths) {
+      std::ifstream test_lld(path);
+      if (test_lld.good()) {
+        lld_executable = path;
+        break;
+      }
+    }
+  }
```
duburcqa, line 212: This seems to be linux-specific? What would happen on Windows?

tests/python/gstaichi/lang/test_dlpack.py
```diff
@@ -287,6 +313,8 @@ def kernel_init_joint_fields(
 
 @test_utils.test(arch=dlpack_arch)
 def test_dlpack_field_memory_allocation_before_to_dlpack():
+    if ti.cfg.arch == ti.amdgpu:
+        pytest.xfail(reason="Cannot run torch kernels on AMDGPU CI")
```
duburcqa: Is there a way to detect why it cannot run?

systematic fail, even for machines that are supported, in a bad practice and may be hiding real issues when running the unit tests manually on a local machine.

gstaichi/runtime/amdgpu/jit_amdgpu.cpp
hughperkins: At a high level, if we want to support AMD on Windows, we should probably look into having an AMD Runner on windows in our CI.

I'm pretty sure that, without that, no matter how much we try to make things work, they won't.

But to answer your question, preusmably one would need to define ROCM_PATH, or else use the same appracoh as on main currnetly. https://github.com/Genesis-Embodied-AI/gstaichi/blob/72f02b0683d25b1795d93cf7c8f666d747e6c1ab/.github/workflows/test_gpu.yml#L186-L188

tests/python/gstaichi/lang/test_dlpack.py
duburcqa, line 45: What about adding some helper method (local to this test module) atop `torch.utils.dlpack.from_dlpack` that would be responsible for moving data to cpu if necessary?

```python
def dlpack_to_torch(dlpack):
    tensor = torch.utils.dlpack.from_dlpack(dlpack)
    if ti.cfg.arch == ti.amdgpu:
        tensor = tensor.cpu()
    return tensor
```

duburcqa:
> use default context for amdgpu, instead of creating a new one, for pytorch compatibility (actually, unclear if this is necessary tbh)

How does this work for CUDA?

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: > Is there a way to detect why it cannot run?

Not obviously.

What I was thinking was : perhaps we can have an env var like `TI_AMDGPU_TESTS_COPY_TORCH_TENSOR` , or something. So by default this will run, and fail if it fails; but with this env var, it'll xfail instead.

Thoughts?

hughperkins: (edited to quote your first line)

hughperkins:
> > [Perf] dlpack with amdgpu
> 
> I don't think this PR is about perf from GsTaichi perspective.

We dont need to dlpack to copy tensors to amd gpu. We only need dlpack in order to improve perf.

hughperkins:
> > use default context for amdgpu, instead of creating a new one, for pytorch compatibility (actually, unclear if this is necessary tbh)
> 
> How does this work for CUDA?

Well, that's a great question!

The answer appears to be...

It retains the existing context https://github.com/Genesis-Embodied-AI/gstaichi/blob/72f02b0683d25b1795d93cf7c8f666d747e6c1ab/gstaichi/rhi/cuda/cuda_context.cpp#L66-L67

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: PS "systematic fail, even for machines that are supported, in a bad practice and may be hiding real issues when running the unit tests manually on a local machine." => yes, I agree. was thinking the same thing earlier. right. Though I was thinking more about the .cpu() bits. But meme combat.

.github/workflows/test_gpu.yml
hughperkins: (note to self: change the comment; and move back to previous location in the test_gpu.yml)

hughperkins:
Note to self: currently the 70 folder contains the original 5x folder with 70 overwriting it.
- I need to clean 70 folder to *only* contain the new 70 files, without older 5x files in it.

hughperkins:
> Note to self: currently the 70 folder contains the original 5x folder with 70 overwriting it.
> 
> * I need to clean 70 folder to _only_ contain the new 70 files, without older 5x files in it.

checked this point. seems already clean in fact.

hughperkins:
Tested at this point on MI300X, without using TI_AMDGPU_V520 env var, and passing:

Screenshot 2026-01-01 at 21 09 36
Screenshot 2026-01-01 at 21 10 33

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: I'd rather have it explicit I feel.

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: added var TI_AMDGPU_V520=1

gstaichi/runtime/amdgpu/jit_amdgpu.cpp
duburcqa: It is not supported, maybe we should disable this backend on Windows? If not, it would be nice to add least throw a warning stating that this backend is experimental on Windows. Moreover, when failing to find ld.lld, it would be nice to suggest setting ROCM_PATH when throwing an exception.

tests/python/gstaichi/lang/test_dlpack.py
duburcqa: What is the criteria for the tests to fail or pass? Is there any? I guess so?

tests/python/gstaichi/lang/test_dlpack.py
duburcqa: I think it is distracting the reader from what really matters in the unit tests. Unit tests serve as documentation of our specifications/requirements. The easiest it is to understand what a unit test is enforcing, the better it is I think.

duburcqa:
> We only need dlpack in order to improve perf.

Dlpack does not improve the performance of GsTaichi.

duburcqa:
> Well, that's a great question! The answer appears to be... It retains the existing context

In that case, I think it is a good thing to do the same for AMDGPU!

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: I'm not sure I follow. Isnt the 'assert's the criteria?

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: I mean, I agree that "Unit tests serve as documentation of our specifications/requirements. The easiest it is to understand what a unit test is enforcing", which is why I feel I'd like the `tensor = torch.utils.dlpack.from_dlpack(dlpack)` to be directly visible in the test, ideally.

But I guess you are a sample of a user viewing these tests from fairly fresh.

You're thinking a method with a name like `torch_utils_dlpack_from_dlpack`, somethign like that?
```
def torch_utils_dlpack_from_dlpack(dlpack):
    t = torch.utils.dlpack.from_dlpack(dlpack)
    if ti.cfg.arch == ti.amdgpu:
        t = t.cpu()
    return t
```
?

hughperkins:
> > We only need dlpack in order to improve perf.
> 
> Dlpack does not improve the performance of GsTaichi.

It improves the performance for the user of gstaichi I feel.

What is your own suggestion for the category/tag?

.github/workflows/test_gpu.yml
hughperkins: (done)

gstaichi/runtime/amdgpu/jit_amdgpu.cpp
hughperkins: - added warning on init with windows and amdgpu
- added note about ROCM_PATH to the exception

tests/python/gstaichi/lang/test_dlpack.py
duburcqa: I mean, something you could infer from the hardware and software stuck, without running anything.

tests/python/gstaichi/lang/test_dlpack.py
duburcqa: > You're thinking a method with a name like torch_utils_dlpack_from_dlpack, somethign like that?

Yes.

> which is why I feel I'd like the tensor = torch.utils.dlpack.from_dlpack(dlpack) to be directly visible in the test, ideally.

Well, I suppose you are right. But all these `.cpu()` are also very distracting when there are numerous :/

tests/python/gstaichi/lang/test_dlpack.py
hughperkins: oh you mean, v520 or not?

Either way, most of the tests will still pass, because we use .cpu() to copy to cpu. There are a couple of tests which use torch kernels which are missing on v520, eg `.zeros_like()`, which we skip.

We could try runnig .zeros_like(), but I think this is *more* dagnerous: we skip tests automatiaclly, without the user asking for it. Better to fail, unless the user clearly asks to skip them by expliclty defining `TI_AMDGPU_V520=1` I feel?

hughperkins:
> > It improves the performance for the user of gstaichi I feel.
> 
> Not really. It opens up a potential for performance increase. But nothing will happen without a fairly complex on user side.

Sure. But anyone using to_dlpack is doing this. We aren't improving anything else other than perforance I feel?

What are your thoguths on what tag we might use?

duburcqa (approved):

PR #314, opened 2025-12-10
[cuda,amdgpu,cpu,vulkan] Add public API for kernel clock counter.

duburcqa:


tests/python/test_simt.py
```diff
@@ -228,6 +228,28 @@ def foo():
         assert a[i] == 1
 
 
+@test_utils.test(arch=ti.cuda)
```
hughperkins: curious if it works on amdgpu too?

duburcqa: It relies on nvcc under the hood. I guess it cannot be used on AMD GPUs ?

hughperkins: per chatgpt, you can do wall_clock64 on amd:

Screenshot 2025-12-10 at 09 03 23
Screenshot 2025-12-10 at 09 03 11

I agree though that it's iffy whether we want to spend time adding this to amd. On the whole, I think we probably should not, because:
1. we'll use cuda for benchmarking
2. testing on amd is a pita, because basically means making chnages, then waiting 2 hours for CI to run

Still, here's a question: what happens if we add cuda_clock_i64 to a kernel, and then try to run on amd? Could we at least ensure that it `nop`s on non-cuda architectures - metal and amdgpu - and test that this means it at least wont crash on these architectures?

Actually.. I wonder if we should make the taichi command architecture-independent?

instead of

```
ti.simt.timer.cuda_clock_i64()
```

perhaps we can have:

```
ti.simt.timer.wallclock()
```
?

(Then 1. we can easily add other architectures later on, if we choose, 2. it makes sense to have this added to kernels that can run on all architectues, even if it `nop`s out in practice)

hughperkins: (nuance: perhaps it shouldnt `nop` so much as return `0`?)

hughperkins: (or `-1`. no strong opiinon either way)

hughperkins: Note: I'm not entirely convinced we want this command under `simt` namespace 🤔  I'm not saying we don't, just pondering the question. But, I mean, personally I was imagining simply `ti.clock()`, or similar. What do you see as advantages to placing the clock method under the simt namespace?

duburcqa: I'm just following the existing pattern. All intrinsic stuffs are going under SIMT, even `sync`.

duburcqa: > Actually.. I wonder if we should make the taichi command architecture-independent?

I'm just following the same pattern as other intrinsic. We can change if needed.

hughperkins: Yeah, I'd prefer to make the command architecture-independent.

Location under simt or not, I have a softer opinion. I'd prefer to go with the location we prefer, than consistency, on the whole. Do you prefer writing something like `ti.simt.timer.clock()`, or  something like `ti.clock()` ?

duburcqa: `ti.clock` is way better if we went to encourage people to use it. But if we want to keep it kind of hidden to mitigate people expectations then it is better under simt.

hughperkins: Well, I think we just need to target ourselves for now :)

But, maybe I would change my mind if I knew first hand how flaky it is.

Perhaps I use it myself for a bit first, and then can form an opinion about how flaky it is?

(I assume by "mitigate peoples expectations", you mean, it is flaky in some way?)

duburcqa: Well, it is experimental, but I did not observed any particular issue. I don't have any major concerns.

hughperkins: Ok. Then let's use `ti.clock()` then perhaps?

(Another option could be `ti.experimental.clock()`, if you prefer, perhaps?)

hughperkins: what is your preference between `ti.experimental.clock()` and `ti.clock()` ?

duburcqa: Resolved.

duburcqa:
> let's make the method name architecture-independent

Done.

> let's check that it nops out on unsupported architectures (metal, amdgpu)

Done. Well, it is not supported on metal though, because it requires i64 support, which is not available.

duburcqa:
> Couldn't we use i32 on metal? We only care about relative times anyway right? (But I don't think we need to implement that right away)

No. Because clock does not measure a difference. You need it to fetch the clock value twice, then to compute the elapsed time. Once done, you can cast to i32, but not before.

duburcqa:
> How would this be incompatible with using i32?

Because `Start = ti.clock()` or `End = ti.clock()` may be out of range for i32.

hughperkins:
It might be. How long will 4 billion correspond to?

hughperkins:
What is MTLCommandEncoder, in your screenshot? Can we use that?

tests/python/test_intrinsics.py
```diff
@@ -0,0 +1,46 @@
...
+    arch = ti.lang.impl.get_runtime().prog.config().arch
+
+    dtype = ti.i64 if ti.lang.is_extension_enabled(ti.extension.int64) else ti.i32
+    a = ti.field(dtype=dtype, shape=32)
+
+    @ti.kernel
+    def foo():
+        ti.loop_config(serialize=True, block_dim=1)
+        for i in range(32):
+            start = ti.clock()
```
hughperkins: Thanks for giving this a platform-independent name.

After our discussion, it occurs to me that would be useful to add the units to the name, eg `ti.clock_nsecs()`. What are the units in fact? (And I think it is best if the units are consistent across all platforms, ideally)

duburcqa: It is a number of clock cycles for CUDA, and a number of nanoseconds on CPU.

duburcqa:
> we only need relative time. It's ok if it wraps
around, as long as it only wraps around once. This means we can measure up
to 4 billion

No it does not mean if we can measure up to 4 billion. If time_start = max(i32) - 1, then you can only measure 1.

duburcqa:
> What is MTLCommandEncoder, in your screenshot? Can we use that?

No. This is not available in shader code.

tests/python/test_intrinsics.py
hughperkins: That makes sense I guess. Would be good to have it platform independent I feel, ideally. Do we have some way to obtain the clock frequency of the gpu, and convert into nanoseconds?

hughperkins: (though I suppose that might add latency. division is pretty slow...)

hughperkins: I suppose using the fastest approach might be the best approach. Looks like you've added comprehensive doc string https://github.com/Genesis-Embodied-AI/gstaichi/pull/314/files#diff-af0eb041e0962c0346c7b2d3e285f09b416628d7948605bd4c7d58e8c928ad50R7-R23 .

duburcqa: > Would be good to have it platform independent I feel, ideally. Do we have some way to obtain the clock frequency of the gpu, and convert into nanoseconds?

I MUCH MUCH prefer clock cycle. It think real time clock is just garbage, but it is all we have for some devices.

duburcqa: > Do we have some way to obtain the clock frequency of the gpu, and convert into nanoseconds?

Well, I guess it is doable and it would be more intelligible.

hughperkins: I don't have a strong opinion at this point. I'm ok with clock cycles. I'll leave you to decide which you prefer to use.

duburcqa: Actually I was wrong. it is always a counter but not always about clock cycles. Updated the name of the method and documentation accordingly.

hughperkins: I think not doing the division inside the kernel makes sense. I wonder if it might be worth creating a non-kernel function that returns the number of counter ticks per second, on the current hardware? This would make it easy to do the conversion later, outside of the kernel, in our testing code.

I can live with that being out of scope for the current PR though.

gstaichi/inc/internal_ops.inc.h
```diff
@@ -18,7 +18,7 @@ PER_INTERNAL_OP(do_nothing)
 PER_INTERNAL_OP(refresh_counter)
 PER_INTERNAL_OP(test_internal_func_args)
 
-// Vulkan
+// Vulkan/SPIRV
```
hughperkins: Are we adding SPIRV because this generalizes to Metal? If so maybe we can simply put `// SPIRV`.

python/gstaichi/lang/intrinsics.py
```diff
@@ -0,0 +1,43 @@
...
+from gstaichi._lib import core as _ti_core
+from gstaichi.lang import impl
+
+
+def clock_counter():
+    """
+    Returns the current value of a hardware cycle counter.
+
+    All backends return raw clock cycles or ticks, NOT nanoseconds.
+    The counter frequency varies by hardware and may change dynamically
```
hughperkins, line 12: interesting

python/gstaichi/lang/intrinsics.py
```diff
@@ -0,0 +1,43 @@
...
+      See https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#time-function
+    - AMDGPU: GPU cycle counter
+    - Vulkan: Device clock in cycles (requires VK_KHR_shader_clock, else returns 0)
+    - CPU (x64/arm64): Processor timestamp counter (constant rate on modern CPUs)
+
+    Unsupported backends (returns 0):
+    - Metal
+
+    Use this for relative timing measurements within the same backend and run.
+    Comparing cycle counts across different backends or hardware is not meaningful.
```
hughperkins, line 26: docstring seems comprehensive. thank you :)

hughperkins (approved):
Great! Thank you 🙌

gstaichi/inc/internal_ops.inc.h
duburcqa: > Are we adding SPIRV because this generalizes to Metal? 

Yes

duburcqa: done.

hughperkins: Great! 🙌

hughperkins:
Interesting. Does that mean we should start supporting i64 on metal, and
stop disabling all the i64 tests on metal?

PR #315, opened 2025-12-12
[Lang] Allow return void inside functions

hughperkins:
Issue: #

just so I can see what files are changed.

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/gstaichi/analysis/build_cfg.cpp
+++ b/gstaichi/analysis/build_cfg.cpp
@@ -140,14 +140,12 @@ class CFGBuilder : public IRVisitor {
       // For function returns (unwind), the control flow exits the current block
       // and jumps to the unwind target. We track these nodes so we can connect
       // them to the final node after the CFG is fully built.
-      TI_INFO("[CFG DEBUG] Found unwind with from_function_return=true at stmt {}, adding to unwind_nodes_", current_stmt_id_);
       unwind_nodes_.push_back(node);
       // We do NOT add this node to prev_nodes_ because statements after the
       // unwind in the same block are unreachable from this path.
     } else {
       // For normal loop continues, add to continues_in_current_loop_
       // so visit_loop() can connect them back to the loop beginning.
-      TI_INFO("[CFG DEBUG] Found continue with from_function_return=false at stmt {}, adding to continues_in_current_loop", current_stmt_id_);
       continues_in_current_loop_.push_back(node);
     }
   }
--- a/gstaichi/codegen/llvm/codegen_llvm.cpp
+++ b/gstaichi/codegen/llvm/codegen_llvm.cpp
@@ -893,7 +893,6 @@ void TaskCodeGenLLVM::visit(IfStmt *if_stmt) {
   }
   if (!returned) {
     builder->CreateBr(after_if);
-    // } else {
   }
 
   builder->SetInsertPoint(false_block);
@@ -903,8 +902,6 @@ void TaskCodeGenLLVM::visit(IfStmt *if_stmt) {
   }
   if (!returned) {
     builder->CreateBr(after_if);
-    // } else {
-    //   returned = false;
   }
   returned = false;
   builder->SetInsertPoint(after_if);


gstaichi/ir/frontend_ir.cpp
```diff
@@ -1616,7 +1616,12 @@ void ASTBuilder::insert_break_stmt(const DebugInfo &dbg_info) {
 }
 
 void ASTBuilder::insert_continue_stmt(const DebugInfo &dbg_info) {
-  this->insert(Stmt::make<FrontendContinueStmt>(dbg_info));
+  this->insert(Stmt::make<FrontendContinueStmt>(0, dbg_info));
+}
+
+void ASTBuilder::insert_function_continue_stmt(int loop_depth,
```
hughperkins, line 1622: Not sure if we need both of these. I think one of these is from an earlier attempt.

gstaichi/ir/statements.h
```diff
@@ -1220,8 +1225,10 @@ class ReturnStmt : public Stmt {
     for (auto &x : values) {
       names += x->raw_name() + ", ";
     }
-    names.pop_back();
-    names.pop_back();
+    if (!names.empty()) {
```
hughperkins, line 1228: this fixes the printer for a return void, which otherwise prints a bunch of garbage after the `return` statement.

gstaichi/transforms/compile_to_offloads.cpp
```diff
@@ -262,6 +262,12 @@ void offload_to_executable(IRNode *ir,
     print("Demote mesh statements");
   }
 
+  // Structure continues for SPIRV-based backends (Metal, Vulkan)
+  if (arch_uses_spirv(config.arch)) {
+    irpass::structure_continues(ir, config);
+    print("Structure continues");
```
hughperkins, line 268: I guess this satement is debug, and can be removed.

gstaichi/transforms/simplify.cpp
```diff
@@ -311,6 +311,15 @@ class BasicBlockSimplify : public IRVisitor {
 
       bool global_state_changed = false;
       for (int i = 0; i < (int)clause.size() && plain_clause; i++) {
+        // Control flow statements (continue/break/return/unwind) should never
+        // be moved out of their original scope, as this changes program
```
hughperkins, line 315: I wonder if this is actually true 🤔
```
if a:
   do_something()
   unwind(depth=2)
else:
   do_something_else()
   unwind(depth=2)
```
and:
```
if a:
   do_something()
else:
   do_something_else()
unwind(depth=2)
```
seem equivalent to me?

hughperkins: (not sure if the 'if' counts as a depth 🤔 I dont think so, but ... maybe. In which case we'd want to decrement the depth when moving)

gstaichi/transforms/unreachable_code_elimination.cpp
```diff
@@ -147,6 +229,64 @@ class UnreachableCodeEliminator : public BasicStmtVisitor {
...
+        if (cont->from_function_return) {
+          modifier.erase(cont);
+          modified = true;
+        }
+      }
+      // Don't recurse into if-statements - unwinds there are meaningful
+    }
+  }
+
+  // This method is no longer used but kept for reference
```
hughperkins, line 248: if it's no longer used, we should probbably remove it...

gstaichi/transforms/unreachable_code_elimination.cpp
```diff
@@ -147,6 +229,64 @@ class UnreachableCodeEliminator : public BasicStmtVisitor {
...
+        if (if_stmt->false_statements)
+          eliminate_kernel_scope_unwinds_recursive(
+              if_stmt->false_statements.get());
+      }
+    }
+    if (any_modified)
+      modified = true;
+  }
+
+  void eliminate_kernel_scope_unwinds_recursive(Block *block) {
```
hughperkins, line 272: ditto

gstaichi/transforms/unreachable_code_elimination.cpp
```diff
@@ -147,6 +229,64 @@ class UnreachableCodeEliminator : public BasicStmtVisitor {
     }
     return modified;
   }
+
+ private:
+  // Eliminate unwind statements from function returns at the TOP LEVEL only.
+  // Unwinds inside if-statements are meaningful for control flow.
```
hughperkins, line 235: Not sure I follow why we are iterating over statements? Terminators (and a continue/unwind is a terminator) are only valid as the last statement in a basic block I think? Also, shouldn't we replace it with a new terminator 🤔

gstaichi/transforms/unreachable_code_elimination.cpp
```diff
@@ -51,7 +73,62 @@ class UnreachableCodeEliminator : public BasicStmtVisitor {
...
+
+        // Check if false branch ends with return/unwind
+        if (if_stmt->false_statements &&
+            !if_stmt->false_statements->statements.empty()) {
+          auto *last_stmt = if_stmt->false_statements->statements.back().get();
+          false_branch_returns =
+              last_stmt->is<ContinueStmt>() || last_stmt->is<ReturnStmt>();
+        }
+
+        // If both branches return, code after the if-statement is unreachable
```
hughperkins, line 98: I wonder if we should move the continue out of both branches in this case? Otherwise our current block has no terminator satement perhaps? 🤔

gstaichi/transforms/unreachable_code_elimination.cpp
```diff
@@ -51,7 +73,62 @@ class UnreachableCodeEliminator : public BasicStmtVisitor {
...
+            stmt_list->erase(j);
+          modified = true;
+          break;
+        }
+      }
     }
+
+    // After the main loop, check if there are any function-return unwinds
+    // that ended up as siblings to if-statements (not inside them).
+    // These can happen when the simplify pass moves code around.
```
hughperkins, line 111: I wonder if the simplify should notice that it's moving a terminator into the block containing the if, which presumably already has a terminator, and possibly other satemetns, and should eliminate both the existing terminator and any existing statements?

Also, I don think the continue should be eliminated: the terminator after the continue should be eliminated I think?

hughperkins: (I'm not sure whether optimizer steps should leave the IR in an invalid state? (which presumaby is the case currently, since a basic block is not allowed to contain either multipl terminators, or a terminator that is not in the last position).

PR #319, opened 2025-12-16
[Misc] Refactorizing kernel_impl.py

hughperkins:
Issue: #

### Brief Summary
- refactorize kernel_imply.py, which was getting a bit unwieldy I feel
- gstaichi_callable.py
- kernel_types.py
- move Kernel into kernel.py
- move Func into func.py
- create new base class FuncBase, from which both Kernel and Func derive
- merge extract_args from func nad kernel together (and rename to check_parameter_anntoations)
- move many functions from kernel_imp.py into func_base.py
- factorize out ASTTransformer, from Kernel.py nested function
- factorize out _ast_to_dict, and dump_ast, from ast transformer call
- remove dumping of kernel csvs, which I added, but never use
- fix TI_DUMP_IR to actually use compile_config.debug_dump_path, so that test_config.py test passes again (broke after removing dump of kernel checksums)
- factorize out _try_load_fastcache
- factorize out LaunchContextBufferCache

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_config.py
+++ b/tests/python/test_config.py
@@ -57,7 +57,6 @@ def test_config_debug_dump_path(specify_path: bool, tmp_path: pathlib.Path):
         print(proc.stderr)
     assert proc.returncode == RET_SUCCESS
     if specify_path:
-        print("tmp_path", tmp_path)
         os.system(f"ls {tmp_path}")
         assert len(list(tmp_path.glob("*"))) > 0
     else:


python/gstaichi/lang/ast/ast_transformer_utils.py
```diff
@@ -20,14 +20,10 @@
...
 
-AutodiffMode = _ti_core.AutodiffMode
-
-
 if TYPE_CHECKING:
-    from gstaichi.lang.kernel_impl import (
-        Func,
-        Kernel,
-    )
+    from .._func_base import FuncBase
```
erizmr, line 24: nit: is there any easy way we can avoid this relative path import?

python/gstaichi/lang/func.py
```diff
@@ -0,0 +1,173 @@
...
+from gstaichi.lang.matrix import MatrixType
+from gstaichi.lang.struct import StructType
+from gstaichi.types import (
+    ndarray_type,
+    primitive_types,
+    template,
+)
+from gstaichi.types.enums import AutodiffMode
+
+from ._func_base import FuncBase
```
erizmr, line 30: ditto

erizmr (commented):
Thanks for this refactorizing! I have no major concerns if the tests are all passed. This perf benchmark https://github.com/Genesis-Embodied-AI/Genesis/pull/2119/checks?check_run_id=58239021430 looks good in general. The `batched_franka_accessors`[field][cuda] and `duck_in_box_hard`[cpu][field] https://github.com/Genesis-Embodied-AI/Genesis/runs/58278423259 looks have a relative large difference perf comparing to the previous run, do we know the reason or do you think they are noises?

python/gstaichi/lang/ast/ast_transformer_utils.py
hughperkins: @duburcqa prefers relatives imports I believe.

So do I, unless there's some strong reason why relative imports won't work?

What are your reasons for preferring absolute imports over relative imports?

tests/python/test_config.py
```diff
@@ -56,7 +57,11 @@ def test_config_debug_dump_path(specify_path: bool, tmp_path: pathlib.Path):
         print(proc.stderr)
     assert proc.returncode == RET_SUCCESS
     if specify_path:
+        os.system(f"ls {tmp_path}")
```
duburcqa: Why are you calling system commands? I don't think it is a good idea.

python/gstaichi/lang/ast/ast_transformer_utils.py
duburcqa: > nit: is there any easy way we can avoid this relative path import?

Yes, I recommend using relative import. They are not equivalent. First, they are les sensitive to circular import issues. Second, it clearly states that these dependencies are part of your own project.

python/gstaichi/lang/_gstaichi_callable.py
```diff
@@ -0,0 +1,125 @@
...
+        self.wrapper = gstaichi_callable.wrapper
+        self.instance = instance
+        self.gstaichi_callable = gstaichi_callable
+
+    def __call__(self, *args, **kwargs):
+        return self.wrapper(self.instance, *args, **kwargs)
+
+    def __getattr__(self, k: str) -> Any:
+        res = getattr(self.gstaichi_callable, k)
+        return res
```
duburcqa, line 114: Avoid this temporary. It is not free of cost in Python. 

Moreover, it would be more efficient to just do:

```python
class BoundGsTaichiCallable:
    def __init__(self, instance: Any, gstaichi_callable: GsTaichiCallable):
        self.wrapper = gstaichi_callable.wrapper
        self.instance = instance
        self.gstaichi_callable = gstaichi_callable
        self.__getattr__ = self.gstaichi_callable.__getattr__
    
    [...]
```

as it would avoid one extra level of interaction, which is quite slow (like 30ns).

duburcqa: Well, the second trick will not work if `self.gstaichi_callable` does not implement `__getattr__`.

duburcqa: You could probably do:
```python
    def __getattr__(self, k: str) -> Any:
        return self.gstaichi_callable.__dict__[k]
```

But I don't think the performance gain would justify this.

python/gstaichi/lang/func.py
```diff
@@ -0,0 +1,173 @@
...
+        if not self.is_real_function:
+            if self.return_type and ctx.returned != ReturnStatus.ReturnedValue:
+                raise GsTaichiSyntaxError("Function has a return type but does not have a return statement")
+        return ret
+
+    def func_call_rvalue(self, key: FunctionKey, args: tuple[Any, ...]) -> Any:
+        # Skip the template args, e.g., |self|
+        assert self.is_real_function
+        non_template_args = []
+        dbg_info = _ti_core.DebugInfo(impl.get_runtime().get_current_src_info())
```
duburcqa, line 107: If this part of the hot path? If so, it would be great to evaluate this only once if possible.

python/gstaichi/lang/func.py
```diff
@@ -0,0 +1,173 @@
...
+                    non_template_args.append(args[i])
+        non_template_args = impl.make_expr_group(non_template_args)
+        compiling_callable = impl.get_runtime().compiling_callable
+        assert compiling_callable is not None
+        func_call = compiling_callable.ast_builder().insert_func_call(
+            self.cxx_function_by_id[key.instance_id], non_template_args, dbg_info
+        )
+        if self.return_type is None:
+            return None
+        func_call = Expr(func_call)
```
duburcqa, line 131: I like to skip line after early return for readability.

python/gstaichi/lang/func.py
```diff
@@ -0,0 +1,173 @@
...
+        ret = []
+
+        for i, return_type in enumerate(self.return_type):
+            if id(return_type) in primitive_types.type_ids:
+                ret.append(
+                    Expr(
+                        _ti_core.make_get_element_expr(
+                            func_call.ptr, (i,), _ti_core.DebugInfo(impl.get_runtime().get_current_src_info())
```
duburcqa: You already computed `_ti_core.DebugInfo(impl.get_runtime().get_current_src_info())`

python/gstaichi/lang/_kernel_types.py
```diff
@@ -0,0 +1,28 @@
+from dataclasses import dataclass
+from enum import IntEnum
+
+
+class _KernelBatchedArgType(IntEnum):
```
duburcqa, line 5: Moving this enum here is a good idea, as it makes it "official". I would rather keep it "hidden" close to where it is used, more or less like an anonymous struct in C++. But I don't have a strong opinion. Maybe in this case the enum should be renamed to clarify its use case and remove '_' prefix: `KernelLaunchContextBatchedArgType`

python/gstaichi/lang/kernel.py
```diff
@@ -0,0 +1,556 @@
...
+            _, arrs, arrs_grad = zip(*launch_ctx_args)
+            launch_ctx_cache_tracker_ += [ReferenceType(arr, clear_callback) for arr in arrs]
+            launch_ctx_cache_tracker_ += [ReferenceType(arr_grad, clear_callback) for arr_grad in arrs_grad]
+        self._launch_ctx_cache_tracker[args_hash] = launch_ctx_cache_tracker_
+
+    def populate_launch_ctx_from_cache(self, args_hash, launch_ctx) -> tuple[Program, bool]:
+        if self._prog_weakref is None or self._prog_weakref() is None:
+            prog = impl.get_runtime().prog
+            assert prog is not None
+            self._prog_weakref = ReferenceType(prog, self._destroy_callback)
```
duburcqa: This is not ok. You are creating circular references, which would cause `LaunchContextBufferCache` instances to be impossible to garbage collect.

You should do the same as the original code:

```python
class LaunchContextBufferCache:
    [...]
    
    @staticmethod
    def _destroy_callback(ref: ReferenceType):
        self._launch_ctx_cache.clear()
        self._launch_ctx_cache_tracker.clear()
        self._prog_weakref = None
        self.prog = None
   
    [...]

    def populate_launch_ctx_from_cache(self, args_hash, launch_ctx) -> tuple[Program, bool]:
        if self._prog_weakref is None or self._prog_weakref() is None:
            assert prog is not None
            self._prog_weakref = ReferenceType(prog, partial(self._destroy_callback, ReferenceType(self)))
        [...]
```

python/gstaichi/lang/kernel.py
```diff
@@ -0,0 +1,556 @@
...
+        cached_launch_ctx.copy(launch_ctx)
+        self._launch_ctx_cache[args_hash] = cached_launch_ctx
+
+        # Note that the clearing callback will only be called once despite being registered for each tracked
+        # objects, because all the weakrefs get deallocated right away, and their respective callback vanishes
+        # with them, without even getting a chance to get called. This means that registring the clearing
+        # callback systematically does not incur any cumulative runtime penalty yet ensures full memory safety.
+        # Note that it is important to prepend the cache tracker with 'None' to avoid misclassifying no argument
+        # with expired cache entry caused by deallocated argument.
+        launch_ctx_cache_tracker_: list[ReferenceType | None] = []
```
duburcqa: You did not backport my fix.

python/gstaichi/lang/kernel.py
```diff
@@ -0,0 +1,556 @@
...
+            )
+            if _pass == 1:
+                assert key not in self.materialized_kernels
+                self.materialized_kernels[key] = gstaichi_kernel
+
+    def launch_kernel(self, t_kernel: KernelCxx, compiled_kernel_data: CompiledKernelData | None, *args) -> Any:
+        assert len(args) == len(self.arg_metas), f"{len(self.arg_metas)} arguments needed but {len(args)} provided"
+
+        callbacks: list[Callable[[], None]] = []
+        args_hash: ArgsHash = (id(t_kernel), *[id(arg) for arg in args if type(arg) is not template])
```
duburcqa: must back port fix.

python/gstaichi/lang/kernel.py
```diff
@@ -0,0 +1,556 @@
...
+            self._last_compiled_kernel_data = compiled_kernel_data
+            prog.launch_kernel(compiled_kernel_data, launch_ctx)
+        except Exception as e:
+            e = handle_exception_from_cpp(e)
+            if impl.get_runtime().print_full_traceback:
+                raise e
+            raise e from None
+
+        for c in callbacks:
+            c()
```
duburcqa:
```python
for callback in callbacks:
    callback()
```

python/gstaichi/lang/kernel.py
```diff
@@ -0,0 +1,556 @@
...
+
+        self.currently_compiling_materialize_key = None
+
+        return_type = self.return_type
+        if return_type or self.has_print:
+            runtime_ops.sync()
```
duburcqa, line 486: By the way, all the sync in `to_numpy` / `to_torch` should be removed. It is pointless and harming performance.

python/gstaichi/lang/_func_base.py
```diff
@@ -0,0 +1,565 @@
...
+if TYPE_CHECKING:
+    from .kernel import Kernel
+
+MAX_ARG_NUM = 512
+
+# Define proxies for fast lookup
+# _NONE, _REVERSE = (
+#     AutodiffMode.NONE,
+#     AutodiffMode.REVERSE,
+# )
```
duburcqa, line 61: Remove if you don't need.

duburcqa:
Done with review.

tests/python/test_config.py
hughperkins: removed.

hughperkins:
> I would suggest moving all the changes related to debug dump path in a dedicated PR.

Moved to separate PR at https://github.com/Genesis-Embodied-AI/gstaichi/pull/322, and stacked this PR onto that one.

python/gstaichi/lang/_gstaichi_callable.py
hughperkins: removed temporary

python/gstaichi/lang/func.py
hughperkins: I have no idea what this code does. I'm wary of moving it. It's not soething I added or modified in this PR.

hughperkins: oh, it doesnt go into c++, only into python perhaps 🤔 
```
    def get_current_src_info(self):
        return self.src_info_stack[-1]
```

hughperkins: still, I'm not clear when src_info_stack is modified. I'm not clear that it's modifeid during Kernel init. Though perhaps you mean, retrieve it only in this place in this fucntion, b ut cache that?

But anyway, there is no pybuild crossing here. It's just looking up a value on a python list. Should be relatively fast?

python/gstaichi/lang/func.py
hughperkins: addressed in c65d35def

python/gstaichi/lang/func.py
hughperkins: well, the runtime fetch uses pybind. I guess....

hughperkins: oh and the entire DebugInfo bit. I see.

python/gstaichi/lang/func.py
hughperkins: addressed in b9cce8a49

python/gstaichi/lang/_kernel_types.py
hughperkins: It's used across multiple files. I'll remove the _ prefix.

hughperkins: renamed in ff240d010e43e8701d099e0f4c3884c9ad16d2c2 ish

hughperkins: oh wait, I renamed the wrong arg type 😓

hughperkins: Oh, I renamed the right type, but didnt realize you were proposign to rename to KernelLaunchContextBatchedArgType 🤔

hughperkins: I'm not really a fan of long names. Would prefer t he shorter name, just without the _, on the whole.

By the way, I already use this eg in pyr dataclass tests:
```
    assert u1[2] == 0
    kernel_args_count_by_type = k1._primal.launch_stats.kernel_args_count_by_type
    assert kernel_args_count_by_type[_KernelBatchedArgType.TI_ARRAY] == 3
    assert kernel_args_count_by_type[_KernelBatchedArgType.INT] == 1
```

python/gstaichi/lang/kernel.py
hughperkins: updated in 29b0818b4

python/gstaichi/lang/kernel.py
hughperkins: You mean https://github.com/Genesis-Embodied-AI/gstaichi/pull/321 ? Yeah, I'm intending to merge that once it's merged to main. You said the fix is not critical in any case right?

python/gstaichi/lang/kernel.py
hughperkins: See above

python/gstaichi/lang/kernel.py
hughperkins: addressed in db70f3fd1

python/gstaichi/lang/kernel.py
hughperkins: That sounds like out of scope for this PR. Do you mean, we should create a ticket to address this?

python/gstaichi/lang/_func_base.py
hughperkins: addressed in d541ea6bb

python/gstaichi/lang/_kernel_types.py
duburcqa: Well, then just remove _ prefix. The short name is a bit confusing but nothing very concerning.

python/gstaichi/lang/kernel.py
duburcqa: No this one https://github.com/Genesis-Embodied-AI/gstaichi/pull/320. It is critical. I would be surprised if it passes the unit test. All you are missing it `[None]` instead of `[]`

duburcqa: > Yeah, I'm intending to merge that once it's merged to main. 

What do you mean? It has a sizeable impact on performance, but not completely change changer. I'm not a big fan of silently reverting an improvement to add it back later on, but if you have a strong opinion I'm ok.

python/gstaichi/lang/kernel.py
duburcqa: Yes.

python/gstaichi/lang/kernel.py
hughperkins: well, I was responding in the context of 321. Will look at the missing [None] for 320.

hughperkins: addressed in 3f2cea3af

python/gstaichi/lang/kernel.py
hughperkins: For this comment, do you mean 321, or 320? For 321, see my earlier comment, above.

python/gstaichi/lang/_kernel_types.py
hughperkins: addressed in ef8b1a791

python/gstaichi/lang/kernel.py
duburcqa: I mean 321 for this one.

python/gstaichi/lang/_func_base.py
```diff
@@ -0,0 +1,560 @@
...
+                    msg_l.append(f" - {i} arg meta: {self.arg_metas_expanded[i].name} arg type: {type(args[i])}")
+                else:
+                    msg_l.append(f" - {i} arg meta: <out of arg metas> arg type: {type(args[i])}")
+            msg_l.append(f"In function: {self.func}")
+            raise GsTaichiSyntaxError("\n".join(msg_l))
+
+        missing_arg_metas = self.arg_metas_expanded[num_args:]
+        num_missing_args = len(missing_arg_metas)
+        fused_args: list[Any] = [*args, *[arg_meta.default for arg_meta in missing_arg_metas]]
+        if kwargs:
```
duburcqa: https://github.com/Genesis-Embodied-AI/gstaichi/pull/323 is missing

python/gstaichi/lang/_func_base.py
```diff
@@ -0,0 +1,560 @@
...
+            if num_invalid_kwargs_args:
+                for key, value in kwargs.items():
+                    for i, arg_meta in enumerate(self.arg_metas_expanded):
+                        if key == arg_meta.name:
+                            if i < num_args:
+                                raise GsTaichiSyntaxError(f"Multiple values for argument '{key}'.")
+                            break
+                    else:
+                        raise GsTaichiSyntaxError(f"Unexpected argument '{key}'.")
+        elif num_missing_args:
```
duburcqa: #323

python/gstaichi/lang/_func_base.py
hughperkins: good spot. addresed in 74228405c

python/gstaichi/lang/_func_base.py
hughperkins: addressed in 74228405c

hughperkins:
Or maybe we can just do this pr in smaller prs perhaps. I think I might do that.

PR #320, opened 2025-12-17
[Perf] Fix edge-cases defeating caching mechanism

duburcqa:
benchmark:
```python
import genesis as gs

gs.init(backend=gs.cpu)
scene = gs.Scene()
plane = scene.add_entity(gs.morphs.Plane())
franka = scene.add_entity(
    gs.morphs.MJCF(file="xml/franka_emika_panda/panda.xml"),
)
scene.build()
for i in range(20000):
    scene.step()
```

BEFORE:
* field: 6700FPS
* ndarray: 5800FPS

AFTER:
* field: 8200FPS
* ndarray: 6200FPS

### DIFF
--- a/tests/python/test_cache.py
+++ b/tests/python/test_cache.py
@@ -20,7 +20,9 @@ def test_cache_primitive_args():
         else:
             assert "Invalid 'static_args.flag_1' branch"
 
-    assert len(fun._primal.mapper.mapping) == 0
+    assert len(fun._primal.mapper._mapping_cache) == 0
+    assert len(fun._primal.mapper._mapping_cache_tracker) == 0
+    assert len(fun._primal._launch_ctx_cache) == 0
     assert len(fun._primal._launch_ctx_cache_tracker) == 0
 
     static_args = StructStaticConfig()
@@ -30,12 +32,16 @@ def test_cache_primitive_args():
 
     fun(static_args, flag_2, value)
     assert value[None] == 2
-    assert len(fun._primal.mapper.mapping) == 1
+    assert len(fun._primal.mapper._mapping_cache) == 1
+    assert len(fun._primal.mapper._mapping_cache_tracker) == 1
+    assert len(fun._primal._launch_ctx_cache) == 1
     assert len(fun._primal._launch_ctx_cache_tracker) == 1
 
     fun(static_args, flag_2, value)
     assert value[None] == 3
-    assert len(fun._primal.mapper.mapping) == 1
+    assert len(fun._primal.mapper._mapping_cache) == 1
+    assert len(fun._primal.mapper._mapping_cache_tracker) == 1
+    assert len(fun._primal._launch_ctx_cache) == 1
     assert len(fun._primal._launch_ctx_cache_tracker) == 1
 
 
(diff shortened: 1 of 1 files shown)


python/gstaichi/lang/impl.py
```diff
@@ -519,10 +519,10 @@ def materialize(self):
         self._check_gradient_field_not_placed("dual")
         self._check_matrix_field_member_shape()
         self._calc_matrix_field_dynamic_index_stride()
-        self.global_vars = []
-        self.grad_vars = []
-        self.dual_vars = []
-        self.matrix_fields = []
+        self.global_vars.clear()
```
hughperkins, line 522: can you clarify what si the reason taht we need to change from creating a list to clearing the existing one? (not saying we shouldnt, Im just curious)

python/gstaichi/lang/_template_mapper.py
```diff
@@ -79,13 +84,17 @@ def lookup(self, raise_on_templated_floats: bool, args: tuple[Any, ...]) -> tupl
...
-            mapping_cache_tracker_ += [ReferenceType(arg, clear_callback) for arg in args]
+            mapping_cache_tracker_ += [
+                ReferenceType(arg, clear_callback) for arg in args if type(arg) not in _primitive_types
+            ]
             self._mapping_cache_tracker[args_hash] = mapping_cache_tracker_
             self._mapping_cache[args_hash] = (count, key)
-        except TypeError:
-            pass
+        except TypeError as e:
+            warnings_helper.warn_once(f"{e}. Template mapper caching disabled.")
```
hughperkins, line 98: I take it is was siliently disabled, and this took a while to track down 😅

hughperkins (approved):
Awesome! 🙌

python/gstaichi/lang/_template_mapper.py
duburcqa: Indeed x) I should have been more thoughtful when I implemented this caching mechanism in the first place!

python/gstaichi/lang/impl.py
duburcqa: Yes! First, it is slightly faster. Second, I don't like continuously reallocated memory when there is a dedicated in-place function that does the same thing. Third, more generally, it helps to share data efficiently: if for some reason we when to get a reference to these list at some point, now it is doable.

duburcqa: Note that the true motivation in this case is the first argument.

hughperkins: Thanks! 🙌

python/gstaichi/lang/_template_mapper.py
hughperkins: :)

duburcqa:
Non template integers are not compatible with kernel caching because their value is hard coded in cache context.

As a result, we avoid them at all costs currently in genesis, we don’t have any.

hughperkins:
> > You know, it would be good to have some unit tests
> 
> At least we are checking that we do not break anything :p

True. But I think that sooner or later would be good to have some tests on this cache mechanism
1. means we are sure it is working now
2. ensures I dont break it accidentally in the future

PR #321, opened 2025-12-17
[Perf] Speed up computation of cache key.

duburcqa:
If stacked on top of https://github.com/Genesis-Embodied-AI/gstaichi/pull/323:

Before: 9100 FPS, After: 9500 FPS

### DIFF
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -1252,7 +1252,7 @@ class Kernel:
         launch_ctx = t_kernel.make_launch_context()
         launch_ctx_cache: KernelLaunchContext | None = None
         launch_ctx_cache_tracker: list[ReferenceType | None] | None = None
-        # Special treatment for primitive types is unecessary and detrimendal. See 'TemplateMapper.lookup' for details.
+        # Special treatment for primitive types is unecessary and detrimental. See 'TemplateMapper.lookup' for details.
         args_hash: ArgsHash = (id(t_kernel), *[id(arg) for arg in args])
         try:
             launch_ctx_cache_tracker = self._launch_ctx_cache_tracker[args_hash]


hughperkins:
Thanks! I'm curiosu why this wasn't picked up by the unit tests? Is there something we need to tweak on the unit tests that would have detected this?

hughperkins (approved):
Thank you!

duburcqa:
> Thanks! I'm curiosu why this wasn't picked up by the unit tests? Is there something we need to tweak on the unit tests that would have detected this?

Because I don't think the current version is buggy. Just more restrictive than it should. I will write a unit test to check that we are just as restrictive as we should.

duburcqa:
Update to add more unit tests.

hughperkins:
> > Thanks! I'm curiosu why this wasn't picked up by the unit tests? Is there something we need to tweak on the unit tests that would have detected this?
> 
> Because I don't think the current version is buggy. Just more restrictive than it should. I will write a unit test to check that we are just as restrictive as we should.

Thank you! 🙌

erizmr (approved):

hughperkins:
Note: @duburcqa  I've set this to auto-merge. If this isn't what you intend, then please cancel that :)

duburcqa:
> Note: @duburcqa I've set this to auto-merge. If this isn't what you intend, then please cancel that :)

I wanted you to do it yourself because this PR changed quite a bit since your initial approval.

hughperkins:
Oh it did? 🤔 Oh ok, I might cancel the auto merge and recheck then.

python/gstaichi/lang/kernel_impl.py
```diff
@@ -1254,7 +1254,8 @@ def launch_kernel(self, t_kernel: KernelCxx, compiled_kernel_data: CompiledKerne
         launch_ctx = t_kernel.make_launch_context()
         launch_ctx_cache: KernelLaunchContext | None = None
         launch_ctx_cache_tracker: list[ReferenceType | None] | None = None
-        args_hash: ArgsHash = (id(t_kernel), *[id(arg) for arg in args if type(arg) is not template])
+        # Special treatment for primitive types is unecessary and detrimendal. See 'TemplateMapper.lookup' for details.
```
hughperkins: detrimental

PR #324, opened 2025-12-18
[Perf] Early return earlier when materializing kernels.

duburcqa:
BEFORE: 9500FPS, AFTER: 10400FPS

### DIFF
--- a/python/gstaichi/lang/kernel_impl.py
+++ b/python/gstaichi/lang/kernel_impl.py
@@ -1066,13 +1066,13 @@ class Kernel:
     def materialize(self, key: CompiledKernelKeyType | None, args: tuple[Any, ...], arg_features=None):
         if key is None:
             key = (self.func, 0, self.autodiff_mode)
-        self.runtime.materialize()
         self.fast_checksum = None
-
         self.currently_compiling_materialize_key = key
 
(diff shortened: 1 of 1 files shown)


python/gstaichi/lang/kernel_impl.py
```diff
@@ -1066,14 +1066,14 @@ def extract_arguments(self) -> None:
...
             key = (self.func, 0, self.autodiff_mode)
-        self.runtime.materialize()
         self.fast_checksum = None
-
         self.currently_compiling_materialize_key = key
 
         if key in self.materialized_kernels:
             return
 
+        self.runtime.materialize()
```
hughperkins, line 1075: Can you provide some insight into why this accelerates? Is it because runtime.materialize() cross pybind11 boundary?

duburcqa: Yes, but also because `materialize` is very complex, it calls like 8 functions, each of them calling functions and functions...etc

hughperkins:
Looks good. Will approve once https://github.com/Genesis-Embodied-AI/gstaichi/pull/319 is merged.

hughperkins (approved):
Thanks! 🙌

PR #325, opened 2025-12-18
[Perf] Cache compiled kernel data systematically.

duburcqa:
Benchmark after rebased on https://github.com/Genesis-Embodied-AI/gstaichi/pull/324

BEFORE: 10400FPS, AFTER: 13500FPS

PR #329, opened 2025-12-19
[Misc] Factorize out launch context buffer cache

hughperkins:
Issue: #

### Brief Summary

anymal ndarray cpu mac:
```
	env=anymal_c 	| batch_size=0 	| use_contact_island=False 	| dtype=ndarray 	| backend=cpu 	| compile_time=17.3 	| runtime_fps=7670.0 	| realtime_factor=76.7
```
compared to on #325 :
```
mac origin/main (#325)
    env=anymal_c    | batch_size=0  | use_contact_island=False  | dtype=ndarray     | backend=cpu   | compile_time=17.3     | runtime_fps=7694.0    | realtime_factor=76.9
```
7670 vs 7694, seems ~similar.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/kernel.py
+++ b/python/gstaichi/lang/kernel.py
@@ -1,15 +1,9 @@
 import ast
-import inspect
 import json
 import os
 import pathlib
 import time
-import types
-import typing
 from collections import defaultdict
-from dataclasses import (
-    is_dataclass,
-)
 
 # Must import 'partial' directly instead of the entire module to avoid attribute lookup overhead.
 from functools import partial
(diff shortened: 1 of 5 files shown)


duburcqa (approved):

python/gstaichi/lang/kernel.py
```diff
@@ -71,6 +71,104 @@
...
+        try:
+            launch_ctx_cache_tracker = self._launch_ctx_cache_tracker[args_hash]
+        except KeyError:
+            pass
+        if not launch_ctx_cache_tracker:  # Empty or none
+            return prog, False
+
+        assert args_hash is not None
+        launch_ctx.copy(self._launch_ctx_cache[args_hash])
+        return prog, True
```
duburcqa: Do not return prog. Just do `self._prog_weakref()` right after calling `populate_launch_ctx_from_cache`

hughperkins: the logic to manage the prog weakref takes up ~6-7 lines or so. I'd prefer to keep that out of the launch_kernel method if we can, ideally.

I'm happy to factorize that into a separate class though (can be a separate PR), and then pass prog from launch kernel into populate_launch_ctx_from_cache.

Just to check, when you say do `self._prog_weakref()`, you are imaginging storing prog_weakref in Kernel object, rather than in the kernel launch context buffer cache?

erizmr (approved):

PR #333, opened 2025-12-23
[Type] Can rename py dataclass structs when calling sub functions

hughperkins:
Issue: #

### Brief Summary

Key concept is that we need to store, during compilation:
- the used parameters per func, not just globally
- the relationship between parameters in one func, and in each func that it is calling

We do this in the new pruning.py file, with each dict indexed first by func_id, which is unique per function.

We assume that the parameters used for a func are consistent across all calls to that func within the same compiled kernel. However, different calls to the same kernel can have different used parameters for each func in that kernel call, compared to other calls into the same kernel.

We only need to store the used parameters during the compilation of the kernel. Once the kernel is compiled, we throw away the function-specific lists of parameters - by discarding the ASTTransformer*Context objects - and store only the parametesr needed on entry to the top level kernel. We store these parameters in the Kernel object, indexed by calling key.

Old arch:

Screenshot 2025-12-23 at 13 53 35

New arch:

Screenshot 2025-12-23 at 13 53 40

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/func.py
+++ b/python/gstaichi/lang/func.py
@@ -74,7 +74,6 @@ class Func(FuncBase):
             if key.instance_id not in self.compiled:
                 self.do_compile(key=key, args=py_args, arg_features=arg_features)
             return self.func_call_rvalue(key=key, args=py_args)
-
         tree, ctx = self.get_tree_and_ctx(
             is_kernel=False,
             py_args=py_args,


hughperkins:
Hmmm, error on anymal:
```
E               return self.wrapper.__call__(*args, **kwargs)
E             File "/Users/hugh/git/gstaichi/python/gstaichi/lang/func.py", line 59, in __call__
E               py_args = self.process_args(
E             File "/Users/hugh/git/gstaichi/python/gstaichi/lang/_func_base.py", line 335, in process_args
E               raise GsTaichiSyntaxError(f"Unexpected argument '{key}'.")
E           gstaichi.lang.exception.GsTaichiSyntaxError: Unexpected argument '__ti_dofs_state__ti_cdofd_vel'.
E
E           ===
E
E           Your code:
E           File "/Users/hugh/git/genesis/genesis/engine/solvers/rigid/rigid_solver_decomp.py", line 4617, in kernel_forward_kinematics_links_geoms:
E                   func_update_cartesian_space(
E                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                       i_b=i_b,
E                       ^^^^^^^^
E                       links_state=links_state,
E                       ^^^^^^^^^^^^^^^^^^^^^^^^
E                       links_info=links_info,
E                       ^^^^^^^^^^^^^^^^^^^^^^
E                       joints_state=joints_state,
E                       ^^^^^^^^^^^^^^^^^^^^^^^^^^
E                       joints_info=joints_info,
E                       ^^^^^^^^^^^^^^^^^^^^^^^^
E                       dofs_state=dofs_state,
E                       ^^^^^^^^^^^^^^^^^^^^^^
E                       dofs_info=dofs_info,
E                       ^^^^^^^^^^^^^^^^^^^^
E                       geoms_info=geoms_info,
E                       ^^^^^^^^^^^^^^^^^^^^^^
E                       geoms_state=geoms_state,
E                       ^^^^^^^^^^^^^^^^^^^^^^^^
E                       entities_info=entities_info,
E                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                       rigid_global_info=rigid_global_info,
E                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                       static_rigid_sim_config=static_rigid_sim_config,
E                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                       force_update_fixed_geoms=True,
E                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E                       is_backward=False,
E                       ^^^^^^^^^^^^^^^^^^
E                   )
E                   ^
E           Unexpected argument '__ti_dofs_state__ti_cdofd_vel'.

../gstaichi/python/gstaichi/lang/kernel_impl.py:140: GsTaichiSyntaxError
---------------------------------------------------------------------------- Captur
```

hughperkins:
Runs on anymal now, conditional on Genesis PR at https://github.com/Genesis-Embodied-AI/Genesis/pull/2143/files 

```
	env=anymal_c 	| batch_size=0 	| use_contact_island=False 	| dtype=ndarray 	| backend=cpu 	| compile_time=17.7 	| runtime_fps=7463.0 	| realtime_factor=74.6
```

7463fps. Seems ok-ish?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -268,22 +265,44 @@ def build_Call(ctx: ASTTransformerFuncContext, node: ast.Call, build_stmt, build
         example ast:
         Call(func=Name(id='f2', ctx=Load()), args=[Name(id='my_struct_ab', ctx=Load())], keywords=[])
         """
+        is_func_base_wrapper = False
+        # split them off into locals, otherwise iterated function calls, with different pruning,
+        # will break, if we modify node.args/node.keywords in-place
+        args, keywords = node.args, node.keywords
```
hughperkins, line 271: tempted to rename these node_args and node_keywords, like I do in pruning.py

hughperkins: (I suppose another approach could be to back up the old ones, at the start of the function, then restore them at the end. Both approaches have good/bad points.

Another approach could be, when there is a for loop over static something, duplicate the entire contents of the for loop for each iteration. But that sounds like it might have performance implications, for compilation time, perhaps).

python/gstaichi/lang/_func_base.py
```diff
@@ -247,78 +264,110 @@ def get_tree_and_ctx(
...
-                current_kernel.used_py_dataclass_leaves_by_key_enforcing.get(currently_compiling_materialize_key),
+                used_by_dataclass_parameters_enforcing,
                 self.arg_metas,
             )
         else:
             self.arg_metas_expanded = list(self.arg_metas)
 
-        num_args = len(args)
-        num_arg_metas = len(self.arg_metas_expanded)
+        arg_metas_pruned = self.arg_metas_expanded
```
hughperkins: should probably remove this

python/gstaichi/lang/_pruning.py
```diff
@@ -0,0 +1,185 @@
...
+            dotted_by_func_id[func_id] = set([tuple(p.split("__ti_")[1:]) for p in used_parameters])
+        self.dotted_by_func_id = dotted_by_func_id
+
+    def record_after_call(
+        self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCallable", node, node_args, node_keywords
+    ) -> None:
+        """
+        called from build_Call, after making the call, in pass 0
+
+        note that this ONLY handles args, not kwargs
```
hughperkins: thsi comments seems not accurate.

hughperkins:
> * Make sure we do not call the same func with different template parameters in the same high-level taichi kernel, because this is not supported
> * Add proper exception with clean guideline instead of assert for *arg being required to be the last input argument

Addressed both these points.

PR #341, opened 2026-01-09
[MISC] Improve information in exceptions during fuse_args

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_func_base.py
+++ b/python/gstaichi/lang/_func_base.py
@@ -257,7 +257,7 @@ class FuncBase:
             is_real_function=is_real_function,
             autodiff_mode=autodiff_mode,
             raise_on_templated_floats=raise_on_templated_floats,
-            used_py_dataclass_parameters_collecting=current_kernel.used_py_dataclass_leaves_by_key_collecting[
+            used_py_dataclass_parameters_collecting=current_kernel.used_py_dataclass_parameters_by_key_collecting[
                 args_instance_key
             ],
             used_py_dataclass_parameters_enforcing=used_py_dataclass_parameters_enforcing,
@@ -302,7 +302,7 @@ class FuncBase:
             current_kernel = global_context.current_kernel
             assert current_kernel is not None
             self.arg_metas_expanded = _kernel_impl_dataclass.expand_func_arguments(
-                current_kernel.used_py_dataclass_leaves_by_key_enforcing.get(
+                current_kernel.used_py_dataclass_parameters_by_key_enforcing.get(
                     global_context.currently_compiling_materialize_key
                 ),
                 self.arg_metas,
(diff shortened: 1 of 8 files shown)


python/gstaichi/lang/_func_base.py
```diff
@@ -339,21 +340,26 @@ def fuse_args(
...
+                        errors_l.append(f"Unexpected argument '{key}'.")
         else:
             for i in range(num_args, num_arg_metas):
                 if fused_py_args[i] is _ARG_EMPTY:
                     arg_meta = self.arg_metas_expanded[i]
-                    raise GsTaichiSyntaxError(f"Missing argument '{arg_meta.name}'.")
+                    errors_l.append(f"Missing argument '{arg_meta.name}'.")
+                    continue
+
+        if len(errors_l) > 0:
```
duburcqa: `if errors_l:` is faster

duburcqa: The current implementation is not designed to support this. Each successive block is assuming everything went well on the previous one. Postponing the error may lead to wrongly classified errors.

I'm not in favour of complexifying the default codepath just to improve error reporting, in a way that does not align with what Python itself is doing by default when such error occurs. This code path is under extreme pressure and can easily induce 20% slowdown for non-batched cpu-based simulation if not done properly.

What you could do is using the flag `__debug__` to determine whether you are running in debug mode of not. This way it would be optimal in most cases. See: https://github.com/Genesis-Embodied-AI/gstaichi/blob/5cc46bca9f2280e500e0f07053027308272a4eff/python/gstaichi/lang/_template_mapper_hotpath.py#L22

duburcqa: If you can manage to achieve what you want (i.e. relax "successive block is assuming everything went well on the previous one") without slowing down the usual codepath then I'm all for it, but I doubt it is possible.

hughperkins: "Postponing the error may lead to wrongly classified errors." => possibly, but errors_l will be greater than zero length, so an error will be thrown.

Note that almost all changes are to code paths throwing exceptions anyway. The only changes to non-exception code path are:
```
errors_l = []
```
(creating an empty list)
And:
```
if len(errors_l) > 0:
```
(checking the length of an empty list).

duburcqa: > possibly, but errors_l will be greater than zero length, so an error will be thrown.

Sure, but a possibly wrong error report is not better than a partial one, or is it?

> Note that almost all changes are to code paths throwing exceptions anyway. 

I'm not afraid that your modified implementation slows things down. I'm afraid that refactoring the code so that catching all issues and reporting all of them would make things slower.

PR #345, opened 2026-01-09
[Type] Implement basic func local pruning kwargs only

hughperkins:
Issue: #

### Brief Summary

subset of https://github.com/Genesis-Embodied-AI/gstaichi/pull/344/files#diff-9a7de088688708a1efe6739cafaabb3b8fc484a0676f52147e6b1e7c522ce841

- but: remove args support

In more detail:
- add Pruning class/object
    - store in Global Context, during ast transformer walk
    - contains set of per-function used parameters

How the pruning works:
- populate the used parameters in build_Name, as before, but using hte _pruning object, and specific to the current function, via func_id
- copy used parameters from called function into current function, also inside Pruning object
- Kernel object still stores used parameters, but used only at launch time, not during ast transformer walk time
     - fastcache will contain this set too

We still have two passes as before:
- in pass 0, we populate the used parameters, ~and the mapping~
- in pass 1, we enforce the used parameters, to do the actual pruning
- we set a boolean parameter on the global context, to switch between the two modes.

Pruning object indexes the used parameters by func id
- for kernel, the func is set to -1

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_tools_markdown_check.py
```diff
@@ -6,7 +6,7 @@
 import pytest
 
 # I don't want to make python/tools a module, and I don't want to move this tool
-# into `gstaichi` namespace, so that parameters temporarily importing it somehow
+# into `gstaichi` namespace, so that leaves temporarily importing it somehow
```
hughperkins, line 9: yes, the search replace from leaves => parameters earlier was a bit too aggressive :P

tests/python/test_tools_run_modified_files_integration.py
```diff
@@ -7,7 +7,7 @@
 import pytest
 
 # I don't want to make python/tools a module, and I don't want to move this tool
-# into `gstaichi` namespace, so that parameters temporarily importing it somehow
+# into `gstaichi` namespace, so that leaves temporarily importing it somehow
```
hughperkins, line 10: search replace from leaves => parameters earlier was a bit too aggressive

python/gstaichi/lang/kernel.py
```diff
@@ -398,16 +396,30 @@ def materialize(self, key: "CompiledKernelKeyType | None", py_args: tuple[Any, .
...
             )
-            used_py_dataclass_parameters = self.used_py_dataclass_parameters_by_key_collecting[key]
             gstaichi_kernel = impl.get_runtime().prog.create_kernel(
                 gstaichi_ast_generator, kernel_name, self.autodiff_mode
             )
             if _pass == 1:
                 assert key not in self.materialized_kernels
                 self.materialized_kernels[key] = gstaichi_kernel
+            else:
+                for used_parameters in pruning.used_parameters_by_func_id.values():
```
hughperkins, line 400: I wonder if this paragraph should be extracted into a separate method?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -212,7 +209,9 @@ def _expand_Call_dataclass_args(
 
     @staticmethod
     def _expand_Call_dataclass_kwargs(
-        ctx: ASTTransformerFuncContext, kwargs: list[ast.keyword]
+        ctx: ASTTransformerFuncContext,
+        kwargs: list[ast.keyword],
+        called_needed: set[str] | None,
```
duburcqa: What is `called_needed` ? I only know the terminology `caller` and `callee`. It seems that `called_needed` here refers to `callees`, but it is not consistent with `child_name not in called_needed` I think...

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -212,7 +209,9 @@ def _expand_Call_dataclass_args(
 
     @staticmethod
     def _expand_Call_dataclass_kwargs(
```
duburcqa, line 211: Uppercase in the name of this function?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -282,10 +284,27 @@ def build_Call(ctx: ASTTransformerFuncContext, node: ast.Call, build_stmt, build
             build_stmts(ctx, node.keywords)
 
             func = node.func.ptr
+            func_type = type(func)
+            is_func_base_wrapper = func_type is GsTaichiCallable or func_type is BoundGsTaichiCallable
```
duburcqa: Note that you could do `is_func_base_wrapper = func_type in {GsTaichiCallable, BoundGsTaichiCallable}` if you want, because for set it requires on hash, which corresponds to identity if not customised.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: Why not just `called_args` ? They are known to be needed by definition, otherwise they would not be called in the first place?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: Why not to move:
```python
            func = node.func.ptr
            func_type = type(func)
            is_func_base_wrapper = func_type is GsTaichiCallable or func_type is BoundGsTaichiCallable
```

on top, outside the if-else clauses?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -282,10 +284,27 @@ def build_Call(ctx: ASTTransformerFuncContext, node: ast.Call, build_stmt, build
...
+            _pruning = ctx.global_context.pruning
```
duburcqa: Why the `_` ? it usually means that this variable is unused but kept for "documentation purposes". You avoid name shadowing, I recommend, `_` suffix. But in this context `pruning` would be just fine because there is no shadowing whatsoever?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -282,10 +284,27 @@ def build_Call(ctx: ASTTransformerFuncContext, node: ast.Call, build_stmt, build
...
+                called_needed = _pruning.used_parameters_by_func_id[_called_func_id]
+
             added_args, node.args = CallTransformer._expand_Call_dataclass_args(ctx, node.args)
-            added_keywords, node.keywords = CallTransformer._expand_Call_dataclass_kwargs(ctx, node.keywords)
+            added_keywords, node.keywords = CallTransformer._expand_Call_dataclass_kwargs(
+                ctx, node.keywords, called_needed
+            )
 
             # create variables for the now-expanded dataclass members
+            # we don't want to include these in the list of variables to not prune
```
duburcqa: "to include these" -> these what?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -282,10 +284,27 @@ def build_Call(ctx: ASTTransformerFuncContext, node: ast.Call, build_stmt, build
...
+            # since these will contain *all* declared fields, instead of all used fields
+            # so we set expanding_dataclass_call_parameters to True during this expansion
+            # We don't build these until this point in the function, because we need
+            # to prune when enforcing is True, and to do that, we need to know what to prune,
+            # which we get from the child func, and so we want to have filtered out all the
+            # cases first where we aren't calling a ti.func
```
duburcqa: I'm sorry but I did not understand anything to this comment x)

duburcqa: I get the idea, but for this 1 line would be enough. Beyond that, the details do not help I'm afraid :/

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -346,6 +365,9 @@ def build_Call(ctx: ASTTransformerFuncContext, node: ast.Call, build_stmt, build
         CallTransformer._warn_if_is_external_func(ctx, node)
         try:
             node.ptr = func(*py_args, **py_kwargs)
+            _pruning = ctx.global_context.pruning
+            if not _pruning.enforcing:
+                _pruning.record_after_call(ctx, func, node)
```
duburcqa: Do as you want, but I don't think calling `ctx.global_context.pruning` systematically is problematic here? I don't know how much of a bottleneck is this part of the code but doubly nested attribute lookup should add like 20ns overhead, nothing more.

python/gstaichi/lang/_pruning.py
```diff
@@ -0,0 +1,94 @@
...
+
+    from ._gstaichi_callable import GsTaichiCallable
+    from .ast.ast_transformer_utils import ASTTransformerFuncContext
+
+
+class Pruning:
+    """
+    Note: this assumes when compiling a kernel that each function will only have
+    one set of used parameters within the compiled kernel, even if called in multiple
+    places.
```
duburcqa: I thought you said this was actually not the case. This restriction is problematic, so it should at least be a warning, not a simple note.

python/gstaichi/lang/_pruning.py
```diff
@@ -0,0 +1,94 @@
...
+
+    To be clear, there is no restriction that a function needs to have the same set of
+    used parameters between kernels, or between calls to the same kernel.
+
+    This assumption allows us to use the func id to uniquely identify each kernel, without
+    some additional index based on used parameters or similar.
+
+    Note that we unify handling of func and kernel by using func_id -1 to denote kernel.
```
duburcqa: I would recommend 0, in a nullptr fashion.

python/gstaichi/lang/_pruning.py
```diff
@@ -0,0 +1,94 @@
...
+    def mark_used(self, func_id: int, parameter_flat_name: str) -> None:
+        """
+        func_id -1 means kernel
+        """
+        assert not self.enforcing
+        self.used_parameters_by_func_id[func_id].add(parameter_flat_name)
+
+    def enforce(self) -> None:
+        self.enforcing = True
+        self._calc_dotted()
```
duburcqa: Why not inline `_calc_dotted` since `enforce` is basically just `_calc_dotted`?

python/gstaichi/lang/_pruning.py
```diff
@@ -0,0 +1,94 @@
...
+            - desigend to not conflict with customer variable names
+            - cannot contain "."
+        - tuple of str, like ("struct", "some_member")
+            - named "dotted", for historical reasons. We should probably rename...
+
+        The latter is called "dotted" because the format used to be "struct.some_member", but
+        was changed to tuple notation. It's used in _recursive_set_args method, as parameter
+        used_py_dataclass_parameters.
+
+        For speed we pre-calculate dotted here.
```
duburcqa: The speed of what?

python/gstaichi/lang/_pruning.py
```diff
@@ -0,0 +1,94 @@
...
+    def enforce(self) -> None:
+        self.enforcing = True
+        self._calc_dotted()
+
+    def is_used(self, func_id: int, parameter_flat_name: str) -> bool:
+        return parameter_flat_name in self.used_parameters_by_func_id[func_id]
+
+    def _calc_dotted(self) -> None:
+        """
+        There are two formats we need:
```
duburcqa: What and where there is 2 formats? In AST transformers?

python/gstaichi/lang/_pruning.py
```diff
@@ -0,0 +1,94 @@
...
+        self,
+        ctx: "ASTTransformerFuncContext",
+        func: "GsTaichiCallable",
+        node,
+    ) -> None:
+        """
+        called from build_Call, after making the call, in pass 0
+        """
+        if not hasattr(func, "wrapper"):
+            return
```
duburcqa: I don't like this. There is no better way to avoid `hasattr`? Using `type(func) in {...}` typically?

python/gstaichi/lang/_pruning.py
duburcqa: "denote kernel." -> Denote THE top-level kernel. There is only one kernel involved in this context. This is why id collapse is not a problem.

python/gstaichi/lang/kernel.py
duburcqa: As you want. I'm fine with it.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: > What is called_needed ? I only know the terminology caller and callee. It seems that called_needed here refers to callees, but it is not consistent with child_name not in called_needed I think...

It means, the called function needs these parameters.

Feel free to propose a better name.

> Why not just called_args ? They are known to be needed by definition, otherwise they would not be called in the first place?

You can call with the args without the called function needing them. In
fact, this is what we do, when pruning is turned off.

"Needed" is much stronger than "called".

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: Call is a node class name. This function is called from build_Call, which
is also named after the node class name, and is called from
ast_transformer_utils

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: I actually wrote these lines of doc for myself, and all of these lines are
needed by me. I wrote this doc as I was trying to figure out why I had
written this code earlier, in this way, a few weeks ago.

python/gstaichi/lang/_pruning.py
hughperkins: I prefer to factorize out functions on the whole. Makes it easier for me to read what is happening.

python/gstaichi/lang/_pruning.py
hughperkins: I will update the comment.

python/gstaichi/lang/_pruning.py
hughperkins: Runtime speed. Kernel launch latency. I can update teh comment to clarify.

python/gstaichi/lang/_pruning.py
hughperkins: The tuple format is used by set recursive args. Everywhere else uses the
name of the expanded variable.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: Pruning object is used to call an additional method in a later pr.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: "these now-expaned dataclass members". I can upudate the comment if you want?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: It means "I'm too lazy to figure out whether it collides with anything" 😅 I can remove the `_`.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: `func = node.func.ptr` isn't valid until we do `build_stmt(ctx, node.func)`

hughperkins: (like, built_stmt populates ,`.ptr`)

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: well, actually, it means "this is a really long function, and I don't wnat to imply that people can use this variable outside of this tiny block". Generally it's a symptom that the containing function should probably b e fatorized a bit, I feel. I'll remove the `_` anyway

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: removed the last 4 lines which were no longer valid. (the risk of doc strings / comments in general...)

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: updated comment

python/gstaichi/lang/_pruning.py
hughperkins: updated comment

python/gstaichi/lang/_pruning.py
hughperkins: updated comment

python/gstaichi/lang/_pruning.py
hughperkins: updated comment

python/gstaichi/lang/_pruning.py
```diff
@@ -0,0 +1,97 @@
...
+        was changed to tuple notation. It's used in _recursive_set_args method, as parameter
+        used_py_dataclass_parameters.
+
+        To optimize runtime speed we pre-calculate dotted here. (I tried calculating it on the fly,
+        and it was noticeably slower, so I took the time to pre-calculate/cache it instead)
+        """
+        assert self.enforcing
+        dotted_by_func_id = {}
+        for func_id, used_parameters in self.used_parameters_by_func_id.items():
+            dotted_by_func_id[func_id] = set([tuple(p.split("__ti_")[1:]) for p in used_parameters])
```
erizmr: Do we want to implement a `_flat_to_dotted` helper for this `set([tuple(p.split("__ti_")[1:]) for p in used_parameters])`? And do we ensure that every `p` in `used_parameters` has "__ti_"?

hughperkins: @duburcqa opinions on this? I don't mind either way really. I lean mildly towards, since it fits on one line anyway, maybe it's clearer to write it out explicitly?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: I see. Not very standard but ok.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: then `needed_args` or `used_args` ?

duburcqa: No matter what? Could you document what this parameter corresponds to in the docstring?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: I see. Still. Is this really a bottleneck? Otherwise just going through `ctx.global_context.pruning` sounds fine to me. As you want.

python/gstaichi/lang/_pruning.py
duburcqa: Wait, so within a single kernel call it is still not supported? This may be problematic. Let me double check.

hughperkins: I think maybe you are not reading the rest of the comment https://github.com/Genesis-Embodied-AI/gstaichi/pull/345/files#diff-62749e2157341e37ccef443171f78e734d9f6919520e13b6035655a90983102fR13-R32

duburcqa: From what I see we are lucky and not hitting this issue. Would it be possible to detect this conflict and raise an exception? If not, I guess I can live with it. Not very common. But still it would be nice.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: It's higher cognitive load to read I feel. It will make more sense when both fucntion calls are included, in later PR.

python/gstaichi/lang/_pruning.py
duburcqa: > I think maybe you are not reading the rest of the comment 

Yes I did read the entire updated comment.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: `needed_args` sounds good to me.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: Yeah, that is what I guessed. Could you do the same for the other places where `_pruning` is used?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: or maybe used_args

hughperkins: (put used_args)

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: > I actually wrote these lines of doc for myself, and all of these lines are needed by me. I wrote this doc as I was trying to figure out why I had written this code earlier, in this way, a few weeks ago.

In that case, it would be great if it could be understood by everybody having a vague idea of what gstaichi is, like me?

python/gstaichi/lang/_pruning.py
hughperkins: We are lucky that the implementation solve s the issue.

I will update the comment to frame it as by design.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: Could you elaborate? I don't understand? Then you `_` suffix as I suggest. `_` prefix means unused in the many project (even ruff will interpret it this way from what I know).

python/gstaichi/lang/_pruning.py
duburcqa: I don't understand? It is just twice the same function with different names. How merging them hurts factorisation?

python/gstaichi/lang/_pruning.py
hughperkins: updated in c2b4d4737

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: updated

python/gstaichi/lang/_pruning.py
duburcqa: I think it would be nice to clarify this, or cross-reference where it is introduced if already explained somewhere else.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: "removed the last 4 lines which were no longer valid. (the risk of doc strings / comments in general...)"

The comments now look liek:

"            # Create variables for the now-expanded dataclass members.
            # We don't want to include these now-expanded dataclass members
            # in the list of variables to not prune,
            # since these will contain *all* declared fields, instead of all used fields.
            # So we set expanding_dataclass_call_parameters to True during this expansion."

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: there are two function calls: https://github.com/Genesis-Embodied-AI/gstaichi/pull/333/files#diff-9a7de088688708a1efe6739cafaabb3b8fc484a0676f52147e6b1e7c522ce841R367-R374

python/gstaichi/lang/_pruning.py
duburcqa: I'm fine with explicit but no strong opinion. I don't think checking is necessary because it is internal and we are already ensuring by design that __ti_ is properly added.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: (I've removed the `_` suffix in 0c4abf27fc72df3c092 )

python/gstaichi/lang/_pruning.py
hughperkins: This function doesnt exist any more. You ar elooking at outdated code.

python/gstaichi/lang/_pruning.py
hughperkins: This function no longer exists any more. I removed it on sunday morning, because you said that 1.3% runtime regression seems ok to you.

python/gstaichi/lang/_pruning.py
hughperkins: this code no longer exists. it was removed on sunday morning.

python/gstaichi/lang/_pruning.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCallable", node: "ast.Call") -> None:
+        """
+        called from build_Call, after making the call, in pass 0
+        """
+        if type(func) not in {GsTaichiCallable, BoundGsTaichiCallable}:
+            return
+
+        _my_func_id = ctx.func.func_id
+        _called_func_id = func.wrapper.func_id  # type: ignore
```
duburcqa: Same issue with `_` prefix that is commonly reserved for unused variables by convention.

python/gstaichi/lang/kernel.py
```diff
@@ -357,34 +358,23 @@ def materialize(self, key: "CompiledKernelKeyType | None", py_args: tuple[Any, .
...
 
         self.runtime.materialize()
-        used_py_dataclass_parameters = self._try_load_fastcache(py_args, key)
+        _used_py_dataclass_parameters = self._try_load_fastcache(py_args, key)
         kernel_name = f"{self.func.__name__}_c{self.kernel_counter}_{key[1]}"
         _logging.trace(f"Materializing kernel {kernel_name} in {self.autodiff_mode}...")
 
-        range_begin = 0 if used_py_dataclass_parameters is None else 1
+        pruning = Pruning(kernel_used_parameters=_used_py_dataclass_parameters)
+        range_begin = 0 if _used_py_dataclass_parameters is None else 1
```
duburcqa: `_` prefix.

python/gstaichi/lang/_pruning.py
duburcqa: Ok I did not remember it was merged.

python/gstaichi/lang/_pruning.py
duburcqa: Yes it was. Did not remember it was merged.

python/gstaichi/lang/_pruning.py
hughperkins: It was not a PR. I chnged direclty on this PR.

hughperkins: since this PR was the trigger about questions/objections to using dotted notation, and is the reason i ran the experimetn.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: I still don't understand, sorry.

duburcqa: "We don't want to include these now-expanded dataclass members in the list of variables to not prune, since these will contain all declared fields, instead of all used fields."

-> these will contain all declared fields if what, you to include them or no matter what? No clear

hughperkins: when we build ast nodes that use some variable, at some point, we'll call build_Name on an ast.Name node. the build_Name function then calls pruning.mark_used on that variable, as long as the variable name starts with __ti_, and we are not yet in enforicing mode (which freezes the used names, does not allow new names to be added).

build_stmt and build_stmts are wrappers around ASTTransformer, that call into the build function in ast_transformer_utils, which builds a node and its children.

When we build the now-expanded dataclass members, we do NOT want them to be added to the list of 'used' variables. Otherwise they'll never be pruned. Since *all* dataclass members will be in the list of expanded dataclass members, during pass 0. And so nothing would ever get pruned.

Therefore, before we call build_stmt on these expanded dataclass members, we set this pseudo-global variable 'expanding dataclass members'. When build_Name is run against the children of these nodes, it will see that 'expanding cataclass members' is set, and not call `pruning.mark_used` method.

To what extent does this clarify some of your questions? Waht additional steps should I flesh out a little?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: I see!

hughperkins: happy to add this to a comment

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: > Since all dataclass members will be in the list of expanded dataclass members, during pass 0.

I don't get this part. If I understand correctly:
* non-expanded dataclass members that are used will be marked as used when build ast nodes
* we will generate expended dataclass members at some point in the process and build ast nodes for them accordingly
* this is done for all nodes, so we need to make sure that these variables are filtered out from the pruning mechanism

If this is correct, which I doubt, I don't understand what would happen if when considering the expended dataclass members as used. These variables does not even exist as input argument of the kernel. Is it because of the way it is determined that a variable is used: we flatten its nested name, so whether you flatten ("a", "b") or ("__ti_a__ti_b",) you get the same, even if it is actually different since one indicates truly flat hierarchy but not the same (kind of name collapse issue)?

hughperkins: so, let's say we have:

```
@dataclasses.dataclass
class MyStruct:
     a: ti.types.NDArray[ti.i32, 1]
     unused1: ti.types.NDArray[ti.i32, 1]

@ti.func
def f1(my_struct: MyStruct):
    my_struct.a[0] = 5
```

this bit `my_struct.a[0] = 5` is represented by an ast tree, with a single node at the apex. When we `build_stmt` on that node, `build_Name` will be called on one of `my_struct` or `a` (I forget which, thought it's back to front, so most likely on `my_struct`. At that point, the entire flat name will be created, giving `__ti_my_struct__ti_a`, which will be added to the set of used parameters, by calling pruning.mark_used("__ti_my_struct__ti_a").

So far so good.

Now when we build the call into f1, eg from some kernel k1:
```
@ti.kernel
def k1(my_struct: MyStruct):
    f1(my_struct=my_struct)
```
... at this point we have to build the various ast nodes that represent the kernel function, and notably the call into f1 `f1(my_struct=my_struct)`. Building a node of some class NodeClass causes a call to the corresponding build method `build_NodeClass`. In this case, buliding the `Call` node causes a call to `build_Call`, where we find ourselves now.

As part of the build of the Call node, we need to build the parameters `my_struct=my_struct`. We'll build these using build_stmt and/or build_stmts just as for every other ast node we walk. When we build this, we are going to notice it's a dataclass, and we're going to expand it out so what *acutally* gets generated is:

```
f1(__ti_my_struct__ti_a=__ti_my_struct__ti_a, __ti_my_struct__ti_unused1=__ti_my_struct__ti_unused1)
```

Now, in order to build these, these will ultimatley end up inside calls to build_Name too.

But `__ti_my_struct__ti_unused1` should NOT be added to the list of used variables: it's not acutlaly used! I mean, it is. It's used in this bit of ast, to make the call into f1. But f1 doesn't use it. So we should prune it. But pruning won't happen until pass 1. And marking variables as used happens in pass 0. So, no pruning yet, and yet a call to build_Name, with __ti_my_struct__ti_unused1, is gong to happen.

What to do?

We first set `expanding_dataclass_call_parameters` to True, before expanding the various parameters being used in the call. Then we build those parameters. Then we unset this.

It's a bit ugly, but we are 100% single-threaded, so it works ok 😌

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: Could be helpful to avoid any issue during a potential future refactoring. But necessary.

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
duburcqa: Oh I see! Now it is clear! I think I could write a much more newbie friendly comment based on what you just explained.

PR #346, opened 2026-01-11
[Misc] Add ti.clock_speed_hz()

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/intrinsics.py
+++ b/python/gstaichi/lang/intrinsics.py
@@ -38,7 +38,7 @@ def clock_counter():
     return 0
 
 
-def clock_speed_hz():
+def clock_freq_hz():
     """
     Returns the clock speed in Hz of the compute device, i.e. GPU. Throws NotImplementedError on
     unsupported architectures.
@@ -58,10 +58,10 @@ def clock_speed_hz():
     if arch == _ti_core.cuda:
         clock_rate_khz = _ti_core.query_int64("cuda_clock_rate_khz")
         return float(clock_rate_khz * 1000)
-    raise NotImplementedError(f"{clock_speed_hz.__name__} not implemented for arch {arch.name}")
+    raise NotImplementedError(f"{clock_freq_hz.__name__} not implemented for arch {arch.name}")
 
 
 __all__ = [
     "clock_counter",
-    "clock_speed_hz",
+    "clock_freq_hz",
 ]
--- a/tests/python/test_api.py
+++ b/tests/python/test_api.py
@@ -114,7 +114,7 @@ user_api[ti] = [
     "cast",
     "ceil",
     "clock_counter",
-    "clock_speed_hz",
+    "clock_freq_hz",
     "cos",
     "cpu",
     "cuda",
--- a/tests/python/test_intrinsics.py
+++ b/tests/python/test_intrinsics.py
@@ -4,7 +4,7 @@ import gstaichi as ti
 
 from tests import test_utils
 
-clock_speed_supported_archs = ti.cuda
+clock_freq_supported_archs = ti.cuda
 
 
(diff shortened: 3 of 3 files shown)


tests/python/test_clock_speed.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+"""Simple test script for ti.clock_speed_hz()"""
```
erizmr: Shall we move this file into `examples` or `test`?

hughperkins: good spot :)

hughperkins: or delete perhaps

hughperkins: (deleted)

erizmr (approved):

python/gstaichi/lang/intrinsics.py
```diff
@@ -38,6 +38,50 @@ def clock_counter():
     return 0
 
 
+def clock_speed_hz():
```
duburcqa: In this case I’m not convinced that returning 0.0 is better than an exception if not supported. Quite the opposite.

python/gstaichi/lang/intrinsics.py
duburcqa, line 1: It should be specified in the doc how this property behaves when the real frequency of the GPU is changing. Is this the nominal frequency? The current one, the rated one?

python/gstaichi/lang/intrinsics.py
hughperkins: addressed in 51eef6456d

python/gstaichi/lang/intrinsics.py
hughperkins: addressed in 0302bd5f851dc

python/gstaichi/lang/intrinsics.py
```diff
@@ -40,45 +40,24 @@ def clock_counter():
 
 def clock_speed_hz():
     """
-    Returns the clock speed in Hz corresponding to the clock counter.
+    Returns the clock speed in Hz of the compute device, i.e. GPU.
```
duburcqa: "clock speed" is weird I feel. I'm more used to clock frequency. (which does not require specifying _hz since it is implicit). But as you want.

python/gstaichi/lang/intrinsics.py
```diff
@@ -40,45 +40,24 @@ def clock_counter():
...
 
-    This function runs on the host side and queries the device for its clock rate.
-    The returned value can be used to convert clock cycles from clock_counter() to time.
+    Note that this is the nominal speed, NOT the current dynamic speed.
 
-    Supported backends:
-    - CUDA: Returns the GPU clock rate in Hz
```
duburcqa: I think you should specifying that calling this function on supported backends will raise an exception. No need to mention which one if you want to keep your comment generic.

python/gstaichi/lang/intrinsics.py
hughperkins: it would still require units, because the underlying call is in kHz, and that is the reason why I'm explicitly giving the units.

I can change to `clock_freq_hz` if you prefer though?

python/gstaichi/lang/intrinsics.py
hughperkins: addressed in e684b2e48

python/gstaichi/lang/intrinsics.py
duburcqa: yes.

hughperkins: addressed in 9faf9e4fb

PR #347, opened 2026-01-12
[Type] Add args handling to py dataclass parameter renaming

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/ast/ast_transformers/call_transformer.py
+++ b/python/gstaichi/lang/ast/ast_transformers/call_transformer.py
@@ -8,6 +8,7 @@ import re
 import warnings
 from ast import unparse
 from collections import ChainMap
+from contextlib import nullcontext
 from typing import Any
 
 from gstaichi.lang import (
@@ -271,22 +272,18 @@ class CallTransformer:
         Call(func=Name(id='f2', ctx=Load()), args=[Name(id='my_struct_ab', ctx=Load())], keywords=[])
         """
         is_func_base_wrapper = False
-        if get_decorator(ctx, node) in ["static", "static_assert"]:
-            with ctx.static_scope_guard():
-                build_stmt(ctx, node.func)
-                build_stmts(ctx, node.args)
-                build_stmts(ctx, node.keywords)
-            func = node.func.ptr
-            func_type = type(func)
-        else:
+        is_static = get_decorator(ctx, node) in ("static", "static_assert")
+
+        with ctx.static_scope_guard() if is_static else nullcontext():
             build_stmt(ctx, node.func)
             # creates variable for the dataclass itself (as well as other variables,
             # not related to dataclasses). Necessary for calling further child functions
             build_stmts(ctx, node.args)
             build_stmts(ctx, node.keywords)
+        func = node.func.ptr
+        func_type = type(func)
 
-            func = node.func.ptr
-            func_type = type(func)
+        if not is_static:
             is_func_base_wrapper = func_type in {GsTaichiCallable, BoundGsTaichiCallable}
             pruning = ctx.global_context.pruning
             called_needed = None


python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -313,6 +313,8 @@ def build_Call(ctx: ASTTransformerFuncContext, node: ast.Call, build_stmt, build
                 build_stmt(ctx, arg)
             ctx.expanding_dataclass_call_parameters = False
 
+        # check for pure violations
```
hughperkins: this is just some docstring I added au passage. Doesn't directly relate to this PR.

python/gstaichi/lang/_pruning.py
```diff
@@ -46,17 +49,35 @@ def is_used(self, func_id: int, parameter_flat_name: str) -> bool:
...
         to_unprune: set[str] = set()
         arg_id = 0
+        # node.args ordering will match that of the called function's metas_expanded,
+        # because of the way calling with sequential args works.
+        # We need to look at the child's declaration - via metas - in order to get the name they use.
+        # We can't tell their name just by looking at our own metas.
+        for i, arg in enumerate(node.args):
+            if hasattr(arg, "id"):
+                calling_name = arg.id
+                called_name = node.func.ptr.wrapper.arg_metas_expanded[arg_id].name
```
hughperkins: this is pretty ugly 😅. Open to suggestions on how we can improve over this long list of unintuitive dots.

hughperkins: (maybe just needs some docstring/comment?)

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -313,6 +313,8 @@ def build_Call(ctx: ASTTransformerFuncContext, node: ast.Call, build_stmt, build
...
+        # we have to do this after building the statements
         # if any arg violates pure, then node also violates pure
```
duburcqa: It would probably help to add some uppercase and dot. otherwise the two last sentences appear like one.

python/gstaichi/lang/_pruning.py
duburcqa: I don't find it so unintuitive. I'm probably used to it already.

duburcqa: What about parent_arg_name / child_arg_name instead calling_name / called_name? It is more clear to me and I think you are already using the terminology at other places.

python/gstaichi/lang/_pruning.py
```diff
@@ -46,17 +49,35 @@ def is_used(self, func_id: int, parameter_flat_name: str) -> bool:
...
+        note that this handles both args and kwargs
         """
         if type(func) not in {GsTaichiCallable, BoundGsTaichiCallable}:
             return
 
         my_func_id = ctx.func.func_id
         called_func_id = func.wrapper.func_id  # type: ignore
-
         # Copy the used parameters from the child function into our own function.
         called_unpruned = self.used_parameters_by_func_id[called_func_id]
```
duburcqa: What about `used_arg_unpruned`? It sounds more natural I think. Called for an arg does not mean much.

python/gstaichi/lang/_pruning.py
```diff
@@ -46,17 +49,35 @@ def is_used(self, func_id: int, parameter_flat_name: str) -> bool:
...
+                calling_name = arg.id
+                called_name = node.func.ptr.wrapper.arg_metas_expanded[arg_id].name
+                if called_name in called_unpruned:
+                    to_unprune.add(calling_name)
+            arg_id += 1
+        # Note that our own arg_metas ordering will in general NOT match that of the child's. That's
+        # because our ordering is based on the order in which we pass arguments to the function, but the
+        # child's ordering is based on the ordering of their declaration; and these orderings might not
+        # match.
+        # Luckily, for keywords, we don't need to look at the child's metas, because we can get the
```
duburcqa: I don't like the use of luckily in programming and research in General. Stuffs have properties and you leverage them. No luck involved here. Just hard work and cleaver design choices. I hope your code is not working out of luck.

python/gstaichi/lang/_pruning.py
```diff
@@ -65,3 +86,61 @@ def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCal
...
+        child_arg_id = 0
+        child_metas: list[ArgMetadata] = node.func.ptr.wrapper.arg_metas_expanded
+        child_name_by_our_name = self.child_name_by_caller_name_by_func_id[called_func_id]
+        for i, arg in enumerate(node.args):
+            if hasattr(arg, "id"):
+                calling_name = arg.id
+                if calling_name.startswith("__ti_"):
+                    called_name = child_metas[child_arg_id].name
+                    if called_name in called_needed or not called_name.startswith("__ti_"):
+                        child_name_by_our_name[calling_name] = called_name
```
duburcqa: Let's stick to (child|parent)_arg_name or (child|parent)_name?

python/gstaichi/lang/_pruning.py
hughperkins: the thing I dont like about parent is that, in my mind things look like this:

```
parent
|
me
|
child
```

...but in this case, 'parent' refers to 'me'. In general, I find this very confusing/ambiguous.

calling vs called is very unambiguous I feel. We have one function (us) calling another function (the called one).

python/gstaichi/lang/_pruning.py
```diff
@@ -65,3 +86,61 @@ def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCal
                     to_unprune.add(calling_name)
             arg_id += 1
         self.used_parameters_by_func_id[my_func_id].update(to_unprune)
+
+        called_needed = self.used_parameters_by_func_id[called_func_id]
+        child_arg_id = 0
+        child_metas: list[ArgMetadata] = node.func.ptr.wrapper.arg_metas_expanded
+        child_name_by_our_name = self.child_name_by_caller_name_by_func_id[called_func_id]
```
duburcqa: I think it would be more clear if `_arg` could be part of all these variable names (`self.child_name_by_caller_name_by_func_id`, `child_name_by_our_name`...etc) because it is a bit misleading `caller_name` is grammatically the name of the function, not the name of the arg from the function.

What I usually do the shorten the names is avoiding repeating the common part of the name:
`child_arg_name_by_caller_arg_name` -> `child_by_caller_arg_name`

python/gstaichi/lang/_pruning.py
```diff
@@ -65,3 +86,61 @@ def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCal
...
+        for i, arg in enumerate(node.args):
+            if hasattr(arg, "id"):
```
duburcqa: Do we need to rely on `hasattr` here?

python/gstaichi/lang/_pruning.py
```diff
@@ -65,3 +86,61 @@ def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCal
...
+                calling_name = arg.id
+                if calling_name.startswith("__ti_"):
+                    called_name = child_metas[child_arg_id].name
+                    if called_name in called_needed or not called_name.startswith("__ti_"):
```
duburcqa: A short comment to explain this pattern could be nice.

python/gstaichi/lang/_pruning.py
hughperkins: "Happily"? "Usefully"? Needs some conjunction I feel?

python/gstaichi/lang/_pruning.py
```diff
@@ -65,3 +86,61 @@ def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCal
...
+        node: "ast.Call",
+        py_args: list[Any],
+    ) -> list[Any]:
+        """
+        used in build_Call, before making the call, in pass 1
+
+        note that this ONLY handles args, not kwargs
+        """
+        if not (hasattr(func, "wrapper") and hasattr(func.wrapper, "func_id")):
+            return py_args
```
duburcqa: No way to use inheritance instead of `hasattr`?

python/gstaichi/lang/_pruning.py
```diff
@@ -65,3 +86,61 @@ def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCal
...
+                if calling_name.startswith("__ti_"):
+                    called_name = self.child_name_by_caller_name_by_func_id[called_func_id].get(calling_name)
+                    if called_name is not None and (
+                        called_name in called_needed or not called_name.startswith("__ti_")
+                    ):
+                        new_args.append(py_args[i])
+                else:
+                    new_args.append(py_args[i])
+            else:
+                new_args.append(py_args[i])
```
duburcqa: Why not to determine the single case that must be filtered out and do:
```python
        for i, arg in enumerate(node.args):
            if hasattr(arg, "id"):
                calling_name = arg.id  # type: ignore
                if calling_name.startswith("__ti_"):
                    called_name = self.child_name_by_caller_name_by_func_id[called_func_id].get(calling_name)
                    if called_name is None or (
                        called_name not in called_needed and called_name.startswith("__ti_")
                    ):
                        continue
            new_args.append(py_args[i])
```

duburcqa (approved):

python/gstaichi/lang/_pruning.py
duburcqa: You are using our. At other places.

duburcqa: The issue is that I don't like switching between calling / called vs parent / child, vs our. Would be nice to find some consistent names. child vs our as you use at some other places?

Personally I don't think calling vs called is very clear by maybe it is just me. You are the main maintainer so what you are the most comfortable with prevails.

python/gstaichi/lang/_pruning.py
duburcqa: I would stick to fact: This is not an issue because, for keywords, we don't need to look at the child's metas. Instead, we can get the [...]

python/gstaichi/lang/_pruning.py
hughperkins: our vs child satisfies the constraint I cite above.

So, just now I asked chatgpt. Here's waht it said:

Screenshot 2026-01-12 at 16 15 37
Screenshot 2026-01-12 at 16 16 01
Screenshot 2026-01-12 at 16 16 05
Screenshot 2026-01-12 at 16 16 09
Screenshot 2026-01-12 at 16 16 13
Screenshot 2026-01-12 at 16 16 17
Screenshot 2026-01-12 at 16 16 21

hughperkins: https://chatgpt.com/share/696564ee-a4c8-8012-8ba4-72521a7597d4

hughperkins: How would you feel about changing from calling/called to caller/callee ?

python/gstaichi/lang/_pruning.py
hughperkins: addressed in a4563f330

python/gstaichi/lang/_pruning.py
hughperkins: Updated

python/gstaichi/lang/_pruning.py
hughperkins: Addressed in a4563f330

python/gstaichi/lang/_pruning.py
hughperkins: addresed in c08a053ac

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -184,7 +184,7 @@ def _expand_Call_dataclass_args(
                         child_name = create_flat_name(arg.id, field.name)
                     except Exception as e:
                         raise RuntimeError(f"Exception whilst processing {field.name} in {type(dataclass_type)}") from e
-                    if pruning.enforcing and child_name not in pruning.used_parameters_by_func_id[func_id]:
+                    if pruning.enforcing and child_name not in pruning.used_vars_by_func_id[func_id]:
```
duburcqa, line 188: vars is less specific than `args` I feel. I have never seen vars being used in place of args actually.

python/gstaichi/lang/_pruning.py
```diff
@@ -29,39 +34,115 @@ class Pruning:
...
+        # We can get the child's name directly from our own keyword node.
         for kwarg in node.keywords:
-            if hasattr(kwarg.value, "id"):
-                calling_name = kwarg.value.id  # type: ignore
-                called_name = kwarg.arg
-                if called_name in called_unpruned:
-                    to_unprune.add(calling_name)
+            if type(kwarg.value) in {Name}:
+                caller_arg_name = kwarg.value.id  # type: ignore
+                callee_param_name = kwarg.arg
```
duburcqa, line 88: Why mixing `arg` and `param` ? There is any difference in meaning?

python/gstaichi/lang/_pruning.py
```diff
@@ -29,39 +34,115 @@ class Pruning:
...
+                if callee_param_name in callee_used_vars:
+                    vars_to_unprune.add(caller_arg_name)
```
duburcqa, line 90: Same. Why mixing vars / param / arg?

python/gstaichi/lang/ast/ast_transformers/call_transformer.py
hughperkins: yes, but vars is actually correct.

we have a funtion:

```
@ti.func
def f1(a, b):
    a = 5
```
a is now a used var.

This used var can then be used as an argument to another function:
```
@ti.func
def f1(a, b):
    a = 5
    f2(a)
```

python/gstaichi/lang/_pruning.py
hughperkins: argument is what is passed into a function. parameter is what is received by the function. Per chatgpt.

python/gstaichi/lang/_pruning.py
hughperkins: see above

python/gstaichi/lang/_pruning.py
hughperkins: see second quesion in https://chatgpt.com/share/696564ee-a4c8-8012-8ba4-72521a7597d4

python/gstaichi/lang/_pruning.py
duburcqa: I understand what you mean, but what you are actually pruning / unpruning are the arguments (eventually parameters) of the top-level kernel, not some random intermediary function vars?

erizmr (approved):

python/gstaichi/lang/_pruning.py
duburcqa: I’m completely fine

python/gstaichi/lang/_pruning.py
hughperkins: Sure, but the top-level parameters are being stored in Kernel object, after compilation.

This dict is storing func-specific "things", during the tree walk phase.

PR #349, opened 2026-01-13
[Type] Handle *-calling for Genesis test_utils

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_exceptions.py
+++ b/python/gstaichi/lang/_exceptions.py
@@ -1,10 +1,8 @@
 from typing import Type
 
 
-def raise_exception(
-    ExceptionClass: Type[Exception], msg: str, category: str, err_code: str, orig: Exception | None = None
-):
-    err = ExceptionClass(f"[{category}] {msg} (error code: {err_code}).")
+def raise_exception(ExceptionClass: Type[Exception], msg: str, err_code: str, orig: Exception | None = None):
+    err = ExceptionClass(f"{msg} (error code: {err_code}).")
     if orig:
         raise err from orig
     raise err
--- a/python/gstaichi/lang/_pruning.py
+++ b/python/gstaichi/lang/_pruning.py
@@ -156,10 +156,9 @@ class Pruning:
             if is_starred:
                 if i != len(node_args) - 1 or len(node_keywords) != 0:
                     raise_exception(
-                        GsTaichiSyntaxError,
-                        "* args can only be present as the last argument of a function",
-                        "FUNCDECLARATION",
-                        "STARNOTLAST",
+                        ExceptionClass=GsTaichiSyntaxError,
+                        msg="* args can only be present as the last argument of a function",
+                        err_code="STARNOTLAST",
                     )
 
                 # we'll just dump the rest of the py_args in:


python/gstaichi/lang/_pruning.py
```diff
@@ -132,6 +131,17 @@ def filter_call_args(
                 callee_metas_pruned.append(_callee_meta)
         callee_metas = callee_metas_pruned
         for i, arg in enumerate(node.args):
+            is_starred = type(arg) is Starred
+            if is_starred:
+                if i != len(node.args) - 1 or len(node.keywords) != 0:
+                    raise GsTaichiSyntaxError(
+                        "STARNOTLAST * args can only be present as the last argument of a function"
```
duburcqa: What is `STARNOTLAST` ? I would suggest to either use `[classifier/category] msg` if you want to add some category prefix. It is just make the exception message fairly cryptic for the end user. I think it is important to be as user friendly as possible, even if it means complexifying the unit tests.

duburcqa: This limitation / feature should be clearly documented in read-the-doc. I think it is very important.

hughperkins: You mean like `[STARNOTLAST] * args can only be present as the last argument of a function` ?

hughperkins: added some doc in 9f3110e67

duburcqa: It is called argument unpacking, not star calling x)

duburcqa:
```
- `*args` must be the last argument in the function call
- `*args` must not contain any `dataclasses.dataclass` objects
```

Or what? Are you raising an exception clearly stating what is the issue for the end user?

duburcqa: `"STARNOTLAST * args can only be present as the last argument of a function"`

-> Your exception is not specific enough apparently?

hughperkins: in waht sense?

hughperkins: oh I see.

hughperkins: So, I never found a way to test calling a function with a dataclass in *args. Like, it didnt seem possible to do. I got other wierd exceptions before getting anywhere my own code here.

duburcqa: I'm not convinced `STARNOTLAST` is a good name of a category of error, since you only have one.

hughperkins: what are you imagining in your mind for this?

duburcqa: > Like, it didnt seem possible to do. I got other wierd exceptions before getting anywhere my own code here.

Then I would not mention in the doc that it is not supported if it cannot occur. It is just confusing otherwise.

hughperkins: wait, I want to be able to detect the specific error in tests. But without having to change the test every time I change the messages. That's my intention. Category seems not sepcific enough.

hughperkins: here si the approach we were both ok with previously: https://github.com/Genesis-Embodied-AI/gstaichi/blob/main/python/gstaichi/lang/_fast_caching/src_hasher.py#L37

```
            f"[FASTCACHE][INVALID_FUNC] The pure function {kernel_source_info.function_name} could not be "
            "fast cached, because one or more parameter types were invalid"
```

duburcqa: > wait, I want to be able to detect the specific error in tests. But without having to change the test every time I change the messages. That's my intention. Category seems not sepcific enough.

yes. And I already said I don't consider this excuse as legitimate: "I think it is important to be as user friendly as possible, even if it means complexifying the unit tests."

> here si the approach we were both ok with previously:

Yes, because `[FASTCACHE][INVALID_FUNC]` is precisely a valid category of error. But I'm not convinced `[STARNOTLAST]` is. For example, `[KERNEL][INVALID_ARG]` would be valid.

hughperkins: I've put [FUNCDECLARATION][STARNOTLAST]

(categoy is [FUNCDECLARATION].)

hughperkins: I would prefer to put it, for completeness, even if we haven't discovered a test for it. I'll add it to the exception message.

hughperkins: Well, it isn't needed in the xception message, because the exception message corresponds precisely to "not last" check.

hughperkins: You can consider this message to follow the pattern:
[category]message
where:
category = FUNCDECLARATION
message= [STARNOTLAST] Some other stuff

duburcqa: Still, I don't think it makes sense for the end user to have an error message starting with `[STARNOTLAST]`. It is not an actual category of error. it is one single error, like an error code. If you want to mention some error code in the error message, why not. But you need to clearly state that it corresponds to the error code and keep track of all the error code that are being defined in the global enum or something.

duburcqa: > Well, it isn't needed in the xception message, because the exception message corresponds precisely to "not last" check.

yes. That is why I don't like it. You are not guarding in your code against something you are saying is not supported which you cannot even reproduce. At this point you can just say that passing classes whose define is kernel-local is not supported, for completeness?

```python

@ti.kernel
def my_kernel(struct: MyStruct) -> None:
    @ti.data_oriented
    class my_nested_class:
        flag: bool = True

    my_nested_value = my_nested_class()
    
    if ti.static(my_nested_value):
        struct.a += 1
    else:
        struct.b += 1
```

duburcqa: Let's focus on what the average user should pay attention to? You can add these extra information somewhere in the code, either in comments or in docstring. I don't know where exactly.

hughperkins: > > Well, it isn't needed in the xception message, because the exception message corresponds precisely to "not last" check.
> 
> yes. That is why I don't like it. You are not guarding in your code against something you are saying is not supported which you cannot even reproduce. At this point you can just say that passing classes whose define is kernel-local is not supported, for completeness?
> 
> ```python
> @ti.kernel
> def my_kernel(struct: MyStruct) -> None:
>     @ti.data_oriented
>     class my_nested_class:
>         flag: bool = True
> 
>     my_nested_value = my_nested_class()
>     
>     if ti.static(my_nested_value):
>         struct.a += 1
>     else:
>         struct.b += 1
> ```

Honestly, I don't even want to say they're supported at all. I just enabled *args because the lift to modify Genesis test_utils to not use them would have been far greater, and the result really ugly, and hard to PR. If we are going to say they are supported, I want it to be a very narrow list of things we support.

Let's say someone tries to pass a dataclass to *args, and gets some weird esoteric error message, and our doc doesnt say it's not allowed. Then it just looks like a bug.

hughperkins: I can move it to the end of the message if you prefer? `blah lbah [internal error code: BLAHBLAH]`

duburcqa: > Let's say someone tries to pass a dataclass to *args, and gets some weird esoteric error message, and our doc doesnt say it's not allowed. Then it just looks like a bug.

The thing is, your single line about dataclass is either too specific or not enough. It gives the feeling that it is complete, while it is not.

I would just say that this feature is experimental and should be used with caution. It should work fine in simple cases, which is the by far the most common, but may fail miserably with some weird esoteric error message in others. No need to mention dataclass or anything.

duburcqa: `[CATEGORY] blah blah (error code: BLAHBLAH).` syntax would do fine.

duburcqa: By the way, raise some helper to automatically format exception may be convenient:
```python
def raise_exception(err_type, msg="", category=None, errno=None, orig=None):
    cat_msg = "" if category is None else f"[{msg}] "
    errno_msg = "" if errno is None else f" (error code: {errno})"
    err = err_type(f"{cat_msg}{msg}{errno_msg}.")
    if from is None:
        raise err
    raise err from orig
```

hughperkins: well, I'd argue it's not experimental. It's only supported for this very narrow use case. "experimental" implies to my mind that we are conssidering extending it.

I'll rephrase to say it's not supported, except for this very narrow use-case.

hughperkins: Note that the stated requireements are specific though. Non-dataclass things should work ok. At least, the renamable py dataclasses does not break them.

hughperkins: Actually, I'm just going to leave it as it is. It is correct as-is I feel.
- does not need to be marked experimental
- does work if:
     - last arg in the call
     - no dataclasses contained
 
This appears to me to be accurate and correct.

Oh, here the context is the exception.

Let's factorize what we are dsicusssing:

# Exception

The exception is raised when someone doesnt put the *args as the last argument. It states that *args must be the last argument. It is not intended to imply that this is sufficient condition, only that it is a necessary condition.

# doc

## Argument unpacking

i.e. syntax like `some_function(*args)`

Argument unpacking is supported in a very narrow case, in order for test_utils in Genesis to work, https://github.com/Genesis-Embodied-AI/Genesis/blob/f711ef981a71558f8961f308066023d369405df4/tests/test_utils.py#L63:
- `*args` must be the last argument in the function call
- `*args` must not contain `dataclasses.dataclass` objects

hughperkins: addressed in [c5384f4](https://github.com/Genesis-Embodied-AI/gstaichi/pull/349/commits/c5384f4aed0e4f4f90e2da61f814af8c1bbd7a9f)

python/gstaichi/lang/_pruning.py
```diff
@@ -149,6 +152,20 @@ def filter_call_args(
...
         callee_metas = callee_metas_pruned
         for i, arg in enumerate(node_args):
+            is_starred = type(arg) is Starred
+            if is_starred:
+                if i != len(node_args) - 1 or len(node_keywords) != 0:
+                    raise_exception(
+                        GsTaichiSyntaxError,
+                        "* args can only be present as the last argument of a function",
+                        "FUNCDECLARATION",
+                        "STARNOTLAST",
```
duburcqa: I recommend using keywords for anonymous arguments.

python/gstaichi/lang/_exceptions.py
```diff
@@ -0,0 +1,8 @@
+from typing import Type
+
+
+def raise_exception(
+    ExceptionClass: Type[Exception], msg: str, category: str, err_code: str, orig: Exception | None = None
+):
+    err = ExceptionClass(f"[{category}] {msg} (error code: {err_code}).")
+    raise err from orig if orig else err
```
duburcqa: Is specifying a category really necessary?

python/gstaichi/lang/_pruning.py
duburcqa: > `*args` must not contain `dataclasses.dataclass` objects

Still not ok nothing in the code is guarding against `args` containing `dataclasses.dataclass` objects.

python/gstaichi/lang/_exceptions.py
duburcqa: Same for message

duburcqa: `raise err from orig if orig else err`

This line is too cryptic, I don't like it personally.

python/gstaichi/lang/_pruning.py
hughperkins: what do you mean by 'anonymous argument'? The function parameters all have names and types.

python/gstaichi/lang/_exceptions.py
hughperkins: Not sure I follow. I feel the requirement for a category is something you were asking for. Areyou saying, you don't require a category? Or, at any rate, if we want to throw an exception without an errorcode, then we don't need to use this function at all I feel?

hughperkins: Normally you like 1-liners, but ok. What is your preferred formulation?

duburcqa: I just having a generic implementation that you use systematically when raising exceptions:
```python
def raise_exception(err_type, msg="", category=None, errno=None, orig=None):
    cat_msg = "" if category is None else f"[{msg}] "
    errno_msg = "" if errno is None else f" (error code: {errno})"
    err = err_type(f"{cat_msg}{msg}{errno_msg}.")
    if from is None:
        raise err
    raise err from orig
```


This is what we are doing in Genesis. At this point, it is up to you to determine whether msg, category and errno should be mandatory.

hughperkins: I feel like using normal exceptions is fine when we don't wish to specify an errorcode. And if we do wish to specify an error code, then both category and errorcode are mandatory, but not orig.

duburcqa: > Normally you like 1-liners, but ok. 

Yes, but `raise err from orig` is not "standard" (like assert). So for me it is a strange beast I prefer to isolate on its own line.

I prefer:
```python
if orig:
    raise err from orig
raise err
```

hughperkins: addressed in 68d255e83

duburcqa: I don't share this opinion. Formatting for category should be standardised by design. You want to enforce structure in your message (with category and error code), if you want this you need something that actually enforces this structure, not just your good will.

hughperkins: I have no use-case for category personally. The only constraint I have personally is to have a short sequence of letters I can put in an exception message, that I can check for in unit tests, and that won't be modified every time we tweak the message slightly.

What is your own use case for exception categories? Is this something you would like to be able to use from Genesis? If so, how, and when/hwere, would you use it?

duburcqa: We don't use category in Genesis. Then remove it and just go for error code. But try to stick to the same formatting everywhere.

python/gstaichi/lang/_pruning.py
duburcqa: "STARNOTLAST" is not assigned to a variable (same for the others).

Either do:
```python
a = 1
b = 2
f(a, b)
```
or 
```python
f(a=1, b=2)
```

I find it much easier to understand what are the arguments that are being passed to a function without looking at the declaration systematically.

hughperkins: oh I see. 👍

python/gstaichi/lang/_pruning.py
hughperkins: yes, but I have no way to guard against them, becauase execution crasehse before my code.

They're not legal either way, I just don't have any obvious way of throwing a meaningful exception, in the scope of this PR.

hughperkins: so:
- the doc is correct
- the current exception is correct
- a meaningful exception if you try to use a dataclass might be nice, but is outside the scope of this PR I feel (this PR isn't triggering the crash: it alrady exists)

python/gstaichi/lang/_exceptions.py
hughperkins: addressed in 681e1f399

python/gstaichi/lang/_pruning.py
hughperkins: addressed in 681e1f399

python/gstaichi/lang/_pruning.py
duburcqa: > but is outside the scope of this PR I feel

No it is not. This PR is specifically about handling arg packing

hughperkins: Ok, let's describe the situation, step by step:

1. main, as now:
    - test_utils *-calling works ok
    - *-calling works even when not in the last position in the arguments
    - trying to pass in a dataclass in *-calling does NOt work
2. `hp/renaming-py-dataclass-merge-staging-target` branch, before this PR
     - *-calling of any form is broken, because we broke it, when we started filtering the *args as part of pruning https://github.com/Genesis-Embodied-AI/gstaichi/blob/5700036161820df77e1706e417de638e6cd43543/python/gstaichi/lang/_pruning.py#L151-L164
3. this PR fixes the break in 2, by adding partial handling of *-calling, when the *-call is in the last argument position
    - however:
         - it cannot fix passing in dataclasses, because they are already broken in main, for *-args
         - execution never reaches the pruning funciton in question: crashes in a completely different part of the code
             - not obvoius how to detect and throw a more appropriate exception
         - that said, if dataclasses did arrive our way, we are explicitly not handling them, in this bit of the code
             - however, there seems little point in throwing an exception if they did arrive into this bit of code, since they don't

I suppose one compromise could be to add an exception into this bit of code if a dataclass does arrive to them. Howeer, I have no way of testing such an exception, because the dataclass never gets this far.

hughperkins: In addition, detecting this theoretical case, that I cannot repreoduce, will use up "hot-path" compute, needlessly, I feel.

hughperkins: Running on `main` branch:

This code:
```
import gstaichi as ti

ti.init(ti.cpu)

@ti.func
def f1(a, b):
	print(a, b)

@ti.kernel
def k1():
	f1(*(2, 3))

k1()
```

Output:

```
[GsTaichi] version 4.6.0, llvm 20.1.0, commit 69cc2417, osx, python 3.10.18
[GsTaichi] Starting on arch=arm64
2 3
```

Running on main still, this code:
```
# import gstaichi as ti

# ti.init(ti.cpu)

# @ti.func
# def f1(a, b):
# 	print(a, b)

# @ti.kernel
# def k1():
# 	f1(*(2, 3))

# k1()

import dataclasses
import gstaichi as ti

ti.init(ti.cpu)

@dataclasses.dataclass
class Struct:
	c: ti.i32

@ti.func
def f2(a, b: Struct):
	print(a, b.c)

@ti.kernel
def k2(struct: Struct):
	f2(*(2, struct))

struct = Struct(c=3)
k2(struct)


```

Output:
```
(gstaichi) ~/git/gstaichi (main|…6) $ python test_star_calling.py
[GsTaichi] version 4.6.0, llvm 20.1.0, commit 69cc2417, osx, python 3.10.18
[GsTaichi] Starting on arch=arm64
Traceback (most recent call last):
  File "/Users/hugh/git/gstaichi/test_star_calling.py", line 33, in <module>
    k2(struct)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/_gstaichi_callable.py", line 96, in __call__
    return self.wrapper.__call__(*args, **kwargs)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/kernel_impl.py", line 140, in wrapped_func
    raise type(e)("\n" + str(e)) from None
gstaichi.lang.exception.GsTaichiTypeError:
gstaichi stack trace:
===
Traceback (most recent call last):
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/call_transformer.py", line 348, in build_Call
    node.ptr = func(*py_args, **py_kwargs)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/_gstaichi_callable.py", line 96, in __call__
    return self.wrapper.__call__(*args, **kwargs)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/func.py", line 93, in __call__
    ret = transform_tree(tree, ctx)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/transform.py", line 8, in transform_tree
    ASTTransformer()(ctx, tree)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformer_utils.py", line 50, in __call__
    raise e.with_traceback(None)
gstaichi.lang.exception.GsTaichiTypeError: gstaichi stack trace:
===
Traceback (most recent call last):
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformer_utils.py", line 39, in __call__
    res = method(ctx, node)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformer.py", line 454, in build_FunctionDef
    FunctionDefTransformer.build_FunctionDef(ctx, node, build_stmts)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py", line 297, in build_FunctionDef
    FunctionDefTransformer._transform_as_func(ctx, node, args)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py", line 260, in _transform_as_func
    FunctionDefTransformer._transform_func_arg(ctx, argument.name, argument.annotation, py_arg)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py", line 243, in _transform_func_arg
    ctx.create_variable(argument_name, impl.expr_init_func(ti_ops.cast(data, argument_type)))
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ops.py", line 84, in cast
    return expr.Expr(_ti_core.value_cast(expr.Expr(obj).ptr, dtype))
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/expr.py", line 46, in __init__
    self.ptr = make_constant_expr(arg, dtype).ptr
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/expr.py", line 139, in make_constant_expr
    raise GsTaichiTypeError(f"Invalid constant scalar data type: {type(val)}")
gstaichi.lang.exception.GsTaichiTypeError: Invalid constant scalar data type: <class 'type'>

===

Your code:
File "/Users/hugh/git/gstaichi/test_star_calling.py", line 25, in f2:
def f2(a, b: Struct):
^^^^^^^^^^^^^^^^^^^^^
Invalid constant scalar data type: <class 'type'>


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformer_utils.py", line 39, in __call__
    res = method(ctx, node)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformer.py", line 450, in build_Call
    return CallTransformer.build_Call(ctx, node, build_stmt, build_stmts)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/call_transformer.py", line 364, in build_Call
    raise GsTaichiTypeError(msg)
gstaichi.lang.exception.GsTaichiTypeError: TypeError when calling `f2`: gstaichi stack trace:
===
Traceback (most recent call last):
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformer_utils.py", line 39, in __call__
    res = method(ctx, node)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformer.py", line 454, in build_FunctionDef
    FunctionDefTransformer.build_FunctionDef(ctx, node, build_stmts)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py", line 297, in build_FunctionDef
    FunctionDefTransformer._transform_as_func(ctx, node, args)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py", line 260, in _transform_as_func
    FunctionDefTransformer._transform_func_arg(ctx, argument.name, argument.annotation, py_arg)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py", line 243, in _transform_func_arg
    ctx.create_variable(argument_name, impl.expr_init_func(ti_ops.cast(data, argument_type)))
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ops.py", line 84, in cast
    return expr.GsTaichi Expression(_ti_core.value_cast(expr.GsTaichi Expression(obj).ptr, dtype))
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/expr.py", line 46, in __init__
    self.ptr = make_constant_expr(arg, dtype).ptr
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/expr.py", line 139, in make_constant_expr
    raise GsTaichiTypeError(f"Invalid constant scalar data type: {type(val)}")
gstaichi.lang.exception.GsTaichiTypeError: Invalid constant scalar data type: <class 'type'>

===

Your code:
File "/Users/hugh/git/gstaichi/test_star_calling.py", line 25, in f2:
def f2(a, b: Struct):
^^^^^^^^^^^^^^^^^^^^^
Invalid constant scalar data type: <class 'type'>
.

===

Your code:
File "/Users/hugh/git/gstaichi/test_star_calling.py", line 30, in k2:
    f2(*(2, struct))
    ^^^^^^^^^^^^^^^^
TypeError when calling `f2`: gstaichi stack trace:
===
Traceback (most recent call last):
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformer_utils.py", line 39, in __call__
    res = method(ctx, node)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformer.py", line 454, in build_FunctionDef
    FunctionDefTransformer.build_FunctionDef(ctx, node, build_stmts)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py", line 297, in build_FunctionDef
    FunctionDefTransformer._transform_as_func(ctx, node, args)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py", line 260, in _transform_as_func
    FunctionDefTransformer._transform_func_arg(ctx, argument.name, argument.annotation, py_arg)
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ast/ast_transformers/function_def_transformer.py", line 243, in _transform_func_arg
    ctx.create_variable(argument_name, impl.expr_init_func(ti_ops.cast(data, argument_type)))
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/ops.py", line 84, in cast
    return expr.GsTaichi Expression(_ti_core.value_cast(expr.GsTaichi Expression(obj).ptr, dtype))
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/expr.py", line 46, in __init__
    self.ptr = make_constant_expr(arg, dtype).ptr
  File "/Users/hugh/git/gstaichi/python/gstaichi/lang/expr.py", line 139, in make_constant_expr
    raise GsTaichiTypeError(f"Invalid constant scalar data type: {type(val)}")
gstaichi.lang.exception.GsTaichiTypeError: Invalid constant scalar data type: <class 'type'>

===

Your code:
File "/Users/hugh/git/gstaichi/test_star_calling.py", line 25, in f2:
def f2(a, b: Struct):
^^^^^^^^^^^^^^^^^^^^^
Invalid constant scalar data type: <class 'type'>
.
```
```

PR #350, opened 2026-01-13
[Type] Enable iterating over static list of functions

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_pruning.py
+++ b/python/gstaichi/lang/_pruning.py
@@ -1,4 +1,4 @@
-from ast import AST, Name
+from ast import Name, expr, keyword
 from collections import defaultdict
 from typing import TYPE_CHECKING, Any
 
@@ -53,8 +53,8 @@ class Pruning:
         ctx: "ASTTransformerFuncContext",
         func: "GsTaichiCallable",
         node: "ast.Call",
-        node_args: list[AST],
-        node_keywords: list[AST],
+        node_args: list[expr],
+        node_keywords: list[keyword],
     ) -> None:
         """
         called from build_Call, after making the call, in pass 0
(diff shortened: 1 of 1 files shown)


python/gstaichi/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -283,48 +283,45 @@ def build_Call(ctx: ASTTransformerFuncContext, node: ast.Call, build_stmt, build
...
+            called_needed = pruning.used_vars_by_func_id[called_func_id_]
+
+        added_args, node_args = CallTransformer._expand_Call_dataclass_args(ctx, node.args)
+        added_keywords, node_keywords = CallTransformer._expand_Call_dataclass_kwargs(ctx, node.keywords, called_needed)
+
+        # Create variables for the now-expanded dataclass members.
+        # We don't want to include these now-expanded dataclass members
+        # in the list of variables to not prune,
+        # since these will contain *all* declared fields, instead of all used fields.
+        # So we set expanding_dataclass_call_parameters to True during this expansion.
```
duburcqa, line 300: Here we are!

```
        # Create variables for the now-expanded dataclass members.
        # Note that we don't want to include these now-expanded dataclass members in 
        # the list of used variables (ie to not prune), because passing input arguments to 
        # a function does not provide any guarantees that they are actually used anywhere 
        # in that function. Setting 'expanding_dataclass_call_parameters' to True during 
        # this expansion is used to achieve the desired behaviour, although it is kind of 
        # hacky and not thread-safe. If parameters are actually used in that function, it 
        # will be added to the list of used variables later on, when actually traversing 
        # the source code of this function.
```

hughperkins: put in separate PR

PR #351, opened 2026-01-13
[Type] Enable data oriented, in the context of renamable py dataclass arguments

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/gstaichi/lang/_pruning.py
+++ b/python/gstaichi/lang/_pruning.py
@@ -1,12 +1,10 @@
-from ast import Name
+from ast import Name, expr, keyword
 from collections import defaultdict
 from typing import TYPE_CHECKING, Any
 
 from ._gstaichi_callable import BoundGsTaichiCallable, GsTaichiCallable
 from .func import Func
 from .kernel_arguments import ArgMetadata
-
-# from .kernel import Kernel, fu
 
 if TYPE_CHECKING:
     import ast
@@ -50,7 +48,14 @@ class Pruning:
     def is_used(self, func_id: int, var_flat_name: str) -> bool:
         return var_flat_name in self.used_vars_by_func_id[func_id]
 
-    def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCallable", node: "ast.Call") -> None:
+    def record_after_call(
+        self,
+        ctx: "ASTTransformerFuncContext",
+        func: "GsTaichiCallable",
+        node: "ast.Call",
+        node_args: list[expr],
+        node_keywords: list[keyword],
+    ) -> None:
         """
         called from build_Call, after making the call, in pass 0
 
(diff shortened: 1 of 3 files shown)


python/gstaichi/lang/_pruning.py
```diff
@@ -69,10 +69,15 @@ def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCal
         # because of the way calling with sequential args works.
         # We need to look at the child's declaration - via metas - in order to get the name they use.
         # We can't tell their name just by looking at our own metas.
+        #
+        # One issue is when calling data-oriented methods, there will be a `self`. We'll detect this
+        # by seeing if the childs arg_metas_expanded is exactly 1 longer than len(node.args) + len(node.kwargs)
+        has_self = len(node.args) + len(node.keywords) + 1 == len(node.func.ptr.wrapper.arg_metas_expanded)
```
duburcqa: I don't like it but it is reasonable

python/gstaichi/lang/_pruning.py
```diff
@@ -116,9 +122,13 @@ def filter_call_args(
 
         note that this ONLY handles args, not kwargs
         """
-        if type(func) not in {GsTaichiCallable, BoundGsTaichiCallable} or type(func.wrapper) != Func:
+        if (
+            type(gstaichi_callable) not in {GsTaichiCallable, BoundGsTaichiCallable}
+            or type(gstaichi_callable.wrapper) != Func
+        ):
```
duburcqa, line 135: Could be worth adding a comment to explain what this condition is detecting. Not very clear at this point.

python/gstaichi/lang/_pruning.py
```diff
@@ -69,10 +69,16 @@ def record_after_call(self, ctx: "ASTTransformerFuncContext", func: "GsTaichiCal
         # because of the way calling with sequential args works.
         # We need to look at the child's declaration - via metas - in order to get the name they use.
         # We can't tell their name just by looking at our own metas.
+        #
+        # One issue is when calling data-oriented methods, there will be a `self`. We'll detect this
+        # by seeing if the childs arg_metas_expanded is exactly 1 longer than len(node.args) + len(node.kwargs)
+        callee_func: Func = node.func.ptr.wrapper  # type: ignore
+        has_self = len(node.args) + len(node.keywords) + 1 == len(callee_func.arg_metas_expanded)
+        self_offset = 1 if has_self else 0
```
duburcqa, line 82: Isn't `BoundGsTaichiCallable` used to sort out functions expecting self as input argument from others? If so, why not just use is isinstance to determine self offset?

hughperkins: interseting. seems like that might work

PR #356, opened 2026-01-27
[Perf] Add ti.perf_dispatch

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_perf_dispatch.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@test_utils.test()
+def test_perf_dispatch() -> None:
+    class ImplEnum(IntEnum):
+        serial = 0
+        a_shape0_lt2 = 1
+        a_shape0_ge2 = 2
+
+    @ti.perf_dispatch(get_geometry_hash=lambda a, b: hash(a.shape + c.shape))
```
erizmr: `lambda a, b: hash(a.shape + c.shape)` is it `b.shape` instead of `c.shape` here? And is python built-in `hash` deterministic?

tests/python/test_perf_dispatch.py
```diff
@@ -0,0 +1,63 @@
...
+        my_func1(a, c)
+        assert (a.to_numpy()[:5] == [0, 5, 20, 45, 80]).all()
+        if it <= NUM_WARMUP:
+            assert c[ImplEnum.serial] == 1
+            assert c[ImplEnum.a_shape0_lt2] == 0
+            assert c[ImplEnum.a_shape0_ge2] == 1
+        else:
+            assert c[ImplEnum.a_shape0_ge2] == 1
+            assert c[ImplEnum.a_shape0_lt2] == 0
+            assert c[ImplEnum.a_shape0_ge2] == 0
```
erizmr: looks a typo here `assert c[ImplEnum.a_shape0_ge2] == 0` and `assert c[ImplEnum.a_shape0_ge2] == 1`? Is it `c[ImplEnum.serial] == 0`? As after warming up, we only want to run the fastest one given the geometry hash

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,74 @@
...
+                continue
+            runtime.sync()
+            start = time.time()
+            res = underlying.underlying(*args, **kwargs)
+            runtime.sync()
+            end = time.time()
+            elapsed = end - start
+            speeds_l.append((elapsed, underlying))
+        self._trial_count_by_geometry_hash[geometry_hash] += 1
+        if self._trial_count_by_geometry_hash[geometry_hash] > NUM_WARMUP:
```
erizmr: Do we intend to start to sort the different impls in the 3rd calls not the 2nd?

tests/python/test_perf_dispatch.py
hughperkins: addressed in 1045d4aa3

tests/python/test_perf_dispatch.py
hughperkins: This is a very good spot. And also led me to notice this bit of code wasnt being called, and my loop was too short. And when I made it longer, I noticed this PR assumed idmpotent kernels. 😅 Anyway, fixed now, in fec23726a

python/gstaichi/lang/_perf_dispatch.py
hughperkins: yes, that was intentional.  if num warmup is 2, then the first 2 iterations are warmups, and iteration idx 2, ie third iteartion is measurement

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
+import time
+from collections import defaultdict
+from typing import Callable
+
+from . import impl
+from ._gstaichi_callable import GsTaichiCallable
+
+
+class DispatchKernelImpl:
+    kernel_impl_idx = 1
```
duburcqa: I'm in favour of explicit typing for class attributes:
```
kernel_impl_idx: int = 1
```

tests/python/test_perf_dispatch.py
```diff
@@ -0,0 +1,69 @@
...
+    @ti.kernel
+    def my_func1_impl_serial(a: ti.types.NDArray[ti.i32, 1], c: ti.types.NDArray[ti.i32, 1]) -> None:
+        B = a.shape[0]
+        ti.loop_config(serialize=True)
+        for b in range(B):
+            a[b] = a[b] * b
+            c[ImplEnum.serial] = 1
+
+    @my_func1.register(is_compatible=lambda a, c: a.shape[0] < 2)
+    @ti.kernel
```
duburcqa: Is the order of the decorator important? In any case, the behaviour went flipping the order should be assessed in this unit test, whether it is checking that nothing special happens, or a clear exception is thrown.

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __init__(self, underlying: GsTaichiCallable, is_compatible: Callable | None) -> None:
+        self.is_compatible: Callable | None = is_compatible
+        self._underlying: GsTaichiCallable = underlying
+        self.kernel_impl_idx = DispatchKernelImpl.kernel_impl_idx
+        DispatchKernelImpl.kernel_impl_idx += 1
+
+    @property
+    def underlying(self) -> GsTaichiCallable:
+        return self._underlying
```
duburcqa: I find it a bit unusual to use a class to store these information instead of a pair of plain list for kernels and compatibility callbacks.

duburcqa: I would suggest to rather adopt the terminology `__wrapped__` that is ubiquitous in standard libraries for thin wrappers, like this one in my view.

duburcqa: What about implementing `__call__`, so that it is never necessary to extract the raw pointer to call the kernel? It feels more intuitive and convenient to me, and much more in line with standard thin wrappers.

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class KernelSpeedChecker:
+    num_warmup = 2
```
duburcqa: Same. I would recommend: `num_warmup: int = 2`

duburcqa: I would recommend defining a global constant in uppercase right after import `NUM_WRAMUP = 2` or `NUM_WARMUP_DEFAULT = 2` and do `num_warmup = NUM_WARMUP_DEFAULT`, instead of hard-coding 2 directly here.

tests/python/test_perf_dispatch.py
hughperkins: I'm pretty sure it will break, but I can check this point.

hughperkins: oh, "clear exceptoin is thrown" 🤔

hughperkins: addressed in 01554f86d

python/gstaichi/lang/_perf_dispatch.py
hughperkins: Understood. Different programming styles. Personally I like using objects for storing related informations.

python/gstaichi/lang/_perf_dispatch.py
hughperkins: That's admittedly what I had initially. Ok, I can revert back to tath.

hughperkins: addressed in fbff1f62d

python/gstaichi/lang/_perf_dispatch.py
hughperkins: addressed in f2169d6da

hughperkins: addressed in f2169d6da

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __init__(self, get_geometry_hash: Callable) -> None:
+        self._get_geometry_hash: Callable = get_geometry_hash
+        self._underlying_by_idx: dict[int, DispatchKernelImpl] = {}
```
duburcqa: The notion of "underlying" is confusing I feel. For me it should either be simply "kernel" or "implementation". Eventually "variant", or "overload". Something like this. If you want to be explicit, maybe "wrapped_implementations". But I think "implementation" is just fine and consistent with standard library for overload mechanism.

python/quadrants/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+        if self._trial_count_by_underlying_idx_by_geometry_hash[geometry_hash][underlying_idx] >= self.num_warmup:
+            self._times_by_underlying_idx_by_geometry_hash[geometry_hash][underlying_idx].append(elapsed)
+        if self._finished_trials(geometry_hash=geometry_hash):
+            self._calculate_fastest(geometry_hash)
+            speeds_l.sort(key=lambda x: x[0], reverse=False)
+            self._fastest_by_geometry_hash[geometry_hash] = speeds_l[0][1]
+        return res
+
+
+def perf_dispatch(*, get_geometry_hash: Callable):
```
duburcqa, line 186: I think it is very important to document this entrypoint, and introduce the notion of "geometry", which is quite uncommon to me (it was the first time I heard about it). Moreover, it would be interesting to also provide an example and explain the rational behind it (what is it used for and what users should be careful about). Similar, it would be nice to also mention the trick of using hash of tuple of addresses for all taichi tensor if memory is preallocated.

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+            lambda: defaultdict(int)
+        )
+        self._fastest_by_geometry_hash: dict[int, DispatchKernelImpl] = defaultdict(None)
+        self._times_by_underlying_idx_by_geometry_hash: dict[int, dict[int, list[float]]] = defaultdict(
+            lambda: defaultdict(list)
+        )
+
+    def register(
+        self, kernel: GsTaichiCallable | None = None, *, is_compatible: Callable[[dict], bool] | None = None
+    ) -> Callable[[GsTaichiCallable], GsTaichiCallable]:
```
duburcqa: Similarly, it is very important to add docstring to this method, and details about the expected use for `is_compatible`, ideally with the (theoretical) use cases for users to better understand in which context it is relevant.

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+        self._underlying: GsTaichiCallable = underlying
+        self.kernel_impl_idx = DispatchKernelImpl.kernel_impl_idx
+        DispatchKernelImpl.kernel_impl_idx += 1
+
+    @property
+    def underlying(self) -> GsTaichiCallable:
+        return self._underlying
+
+
+class KernelSpeedChecker:
```
duburcqa: This class is not a kernel speed checker. Why calling it like this? It is a `KernelDispatcher`, or to be precise `ImplementationsDispatcherSamplingBasedPerformance` ?

python/gstaichi/lang/_perf_dispatch.py
hughperkins: addressed in 8646e34a9b78c1d3a2bb0cd6dba086ae8961e3d3

python/gstaichi/lang/_perf_dispatch.py
hughperkins: renamed to PerformanceDispatcher in dac8e2e80

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+    num_warmup = 2
+
+    def __init__(self, get_geometry_hash: Callable) -> None:
+        self._get_geometry_hash: Callable = get_geometry_hash
```
duburcqa: You should make this class Generic, so that you can strongly type `geometry_hash` instead of weakly typing it as a callable. It would make best use of static type checker:
```python
TV = TypeVar("TV")


class MyClass(Generic[TV]):
    init_kwargs: TV

    def __init__(self, **kwargs: Unpack[TV]) -> None:
        self.init_kwargs = kwargs
```

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+            lambda: defaultdict(list)
+        )
+
+    def register(
+        self, kernel: GsTaichiCallable | None = None, *, is_compatible: Callable[[dict], bool] | None = None
+    ) -> Callable[[GsTaichiCallable], GsTaichiCallable]:
+        def decorator(func: GsTaichiCallable) -> GsTaichiCallable:
+            dispatch_impl = DispatchKernelImpl(underlying=func, is_compatible=is_compatible)
+            self._underlying_by_idx[dispatch_impl.kernel_impl_idx] = dispatch_impl
+            return func
```
duburcqa: I would not recommend capturing self this way. In my experience it tends to create circular references leading to impossible to collect dangling pointers (and therefore memory leak). I would recommend partial as always, or nonlocal if you are against it, but only on the attribute that must be accessible in the callable:
```python
def decorator(
    underlying_by_idx: dict[int, DispatchKernelImpl], 
    is_compatible: Callable[[dict], bool], 
    func: GsTaichiCallable,
) -> GsTaichiCallable:
    dispatch_impl = DispatchKernelImpl(underlying=func, is_compatible=is_compatible)
    underlying_by_idx[dispatch_impl.kernel_impl_idx] = dispatch_impl
     return func
     
 return partial(decorator, self._underlying_by_idx, is_compatible)
```
OR
```python
underlying_by_idx = self._underlying_by_idx

def decorator(func: GsTaichiCallable) -> GsTaichiCallable:
    nonlocal underlying_by_idx, is_compatible
    dispatch_impl = DispatchKernelImpl(underlying=func, is_compatible=is_compatible)
    underlying_by_idx[dispatch_impl.kernel_impl_idx] = dispatch_impl
    return func

return decorator
```

python/gstaichi/lang/_perf_dispatch.py
duburcqa: Same, it should be possible to strongly type `is_compatible`

duburcqa: And `_get_compatible_underlyings`

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _get_compatible_underlyings(self, *args, **kwargs) -> dict[int, DispatchKernelImpl]:
+        compatible = {}
+        for underlying_idx, underlying in self._underlying_by_idx.items():
+            if underlying.is_compatible and not underlying.is_compatible(*args, **kwargs):
+                continue
+            compatible[underlying_idx] = underlying
+        return compatible
+
+    def _pick_next_underlying_idx(self, compatible: dict[int, DispatchKernelImpl], geometry_hash: int) -> int:
```
duburcqa: As mentioned, I prefer `implementation` over `underlying`

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+            return decorator(kernel)
+        return decorator
+
+    def _get_compatible_underlyings(self, *args, **kwargs) -> dict[int, DispatchKernelImpl]:
+        compatible = {}
+        for underlying_idx, underlying in self._underlying_by_idx.items():
+            if underlying.is_compatible and not underlying.is_compatible(*args, **kwargs):
+                continue
+            compatible[underlying_idx] = underlying
+        return compatible
```
duburcqa: A dict using integers as key is weird to me. Why not just return a boolean mask? It is much more standard I feel, unless I'm missing something.
```
return tuple([not underlying.is_compatible or underlying.is_compatible(*args, **kwargs) for underlying_idx, underlying in self._underlying_by_idx.items()])
```

python/quadrants/lang/_perf_dispatch.py
hughperkins: documented in 4f51571d3

python/gstaichi/lang/_perf_dispatch.py
duburcqa: Similarly, why `self._underlying_by_idx` is not just a list of implementation? The index should just be an increasing sequence of integer without missing values? What is the point of using a dict for this?

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+        assert least_trials_idx is not None
+        return least_trials_idx
+
+    def _finished_trials(self, geometry_hash: int) -> bool:
+        return (
+            min(self._trial_count_by_underlying_idx_by_geometry_hash[geometry_hash].values())
+            >= KernelSpeedChecker.num_warmup + 1
+        )
+
+    def _calculate_fastest(self, geometry_hash: int) -> None:
```
duburcqa: For me this method should rather be called `_update_fastest` to make it slightly more obvious that it updates ranking in-place. By the way, I'm more used to `compute` terminology rather than `calculate` in this context.

python/gstaichi/lang/_perf_dispatch.py
hughperkins: addressed in dec13628b

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+        least_trials = None
+        for underlying_idx in compatible.keys():
+            trial_count = self._trial_count_by_underlying_idx_by_geometry_hash[geometry_hash].get(underlying_idx, 0)
+            if least_trials is None or trial_count < least_trials:
+                least_trials_idx = underlying_idx
+                least_trials = trial_count
+        assert least_trials_idx is not None
+        return least_trials_idx
+
+    def _finished_trials(self, geometry_hash: int) -> bool:
```
duburcqa: I like adding `_get_` prefix for this kind of helper methods.

python/gstaichi/lang/_perf_dispatch.py
duburcqa: Similarly, why not just `_get_` instead of `_pick_`?

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+            if underlying.is_compatible and not underlying.is_compatible(*args, **kwargs):
+                continue
+            compatible[underlying_idx] = underlying
+        return compatible
+
+    def _pick_next_underlying_idx(self, compatible: dict[int, DispatchKernelImpl], geometry_hash: int) -> int:
+        least_trials_idx = None
+        least_trials = None
+        for underlying_idx in compatible.keys():
+            trial_count = self._trial_count_by_underlying_idx_by_geometry_hash[geometry_hash].get(underlying_idx, 0)
```
duburcqa: Same, `self._trial_count_by_underlying_idx_by_geometry_hash[geometry_hash]` should be list (initialised with 0)

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+        underlying_idx = self._pick_next_underlying_idx(compatible=compatible, geometry_hash=geometry_hash)
+        underlying = self._underlying_by_idx[underlying_idx]
+        runtime.sync()
+        start = time.time()
+        res = underlying.underlying(*args, **kwargs)
+        runtime.sync()
+        end = time.time()
+        elapsed = end - start
+        speeds_l.append((elapsed, underlying))
+        self._trial_count_by_underlying_idx_by_geometry_hash[geometry_hash][underlying_idx] += 1
```
duburcqa: I would recommend to define a temporary for `self._trial_count_by_underlying_idx_by_geometry_hash[geometry_hash]`

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+            self._calculate_fastest(geometry_hash)
+            speeds_l.sort(key=lambda x: x[0], reverse=False)
+            self._fastest_by_geometry_hash[geometry_hash] = speeds_l[0][1]
+        return res
+
+
+def perf_dispatch(*, get_geometry_hash: Callable):
+    def decorator(fn: GsTaichiCallable):
+        speed_checker = KernelSpeedChecker(get_geometry_hash=get_geometry_hash)
+        return speed_checker
```
duburcqa: I don't think this temporary adds any value here.

python/gstaichi/lang/_perf_dispatch.py
hughperkins: this assumes the list of implementatons wont change. Then if it  does change, the bugs would be harder to identify I feel. I feel the more assumptions we make, the more things that can break.

python/gstaichi/lang/_perf_dispatch.py
duburcqa: Your decorator is not doing anything with `fn` for now? Maybe you could store annotations and check that the annotations for all implementations that are being registered are consistent with the prescribed annotations?

python/quadrants/lang/_perf_dispatch.py
duburcqa, line 186: `perf_dispatch` is a weird name I think, but I get what you have in mind.

Interestingly, this dispatcher may be useful beyond performance, typically when registering one implementation that is CPU only, and one that is GPU only, and nothing more. Explicitly restricting this dispatcher to performance concerns sounds like a missed opportunity to me.

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,118 @@
...
+        end = time.time()
+        elapsed = end - start
+        speeds_l.append((elapsed, underlying))
+        self._trial_count_by_underlying_idx_by_geometry_hash[geometry_hash][underlying_idx] += 1
+        if self._trial_count_by_underlying_idx_by_geometry_hash[geometry_hash][underlying_idx] >= self.num_warmup:
+            self._times_by_underlying_idx_by_geometry_hash[geometry_hash][underlying_idx].append(elapsed)
+        if self._finished_trials(geometry_hash=geometry_hash):
+            self._calculate_fastest(geometry_hash)
+            speeds_l.sort(key=lambda x: x[0], reverse=False)
+            self._fastest_by_geometry_hash[geometry_hash] = speeds_l[0][1]
```
duburcqa: I think you should skip collecting statistics if there is a single compatible implementation available at this point. This would allow using this dispatcher to unify one implementation that is CPU-only and one implementation that is GPU-only, with barely noticeable performance hit.

python/gstaichi/lang/_perf_dispatch.py
hughperkins: What type are you imaginging I would put into the type of the PerformanceDispatcher class?

python/gstaichi/lang/_perf_dispatch.py
hughperkins: addressed in 9427ffc8c

python/gstaichi/lang/_perf_dispatch.py
hughperkins: addressed earlier

python/gstaichi/lang/_perf_dispatch.py
hughperkins: addressed in 26af40250

python/gstaichi/lang/_perf_dispatch.py
hughperkins: Oh I disagree fairly strongly here, because then if the list mutates in some way, we retrieve the wrong count.

python/quadrants/lang/_perf_dispatch.py
hughperkins: I feel like using this instead fo a simple `if ti.static(arch == ti.cpu) ... else:...` is overkill?

Like, imagine I create a PR onto Genesis, where I replace all `if`s of the type above, with `ti.perf_dispatch`, are you saying you would accept that PR?

hughperkins: (also, note that many such conditionals are inside kernels, eg inside the solver, for Darwin, and not at the level of choosing between kernels.)

python/gstaichi/lang/_perf_dispatch.py
hughperkins: 🤔

hughperkins: addressed in 3f1426181

python/gstaichi/lang/_perf_dispatch.py
hughperkins: If the list mutates in some way then we will get weird and hard to diagnose bugs I feel.

python/gstaichi/lang/_perf_dispatch.py
hughperkins: I feel like comparing annotations is a big can of worms. Our annotations are fairly broken right now, in general, and I don't think it's a high priority task to fix that.

What I could do is to check that at least the names of the parameters match. Thoughts?

Note: an experiment:
```
import gstaichi as ti

ti.init(ti.gpu)

a = ti.types.NDArray
a1 = ti.types.NDArray
a2 = ti.types.NDArray()
a3 = ti.types.NDArray[ti.i32, 1]
a3b = ti.types.NDArray[ti.i32, 1]
a4 = ti.types.NDArray[ti.i32, 2]
a4b = ti.types.NDArray[ti.i32, 2]

print(a == a1)
print(a == a2)
print(a == a3)
print(a3 == a3b)
print(a4 == a4b)
print(a3 == a4)
print(type(a) == type(a1), type(a), type(a1))
print(type(a) == type(a2), type(a2))
print(type(a) == type(a3), type(a3))
print(type(a) == type(a3b), type(a3b))
print(type(a) == type(a4))
print(type(a) == type(a4b))
```

output:
```
[GsTaichi] version 4.6.1, llvm 20.1.0, commit 025cac4d, osx, python 3.10.18
[GsTaichi] Starting on arch=metal
True
False
False
False
False
False
True <class 'type'> <class 'type'>
False <class 'gstaichi.types.ndarray_type.NdarrayType'>
False <class 'gstaichi.types.ndarray_type.NdarrayType'>
False <class 'gstaichi.types.ndarray_type.NdarrayType'>
False
False
```

python/gstaichi/lang/_perf_dispatch.py
duburcqa: > this assumes the list of implementatons wont change. 

No. Why do you think it does?

> Then if it does change, the bugs would be harder to identify I feel. I feel the more assumptions we make, the more things that can break.

I don't think it applies here. Here the only assumption that we are making is that no implementation gets deleted or added between evaluation of the boolean mask and where it is used. I think this assumption is perfectly safe.

> I feel the more assumptions we make, the more things that can break.

I don't share this opinion. I prefer starting with simple implementations designed to be easy to extend, and adding more features / removing restrictions as we start facing limitations, rather than starting with something that is doing more than I need, and eventually ending up with an implementation that is more complex that it has to be and therefore more demanding to maintain and debug. But this means that the assumptions that are made are clearly identified and guarded with proper exception raising, not just ignoring them of course.

> If the list mutates in some way then we will get weird and hard to diagnose bugs I feel.

It returns a tuple, not a list. It is immutable.

python/gstaichi/lang/_perf_dispatch.py
duburcqa: > Oh I disagree fairly strongly here, because then if the list mutates in some way, we retrieve the wrong count.

You don't. When you register a new function you are supposed to expend this list of course. Same if you delete some (which is not supported but still).

python/gstaichi/lang/_perf_dispatch.py
duburcqa: > What I could do is to check that at least the names of the parameters match. Thoughts?

I think it is a good start! Let's ignore annotations for now. But I would recommend storing them, just in case. Maybe it could be useful for the end user when implementing some advanced `get_geometry_hash` or `is_compatible`? It feels like something that should be an attribute of the dispatcher.

python/quadrants/lang/_perf_dispatch.py
duburcqa: > Like, imagine I create a PR onto Genesis, where I replace all ifs of the type above, with ti.perf_dispatch, are you saying you would accept that PR?

Well, why not? I have no specific concern with this.

I still do think it can be used beyond just performance concerns. CPU vs GPU was just a basic example. For example, our accelerated Jacobin implementation does not support more than 96 dofs, so we have to fallback to another implementation in this case.

> (also, note that many such conditionals are inside kernels, eg inside the solver, for Darwin, and not at the level of choosing between kernels.)

This is probably the strongest point. Yes actually you are right, these logics are inside kernels, when selecting the right `ti.func` based on some static conditions. So I guess you are right.

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,208 @@
...
+
+        res = None
+        speeds_l = []
+        runtime = impl.get_runtime()
+        compatible = self._get_compatible_kernels(*args, **kwargs)
+        if len(compatible) == 0:
+            raise GsTaichiRuntimeError("No suitable kernels were found.")
+
+        elif len(compatible) == 1:
+            self._fastest_by_geometry_hash[geometry_hash] = list(compatible.values())[0]
```
duburcqa: I would recommend:
```
self._fastest_by_geometry_hash[geometry_hash] = next(iter(compatible.values()))
```
Or
```
(self._fastest_by_geometry_hash[geometry_hash],) = compatible.values()
```

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,208 @@
...
+            return self._fastest_by_geometry_hash[geometry_hash](*args, **kwargs)
+
+        kernel_idx = self._get_next_kernel_idx(compatible=compatible, geometry_hash=geometry_hash)
```
duburcqa: What about adding a comment on top of this block to explain what is happening here (and probably some docstring for the dispatch decorator):
* What hypothesis are we doing? What are the limitations / drawback / advantages?
* What is the status of idempotent?
* Why sync before and after?
* Are we determining which algorithm is the fastest by sampling or hypotheses based on flop estimate for a given hardware? How many stats we collect? What is the rational to select a kernel? Are we implementing some epsilon-greedy algorithm to keep sampling non-fastest variants just in case the distribution is shifting over time? Is it possible for the user to control exploration vs exploitation?

python/gstaichi/lang/_perf_dispatch.py
hughperkins: I'm 90% sure I tried the former, but it said someting like 'keyvalues are not iterable'.

will try the latter.

hughperkins: oh, I didnt add `iter` 👀

tests/python/test_perf_dispatch.py
```diff
@@ -0,0 +1,84 @@
...
+            my_func1(a, c)
+            assert (a.to_numpy()[:5] == [0, 5, 10, 15, 20]).all()
+        if it <= NUM_WARMUP:
+            assert c[ImplEnum.serial] == 1
+            assert c[ImplEnum.a_shape0_lt2] == 0
+            assert c[ImplEnum.a_shape0_ge2] == 1
+        else:
+            assert c[ImplEnum.a_shape0_ge2] == 1
+            assert c[ImplEnum.a_shape0_lt2] == 0
+            assert c[ImplEnum.serial] == 0
```
duburcqa: Serial not being selected is robust?

tests/python/test_perf_dispatch.py
```diff
@@ -0,0 +1,84 @@
...
+            a.fill(5)
+            my_func1(a, c)
+            assert (a.to_numpy()[:5] == [0, 5, 10, 15, 20]).all()
+        if it <= NUM_WARMUP:
+            assert c[ImplEnum.serial] == 1
+            assert c[ImplEnum.a_shape0_lt2] == 0
+            assert c[ImplEnum.a_shape0_ge2] == 1
+        else:
+            assert c[ImplEnum.a_shape0_ge2] == 1
+            assert c[ImplEnum.a_shape0_lt2] == 0
```
duburcqa: I would recommend making `my_func1_impl_a_shape0_lt_2` a better candidate (faster) than `my_func1_impl_a_shape0_ge_2` if it was compatible, just to make sure the compatibility filter is doing the job.

tests/python/test_perf_dispatch.py
```diff
@@ -0,0 +1,84 @@
...
+        ti.loop_config(serialize=False)
+        for b in range(B):
+            a[b] = a[b] * b
+            c[ImplEnum.a_shape0_ge2] = 1
+
+    N = 10000000
+    a = ti.ndarray(ti.i32, (N,))
+    c = ti.ndarray(ti.i32, (10,))
+
+    for it in range((NUM_WARMUP + 5)):
```
duburcqa, line 72: I would suggest to rather use `my_func1.num_warmup`.

tests/python/test_perf_dispatch.py
hughperkins: No. I fully anticipate needing to migrate this test to use time.sleep soonish. Perhaps now?

tests/python/test_perf_dispatch.py
hughperkins: yeah, ok. migrating to time.sleep will make this both easier, and more obvoius.

python/quadrants/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,208 @@
+import time
+from collections import defaultdict
+from typing import Any, Callable
+
+from . import impl
+from ._gstaichi_callable import GsTaichiCallable
+from .exception import GsTaichiRuntimeError, GsTaichiSyntaxError
+
+NUM_WARMUP: int = 2
```
duburcqa, line 10: Wait, my point what to define `num_warmup` inside the class, plus `NUM_WARMUP_DEFAULT` as a global variable. In the future, we may support more advanced control on the way statistics are computed to pick the best kernel, in this case it would make sense for `num_warmup` to be an optional user-specified argument  for each `perf_dispatch` instance.

tests/python/test_perf_dispatch.py
hughperkins: thats what I had yesterday 🤔  Oh, you mean, I should set that value from `NUM_WARMUP` ?

python/quadrants/lang/_perf_dispatch.py
hughperkins: ok, got it. makes sense. will look into this.

python/gstaichi/lang/_perf_dispatch.py
hughperkins: the trial counts, if they are a list, could become out of sync. strongly prefer to make them a dict, indexed by kernel index.

python/gstaichi/lang/_perf_dispatch.py
duburcqa: `ARGS = TypeVar("ARGS")` should be good enough (basically my example).

python/gstaichi/lang/_perf_dispatch.py
duburcqa: > the trial counts, if they are a list, could become out of sync. 

Why? I don't see any reason for this. You have a register method that is the only entrypoint for users. it is easily to make sure that everything is in sync.

hughperkins: why do you dislike using a dictionary? what downsides do you see to using a dict? Note that this bit of code is not performance critical, since we are only in it for a few iterations.

duburcqa: If you really want to go with dict, then I don't see the point of user kernel indices and I would recommend replacing them by the wrapped kernel or the wrapper itself (either the object directly or its unique address via `.id()`, both should work fine, slightly prefer the object directly to avoid useless additional indirection). Then it is consistent. But incrementing kernel indices manually while not using a list to store information is mixing two concepts in my view.

tests/python/test_perf_dispatch.py
duburcqa: > Perhaps now?

I guess! That would be nice :)

python/gstaichi/lang/_perf_dispatch.py
hughperkins: the kernel ids dont start from 1 anyway. at least not in tests. they dont reset to 1 when you call ti.reset, or run a new test.

The ponit of the ids is that I don't trust .id to be unique across time, beacuse I've seen multiple objects with the same id too many times.

tests/python/test_perf_dispatch.py
duburcqa: Yeah, nothing more x)

python/gstaichi/lang/_perf_dispatch.py
hughperkins: like first test run, kernel ids wil be : 1, 2, 3. Second test run (eg metal instead of cpu), they'll. be: 4,5,6

hughperkins: also, note that kernel ids are global across all instances of perf dispatch.

hughperkins: So, the kernel ids are a lot like 'id', but without the cognitive load of thinking about what happesn if one of the objects gets recreated with the same id, but is a different object. This is why I added kernel id.

duburcqa: > the kernel ids dont start from 1 anyway. at least not in tests. they don't reset to 1 when you call ti.reset, or run a new test.

This is simply not true. Addresses of module-level functions do not change. It is impossible. Once the module has been imported to are doomed. No way to reported it to reallocate function pointer. It is already dead at this point.

hughperkins: It's not impossible for the same id to represent differnet objects. I've seen it happen. Here's how to do it:

```
a = MyCoolObject()
a.id
# 1234578
del a
a = MyCoolObject()
a.id
# still 1234578
```

hughperkins: I'd prefer not to have to think about lots of technical details, when using a kernel id keeps things nice and simple, easy to reason about, safe.

duburcqa: note that I edited my previous message.

duburcqa: If you use the kernel itself as key, then your concern about function addresses that may be changing are moot, whether it is true or not is not really an issue anymore.

hughperkins: wouldnt the kernel have to be hashable to use it as a key? 🤔

duburcqa: Yes. Still, any python native function or custom class instance can be used as dict key:

```python
def fun():
    pass
    
class Class:
    pass

cls = Class()

d = {fun: 2, cls: 1}
assert d[fun] == 2 * d[cls]
```

hughperkins: ok. how does this work? using the id? hashing the object? something else?

python/quadrants/lang/_perf_dispatch.py
hughperkins: addressed in 8b3c7beb1

tests/python/test_perf_dispatch.py
hughperkins: addressed in 8b3c7beb1

tests/python/test_perf_dispatch.py
hughperkins: addresseed

tests/python/test_perf_dispatch.py
hughperkins: addressed

python/gstaichi/lang/_perf_dispatch.py
hughperkins: addressed in ec6593796

python/gstaichi/lang/_perf_dispatch.py
hughperkins: addressed

python/gstaichi/lang/_perf_dispatch.py
hughperkins: sure, but the tuple is based on the underlying list, which is mutable.

python/gstaichi/lang/_perf_dispatch.py
hughperkins: tried something, but typing is pretty broken anyway. I get squiggly red lines everywhere anyway...

Screenshot 2026-02-06 at 08 32 54

python/gstaichi/lang/_perf_dispatch.py
hughperkins: addressed in c89c10ff9

python/gstaichi/lang/_perf_dispatch.py
duburcqa: Under the hood, `dict` is using the output of `hash()` to build the hashing table when inserting a new item, then equality comparison to make sure it has retrieved the expected object. The default hash for python objects is `(id >> 4) | ((id << 60) & 0xffffffffffffffff)` (see https://stackoverflow.com/a/38519187/4820605).

python/gstaichi/lang/_perf_dispatch.py
duburcqa: Maybe Cursor would be able to fix it? If not, you can give up, it is a shame but not a big deal. Also, I quite confident @CharlesMasson should be able to fix it.

tests/python/test_perf_dispatch.py
```diff
@@ -0,0 +1,105 @@
...
+    def my_func1(a: ti.types.NDArray[ti.i32, 1], c: ti.types.NDArray[ti.i32, 1]): ...
+
+    @my_func1.register
+    @ti.kernel
+    def my_func1_impl_slow(a: ti.types.NDArray[ti.i32, 1], c: ti.types.NDArray[ti.i32, 1]) -> None:
+        B = a.shape[0]
+        for b in range(B):
+            a[b] = a[b] * b
+            c[ImplEnum.serial] = 1
+            time.sleep(0.05)
```
duburcqa: `time.sleep` inside kernels is supported ?!

tests/python/test_perf_dispatch.py
duburcqa: Actually no it is not addressed. I said `my_func1_impl_a_shape0_lt_2` a better candidate (faster) than `my_func1_impl_a_shape0_ge_2` if it was compatible.

python/quadrants/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,256 @@
...
+        We are comparing algorithms based on empirical runtime.
+
+        Note that for best results, sets of input arguments that have different runtimes should map to different
+        geometries, otherwise the comparison between runtimes might not be fair, and an inappropriate implementation
+        kernel might be selected.
+
+        We are not implementing an epsilon-greedy algorithm to keep sampling non-fastest variants just in case the
+        distribution is shifting over time.
+
+        It is not possible for you to control exploration vs exploitation.
```
duburcqa, line 149: Information is there. Not enjoyable to read due to both formatting and wording but at least it is there x)

python/gstaichi/lang/_perf_dispatch.py
hughperkins: ok, so it's hashing on id, which I explicitly would like to avoid please, since multiple objects can have identical ids, over time.

tests/python/test_perf_dispatch.py
hughperkins: interesting question. Let me migrate to use linear congruential generator.

hughperkins: addressed in affd5375d

tests/python/test_perf_dispatch.py
hughperkins: addressed in affd5375d

python/gstaichi/lang/_perf_dispatch.py
duburcqa: This is non-sense. I'm not going to approve this.

duburcqa: You cannot have multiple objects with identical ids as keys of the same dict. This is completely impossible. 2 objects CANNOT have the same ids. It is just that ids can be reused if an object gets garbage collected. Nothing new here. But since all your objects are used as keys of the same dict anyway, they are not going anyway and cannot be garbage collected by definition.

Moreover, re-read what I said: it is NOT using id to return value but equality check in case of hash collision. So even if that was true, this would not be a valid reason anyway.

hughperkins: Ok, you're saying:
- previously, when we were using ids, it was in the context of using WeakRef, where it is in fact possible to have multple ojbects with the same id, however
- when storing an object as a key in a dict, then the dict prevents the object being garbage collected, so not an issue?

So, the compromise you are proposing is:
- can still use a dict, but
- rather than using an integer id, use the object itself as the key?

python/gstaichi/lang/_perf_dispatch.py
hughperkins: Oh, I think it's a fundamental issue that involves the fact that we are using a c++ object DataTypeCxx as the dtype.

I spent a bit of time trying to fix that at one point, but very challenging.

Given up for now.

Previous attempts, for example:
- this branch https://github.com/hughperkins/taichi/commits/hp/prim-typing/
    - eg this commit https://github.com/hughperkins/taichi/commit/39d533ac09885bb7e0d9b7bfc3ffca1f32465972

Huge rabbit hole... hard to justify spending much time on it, since most people don't care too much about typing.

python/gstaichi/lang/_perf_dispatch.py
duburcqa: > previously [...] so not an issue?

Exactly!

> can still use a dict, but [...] use the object itself as the key?

Exactly!

python/gstaichi/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,257 @@
...
+        least_trials = None
+        for dispatch_impl in compatible_set:
+            trial_count = self._trial_count_by_dispatch_impl_by_geometry_hash[geometry_hash].get(dispatch_impl, 0)
+            if least_trials is None or trial_count < least_trials:
+                least_trials_dispatch_impl = dispatch_impl
+                least_trials = trial_count
+        assert least_trials_dispatch_impl is not None
+        return least_trials_dispatch_impl
+
+    def _finished_trials(self, geometry_hash: int) -> bool:
```
duburcqa: Why did you removed the `_get_` prefix again?

python/quadrants/lang/_perf_dispatch.py
```diff
@@ -0,0 +1,257 @@
...
+    def _finished_trials(self, geometry_hash: int) -> bool:
+        return min(self._trial_count_by_dispatch_impl_by_geometry_hash[geometry_hash].values()) >= self.num_warmup + 1
+
+    def _update_fastest(self, geometry_hash: int) -> None:
+        speeds_l = []
+        for dispatch_impl, elapsed_time in self._times_by_dispatch_impl_by_geometry_hash[geometry_hash].items():
+            speeds_l.append((dispatch_impl, elapsed_time))
+        speeds_l.sort(key=lambda x: x[1], reverse=False)
+        self._fastest_dispatch_impl_by_geometry_hash[geometry_hash] = speeds_l[0][0]
+
```
duburcqa, line 122: This pattern weird. I would rather recommend:
```python
times_by_dispatch_impl = self._times_by_dispatch_impl_by_geometry_hash[geometry_hash]
fastest_dispatch, _ = min(times_by_dispatch_impl.items(), key=lambda x: x[1])
self._fastest_dispatch_impl_by_geometry_hash[geometry_hash] = fastest_dispatch
```

duburcqa: Note that you could replace the lambda by `operator.itemgetter(1)` if you wanted. It is up to you.
https://docs.python.org/3/library/operator.html#operator.itemgetter

python/gstaichi/lang/_perf_dispatch.py
hughperkins: because get to me implies that it's just doing a get, not doing any work. Basically a getter and setter. Made it harder for me to read and debug the code.

duburcqa: In that case I usually use `_compute_` prefix to make it clear that it is doing computations. But I do think some prefix is needed here.

hughperkins: `_compute_are_trials_finished`?

duburcqa: Sounds good!

hughperkins: addressed in 19a22d8dc

PR #361, opened 2026-02-08
[Bug] Add SM version to ptx cache key

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


setup.py
```diff
@@ -72,7 +72,8 @@ def run(self):
         super().run()
         self.build_temp = os.path.join(root_dir, "_skbuild")
         if os.path.exists(self.build_temp):
-            remove_tree(self.build_temp, dry_run=self.dry_run)
+            if not self.dry_run:
```
hughperkins, line 75: this is needed because of some change in setuptools.

happy to move to separate pr if we prefer?

duburcqa: I'm fine. But could you update the minimum version of setuptools accordingly?

PR #375, opened 2026-02-19
[Misc] Add lock around kernel compile

hughperkins:
Issue: #

### Brief Summary

- prevent issues such as https://github.com/Genesis-Embodied-AI/Genesis/actions/runs/22183349188/job/64178329065?pr=2406 , where two threads try to buidl kernels at same time
- added unit test which fails correctly without the lock, and passes with it

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_concurrent_kernels.py
+++ b/tests/python/test_concurrent_kernels.py
@@ -3,6 +3,7 @@ import threading
 import time
 
 import quadrants as ti
+from quadrants.lang import impl
 from quadrants.lang.ast import transform_tree as _original_transform_tree
 
 from tests import test_utils
@@ -25,6 +26,12 @@ def test_concurrent_kernel_materialization():
     triggering a spurious "nested kernels" error or an assertion failure
     on _compiling_callable.
     """
+
+    # Force the one-time runtime/LLVM struct initialization on the main thread.
+    # Some backends (LLVM on x86/Windows) assert that add_struct_module is
+    # called from the main thread; this ensures that has already happened
+    # before worker threads attempt per-kernel materialization.
+    impl.get_runtime().materialize()
 
     @ti.kernel
     def kernel_a(x: ti.types.NDArray[ti.i32, 1]) -> None:


python/quadrants/lang/impl.py
```diff
@@ -344,6 +345,7 @@ def __init__(self, kernels=None):
         self._prog: Program | None = None
         self.src_info_stack = []
         self.inside_kernel: bool = False
+        self.compilation_lock = threading.RLock()
```
duburcqa, line 348: You Lock instead of RLock. There is no reason to make it re-entrant.

hughperkins: We needRLock so we can detect when someone actually calls a kernel from within a kernel.

hughperkins: (which would otherwise deadlock)

duburcqa: > We needRLock so we can detect when someone actually calls a kernel from within a kernel.

Wait, you are not supposed to deadlock in this case?

duburcqa: How is that supposed to be supported?

hughperkins: re-entrant will allow the call from within the compiling kernel to enter the lock.

hughperkins: hmmm, tons of failing tests 🤔  Let me check those...

duburcqa: > re-entrant will allow the call from within the compiling kernel to enter the lock.

Exactly, which is undesirable from what I can tell?

hughperkins: seems pases on linux arm, fails on linux x86 🤔  passes on mac, fails on windows.

duburcqa: Like it is basically the exact same issue than cross-thread no? This is not supported? Unless I'm missing something.

hughperkins: > > re-entrant will allow the call from within the compiling kernel to enter the lock.
> 
> Exactly, which is undesirable from what I can tell?

No. This means:
- no deadlock
- throw the correct error baout not calling a kernel from within a kernel.

hughperkins: > Like it is basically the exact same issue than cross-thread no? This is not supported? Unless I'm missing something.

It's a different case. A differnet issue. We want to throw an exception with a clear message, not just hang/deadlock.

hughperkins: (We already have this ecxception, message. Seems no reason to intentinoally have a deadlock instead I feel?)

duburcqa: if we already have this message, why we need a special handling cross-thread? They are all sharing the same memory so it should be triggered indifferently?

hughperkins: If you mean:
- why do we need locks?
    - because compliation is not thread safe
- why do we need to allow calling into the same kernel again to bypass the lock?
     - because the exception isnt thrown until a bit later in the build

duburcqa: > why do we need locks? because compliation is not thread safe

Ok now I get it. It is because of some race condition. Everything is fine when first checked, but then some other thread starts messing up with what the first thread is doing and there is no check anymore to detect this.  From what I understand the likelihood if this race condition could be dramatically reduced by just moving the location of the check and the flag being set to indicate that the kernel is being compiled.

Anyway, this RLock mechanism is more robust and explicit.

hughperkins: It's not really a race condition. The new lock means if you have:

```
@ti.kernel
def k2():
    ...

@ti.kernel
def k1():
     k2()

k1()
```
... which would have been corrrectly flagged before, then when k2 is called, you need a reentrant lock, or else you will get a deadlock. The check that k2 is being compiled inside k1 happens after the lock, during the compilation itself.

duburcqa: Race condition is the current failure we observe without lock.

duburcqa (approved):

erizmr (approved):

hughperkins:
Thanks!

PR #385, opened 2026-02-24
[Misc] Add new Python backend for unit tests of funcs, coverage testing and debugging

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/quadrants/lang/_py_tensor.py
+++ b/python/quadrants/lang/_py_tensor.py
@@ -88,6 +88,8 @@ class PyTensor(torch.Tensor):
         key = PyTensor._unpack_key(key)
         # Scalar fields (shape ()) are still indexed with field[0] in kernels;
         # torch would raise IndexError on a 0-d tensor, so return the scalar.
+        # .item() (not `return self`) so the caller gets a snapshot, not a
+        # mutable alias to the field's storage.
         if not isinstance(key, tuple) and key == 0 and self.size() == ():
             return self.item()
         return super().__getitem__(key)


python/quadrants/lang/ast/ast_transformers/call_transformer.py
```diff
@@ -120,7 +120,7 @@ def _warn_if_is_external_func(ctx: ASTTransformerFuncContext, node):
     # origin input: 'qwerty {1} {} {1:.3f} {k:.4f} {k:}'.format(1.0, 2.0, k=k)
     # raw_string: 'qwerty {1} {} {1:.3f} {k:.4f} {k:}'
     # raw_args: [1.0, 2.0]
-    # raw_keywords: {'k': <ti.Expr>}
+    # raw_keywords: {'k': <qd.Expr>}
```
duburcqa, line 123: unrelated but ok

python/quadrants/lang/_ndarray.py
```diff
@@ -280,9 +294,14 @@ class ScalarNdarray(Ndarray):
...
-            self.dtype, arr_shape, layout=Layout.NULL, zero_fill=True, dbg_info=_qd_core.DebugInfo(get_traceback())
-        )
+        if impl.is_python_backend():
+            assert torch_is_available
+            torch_dtype = dtype_to_torch_dtype(dtype)
+            self.arr = torch.zeros(shape=arr_shape, dtype=torch_dtype)
+        else:
+            self.arr = impl.get_runtime().prog.create_ndarray(
+                self.dtype, arr_shape, layout=Layout.NULL, zero_fill=True, dbg_info=_qd_core.DebugInfo(get_traceback())
+            )
```
duburcqa: I would recommend to remove guarded torch import, and move it locally inside branch `if impl.is_python_backend():` directly.

python/quadrants/lang/_ndrange.py
```diff
@@ -5,24 +5,37 @@
...
 from quadrants.lang.expr import Expr
 from quadrants.lang.matrix import Matrix
 from quadrants.types.utils import is_integral
 
 
+def _coerce_to_int(v):
+    """Convert 0-d tensors to int for ndrange bounds in python backend."""
+    if torch is not None and isinstance(v, torch.Tensor) and v.ndim == 0:
+        return int(v.item())
+    return v
```
duburcqa: Same. I think it makes more sense to move torch import inside the functions where it is used.

python/quadrants/lang/impl.py
```diff
@@ -71,6 +74,14 @@
...
 )
 
+from . import _py_tensor as py_tensor
+
+torch = None
+try:
+    import torch
+except Exception:
+    pass
+
```
duburcqa: This block is useless here.

python/quadrants/lang/impl.py
```diff
@@ -1223,6 +1267,10 @@ def current_cfg():
     return get_runtime().prog.config()
 
 
+def is_python_backend() -> bool:
+    return get_runtime()._arch == _qd_core.Arch.python
```
duburcqa: `_qd_core.Arch.python` should be replaced by a proxy, unless efficiency does not matter.

python/quadrants/lang/kernel.py
```diff
@@ -302,6 +304,7 @@ def __init__(self, _func: Callable, autodiff_mode: AutodiffMode, _is_classkernel
         self.launch_observations = LaunchObservations()
 
         self.launch_context_buffer_cache = LaunchContextBufferCache()
+        self.arch_is_python: bool | None = None
```
duburcqa: This flag is unused.

python/quadrants/lang/matrix.py
```diff
@@ -249,6 +267,13 @@ class Matrix(QuadrantsOperations):
     _is_matrix_class = True
     __array_priority__ = 1000
 
+    def __new__(cls, arr, dt=None):
+        if impl.is_python_backend():
+            assert torch is not None
```
duburcqa: This check is useless.

python/quadrants/lang/_py_tensor.py
```diff
@@ -0,0 +1,152 @@
...
+    except Exception:
+        pass
+
+    if torch is not None:
+        _TensorBase = torch.Tensor
+    else:
+        _TensorBase = object
+
+
+class MyTorchTensor(_TensorBase):
```
duburcqa: This is not a correct name.

python/quadrants/lang/_py_tensor.py
```diff
@@ -0,0 +1,152 @@
...
+    torch = None
+    try:
+        import torch
+    except Exception:
+        pass
+
+    if torch is not None:
+        _TensorBase = torch.Tensor
+    else:
+        _TensorBase = object
```
duburcqa: I don't think it should be possible to import this file without torch installed. It suggest to delegate the responsibility to make sure torch is available to the other modules importing this one.

python/quadrants/lang/_py_tensor.py
```diff
@@ -0,0 +1,152 @@
...
+    def x(self):
+        return super().__getitem__(0)
+
+    @property
+    def y(self):
+        return super().__getitem__(1)
+
+    @property
+    def z(self):
+        return super().__getitem__(2)
```
duburcqa: just do `return super()[0]`

python/quadrants/lang/_py_tensor.py
```diff
@@ -0,0 +1,152 @@
...
+                if isinstance(item, float) and item.is_integer():
+                    item = int(item)
+                yield item
+            else:
+                yield val
+
+    def __getitem__(self, key):  # type: ignore[override]
+        key = MyTorchTensor._unpack_key(key)
+        if not isinstance(key, tuple) and key == 0 and self.size() == ():
+            return self.item()
```
duburcqa, line 94: I don't understand why this is needed.

python/quadrants/lang/matrix.py
```diff
@@ -36,9 +37,26 @@
 from quadrants.types.enums import Layout
 from quadrants.types.utils import is_signed
 
+from . import _py_tensor as py_tensor
+
+torch = None
+try:
+    import torch
+except Exception:
+    pass
```
duburcqa: I would rather switch to local import

python/quadrants/lang/misc.py
```diff
@@ -716,6 +742,7 @@ def is_arch_supported(arch):
         metal: _qd_core.with_metal,
         vulkan: _qd_core.with_vulkan,
         cpu: lambda: True,
+        python: lambda: True,
```
duburcqa: This backend should not be available if torch is not installed.

python/quadrants/lang/struct.py
```diff
@@ -501,7 +511,8 @@ def fill(self, val):
 
     def _initialize_host_accessors(self):
         for v in self._members:
-            v._initialize_host_accessors()
```
duburcqa, line 504: I would rather have `_initialize_host_accessors` a no-op for python backend

python/quadrants/lang/util.py
```diff
@@ -41,9 +47,7 @@ def has_pytorch():
     _env_torch = os.environ.get("QD_ENABLE_TORCH", "1")
     if not _env_torch or int(_env_torch):
         try:
-            import torch  # noqa: F401 pylint: disable=C0415
-
-            _has_pytorch = True
+            _has_pytorch = torch is not None
         except:
             pass
     return _has_pytorch
```
duburcqa: I'm not convinced this new pattern is better.

python/quadrants/lang/_py_tensor.py
hughperkins: thoughts on a better name?

python/quadrants/lang/_py_tensor.py
hughperkins: added comment to th ecode.

hughperkins:
nwe review from opus-4.6:

Overview

  This branch adds a pure-Python backend to Quadrants. When you do qd.init(qd.python), kernels and functions execute as plain Python — no C++ compilation, no GPU needed.
  Fields and ndarrays are backed by PyTorch tensors instead of the C++ runtime.
  22 files changed, +1127 / -30 lines across C++, Python, and tests.

  Architecture

  The design is clean — almost all changes are guarded by is_python_backend() checks that short-circuit into Python-native paths:
  • `Kernel.__call__` / `Func.__call__`: When on the python backend, just call self.func(*py_args, **kwargs) directly — skip compilation, AST transforms, all of it.
  • `qd.field()` / `qd.ndarray()`: Return PyTensor instances (torch tensors) instead of C++ data structures.
  • Atomics (atomic_add, atomic_sub, etc.): Become simple read-modify-write, which is correct since the backend is single-threaded.
  • `sync()`, `loop_config()`: Become no-ops.
  • C++ side: Minimal — just adds python to the arch enum, sets program_impl_ = nullptr in the Program constructor, and skips finalization. Very well contained.


  PyTensor

  The central new abstraction is _py_tensor.PyTensor, a torch.Tensor subclass. Key design choices:
  • `_batch_shape`: Separates batch dims from element dims. A vec3 field of shape (N,) has torch size (N, 3) but .shape returns (N,). This lets kernel code like a.shape[0]
    work transparently.
  • `__torch_function__`: Unwraps PyTensor when mixed with foreign tensor subclasses to avoid dispatch issues.
  • `__getattr__`: Delegates unknown attributes (.norm(), .dot(), etc.) to matrix_ops, so Quadrants math APIs work on these tensors.
  • `_tc` / `_T_tc` / `_np` / `_T_np`: Mimics the Field/Ndarray view convention so downstream code that accesses these attributes works unchanged.


  Potential Concerns

  1. Monkey-patching `DataTypeCxx.__call__`: _install_python_backend_dtype_call() patches the C++ DataType class globally so qd.f32(0.0) returns a Python float. There's a
     guard against double-install, which is good. But it's permanent global mutation — if you later qd.init(qd.cpu), the patched method checks is_python_backend() at call
     time, so it should still work, but it's fragile. If the C++ side ever changes how __call__ works, this could silently mask it.
  2. Overriding `.shape` on a torch.Tensor subclass: This is playing with fire. A lot of PyTorch internals and third-party code assumes .shape equals the actual tensor
     dimensions. The __torch_function__ unwrapping helps, but any code path that bypasses __torch_function__ (C++ torch internals, some serialization paths) could see
     mismatched shape/size.
  3. `cast()` uses string matching: The python backend path in cast() does "float" in str(dtype) or "f16" in str(dtype) — this is fragile. If dtype string representations
     change or new types are added, it could misclassify.
  4. `_coerce_to_int` in ndrange: Needed because some code paths produce 0-d tensors instead of plain ints for loop bounds. This is a symptom of the .shape override
     leaking — it works, but indicates the abstraction isn't fully sealed.
  5. `StructField._register_fields()` guard: Now conditioned on all(isinstance(v, Field) ...), silently skipping registration for PyTensors. Works correctly, but is an
     implicit contract.
  6. `grouped(x)` returns [[i] for i in range(x.shape[0])] — I'd want to verify this matches the semantics of the grouped iterator on other backends. It looks like it
     converts to a per-element iteration rather than actual grouping.

python/quadrants/lang/_py_tensor.py
duburcqa: Why not just return self?

python/quadrants/lang/_py_tensor.py
```diff
@@ -0,0 +1,141 @@
...
+
+def _setup_views(tensor, batch_ndim):
+    """Set _tc/_T_tc/_np/_T_np instance attributes matching the qd.Field/Ndarray convention."""
+    tensor._tc = tensor
+    tensor._T_tc = tensor.movedim(batch_ndim - 1, 0) if batch_ndim > 1 else tensor
+    tensor._np = tensor.numpy()
+    tensor._T_np = tensor._T_tc.numpy()
+    tensor._batch_shape = tensor.size()[:batch_ndim]
+    tensor.grad = None
+    tensor.dual = None
```
duburcqa: Why not defining this attributes at init? Instead of hacking your own class?

hughperkins: here is what cursor said:

Screenshot 2026-03-08 at 16 12 33
Screenshot 2026-03-08 at 16 12 38
Screenshot 2026-03-08 at 16 13 38

python/quadrants/lang/_py_tensor.py
hughperkins: cursor said:

Screenshot 2026-03-08 at 16 17 37

Issue #386, opened 2026-02-25
Memory leakage in amdgpu launch_kernel


LuoShaoyinn:
**Describe the bug**
In quadrants/runtime/amdgpu/kernel_launcher.cpp:36, a hipMalloc is call and the result saved in ```device_result_buffer```. But after that in line 84, we do:
```
  if (ctx.result_buffer_size > 0) {
    // Malloc_Async and Free_Async are available after ROCm 5.4
    AMDGPUDriver::get_instance().malloc((void **)&device_result_buffer,
                                        ctx.result_buffer_size);
    ctx.get_context().result_buffer = (uint64 *)device_result_buffer;
  }
```
Where the ```device_result_buffer``` is overwritten and the previous memory is leaked. 

**To Reproduce**
Just launch lots of kernel and the VRAM usage increases. 

```py
import torch
import genesis as gs

gs.init(backend=gs.gpu, performance_mode=True)

scene = gs.Scene(
    show_viewer   = True,
    rigid_options = gs.options.RigidOptions(
        dt                = 0.01,
    ),
)

plane = scene.add_entity(
    gs.morphs.Plane(),
)

franka = scene.add_entity(
    gs.morphs.MJCF(file='xml/franka_emika_panda/panda.xml'),
)

scene.build(n_envs=4)

# control all the robots
franka.control_dofs_position(
    torch.tile(
        torch.tensor([0, 0, 0, -1.0, 0, 1.0, 0, 0.02, 0.02], device=gs.device), (4, 1)
    ),
)

for i in range(10000):
    scene.step()
```
The memory usage increase when running step, and eventually OOM.

PR #387, opened 2026-02-25
Memory leak in amdgpu kernel launching

LuoShaoyinn:
Issue: #386

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/runtime/amdgpu/kernel_launcher.cpp
+++ b/quadrants/runtime/amdgpu/kernel_launcher.cpp
@@ -35,10 +35,12 @@ void KernelLauncher::launch_llvm_kernel(Handle handle,
   char *device_result_buffer{nullptr};
   // Here we have to guarantee the result_result_buffer isn't nullptr
   // It is interesting - The code following
-  // L60:           DeviceAllocation devalloc = executor->allocate_memory_on_device(
-  // call another kernel and it will result in 
-  //   Memory access fault by GPU node-1 (Agent handle: 0xeda5ca0) on address (nil). Reason: Page not present or supervisor privilege.
-  // if you don't allocate it. 
+  // L60:           DeviceAllocation devalloc =
+  // executor->allocate_memory_on_device( call another kernel and it will result
+  // in
+  //   Memory access fault by GPU node-1 (Agent handle: 0xeda5ca0) on address
+  //   (nil). Reason: Page not present or supervisor privilege.
+  // if you don't allocate it.
   AMDGPUDriver::get_instance().malloc(
       (void **)&device_result_buffer,
       std::max(ctx.result_buffer_size, sizeof(uint64)));


quadrants/runtime/amdgpu/kernel_launcher.cpp
```diff
@@ -83,8 +83,6 @@ void KernelLauncher::launch_llvm_kernel(Handle handle,
   char *host_result_buffer = (char *)ctx.get_context().result_buffer;
   if (ctx.result_buffer_size > 0) {
     // Malloc_Async and Free_Async are available after ROCm 5.4
```
duburcqa, line 93: This comment is not longer relevant?

quadrants/runtime/amdgpu/kernel_launcher.cpp
```diff
@@ -117,8 +115,8 @@ void KernelLauncher::launch_llvm_kernel(Handle handle,
   if (ctx.result_buffer_size > 0) {
     AMDGPUDriver::get_instance().memcpy_device_to_host(
         host_result_buffer, device_result_buffer, ctx.result_buffer_size);
-    AMDGPUDriver::get_instance().mem_free(device_result_buffer);
   }
+  AMDGPUDriver::get_instance().mem_free(device_result_buffer);
```
duburcqa: Could you add a comment in the code to explain why it should be implemented as you did? Because apparently it is not obvious, otherwise it would not be necessary to fix it.

quadrants/runtime/amdgpu/kernel_launcher.cpp
```diff
@@ -83,8 +83,6 @@ void KernelLauncher::launch_llvm_kernel(Handle handle,
   char *host_result_buffer = (char *)ctx.get_context().result_buffer;
   if (ctx.result_buffer_size > 0) {
     // Malloc_Async and Free_Async are available after ROCm 5.4
-    AMDGPUDriver::get_instance().malloc((void **)&device_result_buffer,
-                                        ctx.result_buffer_size);
     ctx.get_context().result_buffer = (uint64 *)device_result_buffer;
   }
```
duburcqa, line 95: Could you add a comment in the code to explain why it should be implemented as you did? Because apparently it is not obvious, otherwise it would not be necessary to fix it.

quadrants/runtime/amdgpu/kernel_launcher.cpp
hughperkins: well, we are not yet using async, so it seems like it IS still relevant? (perhaps prefixed with TODO). Unless I'm misunderstanding its meaning?

quadrants/runtime/amdgpu/kernel_launcher.cpp
hughperkins: I asked Cursor for a comment and it wrote:

```
  // Free unconditionally: device_result_buffer is always allocated at the top
  // of this function regardless of result_buffer_size.
```
Thoughts @duburcqa ?

quadrants/runtime/amdgpu/kernel_launcher.cpp
hughperkins: Similary, Cursor wrote:
```
  // Allocated unconditionally — must be freed unconditionally at the end of
  // this function (not only when result_buffer_size > 0).
```
Thoughts @duburcqa ?

quadrants/runtime/amdgpu/kernel_launcher.cpp
LuoShaoyinn: This comment is relevant since amdgpu not using async memory now. And ROCm 7.2 do support async memory. But that's fine (or low priority TODO) because it is already a great work of you to run simulation on none NVIDIA GPUs.

quadrants/runtime/amdgpu/kernel_launcher.cpp
LuoShaoyinn: It is a little tricky here - the code following calls ```executor->allocate_memory_on_device``` which may start a kernel on amdgpu (compares to that just call an api on cuda) so we have to guarantee the result_buffer is always available even if the kernel itself don't write anything to it. It is so strange to me that the rocm stack complaining about
```bash
52              if (on_amdgpu_device(data_ptr)) {
(gdb)
55                DeviceAllocation devalloc = executor->allocate_memory_on_device(
(gdb)
Memory access fault by GPU node-1 (Agent handle: 0xeda5ca0) on address (nil). Reason: Page not present or supervisor privilege.

Thread 65 "python3" received signal SIGABRT, Aborted.
[Switching to Thread 0x7ffdabfff6c0 (LWP 1063959)]
__pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
44            return INTERNAL_SYSCALL_ERROR_P (ret) ? INTERNAL_SYSCALL_ERRNO (ret) : 0;
(gdb) bt 5
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff7d02af3 in __pthread_kill_internal (threadid=<optimized out>, signo=6) at pthread_kill.c:89
#2  0x00007ffff7ca81a0 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff7c8f5fe in __GI_abort () at abort.c:77
#4  0x00007ffee2a96fe3 in rocr::core::Runtime::VMFaultHandler (val=<optimized out>, arg=<optimized out>)
    at /usr/src/debug/hsa-rocr/rocm-systems-rocm-7.2.0/projects/rocr-runtime/runtime/hsa-runtime/core/runtime/runtime.cpp:2257
(More stack frames follow...)
(gdb) q
A debugging session is active.

        Inferior 1 [process 1063885] will be killed.
```
if I don't allocate the memory when no return is wanted - and so weird that the complain disappear if I run the script as root. I have been suspecting it as a ROCm bug.  But I don't go further inside the GPU kernel so there might not be very clear where this error occur in the GPU kernel.

quadrants/runtime/amdgpu/kernel_launcher.cpp
```diff
@@ -129,6 +132,8 @@ void KernelLauncher::launch_llvm_kernel(Handle handle,
       executor->deallocate_memory_on_device(itr->second.second);
     }
   }
+  // Since we always allocating above then we should always free
+  AMDGPUDriver::get_instance().mem_free(device_result_buffer);
```
LuoShaoyinn, line 138: Yes you are right here - my mistake. Simple moving the memfree outside the if may result in a UAF, and the reason is the same as above - deallocate_memory_on_device may start another kernel. But I am not quite sure about this because the ROCm kernel  / gdb / rocgdb don't warn about that. But theoritically, there is UAF. So now I think moving the mem_free here is more suitable. Tested on my 7900xtx, the behavior is the same.

PR #411, opened 2026-03-12
[Type] ndarray typing 1: Add eval_str=True to inspect.signature() calls

hughperkins:
…annotations

  ## Summary
  Enables `quadrants` kernels and `@perf_dispatch` functions to work correctly when their type annotations are stringified — either explicitly (`a: "qd.types.NDArray[qd.i32,
  1]"`) or implicitly via `from __future__ import annotations` (PEP 563).
  Previously, `inspect.signature(func)` returned annotations as raw strings, which broke downstream code that expected real type objects (e.g.
  `is_dataclass(parameter.annotation)` silently returned `False` for stringified dataclass annotations, and `NDArray`-style annotations were not recognized).
  ## Changes
  - **New helper** `get_func_signature(func)` in `python/quadrants/lang/exception.py`:
    - Wraps `inspect.signature(func, eval_str=True)` so annotations are resolved to real objects.
    - Catches `NameError` / `AttributeError` raised during annotation evaluation and re-raises as `QuadrantsSyntaxError` with a clear message, so users get a Taichi-flavored
  error instead of a raw Python traceback.
  - **Call-site migration** — all 5 call sites that introspect a user kernel/func signature now go through the helper:
    - `python/quadrants/lang/_func_base.py` (2 sites: `FuncBase` registration + dataclass template arg expansion)
    - `python/quadrants/lang/_kernel_impl_dataclass.py` (1 site: `extract_struct_locals_from_context`)
    - `python/quadrants/lang/_perf_dispatch.py` (2 sites: `PerformanceDispatcher.__init__` + the `register` decorator)
  - **Test** — `tests/python/test_future_annotations.py` adds an end-to-end test of a kernel with `NDArray[qd.i32, 1]` annotations under `from __future__ import annotations`.
## Good points
  - **Fixes a real correctness bug.** Without this, `from __future__ import annotations` (which is increasingly common / will eventually be the default) silently breaks
  dataclass and ndarray kernel arguments.
  - **Single chokepoint.** Centralizing `eval_str=True` + error translation in one helper means future signature-introspection sites only have one obvious thing to call, and the
   error-conversion behavior stays consistent.
  - **User-facing error quality.** A bad annotation now raises `QuadrantsSyntaxError("Invalid type annotation of Taichi kernel: ...")` instead of a bare `NameError` from deep
  inside `inspect`.
  - **Small, focused diff** (+43 / -8 across 5 files) with a dedicated regression test.
  - **Clean import hygiene** — `import inspect` is removed from modules that no longer need it directly.
  - ~**Helper lives in `exception.py`.** `get_func_signature` is a signature utility, not an exception class. It was placed there to avoid an import cycle with `_func_base` /
  `_perf_dispatch`, but `exception.py` is now misnamed for part of its contents. A dedicated `_signature.py` (or putting it in `util.py`) would be cleaner if the import graph
  allows.~ [fixed]
  - **Narrow `except` clause.** Only `NameError` and `AttributeError` are caught. `inspect.signature(..., eval_str=True)` can also surface `SyntaxError`, `TypeError`, or
  arbitrary exceptions from `__getattr__` / forward-ref resolution, which will still bubble up as raw tracebacks. Consider broadening to `Exception` (with the original chained
  via `from e`) or documenting why only these two.
  - ~**Error message hard-codes "Taichi".** The message says *"Invalid type annotation of Taichi kernel"*, but the helper is also used by `@perf_dispatch` and by non-kernel
  `FuncBase` paths. Message should probably be generic ("Invalid type annotation in `{func.__qualname__}`: ...") or context-specific per call site.~ [fixed]
  - ~**`# type: ignore` at the top of `exception.py` is preserved.** The new helper inherits that blanket ignore, so type errors in `get_func_signature` itself won't be caught by
   the type checker. Worth either typing the helper explicitly (`def get_func_signature(func: Callable) -> inspect.Signature`) or moving it to a typed module.~ [fixed]
  - **Behavioral change for already-evaluated annotations.** `eval_str=True` is a no-op for annotations that are already real objects, *but* it does trigger evaluation for any
  string annotation a user may have written intentionally (e.g. forward references to types that only exist at runtime in another scope). Anyone currently relying on getting raw
   strings back from these signatures will see a behavior change. Low risk, but worth calling out in release notes.
  - **Test coverage is thin.** Only one happy-path test (`NDArray` under `from __future__ import annotations`). Missing coverage for:
    - dataclass arguments with stringified annotations (the main motivating case in `_kernel_impl_dataclass.py`)
    - `@perf_dispatch` register / dispatch with stringified annotations
    - the error path — that a bad annotation produces `QuadrantsSyntaxError`, not `NameError`
  - **Two merge commits from `main`.** History could be rebased before merge for a cleaner log (6 logical commits + 2 merges → 6 commits).

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
Opus 4.6 review:

# PR Review: hp/typing-t4-1-eval-str

## Summary

This PR adds `eval_str=True` to all `inspect.signature()` calls in the kernel/func registration path, enabling proper evaluation of string annotations (e.g. from `from __future__ import annotations`). It also wraps the primary call in `_func_base.check_parameter_annotations` with a try/except that converts `NameError` (undefined type in annotation) into `QuadrantsSyntaxError` for clearer user feedback.

## Issues Found

### 1. Inconsistent `inspect.signature` usage in `_perf_dispatch.py`

**Location:** `python/quadrants/lang/_perf_dispatch.py`, line 102

The `PerformanceDispatcher.register` decorator still uses `inspect.signature(func)` without `eval_str=True`:

```python
def decorator(func: Callable | QuadrantsCallable) -> DispatchImpl:
    sig = inspect.signature(func)  # <-- missing eval_str=True
```

The constructor (line 61) uses `eval_str=True` for the prototype function, but registered implementations do not. For consistency and to handle `from __future__ import annotations` in registered implementations, this call should also use `eval_str=True`. If added, it should be wrapped in the same `try/except NameError` pattern for consistent error handling when annotations reference undefined types.

### 2. Missing error handling at other call sites

**Locations:**
- `_func_base.py` line 195: `_populate_global_vars_for_templates`
- `_kernel_impl_dataclass.py` line 76: `extract_struct_locals_from_context`

These calls use `inspect.signature(..., eval_str=True)` but do not catch `NameError`. If a kernel/func has an invalid type annotation (e.g. typo in a dataclass parameter), these will raise a raw `NameError` instead of `QuadrantsSyntaxError`. The error handling in `check_parameter_annotations` is good, but these downstream paths can still surface `NameError` before or in different code paths. Consider adding the same try/except at these sites for consistent user-facing errors.

### 3. Narrow exception handling

**Location:** `_func_base.py` lines 100â€“103

The PR only catches `NameError`. Per Python docs, when `eval_str=True`, the `eval()` used to resolve annotations "could potentially raise any kind of exception" (e.g. `TypeError`, `AttributeError`). Catching only `NameError` is reasonable for the common "undefined type" case, but other annotation evaluation failures would still propagate as raw exceptions. This is a minor pointâ€”`NameError` is likely the dominant caseâ€”but worth noting.

## Suggestions for Improvement

1. **Add `eval_str=True` to `_perf_dispatch.py` line 102** for consistency with the rest of the codebase. If the registered function has invalid annotations, wrap in try/except and re-raise as `QuadrantsSyntaxError` for consistent UX.

2. **Consider a small helper** to centralize the `inspect.signature(..., eval_str=True)` + `NameError` handling, e.g.:
   ```python
   def get_signature_with_eval(func):
       try:
           return inspect.signature(func, eval_str=True)
       except NameError as e:
           raise QuadrantsSyntaxError(f"Invalid type annotation: {e}") from e
   ```
   This would reduce duplication and ensure consistent behavior across all call sites.

3. **Add a test** for `from __future__ import annotations` with a kernel that uses type hints (e.g. `NDArray`, dataclass params) to lock in the intended behavior and guard against regressions.

## Style & Correctness

- **Correctness:** The change correctly enables string annotation evaluation. Existing tests (`test_kernel_bad_argument_annotation`, `test_func_bad_argument_annotation`) pass and continue to expect `QuadrantsSyntaxError` for invalid annotations.
- **Style:** Matches existing patterns (e.g. `QuadrantsSyntaxError` usage, `from e` in re-raise).
- **Python version:** `eval_str` is available in Python 3.10+; the project supports 3.10â€“3.13, so compatibility is fine.

## Verdict

The PR achieves its goal and improves support for `from __future__ import annotations`. The main follow-ups are consistency (the `_perf_dispatch` call site) and optional hardening of error handling at the other `eval_str=True` call sites.

PR #412, opened 2026-03-12
[Type] ndarray typing 2: Fix NDArray single-arg subscript crash

hughperkins:
NdarrayType.__class_getitem__ crashed when called with a single arg (e.g. NdarrayType[dtype]) because it tried to unpack a non-tuple. Wrap single args in a tuple before passing to __init__.

Issue: #

## Summary
  Fixes a crash when `qd.types.NDArray` is subscripted with a single argument (e.g. `qd.types.NDArray[qd.i32]`) instead of a tuple. Previously this raised a `TypeError` from the
   `__class_getitem__` unpacking logic, even though single-arg subscript is the natural way to express "ndarray of `i32` with unspecified rank" — and is the form most type
  checkers/linters and PEP 585 generic conventions encourage.
  This PR is stacked on top of `hp/typing-t4-1-eval-str` (the `get_func_signature` / `eval_str=True` work), which is included via merge.
  ## Changes
  - **`python/quadrants/types/ndarray_type.py`** — `NdarrayType.__class_getitem__`:
    - Normalize a single non-tuple arg into a 1-tuple before unpacking, so `NDArray[qd.i32]` is equivalent to `NDArray[(qd.i32,)]` → `NdarrayType(qd.i32)`.
    - Drop the unused `**kwargs` parameter (Python's `__class_getitem__` protocol only ever passes a single positional `args`, so `**kwargs` was dead code that masked the real
  signature).
  - **`tests/python/test_ndarray_typing.py`** — Adds `test_ndarray_typing_single_arg` covering `qd.types.NDArray[qd.i32]` and asserting `dtype == qd.i32` / `ndim is None`.
  - **Merge of `hp/typing-t4-1-eval-str`** — pulls in the `get_func_signature` helper and `eval_str=True` plumbing, with one merge-conflict resolution in `exception.py` (this
  branch's old in-place copy of `get_func_signature` was removed in favor of the new dedicated `_signature.py` module).
  ## Good points
  - **Tiny, surgical fix** for a real crash. Net diff (excluding the merge of t4-1) is ~6 LOC of source + a focused regression test.
  - **Aligns with Python conventions.** Built-in generics (`list[int]`, `dict[str, int]`, etc.) accept single-arg subscript without requiring an explicit tuple; matching that
  behavior removes a footgun for users coming from typed Python.
  - **Drops a dead `**kwargs`** that was misleading anyone reading the signature into thinking keyword args were supported via subscript syntax (they aren't — Python doesn't
  pass any).
  - **Has a test.**
  - **Composes cleanly** with the t4-1 `eval_str=True` work: stringified annotations like `"NDArray[i32]"` will now both evaluate (thanks to t4-1) *and* produce a valid
  `NdarrayType` (thanks to this PR).
  - **Test coverage is thin.** Only the happy path (`dtype` and `ndim` correctness) is asserted. Missing:
    - End-to-end test that a kernel with `a: qd.types.NDArray[qd.i32]` annotation actually compiles and runs (currently only checks the type object, not kernel-launch behavior).
    - `field_dim` / `boundary` keyword behavior is unchanged but still untested in the single-arg path.
    - No test for the stringified-annotation form (`"NDArray[i32]"` under `from __future__ import annotations`), which is the combined t4-1 + t4-2 motivating use case.
  - **Removing `**kwargs` is technically a public-API narrowing.** `__class_getitem__` is a documented dunder; if any downstream code was (incorrectly) relying on the old
  signature accepting kwargs, it would now fail. Risk is essentially zero since Python's subscript syntax has no way to pass kwargs, but worth a one-line note in the PR.
  - **Stacked on `hp/typing-t4-1-eval-str`.** Reviewers must merge / review t4-1 first; the diff against `main` looks larger than the actual logical change in this PR. Consider
  reviewing as a stack or rebasing once t4-1 lands.
  - **Merge commit history is noisy** — 6 logical commits on this branch get inflated to 16 commits-vs-main once t4-1 and main merges are folded in. Could be cleaned up via
  rebase before merging.
  - **Doesn't unify with other generic dtype types.** If `MatrixType`, `StructType`, etc. have similar `__class_getitem__` quirks, they'll still crash on single-arg subscript.
  This PR fixes only `NdarrayType`. Worth either calling out as out-of-scope or filing a follow-up.

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
Probably we need a user-facing doc for this

.github/workflows/pyright_linter.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           pip install -e .
       - name: install torch and pyright
         run: |
-          pip install torch pyright
+          pip install torch 'pyright==1.1.408'
```
hughperkins: why is this changing?

erizmr (approved):

hughperkins:
Thanks!

PR #422, opened 2026-03-22
[Perf] Add first warmup config to perf dispatch

hughperkins:
Issue: #

### Brief Summary

- add first warmup, defaulting to 1
- also changes default warmup from 3 to 0

  ## Summary
  Splits the perf-dispatcher warmup into two distinct phases:
  - **`first_warmup`** — warmup iterations for the *initial* evaluation cycle of a given geometry, when kernels still need JIT compilation. Defaults to `1`.
  - **`warmup`** — warmup iterations for *subsequent* re-evaluation cycles, when kernels are already compiled. Defaults to `0`.
  Previously there was a single `NUM_WARMUP = 3` used for every evaluation cycle. The new defaults significantly reduce the per-call overhead of re-evaluation (warmup drops from
   3 → 0) while still giving the first cycle a JIT-warmup buffer.
  ## Changes
  ### `python/quadrants/lang/_perf_dispatch.py`
  - New module constant `NUM_FIRST_WARMUP = 1`; `NUM_WARMUP` changed from `3` → `0`.
  - New `first_warmup` parameter on the `perf_dispatch` decorator and `PerformanceDispatcher.__init__`.
  - New `_first_eval_completed: set[int]` to track which geometries have finished their initial evaluation.
  - New `_get_effective_warmup(geometry_hash)` — returns `num_first_warmup` before the first evaluation completes, `num_warmup` afterward.
  - `_compute_are_trials_finished` now takes a `num_compatible` argument and requires that *every* compatible implementation has at least one entry in `trial_counts` before
  declaring trials finished. This fixes a bug where, with `warmup=0`, the check could fire after only one impl had run.
  - Updated docstring for `perf_dispatch` to document both parameters.
  ### `tests/python/test_perf_dispatch.py`
  - `NUM_WARMUP` import replaced with `NUM_FIRST_WARMUP`.
  - Existing tests now pin `first_warmup=WARMUP, warmup=WARMUP` (local `WARMUP=3`) so they continue to exercise the same behaviour as before.
  - Two new tests:
    - `test_perf_dispatch_first_warmup_vs_warmup` — verifies that the first evaluation uses `first_warmup` and subsequent re-evaluations use `warmup`.
    - `test_perf_dispatch_default_warmup_values` — verifies the new defaults (`1` / `0`) produce the expected call counts.
  ## Good points
  - **Reduces steady-state overhead.** With the old defaults, every re-evaluation triggered 3 warmup calls per implementation; the new default of `0` removes that cost entirely
  once kernels are compiled.
  - **Keeps JIT warmup where it matters.** The first evaluation cycle still gets a warmup pass, so the initial timing measurement is not polluted by compilation time.
  - **Fixes a real bug.** The `num_compatible` check in `_compute_are_trials_finished` closes a hole where, with `warmup=0`, evaluation could be declared finished before all
  compatible implementations had been measured — meaning the "fastest" choice could be made from incomplete data.
  - **Backwards-compatible API surface.** `first_warmup` is a new keyword-only argument with a sensible default; existing callers don't need to change unless they want to.
  - **Test coverage for the new behaviour.** Both the parameter split and the new defaults are covered explicitly.
  ## Bad / risky points
  - **Default behaviour change is silent.** Existing callers that did not pass `warmup=` were getting 3 warmup iterations per cycle and now get 0. Any caller that was implicitly
   relying on warmup runs to settle caches / hide first-call jitter will now see noisier measurements. There is no deprecation path.
  - **`first_warmup=1` may be too low for some kernels.** A single warmup is enough to pay JIT cost, but not enough to amortise things like first-touch allocation or autotuning
  inside the kernel. Workloads that previously got 3 warmups for the first cycle may now make a fastest-impl decision based on a less-stable measurement.
  - **Semantics of "trials finished" subtly changed even when `warmup > 0`.** The new `len(trial_counts) >= num_compatible` requirement is correct, but it's a behavioural change
   that affects every caller, not just the `warmup=0` case. Worth flagging in case any downstream code depends on the old timing.
  - **Naming.** `first_warmup` vs `warmup` reads a bit oddly — `warmup` no longer means "warmup" in the obvious sense, it means "warmup on re-evaluation". Names like
  `initial_warmup` / `recheck_warmup` (or `warmup` / `recheck_warmup`) might be clearer. Worth a quick bikeshed before merge.
  - **Per-geometry state grows.** `_first_eval_completed` is another `set` keyed by geometry hash, alongside the existing per-geometry dicts. Minor, but worth noting for
  long-running processes that see many distinct geometries.
  - **CI noise in history.** The branch contains two empty "retrigger flaky CI" commits and two merge commits from `main`. Consider squashing on merge to keep `main`'s history
  clean.

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_perf_dispatch.py
```diff
@@ -32,12 +33,15 @@ def do_work_py(i_b, amount_work: qd.i32, state: qd.types.NDArray[qd.i32, 1]):
 
 @test_utils.test()
 def test_perf_dispatch_kernels() -> None:
+    WARMUP = 3
```
hughperkins, line 35: this local var should be in lowercase

tests/python/test_perf_dispatch.py
```diff
@@ -290,9 +300,107 @@ def my_func1_impl_impl2(
...
 
+@test_utils.test()
+def test_perf_dispatch_first_warmup_vs_warmup() -> None:
+    """first_warmup is used for the initial evaluation, warmup for re-evaluations."""
+    called = []
+
+    @qd.perf_dispatch(
+        get_geometry_hash=lambda a: hash(a.shape),
+        first_warmup=2,
+        warmup=0,
```
hughperkins, line 322: lets test a non-zero warmup

tests/python/test_perf_dispatch.py
```diff
@@ -290,9 +300,107 @@ def my_func1_impl_impl2(
...
+    called.clear()
+    for _ in range(2):
+        my_func(a)
+    assert len(called) == 2
+    assert set(called) == {"a", "b"}
+
+
+@test_utils.test()
+def test_perf_dispatch_default_warmup_values() -> None:
+    """With defaults (first_warmup=1, warmup=0), first eval uses 1 warmup, re-evals use 0."""
```
hughperkins, line 371: should have atest with non-zero

hughperkins: also, should check something like active + warmup == num calls to non-chosen kernel

tests/python/test_perf_dispatch.py
```diff
@@ -290,9 +300,107 @@ def my_func1_impl_impl2(
...
+    for _ in range(2):
+        my_func(a)
+    assert len(called) == 2
+    assert set(called) == {"a", "b"}
+
+
 @test_utils.test()
 def test_perf_dispatch_sanity_check_register_args() -> None:
-    @qd.perf_dispatch(get_geometry_hash=lambda a, c: hash(a.shape + c.shape), warmup=25, active=25)
+    @qd.perf_dispatch(get_geometry_hash=lambda a, c: hash(a.shape + c.shape), first_warmup=25, warmup=25, active=25)
```
hughperkins, line 414: what is the purpose of such large, and identcal, warmups

erizmr (approved):

hughperkins:
Thanks!

PR #426, opened 2026-03-26
[Docs] Add a new page to help for first time contributors

v01dXYZ:
Issue: No issue

### Brief Summary

My first contribution was a little bit harder than expected. So I wrote down some pieces of advice for new contributors.

It is a WIP and you're *welcome* to edit it!


docs/source/user_guide/contributing.md
```diff
@@ -7,7 +7,8 @@
 Although using `LLVM` from your Linux distribution is possible, it is better to be sure you have the right version and rather do one of the following:
 
 * use `download_llvm.py` to get the right binary compiled by us
-* build `LLVM` yourself. You have to be cautious about the version (look at README.md or git history, take a look at CI scripts to get it). You also have to enable RTTI when building `LLVM`.
+* build `LLVM` yourself. You have to be cautious about the version (look at README.md or git history, take a look at CI scripts to get it).
```
hughperkins: I dont really understand why you are doing this. You should NOT need to do this.

Please can you look at how our CI runs the build, and identify what is different between your current build process and the CI build process?

I have *never* built LLVM myself when running on Mac, or Ubuntu.

hughperkins:
I don't agree with building LLVM oneself. It is very heavy and difficult. I *never* have to do this.

Please can you look at how our CI runs the build, and identify what is different between your current build process and the CI build process?

docs/source/user_guide/contributing.md
v01dXYZ: It should be a possible path as sometimes it is great to have LLVM with some debug output (not debug symbols, just get the intermediate passes and some details about Instruction Selection...).

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+Although using `LLVM` from your Linux distribution is possible, it is better to be sure you have the right version.
+In order to do so, use `download_llvm.py` to get the right binary compiled by us.
+
+Once it is done, you could use the env variable `LLVM_DIR` to specify where the root of the LLVM binaries are.
+
+## Building the package
+
+There are at least two ways to build the wheel:
+
+* `setup.py bdist_wheel`: The method we currently use
```
hughperkins: I think we use `./build.py wheel`? 

https://github.com/Genesis-Embodied-AI/quadrants/blob/0fc25c9db467734aba1ca864f57547f6869e286c/.github/workflows/scripts_new/manylinux_wheel/2_build.sh#L30

(Thought whether we should be doing this is an open question, but let's document what we are doing (and testing in CI) currently, and think about changes to this in a separate PR please).

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+* Using a package manager such as `pip` (`pip wheel`) or `uv` (`uv build`): Currently not supported, but possible if you disable build isolation.
```
hughperkins: I use `uv` all the time. It is supported. Now to be clear I don't know for sure `uv` works for wheel build, but I use `uv` constantly for dev builds.

Though sure, using `pip wheel` and `uv build` is almost certainly not supported.

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+
+The important environment variables are:
+
+* `LLVM_DIR`: root directory for `LLVM` binaries
+* `QUADRANTS_CMAKE_ARGS`: extra arguments to `cmake`
+
+Concerning the extra cmake arguments, the notable ones are:
+
+* `CMAKE_CXX_COMPILER`: it should be `clang++`
```
hughperkins: Are you sure this is needed? I never set this myself.

I simply do:

```
./build.py --shell
python setup.py develop
```

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+* `QD_WITH_CUDA` / `QD_WITH_AMDGPU`: If you want to enable/disable a target because you don't need it.
```
hughperkins: Pesonally, what I do is:
- start running a build (as above)
- then do:
```
(cd _skbuild/linux-x86_64-3.10/cmake-build/; ccmake $PWD)
```
(have to change the second path component, depending on the platform. I use tab completion to find the right path).

Then I select the options I want in the ncurses ui, then 'c' for configure, 'g' for generate. Then restart `python setup.py develop`.

The advantage is, I don't have to remember these env vars.

The main 'messy' bit with my approach is I have to start the build before the cmake-build directory appears. This has not so far annoyed me enough to figure out a mildly cleaner approach.

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Notes
+
+* *testing*: Any new features or modified code should be tested. You have to run the test suite using `./tests/run_tests.py` which set up
```
hughperkins: +1

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+  the right test environment for `pytest`. CLI arguments are forwarded to `pytest`. Do not use `pytest` directly as it behaves differently.
```
hughperkins: +1

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+* *format/linter*: Before pushing any commits, ensure you set up `pre-commit` and run it using `pre-commit run -a`
```
hughperkins: +1

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+* No need to force push to keep a clean history as the merging is eventually done by squashing commits.
```
hughperkins: +1

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Advanced users
```
hughperkins: Lets call it simply `Advanced` rather than implying a judgment on the person themselves? (or perhaps `Advanced Usage`?)

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+
+### CI Convention about compilers/LLVM
+
+`quadrants` is by itself somewhat a compiler targeting CPUs and GPUs altogether. There are at least three pieces that need to be compiled or are themselves part
+of a compiler project. We can easily mix things up. So let' explicitly write down the CI convention about the compilers/LLVM that are used for the three pieces:
+
+1. `quadrants` host runtime: compiled using the OS default C/C++ compiler.
+2. `quadrants` device runtime (bitcode): compiled using `clang` from the distribution/OS. Using `clang` is required as it has to support the same targets as `LLVM` (obviously!).
```
hughperkins: nit: I think it uses `clang++`?

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+3. `LLVM` used by host runtime: statically or dynamically linked, used to lower the kernel's final IR to machine code on the host. The CI uses an LLVM version compiled from source.
```
hughperkins: "llvm libraries"

hughperkins: (cf llvm does include a whole bunch of executables too, like the linker etc, I think)

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+* *testing*: Any new features or modified code should be tested. You have to run the test suite using `./tests/run_tests.py` which set up
+  the right test environment for `pytest`. CLI arguments are forwarded to `pytest`. Do not use `pytest` directly as it behaves differently.
+* *format/linter*: Before pushing any commits, ensure you set up `pre-commit` and run it using `pre-commit run -a`
+* No need to force push to keep a clean history as the merging is eventually done by squashing commits.
+
+## Advanced users
+
+### CI Convention about compilers/LLVM
+
+`quadrants` is by itself somewhat a compiler targeting CPUs and GPUs altogether. There are at least three pieces that need to be compiled or are themselves part
```
hughperkins: Lets keep just the second sentence, to keep as concise as posible "There are three pieces that need to be compiled or are themselves part
of a compiler project.". Actually we can shorten further: "Quadrants comprises three parts:".

But actually, what about the python code that is part of Quadrants? One can argue it is included in the 'host runtime' part. But on the other hand, below it says that the 'host runtime' is compiled, usng a C/C++ compiler. Thoughts?

hughperkins: (I am wary of saying that Quadrants 'comprises 3 parts' to be honest. There are also .pyi stub files from pybind11, for example. Thoughts?)

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,50 @@
...
+of a compiler project. We can easily mix things up. So let' explicitly write down the CI convention about the compilers/LLVM that are used for the three pieces:
+
+1. `quadrants` host runtime: compiled using the OS default C/C++ compiler.
+2. `quadrants` device runtime (bitcode): compiled using `clang` from the distribution/OS. Using `clang` is required as it has to support the same targets as `LLVM` (obviously!).
+3. `LLVM` used by host runtime: statically or dynamically linked, used to lower the kernel's final IR to machine code on the host. The CI uses an LLVM version compiled from source.
+
+### Building LLVM for debugging it
+
+Sometimes, it could be useful to have a LLVM version that allows to print intermediate passes or with debug symbols to find out where and why LLVM fails (for example, when Instruction Selection fails).
```
hughperkins: 👀

docs/source/user_guide/contributing.md
hughperkins: Note: to be clear, I'm neither saying that you should do this, nor that the doc should provide this information. Simply providing this information to you, in case it's useful to you.

docs/source/user_guide/contributing.md
v01dXYZ: Yes, better document `./build.py wheel` (although it invokes `setup.py bdist_wheel`).

docs/source/user_guide/contributing.md
v01dXYZ: I didn't know `ccmake` but it's neat. It's a very good idea to suggest it, but I remark `LLVM_DIR` is not shown (``ccmake 4.11.2``). There is maybe a way to make `LLVM_DIR` discoverable through `ccmake` in another PR: it gets it from the env instead of as an option.

docs/source/user_guide/contributing.md
v01dXYZ: Sorry to ask, but what is the command you use to get a dev build using `uv` (I use `pip` and `pip install -e .` requires to run before `build_ext --inplace`).

docs/source/user_guide/contributing.md
v01dXYZ: What does `--shell`? Does it activate a venv with all the deps?

docs/source/user_guide/contributing.md
v01dXYZ: It is a little self-satisfactory. `Advanced` alone seems off, so go with `Advanced usage` or `Advanced setup`

docs/source/user_guide/contributing.md
v01dXYZ: yes, better be specific as LLVM is an umbrella name for many subprojects.

docs/source/user_guide/contributing.md
hughperkins: what I do:
```
uv venv --python '3.10'
source .venv/bin/activate
uv pip install pip setuptools [+ bunch of other stuff I forget]
./build.py --shell
python setup.py develop
```
whether you count this as 'using uv' is unclear.

docs/source/user_guide/contributing.md
hughperkins: it sets a bunch of env vars

there's also a version for AI. I don't pay close attention to what AI does, but it seems to be doing somethign like:
```
./build.py -w env.sh
source env.sh
python setup.py develop
```

docs/source/user_guide/contributing.md
hughperkins: ah, interesting. Yes, open to a PR to make it visible in cmake options. Not sure what that would involve.

docs/source/user_guide/contributing.md
hughperkins: Yes. By the way, open to a PR to clean up and standardize our build system a bunch. Not sure if there will be any trade-offs, but probably good to have a clean modern standard setup approach.

docs/source/user_guide/contributing.md
v01dXYZ: it is using `uv` to get both a `venv` and a specific python version. That could be a good addition (in particular b/c `pre-commit` requires `3.10`).

But otherwise that's not using `uv` to build the package. Can take a look in another PR. Right now, I should try to follow your workflow.

docs/source/user_guide/contributing.md
v01dXYZ: So actually it was me. There are two ways to hint `find_package` where to look at `LLVM_DIR` through env var or `-DLLVM_ROOT`. You could also use a variable to set PATHS which we don't do.

ref: https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+## Good practice reminder
+
+* *testing*: Any new features or modified code should be tested. You have to run the test suite using `./tests/run_tests.py` which set up
+  the right test environment for `pytest`. CLI arguments are forwarded to `pytest`. Do not use `pytest` directly as it behaves differently.
+* *format/linter*: Before pushing any commits, ensure you set up `pre-commit` and run it using `pre-commit run -a`
+* No need to force push to keep a clean history as the merging is eventually done by squashing commits.
+
+## Getting the right python version with an isolated dependency tree
+
+It is recommended to use a virtual env and the right python version (currently `3.10`). `pre-commit` for instance is configured with a pinned Python version.
```
hughperkins: I develop using 3.10, but we officially support 3.10.3.13 currently, see https://github.com/Genesis-Embodied-AI/quadrants/blob/412108b90ac7ed8cdf29d2d5a5765f0eb3b6f863/.github/workflows/manylinux_wheel.yml#L27

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+
+`build.py` is a python script to automatically set up the build environment for you before invoking the build commands:
+
+* `LLVM libraries`: downloads an archive for `LLVM` libraires, decompress it and set `LLVM_DIR`.
+* `clang`: depending on the platform, download `clang` or just check if available with right version.
+
+`build.py` can be used at least two ways:
+
+* `build.py wheel` to build the wheel currently using `setup.py bdist_wheel`
+* `build.py --shell` to enter a shell with environment variables set up as with `build.py wheel` in order to let you invoke yourself the commands.
```
hughperkins, line 42: there is also the AI way, where it outputs a script to set the env vars.

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+
+For instance, if you want to tinker the `quadrants` python runtime without rebuilding every time, you
```
hughperkins: "`python setup.py develop` provides incremental builds. Use like:"

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+
+```
+./build.py -w env.sh
+source env.sh
+python setup.py develop
+```
+
+## Building the package for release purposes
+
+As previously mentionned, to build the release package, you can invoke:
```
hughperkins: "mentioned"

hughperkins: Actually, just "To build the release package:"

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+For instance, if you want to tinker the `quadrants` python runtime without rebuilding every time, you
+can use `setup.py develop` by yourself.
+
+```
+./build.py --shell # run a new shell with environment variables
+python setup.py develop
+```
+
+If you are interested about the environment variables set up by `build.py`
+(or you need them for one of your script), you can write them down to a file
```
hughperkins: "write them to a file"

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+
+For instance, if you want to tinker the `quadrants` python runtime without rebuilding every time, you
+can use `setup.py develop` by yourself.
+
+```
+./build.py --shell # run a new shell with environment variables
+python setup.py develop
+```
+
+If you are interested about the environment variables set up by `build.py`
```
hughperkins: "To see the environment variable set by `build.py`"

hughperkins: Actually, let's just shorten this paragraph to:

"To write the environment variables to a file use `python setup.py -w [filename]`. For example:"

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Building the package for release purposes
+
+As previously mentionned, to build the release package, you can invoke:
+
+```
+./build.py wheel
+```
+
+We use `cmake` to build the C++ core. The build directory depends on the host architecture and the python version: it could be for instance `_skbuild/linux-x86_64-3.10/cmake-build`.
```
hughperkins: "version: it could be for instance" => "version. For example:"

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+
+You can modify the cmake options to your liking in order to enable or disable some features you need or don't need. To discover them, use `ccache`:
```
hughperkins: "use `ccache`" => "you can use `ccache` or `ccmake`"

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+
+```
+ccache _skbuild/linux-x86_64-3.10/cmake-build
+```
```
hughperkins, line 73: lets include also the ccmake command ie
```
(cd _skbuild/linux-x86_64-3.10/cmake-build; ccmake $PWD)
```

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+
+You could then set the environment variable `QUADRANTS_CMAKE_ARGS` that will be appended to the `cmake` command used to configure the `cmake` build.
```
hughperkins: do you mean, when we run `./build.py --shell`?

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+export QUADRANTS_CMAKE_ARGS="$QUADRANTS_CMAKE_ARGS -DLLVM_ROOT=/path/to/llvm"
+```
+
+## Advanced usage
+
+### CI Convention about compilers/LLVM
+
+Quadrants comprises at least three important parts:
+
+1. `quadrants` host runtime: Made with a mix of Python and C++. The C++ core is compiled using the OS default C/C++ compiler.
```
hughperkins: ✅

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+## Advanced usage
+
+### CI Convention about compilers/LLVM
+
+Quadrants comprises at least three important parts:
+
+1. `quadrants` host runtime: Made with a mix of Python and C++. The C++ core is compiled using the OS default C/C++ compiler. 
+2. `quadrants` device runtime (bitcode): C++ code compiled using `clang++` from the distribution/OS. Using `clang` is required as it has to support the same targets as `LLVM` (obviously!).
```
hughperkins: lets remove `(obviously!)` (since if it was obvious we wouldn't need to say it, I feel).

Arent we using `clang++` rather than `clang`? (correct me if I'm wrong on this point.

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+3. `LLVM` libraries used by host runtime: statically or dynamically linked, used to lower the kernel's final IR to machine code on the host. The CI uses an LLVM version compiled from source.
```
hughperkins, line 98: They're used by a lot more than final lowering. They're used throughout the entire c++ side, that receives the python side AST, via pybind11 api calls, transforms them to CHIIR, optimizes them, then gradually lowers them to LLVM IR.

docs/source/user_guide/contributing.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Building LLVM for debugging it
+
+Sometimes, it could be useful to have a LLVM version that allows to print intermediate passes or with debug symbols to find out where and why LLVM fails (for example, when Instruction Selection fails).
```
hughperkins: To clarify to my own mind, what is possible here that is not possible using QD_DUMP_IR=1?

docs/source/user_guide/contributing.md
v01dXYZ: Well, I made a mistake. It's `ccmake`.

docs/source/user_guide/contributing.md
v01dXYZ: it is appended by `setup.py` before invoking `cmake`.

To me, it doesn't seem related to `build.py --shell`.

docs/source/user_guide/contributing.md
v01dXYZ: In my mind, `clang` refers to the project while `clang++` refers to the CLI command. That's why I left this `clang`. But let's be consistent.

docs/source/user_guide/contributing.md
v01dXYZ: I'll take a look at that to be more accurate.

docs/source/user_guide/contributing.md
v01dXYZ: No because we will only debug `quadrants` not LLVM. If we want to follow after each pass what does `LLVM` (ie when we dohas `module->lookup("symbol")`), it has to be with a LLVM build that kept `LLVM_DEBUG` statements.

docs/source/user_guide/contributing.md
hughperkins: this seems not to have been addressed yet?

docs/source/user_guide/contributing.md
```diff
@@ -37,17 +37,14 @@ uv pip install --group dev test
...
 
 ```
 ./build.py --shell # run a new shell with environment variables
 python setup.py develop
 ```
 
-If you are interested about the environment variables set up by `build.py`
-(or you need them for one of your scripts), you can write them down to a file
-to source it by yourself:
+To write the environment variables to a file, use `./build.py -w [filename]`. For example:
```
hughperkins, line 51: ✅

docs/source/user_guide/contributing.md
```diff
@@ -57,18 +54,18 @@ python setup.py develop
 
 ## Building the package for release purposes
 
-As previously mentionned, to build the release package, you can invoke:
+To build the release package:
```
hughperkins, line 61: ✅

docs/source/user_guide/contributing.md
```diff
@@ -57,18 +54,18 @@ python setup.py develop
...
 
 ```
 ./build.py wheel
 ```
 
-We use `cmake` to build the C++ core. The build directory depends on the host architecture and the python version: it could be for instance `_skbuild/linux-x86_64-3.10/cmake-build`.
+We use `cmake` to build the C++ core. The build directory depends on the host architecture and the python version. For example: `_skbuild/linux-x86_64-3.10/cmake-build`.
```
hughperkins, line 67: ✅

docs/source/user_guide/contributing.md
```diff
@@ -57,18 +54,18 @@ python setup.py develop
...
 
-You can modify the cmake options to your liking in order to enable or disable some features you need or don't need. To discover them, use `ccache`:
+You can modify the cmake options to your liking in order to enable or disable some features you need or don't need. To discover them, you can use `ccmake`:
 
 ```
-ccache _skbuild/linux-x86_64-3.10/cmake-build
+ccmake _skbuild/linux-x86_64-3.10/cmake-build
```
hughperkins, line 72: Does this work? I think the command is:
```
(cd _skbuild/linux-x86_64-3.10/cmake-build; ccmake $PWD)
```
Note: $PWD here does NOT resolve to _skbuild/linux-x86_64-3.10/cmake-build, it resolves to the pwd before running the command.

docs/source/user_guide/contributing.md
hughperkins: I've never run cmake directly to be honest. I've always done `./build.py --shell`. I have no objection to making the build more and more standard, but I would like some explanation please of:
- why are you not simply using ./build.py --shell and python setup.py develop?
- how does calling cmake direclty relate to the above?

docs/source/user_guide/contributing.md
hughperkins: This hasnt been addressed yet right?

docs/source/user_guide/contributing.md
v01dXYZ: On my side, it does work.

docs/source/user_guide/contributing.md
v01dXYZ: No, I've got to take a look at it before answering something relevant.

docs/source/user_guide/contributing.md
v01dXYZ: 1. I didn't know `build.py` existed and I used at first `pip` as it is the way. But now I use `build.py`.
2. I don't call `cmake` directly as I have then to copy manually the shared object. But using the `cmake` options is required to disable some unwanted backends. It's not necessary though.

docs/source/user_guide/contributing.md
hughperkins: Try both approaches. You'll see the options available are different.

hughperkins: oh wait 🤔

hughperkins: Interesting. Seems they give the same result.

docs/source/user_guide/contributing.md
v01dXYZ: Not yet.

docs/source/user_guide/contributing.md
hughperkins: Note that I'm not sure why the version `ccmake _skbuild/linux-x86_64-3.10/cmake-build` works. I worry that it might look like it's the same, but not give equivalent results somehow. Please could you either:
1. provide to me (in these comments) a clear step-by-step explanation/proof for why they are 100% equivalent, or
2. use the (cd ... ) version for now please

v01dXYZ: Sorry for being a little pr*ck but the reason is https://cmake.org/cmake/help/latest/manual/ccmake.1.html

hughperkins: That's a very long document. Please could you isolate the relevant 2-3 lines or so.

v01dXYZ: The synopsis shows the command could be used with `argv[1] := <path_to_cmake_build>`.

hughperkins: Oh I see. And you are confident this won't break/change anything?

v01dXYZ: no, I don't think it breaks things compared to doing a cd + $PWD.

docs/source/user_guide/contributing.md
v01dXYZ: sorry, but I don't understand what you said. You talk about a Chi-IR which is an intermediate representation. Could you talk a bit more about this, b/c there is no mention of this in the source code?

Once it is lowered into LLVM IR, do you have custom passes or do you rely only on the ones provided by LLVM?

hughperkins: custom passes.

Please consider asking Opus 4.6 to walk you trhough how the passes work, what passes there are, etc.

Personally I like Cursor CLI interface https://cursor.com/docs/cli/overview

Most people prefer Claude Code for reasons unclear to me.

docs/source/user_guide/contributing.md
v01dXYZ: There is an env var that is read by `setup.py` to add `cmake` CLI args.

docs/source/user_guide/contributing.md
v01dXYZ: Sorry to state it as plainly but I don't have the money to use `Opus 4.6`.

I grep-ed the code base to look at **LLVM** pass and I hardly found ones. I'm talking about **LLVM** passes not the ones related to higher IRs.

docs/source/user_guide/contributing.md
v01dXYZ: To definitively answer your answer, `QD_DUMP_IR` allows to see the intermediate results produced by `quadrants` but not `LLVM`. To have `LLVM` outputs intermediate steps, we also have to pass so CLI args but it is not yet supported by `quadrants` right now. I created an issue about that.

docs/source/user_guide/contributing.md
hughperkins: Ah, yes, I believe our custom passes are for our custom IR, right.

docs/source/user_guide/contributing.md
hughperkins: ther'es an env var that is read by build.py too I believe.

docs/source/user_guide/contributing.md
v01dXYZ: yes, indeed, it's in `transforms`.

docs/source/user_guide/contributing.md
v01dXYZ: Yes, it is called `CMakeArgsManager`.

hughperkins:
@claude review

hughperkins:
Thank you! 🙌

PR #432, opened 2026-03-29
[SPIRV] Feature Parity Atomics & Shared Array

duburcqa:
### Brief Summary

1. Fix float atomics on shared memory for Metal and Vulkan:
    - allocate shared float arrays as uint, bitcast at load/store
    - only retype arrays targeted by atomics (pre-scan via scan_shared_atomic_allocs)
2. Add f16 shared memory float atomics support:
    - back f16 arrays with u32 (Metal/Vulkan lack 16-bit atomics), with width conversion at load/store/CAS boundaries
3. Add support of shared memory of arbitrary dtype on Metal and Vulkan:
    - flatten nested tensor types (vec3 etc.), handle OpPtrAccessChain for component access
4. Add official support of multiple shared arrays on Metal:
    - enable existing test on Metal
5. Fix shared memory float atomics on GPUs with native float atomic support:
    - handle dest_is_ptr before at_buffer, disable native float atomics for uint-retyped shared arrays

```
┌─────┬────────────────────────────────────────────────────────────────────┬──────────┐
│  #  │                               Commit                               │  Delta   │
├─────┼────────────────────────────────────────────────────────────────────┼──────────┤
│ 1   │ [Test] Expand unit test coverage.                                  │ +96/-6   │
├─────┼────────────────────────────────────────────────────────────────────┼──────────┤
│ 2   │ [SPIRV] Fix float atomic bugs: f16/f64 min/max, CAS type mismatch. │ +85/-7   │
├─────┼────────────────────────────────────────────────────────────────────┼──────────┤
│ 3   │ [SPIRV] Add Int64Atomics and shared float atomic cap detection.    │ +32/-0   │
├─────┼────────────────────────────────────────────────────────────────────┼──────────┤
│ 4   │ [SPIRV] Add CAS-based float atomic emulation for shared memory.    │ +444/-10 │
├─────┼────────────────────────────────────────────────────────────────────┼──────────┤
│ 5   │ [SPIRV] Add nested tensor type flattening.                         │ +40/-25  │
└─────┴────────────────────────────────────────────────────────────────────┴──────────┘
```


quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -268,8 +268,21 @@ void TaskCodegen::visit(AllocaStmt *alloca) {
...
-        ir_->get_primitive_type(tensor_type->get_element_type());
+    DataType elem_dt = tensor_type->get_element_type();
+    // Flatten nested tensor types (e.g., array of vec3 to flat array of f32)
+    if (auto nested = elem_dt->cast<TensorType>()) {
+      elem_num *= nested->get_num_elements();
+      elem_dt = nested->get_element_type();
+    }
+    spirv::SType elem_type = ir_->get_primitive_type(elem_dt);
+    // Allocate shared float arrays as uint so that CAS-based atomic
+    // emulation uses integer atomics (Metal/MoltenVK lack threadgroup
```
hughperkins: nice

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -268,8 +268,21 @@ void TaskCodegen::visit(AllocaStmt *alloca) {
   auto alloca_type = alloca->ret_type.ptr_removed();
   if (auto tensor_type = alloca_type->cast<TensorType>()) {
     auto elem_num = tensor_type->get_num_elements();
-    spirv::SType elem_type =
-        ir_->get_primitive_type(tensor_type->get_element_type());
+    DataType elem_dt = tensor_type->get_element_type();
```
hughperkins: elem_dt and elem_type are very confusing. Could we either give more intuitive names, or at least add a comment on what is the difference between them?

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -268,8 +268,21 @@ void TaskCodegen::visit(AllocaStmt *alloca) {
...
+      elem_num *= nested->get_num_elements();
+      elem_dt = nested->get_element_type();
+    }
+    spirv::SType elem_type = ir_->get_primitive_type(elem_dt);
+    // Allocate shared float arrays as uint so that CAS-based atomic
+    // emulation uses integer atomics (Metal/MoltenVK lack threadgroup
+    // float atomics).
+    if (alloca->is_shared && is_real(elem_dt)) {
+      elem_type =
+          ir_->get_primitive_type(ir_->get_quadrants_uint_type(elem_dt));
```
hughperkins: it's not clear to me from the name what get_quadrants_uint_type does. specifically around nubmer of bits. Could we add a comment to clarify what is happening in this line, specifically around nubmer of bits?

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -291,10 +304,19 @@ void TaskCodegen::visit(MatrixPtrStmt *stmt) {
   spirv::Value origin_val = ir_->query_value(stmt->origin->raw_name());
   spirv::Value offset_val = ir_->query_value(stmt->offset->raw_name());
   auto dt = stmt->element_type().ptr_removed();
+  // Flatten nested tensor types to scalar (e.g., vec3 to f32)
+  if (auto nested = dt->cast<TensorType>()) {
```
hughperkins: this seems very similar to what happens above. Could this be factorized into a helper function?

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -291,10 +304,19 @@ void TaskCodegen::visit(MatrixPtrStmt *stmt) {
...
+    dt = nested->get_element_type();
+  }
   if (stmt->offset_used_as_index()) {
     if (stmt->origin->is<AllocaStmt>()) {
-      spirv::SType ptr_type = ir_->get_pointer_type(
-          ir_->get_primitive_type(dt), origin_val.stype.storage_class);
+      auto elem_type = ir_->get_primitive_type(dt);
+      if (shared_float_retyped_.count(stmt->origin)) {
+        elem_type = ir_->get_primitive_type(ir_->get_quadrants_uint_type(dt));
```
hughperkins: ditto for questoin about helper function.

quadrants/codegen/spirv/detail/spirv_codegen.h
```diff
@@ -200,6 +200,8 @@ class TaskCodegen : public IRVisitor {
   std::unordered_map<int, GetRootStmt *>
       root_stmts_;  // maps root id to get root stmt
   std::unordered_map<const Stmt *, BufferInfo> ptr_to_buffers_;
+  // Shared float arrays retyped to uint (Metal lacks threadgroup float atomics)
```
hughperkins: can we give a bit more detail about what we are storing here, and why.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -291,10 +304,19 @@ void TaskCodegen::visit(MatrixPtrStmt *stmt) {
...
+  // Flatten nested tensor types to scalar (e.g., vec3 to f32)
+  if (auto nested = dt->cast<TensorType>()) {
+    dt = nested->get_element_type();
+  }
   if (stmt->offset_used_as_index()) {
     if (stmt->origin->is<AllocaStmt>()) {
-      spirv::SType ptr_type = ir_->get_pointer_type(
-          ir_->get_primitive_type(dt), origin_val.stype.storage_class);
+      auto elem_type = ir_->get_primitive_type(dt);
+      if (shared_float_retyped_.count(stmt->origin)) {
```
hughperkins: can we add a comment about what this if statement is checking for intuitively

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -306,6 +328,16 @@ void TaskCodegen::visit(MatrixPtrStmt *stmt) {
       spirv::Value offset_bytes = ir_->mul(dt_bytes, offset_val);
       ptr_val = ir_->add(origin_val, offset_bytes);
       ptr_to_buffers_[stmt] = ptr_to_buffers_[stmt->origin];
+    } else if (origin_val.stype.flag == TypeKind::kPtr) {
```
hughperkins: can we add a comment about what this new else if block is checking for

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1519,26 +1564,26 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
...
+  // Shared arrays already have an access chain pointer; at_buffer() only
+  // handles device buffers and would fail on workgroup pointers.
+  if (dest_is_ptr) {
+    addr_ptr = dest_val;
+  } else if (dt->is_primitive(PrimitiveTypeID::f64)) {
     if (caps_->get(DeviceCapability::spirv_has_atomic_float64_add) &&
         stmt->op_type == AtomicOpType::add) {
       addr_ptr = at_buffer(stmt->dest, dt);
     } else {
-      addr_ptr = dest_is_ptr
```
hughperkins, line 1532: this is just refactorizing right? seems like a nice refactorization, if I've undrestood correctly.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1549,22 +1594,26 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
       atomic_fp_op = spv::OpAtomicFAddEXT;
     }
 
+    // Shared float arrays are retyped to uint, so native float atomics
+    // (which require a float pointer) cannot be used on them.
```
hughperkins: I'm not sure I follow. I thought the purpose of changing the backing type to uint was to enable the spirv atomics? Could you give a little more clarification (in the comments) about this point please.

Also, how do we know if we ar edealing with a shared array here?

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,38 @@ def atomic_test(out: qd.types.ndarray()):
     assert arr[224] == sum
 
 
-@test_utils.test(arch=[qd.cuda])
+@pytest.mark.parametrize("op", ["add", "sub", "min", "max"])
+@test_utils.test(arch=[qd.cuda, qd.vulkan, qd.metal, qd.amdgpu])
```
hughperkins: do we have a simpler way of doing this like conceptually:
- not cpu?, or
- gpu?

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,38 @@ def atomic_test(out: qd.types.ndarray()):
...
+def test_shared_array_float_atomics(op):
+    N = 256
+    block_dim = 32
+    total = block_dim * (block_dim - 1) / 2.0
```
hughperkins: total what? `total_threads`?  Why are we dividing by 2.0? Oh, perhaps we are doing some kind of arithmetic progression or similar, and this is the `expected_sum` of that progression? Could we update the name to make the meaning more intuitive please. By the way, this calculation could be done using ints. Could we make this something that needs actual floats? Like, e.g. multiply each term in the progression by 0.333, which is pretty incompatible with binary representation.

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,38 @@ def atomic_test(out: qd.types.ndarray()):
...
+    def make_kernel(atomic_fn):
+        @qd.kernel
+        def kern(out: qd.types.ndarray()):
+            qd.loop_config(block_dim=block_dim)
+            for i in range(N):
+                tid = i % block_dim
+                sharr = qd.simt.block.SharedArray((block_dim,), qd.f32)
+                sharr[tid] = qd.f32(tid)
+                qd.simt.block.sync()
+                atomic_fn(sharr[0], qd.f32(tid))
```
hughperkins: lets multiply qd.f32(tid) by some fractional float, like 0.3333, or maybe something arbitary like 0.1523f

tests/python/test_shared_array.py
```diff
@@ -172,7 +203,7 @@ def test():
     assert (y.to_numpy()[0] == [4.0, 8.0, 12.0, 16.0]).all()
 
 
-@test_utils.test(arch=[qd.cuda], debug=True)
+@test_utils.test(arch=[qd.cuda, qd.vulkan, qd.metal])
```
hughperkins: why is this excluding amdgpu?

again, can we use something simpler?
- 'not cpu', or 
- 'gpu'
?

quadrants/codegen/spirv/detail/spirv_codegen.h
```diff
@@ -200,6 +200,13 @@ class TaskCodegen : public IRVisitor {
   std::unordered_map<int, GetRootStmt *>
       root_stmts_;  // maps root id to get root stmt
   std::unordered_map<const Stmt *, BufferInfo> ptr_to_buffers_;
+  // Tracks AllocaStmt and derived MatrixPtrStmt nodes whose underlying shared
+  // float arrays were retyped to uint of the same bit-width (f32->u32, etc.).
+  // Metal/MoltenVK lack threadgroup float atomics, so we allocate shared floats
+  // as uint and bitcast at load/store boundaries. The set propagates from
+  // AllocaStmt to derived pointer nodes so that atomic codegen uses CAS
```
hughperkins: Can you explain a little more what this means "The set propagates from AllocaStmt to derived pointer nodes"? Maybe give an example?

quadrants/codegen/spirv/spirv_ir_builder.h
```diff
@@ -341,11 +341,18 @@ class IRBuilder {
...
   size_t get_primitive_type_size(const DataType &dt) const;
-  // Get the spirv uint type with the same size of a given Quadrants data type
-  SType get_primitive_uint_type(const DataType &dt) const;
-  // Get the Quadrants uint type with the same size of a given Quadrants data
-  // type
-  DataType get_quadrants_uint_type(const DataType &dt) const;
+  // Return the SPIR-V uint type with the same bit-width as dt (e.g. f32->u32).
+  SType get_bitcast_uint_stype(const DataType &dt) const;
+  // Return the Quadrants uint DataType with the same bit-width as dt.
+  DataType get_bitcast_uint_dtype(const DataType &dt) const;
```
hughperkins: `_dtype` seems fairly ambiguous to me. maybe `_qd_dtype`?

And what kind of dtype is `dt`?

quadrants/codegen/spirv/spirv_ir_builder.h
```diff
@@ -341,11 +341,18 @@ class IRBuilder {
   SType from_quadrants_type(const DataType &dt, bool has_buffer_ptr);
   // Get the size in bytes of a given Quadrants data type
   size_t get_primitive_type_size(const DataType &dt) const;
-  // Get the spirv uint type with the same size of a given Quadrants data type
-  SType get_primitive_uint_type(const DataType &dt) const;
-  // Get the Quadrants uint type with the same size of a given Quadrants data
-  // type
-  DataType get_quadrants_uint_type(const DataType &dt) const;
+  // Return the SPIR-V uint type with the same bit-width as dt (e.g. f32->u32).
+  SType get_bitcast_uint_stype(const DataType &dt) const;
```
hughperkins: `stype` does at least give some clue, but I would prefer the more explciit `_spirv_dtype` I feel. (or `_spirv_dt` is ok for me too, or potentially `_spv_dt`, if you want it really short.)

quadrants/codegen/spirv/spirv_ir_builder.h
```diff
@@ -341,11 +341,18 @@ class IRBuilder {
...
+  // Like get_bitcast_uint_dtype but returns at least u32 so that the result
+  // is usable with atomic ops (Metal/Vulkan lack 16-bit atomics).
+  DataType get_atomic_uint_dtype(const DataType &dt) const;
+  // Convert a value from shared-memory uint backing to float dt. Handles
+  // the width mismatch when the backing type is wider (e.g. u32 for f16).
+  Value shared_uint_to_float(Value val, const DataType &dt);
+  // Convert a value from float dt to shared-memory uint backing.
+  Value float_to_shared_uint(Value val, const DataType &dt);
   // Get the pointer type that points to value_type
   SType get_storage_pointer_type(const SType &value_type);
```
hughperkins, line 350: How does the `SType` here relaet to the `stype` name above?

tests/python/test_shared_array.py
```diff
@@ -59,7 +59,7 @@ def calc_shared_array(
     assert np.allclose(reference, a_arr)
 
 
-@test_utils.test(arch=[qd.cuda, qd.vulkan, qd.amdgpu])
+@test_utils.test(arch=[qd.cuda, qd.vulkan, qd.metal, qd.amdgpu])
```
hughperkins: my question about can we simply exclude cpu, or say to run only on gpu seems not to have been addressed? (or I didnt see the response perhaps)

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,121 @@ def atomic_test(out: qd.types.ndarray()):
     assert arr[224] == sum
 
 
-@test_utils.test(arch=[qd.cuda])
+@pytest.mark.parametrize("op", ["add", "sub", "min", "max"])
+@pytest.mark.parametrize("dtype", [qd.f16, qd.f32])
```
hughperkins: f64? (seems more common than f16 tbh)

hughperkins: ( i mean, we can test both)

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,121 @@ def atomic_test(out: qd.types.ndarray()):
...
+@test_utils.test(arch=qd.gpu)
+def test_shared_array_float_atomics(op, dtype):
+    N = 256
+    block_dim = 32
+    SCALE = 0.1523  # fractional so values are truly non-integer floats
+    # Arithmetic sum: SCALE * (0 + 1 + ... + block_dim-1)
```
hughperkins, line 346: ✅

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,121 @@ def atomic_test(out: qd.types.ndarray()):
...
 
 
-@test_utils.test(arch=[qd.cuda])
+@pytest.mark.parametrize("op", ["add", "sub", "min", "max"])
+@pytest.mark.parametrize("dtype", [qd.f16, qd.f32])
+@test_utils.test(arch=qd.gpu)
+def test_shared_array_float_atomics(op, dtype):
+    N = 256
+    block_dim = 32
+    SCALE = 0.1523  # fractional so values are truly non-integer floats
```
hughperkins, line 345: ✅

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,121 @@ def atomic_test(out: qd.types.ndarray()):
...
+    expected = {
+        "add": expected_sum,
+        "sub": -expected_sum,
+        "min": 0.0,
+        "max": (block_dim - 1) * SCALE,
+    }
+    rtol = 1e-3 if dtype == qd.f16 else 1e-6
+    arr = qd.ndarray(qd.f32, (N))
+    make_kernel(atomic_op)(arr)
+    qd.sync()
```
hughperkins: why do we need a sync? Is this the existing metal bug you've mentioned recenlty?

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,121 @@ def atomic_test(out: qd.types.ndarray()):
...
+    assert arr[0] == test_utils.approx(expected[op], rel=rtol)
+    assert arr[32] == test_utils.approx(expected[op], rel=rtol)
```
hughperkins: can we check also 31 and 255?

tests/python/test_shared_array.py
hughperkins: oh i guess f64 doesnt work on metal right?

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,121 @@ def atomic_test(out: qd.types.ndarray()):
...
+
+
+@pytest.mark.parametrize("dtype", [qd.i8, qd.i16, qd.i32, qd.u8, qd.u16, qd.u32])
```
hughperkins: ✅

tests/python/test_shared_array.py
```diff
@@ -147,7 +147,121 @@ def atomic_test(out: qd.types.ndarray()):
...
+@test_utils.test(arch=qd.gpu)
```
hughperkins, line 383: ✅

quadrants/codegen/spirv/spirv_ir_builder.h
duburcqa: `stype` is already at many places before this PR. I think it is better to be consistent with the existing naming conventions.

quadrants/codegen/spirv/spirv_ir_builder.h
duburcqa: These naming conventions are already used at many places.

quadrants/codegen/spirv/spirv_ir_builder.h
duburcqa: `SType` is the SPIR-V type struct (defined in `spirv_ir_builder.h:48`). The _stype suffix in `get_bitcast_uint_stype` indicates it returns an `SType`, as opposed to _dtype which returns a Quadrants `DataType`. `get_storage_pointer_type` is pre-existing code, not part of this PR.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: It should be better now.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: I should be better now.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/detail/spirv_codegen.h
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: Added some comment on the PR itself to clarify this.

tests/python/test_shared_array.py
duburcqa: done.

tests/python/test_shared_array.py
duburcqa: fixed.

tests/python/test_shared_array.py
duburcqa: done.

tests/python/test_shared_array.py
duburcqa: done.

quadrants/codegen/spirv/detail/spirv_codegen.h
duburcqa: done.

tests/python/test_shared_array.py
duburcqa: We can! Done.

tests/python/test_shared_array.py
duburcqa: It does not, but it is easy to skip if unsupported in a clean way.

tests/python/test_shared_array.py
duburcqa: We do not actually. I was being extra cautious for no reason.

tests/python/test_shared_array.py
duburcqa: done.

quadrants/codegen/spirv/detail/spirv_codegen.h
```diff
@@ -200,6 +200,19 @@ class TaskCodegen : public IRVisitor {
...
       root_stmts_;  // maps root id to get root stmt
   std::unordered_map<const Stmt *, BufferInfo> ptr_to_buffers_;
+  // Shared float AllocaStmts that need uint retyping because they are targeted
+  // by atomic operations. Populated by scan_shared_atomic_allocs() before the
+  // main codegen pass. Only these arrays pay the uint-backing + bitcast cost.
+  std::unordered_set<const Stmt *> shared_atomic_allocs_;
+  // Propagated from shared_atomic_allocs_ to derived MatrixPtrStmt nodes
+  // during codegen, so that load/store/atomic visitors know to bitcast.
+  // Example: if `sharr` (AllocaStmt) is in shared_atomic_allocs_, then
+  // `sharr[0]` (MatrixPtrStmt) is added here during visit(MatrixPtrStmt).
```
hughperkins, line 211: Nice explanation thanks. It would reduce my cognitive load if you could give me a case study showing some code (perhaps LLVM IR, or python, no strong preference), and how that maps to these sets, and to the resulting spir-v instructions. Doesn't have to be as comments in the code; could be in the PR description, or in some slides or similar.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -263,14 +265,65 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
   ir_->register_value(const_stmt->raw_name(), val);
 }
 
+const AllocaStmt *TaskCodegen::trace_to_alloca(const Stmt *s) {
```
hughperkins: could we add a comment to this function saying what it does, and providing an example.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -263,14 +265,65 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
...
+  if (auto *alloca = s->cast<AllocaStmt>())
+    return alloca;
+  if (auto *mptr = s->cast<MatrixPtrStmt>())
+    return trace_to_alloca(mptr->origin);
+  return nullptr;
+}
+
+void TaskCodegen::scan_shared_atomic_allocs(Block *block) {
```
hughperkins: So this scans a single block, but writes the results to a class level collection? I wonder if it would be more intuitive/re-usable/testable if we pased that collection in as a function parameter somehow?

hughperkins: Are there tests for this function? Could we add some?

How do we know that this function is complete? It seems fairly complex.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: > So this scans a single block, but writes the results to a class level collection? I wonder if it would be more intuitive/re-usable/testable if we pased that collection in as a function parameter somehow?

Done.

> Are there tests for this function? Could we add some?

A hole would cause a shader compilation error, not a silent bug. If the pre-scan misses an atomic target, the array stays float-typed, but the CAS emulation expects a uint pointer. The resulting type mismatch (OpAtomicLoad(u32, ptr_to_f32)) is invalid SPIR-V and gets rejected at compile time. So I don't think more testing is necessary at this point.

quadrants/codegen/spirv/detail/spirv_codegen.h
duburcqa: Python kernel:
```python
  sharr = qd.simt.block.SharedArray((32,), qd.f32)
  sharr[tid] = qd.f32(tid)
  qd.simt.block.sync()
  qd.atomic_add(sharr[0], qd.f32(tid))
  qd.simt.block.sync()
  out[i] = sharr[0]
```

  IR statements and set membership:
```
  %sharr = AllocaStmt(shared, array<f32, 32>)     # in shared_atomic_allocs_ (pre-scan found atomic_add targets it)
                                                    # in shared_float_retyped_ (visit(AllocaStmt) retypes to u32)
  %ptr0  = MatrixPtrStmt(%sharr, tid)              # in shared_float_retyped_ (propagated from %sharr)
  %ptr1  = MatrixPtrStmt(%sharr, 0)                # in shared_float_retyped_ (propagated from %sharr)
           LocalStoreStmt(%ptr0, f32(tid))          # sees %ptr0 in shared_float_retyped_ -> float_to_shared_uint
           AtomicOpStmt(add, %ptr1, f32(tid))       # dest_is_ptr=true -> CAS with u32 atomics
  %val   = LocalLoadStmt(%ptr1)                    # sees %ptr1 in shared_float_retyped_ -> shared_uint_to_float

  Generated SPIR-V (simplified):
  ; Allocation: u32 array instead of f32 (retyped)
  %sharr = OpVariable Workgroup array<u32, 32>

  ; Store: f32 -> bitcast to u32 -> store
  %u_tid = OpBitcast u32 %f_tid
           OpStore %sharr[tid] %u_tid

  ; Atomic add (CAS loop):
  %old   = OpAtomicLoad u32 %sharr[0]
  %old_f = OpBitcast f32 %old              ; u32 -> f32
  %new_f = OpFAdd f32 %old_f %f_tid        ; float add
  %new   = OpBitcast u32 %new_f            ; f32 -> u32
  %loaded = OpAtomicCompareExchange u32 %sharr[0] %new %old
           ; (loop until %loaded == %old)

  ; Load: load u32 -> bitcast to f32
  %raw   = OpLoad u32 %sharr[0]
  %val   = OpBitcast f32 %raw

  For f16, the only difference is the array is still u32-backed (not u16, since Metal/Vulkan lack 16-bit atomics), with OpUConvert inserted between the bitcast and the atomic:
  ; Store f16: bitcast f16->u16, widen u16->u32, store
  ; Load u32: narrow u32->u16, bitcast u16->f16
  ; CAS: OpAtomicLoad u32, narrow->bitcast->FAdd->bitcast->widen, OpAtomicCompareExchange u32
```

quadrants/codegen/spirv/detail/spirv_codegen.h
```diff
@@ -200,6 +200,21 @@ class TaskCodegen : public IRVisitor {
   std::unordered_map<int, GetRootStmt *>
       root_stmts_;  // maps root id to get root stmt
   std::unordered_map<const Stmt *, BufferInfo> ptr_to_buffers_;
+  // Shared float AllocaStmts that need uint retyping because they are targeted
+  // by atomic operations. Populated by scan_shared_atomic_allocs() before the
+  // main codegen pass. Only these arrays pay the uint-backing + bitcast cost.
+  std::unordered_set<const Stmt *> shared_atomic_allocs_;
```
hughperkins: this name is too general. Since we will hardly ever look at this code, let's make the names very clear.

How about one of `shared_allocas_having_atomic_ops_`, `shared_float_allocas_atomics_touch_`, or `shared_float_allocas_with_atomic_rmw_`.

quadrants/codegen/spirv/detail/spirv_codegen.h
```diff
@@ -200,6 +200,21 @@ class TaskCodegen : public IRVisitor {
...
+  // Propagated from shared_atomic_allocs_ to derived MatrixPtrStmt nodes
+  // during codegen, so that load/store/atomic visitors know to bitcast.
+  // Example: if `sharr` (AllocaStmt) is in shared_atomic_allocs_, then
+  // `sharr[0]` (MatrixPtrStmt) is added here during visit(MatrixPtrStmt).
+  std::unordered_set<const Stmt *> shared_float_retyped_;
```
hughperkins: `uint_backed_shared_float_ptr_stmts_`

quadrants/codegen/spirv/detail/spirv_codegen.h
hughperkins: Nice. Thanks!

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -263,14 +265,76 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
   ir_->register_value(const_stmt->raw_name(), val);
 }
 
+// Follow MatrixPtrStmt::origin chains back to the source AllocaStmt.
+// E.g. for `atomic_add(sharr[0], val)`, the dest is a MatrixPtrStmt for
+// sharr[0] whose origin is the AllocaStmt for sharr.
+// Returns nullptr if the chain does not lead to an AllocaStmt.
```
hughperkins: // assumes we only have MatrixPtrStmt's on the chain to get to the AllocStmt.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -263,14 +265,76 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
...
+// operations into `out`. Only IfStmt, RangeForStmt, and WhileStmt carry
+// sub-blocks at this stage (StructForStmt/MeshForStmt are lowered earlier).
+void TaskCodegen::scan_shared_atomic_allocs(
+    Block *ir_block,
+    std::unordered_set<const Stmt *> &atomic_allocs) {
+  for (auto &s : ir_block->statements) {
+    if (auto *atomic_stmt = s->cast<AtomicOpStmt>()) {
+      if (auto *alloca = trace_to_alloca(atomic_stmt->dest)) {
+        if (alloca->is_shared) {
+          auto alloca_dtype = alloca->ret_type.ptr_removed();
```
hughperkins: Add comment:

// type of the alloca is a pointer to the type being stored
// alloca->ret_type is the type of what is being stored
// eg alloca type is a pointer to an array of 128 floats
// alloca->ret_type is array of 128 floats

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -263,14 +265,76 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
...
+          if (auto *tensor_type = alloca_dtype->cast<TensorType>()) {
```
hughperkins: add comment:
// shared array is *always* modeled as a tensor type (i.e. an array of scalars)

hughperkins: Interestingly, ALL classes here (Block, AllocStmt etc...) are all Quadrants types. None of them are LLVM types. This module only contains Quadrants types, and SPIRV types, no LLVM.

I can't think of a way to add this to a comment anywhere, but it surprised me when I found out.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -263,14 +265,76 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
...
+            auto scalar_dtype = tensor_type->get_element_type();
+            if (auto *nested = scalar_dtype->cast<TensorType>())
```
hughperkins: this only handles one layer of nesting. either handle recursing or throw an exception if one peeling is not enough.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -263,14 +265,76 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
...
+        scan_shared_atomic_allocs(if_stmt->true_statements.get(),
+                                  atomic_allocs);
+      if (if_stmt->false_statements)
+        scan_shared_atomic_allocs(if_stmt->false_statements.get(),
+                                  atomic_allocs);
+    } else if (auto *range_for = s->cast<RangeForStmt>()) {
+      scan_shared_atomic_allocs(range_for->body.get(), atomic_allocs);
+    } else if (auto *while_stmt = s->cast<WhileStmt>()) {
+      scan_shared_atomic_allocs(while_stmt->body.get(), atomic_allocs);
+    }
```
hughperkins: assert that we don't have a StructForStmt or a MeshForStmt, which should never appear here.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -263,14 +265,76 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
...
+
 void TaskCodegen::visit(AllocaStmt *alloca) {
   spirv::Value ptr_val;
   auto alloca_type = alloca->ret_type.ptr_removed();
   if (auto tensor_type = alloca_type->cast<TensorType>()) {
     auto elem_num = tensor_type->get_num_elements();
-    spirv::SType elem_type =
-        ir_->get_primitive_type(tensor_type->get_element_type());
-    spirv::SType arr_type = ir_->get_array_type(elem_type, elem_num);
+    DataType scalar_dtype = tensor_type->get_element_type();
```
hughperkins: I'd prefer to keep this function as unchanged as possible, and move all the shared array retyping stuff to a separate function.

See what I did for cuda graph for an example:

- this bit of code stays relatively compact https://github.com/Genesis-Embodied-AI/quadrants/blob/5c6fe38df5ea9cda20fda6281854b5d896fb7038/quadrants/runtime/amdgpu/kernel_launcher.cpp#L150-L157
- then all the nasty cuda graph stuff goes into some seprate function we mostly can avoid looking at https://github.com/Genesis-Embodied-AI/quadrants/blob/5c6fe38df5ea9cda20fda6281854b5d896fb7038/quadrants/runtime/amdgpu/kernel_launcher.cpp#L22-L37

hughperkins: Related to this, fairly strong preference to move all this retyping stuff to a physically separate file.

hughperkins: (I'll review this bit of code more in detail once the atomic float bits of it have been factorized into a seprate function (and ideally a separate file)

quadrants/codegen/spirv/spirv_codegen.cpp
hughperkins: same here as for the AllocaStmt: lets have a minimal footprint in this function, that peels off thinknig about uint-backed float shared arrays into a separate function in aseparate file pelase.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -319,14 +409,25 @@ void TaskCodegen::visit(MatrixPtrStmt *stmt) {
 void TaskCodegen::visit(LocalLoadStmt *stmt) {
   auto ptr = stmt->src;
   spirv::Value ptr_val = ir_->query_value(ptr->raw_name());
-  spirv::Value val = ir_->load_variable(
-      ptr_val, ir_->get_primitive_type(stmt->element_type()));
+  auto expected_type = ir_->get_primitive_type(stmt->element_type());
```
hughperkins: peel off to separte function in separate uint-backed-shared-float-array file

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -319,14 +409,25 @@ void TaskCodegen::visit(MatrixPtrStmt *stmt) {
...
+  } else {
+    val = ir_->load_variable(ptr_val, expected_type);
+  }
   ir_->register_value(stmt->raw_name(), val);
 }
 
 void TaskCodegen::visit(LocalStoreStmt *stmt) {
   spirv::Value ptr_val = ir_->query_value(stmt->dest->raw_name());
   spirv::Value val = ir_->query_value(stmt->val->raw_name());
+  if (shared_float_retyped_.count(stmt->dest)) {
```
hughperkins: albeit this is very small, but let' peel off to separte function in separate uint-backed-shared-float-array file, for consistency, and to keep the uint-backed-sharedf-loat-array things in one place.

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -378,7 +378,7 @@ size_t IRBuilder::get_primitive_type_size(const DataType &dt) const {
   }
 }
 
-SType IRBuilder::get_primitive_uint_type(const DataType &dt) const {
+SType IRBuilder::get_bitcast_uint_stype(const DataType &dt) const {
```
hughperkins: why are we renaming? Is this something i requested, not realizing the impact of my request?

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1516,29 +1617,29 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
 
   spirv::Value addr_ptr;
   spirv::Value dest_val = ir_->query_value(stmt->dest->raw_name());
```
hughperkins, line 1551: note to self: we are inside `TaskCodegen::visit(AtomicOpStmt *stmt) {` at tihs point

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -412,6 +412,33 @@ DataType IRBuilder::get_quadrants_uint_type(const DataType &dt) const {
   }
 }
 
+DataType IRBuilder::get_atomic_uint_dtype(const DataType &dt) const {
```
hughperkins: move to our special new fiel for uint float atomic thigns

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -412,6 +412,33 @@ DataType IRBuilder::get_quadrants_uint_type(const DataType &dt) const {
...
+  DataType uint_dt = get_bitcast_uint_dtype(dt);
+  if (uint_dt == PrimitiveType::u16 || uint_dt == PrimitiveType::u8) {
+    return PrimitiveType::u32;
+  }
+  return uint_dt;
+}
+
+Value IRBuilder::shared_uint_to_float(Value val, const DataType &dt) {
```
hughperkins: move to our special new fiel for uint float atomic thigns

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -412,6 +412,33 @@ DataType IRBuilder::get_quadrants_uint_type(const DataType &dt) const {
...
+  SType narrow_uint = get_bitcast_uint_stype(dt);
+  SType atomic_uint = get_primitive_type(get_atomic_uint_dtype(dt));
+  if (atomic_uint.id != narrow_uint.id) {
+    val = make_value(spv::OpUConvert, narrow_uint, val);
+  }
+  return make_value(spv::OpBitcast, get_primitive_type(dt), val);
+}
+
+Value IRBuilder::float_to_shared_uint(Value val, const DataType &dt) {
```
hughperkins: move to our special new fiel for uint float atomic thigns

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -1074,33 +1101,36 @@ bool IRBuilder::check_value_existence(const std::string &name) const {
 Value IRBuilder::float_atomic(AtomicOpType op_type,
                               Value addr_ptr,
                               Value data,
-                              const DataType &dt) {
+                              const DataType &dt,
+                              const DataType &atomic_uint_dt) {
```
hughperkins: hmmm, is there some way to avoid polluting the normal function with this? I'm not saying there is, but would be preferable, I feel.

quadrants/codegen/spirv/spirv_codegen.cpp
hughperkins: note to self:

`at_buffer` = “give me the SPIR‑V address inside global/storage memory for this IR access.”

hughperkins: note to self:

workgroup/shared pointers skip at_buffer: they’re already `OpAccessChain` pointers into Workgroup storage; at_buffer is only for buffer-backed (ptr_to_buffers_) globals, not
   threadgroup arrays.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1516,29 +1617,29 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
...
-  // Shared arrays have already created an accesschain, use it directly.
+  // dest_is_ptr is true for shared (workgroup) arrays, which already have a
+  // typed pointer from OpAccessChain. Device buffers go through at_buffer(),
+  // which would fail on workgroup pointers.
   const bool dest_is_ptr = dest_val.stype.flag == TypeKind::kPtr;
```
hughperkins, line 1556: lets rename to `use_dest_pointer_directly`. `dest_is_ptr` is confusing I feel.

quadrants/codegen/spirv/spirv_codegen.cpp
hughperkins: why are we refactorziing by the way? It's just extra codde for me to read. Is this reafctoziing necessary?

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -1074,33 +1101,36 @@ bool IRBuilder::check_value_existence(const std::string &name) const {
...
-        dt);
+        addr_ptr, data, [&](Value lhs, Value rhs) { return mul(lhs, rhs); }, dt,
+        atomic_uint_dt);
   } else if (op_type == AtomicOpType::min) {
+    auto float_type = get_primitive_type(dt);
     return atomic_operation(
         addr_ptr, data,
         [&](Value lhs, Value rhs) {
-          return call_glsl450(t_fp32_, /*FMin*/ 37, lhs, rhs);
+          return call_glsl450(float_type, /*FMin*/ 37, lhs, rhs);
```
hughperkins, line 1103: why renaming?

quadrants/codegen/spirv/spirv_codegen.cpp
hughperkins: scratch that. renaming making the pr longer.

quadrants/codegen/spirv/spirv_ir_builder.cpp
hughperkins: lets not rename stuff. makes the pr longer

quadrants/codegen/spirv/spirv_ir_builder.h
hughperkins: fine. lets not rename stuff. makes the pr longer.

quadrants/codegen/spirv/spirv_ir_builder.h
hughperkins: fine. lets not rename stuff. makes the pr longer.

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -1122,9 +1152,11 @@ Value IRBuilder::integer_atomic(AtomicOpType op_type,
 Value IRBuilder::atomic_operation(Value addr_ptr,
                                   Value data,
                                   std::function<Value(Value, Value)> op,
-                                  const DataType &dt) {
-  SType out_type = get_primitive_type(dt);
-  SType res_type = get_primitive_uint_type(dt);
```
hughperkins, line 1127: put renaming factorizations in a separate PR, or keep the existing names please. This pr is already very long I feel :)

hughperkins: I'll review this function once names have been reverted.

hughperkins:
I've at least done a first pass on all but the tests (which took me ~1.5 hour in this sitting, not including earlier review time). I'll take a look at the tests another day.

quadrants/codegen/spirv/spirv_ir_builder.cpp
hughperkins: let's not rename please. (or move renames to a separte renaming pr)

duburcqa: You specifically requested it.

quadrants/codegen/spirv/spirv_ir_builder.cpp
duburcqa: It is not renaming it is fixing a bug.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: updated.

tests/python/test_shared_array.py
duburcqa: done.

tests/python/test_shared_array.py
duburcqa: done.

tests/python/test_shared_array.py
duburcqa: done.

quadrants/codegen/spirv/detail/spirv_codegen.h
duburcqa: done.

quadrants/codegen/spirv/detail/spirv_codegen.h
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: added.

quadrants/codegen/spirv/spirv_ir_builder.cpp
duburcqa: Reverted.

quadrants/codegen/spirv/spirv_ir_builder.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_ir_builder.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_ir_builder.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_ir_builder.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_ir_builder.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: add comment in the code.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: added.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/detail/spirv_codegen.h
```diff
@@ -200,6 +200,16 @@ class TaskCodegen : public IRVisitor {
   std::unordered_map<int, GetRootStmt *>
       root_stmts_;  // maps root id to get root stmt
   std::unordered_map<const Stmt *, BufferInfo> ptr_to_buffers_;
+  // Shared float AllocaStmts targeted by atomics, populated by
+  // scan_shared_atomic_allocs() before codegen. Value = true means the alloca
+  // has non-add ops (CAS unconditionally needed); false = add-only (native
+  // shared float atomics can be used if the device supports them).
+  std::unordered_map<const Stmt *, bool> shared_float_allocas_with_atomic_rmw_;
```
hughperkins, line 207: ✅

quadrants/codegen/spirv/detail/spirv_codegen.h
```diff
@@ -200,6 +200,16 @@ class TaskCodegen : public IRVisitor {
...
+  // Propagated from shared_float_allocas_with_atomic_rmw_ to derived
+  // MatrixPtrStmt nodes during codegen, so that load/store/atomic visitors
+  // know to bitcast. E.g. if `sharr` (AllocaStmt) is retyped, then
+  // `sharr[0]` (MatrixPtrStmt) is added here during visit(MatrixPtrStmt).
+  std::unordered_set<const Stmt *> uint_backed_shared_float_ptr_stmts_;
```
hughperkins, line 212: ✅

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -270,12 +274,25 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
 
 void TaskCodegen::visit(AllocaStmt *alloca) {
   spirv::Value ptr_val;
+  // alloca->ret_type is a pointer to the stored type; ptr_removed() gives the
```
hughperkins, line 277: ✅

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -270,12 +274,25 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
...
+  // stored type itself (e.g. TensorType<32 x f32> for a 32-element array).
   auto alloca_type = alloca->ret_type.ptr_removed();
+  // Shared array is always modeled as a tensor type, i.e. an array of scalars.
```
hughperkins, line 280: ✅

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -270,12 +274,25 @@ void TaskCodegen::visit(ConstStmt *const_stmt) {
...
   if (auto tensor_type = alloca_type->cast<TensorType>()) {
     auto elem_num = tensor_type->get_num_elements();
-    spirv::SType elem_type =
-        ir_->get_primitive_type(tensor_type->get_element_type());
-    spirv::SType arr_type = ir_->get_array_type(elem_type, elem_num);
+    DataType scalar_dtype = tensor_type->get_element_type();
+    // Flatten nested tensor types (e.g., array of vec3 to flat array of f32)
+    if (auto nested = scalar_dtype->cast<TensorType>()) {
```
hughperkins: this is new code. only needed for the uint backed atomics right? Can we move this into the uint backed atomics module please.

quadrants/codegen/spirv/spirv_codegen.cpp
hughperkins: lots more new code. lets mov to the uint backed shared array file please.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -296,10 +313,17 @@ void TaskCodegen::visit(MatrixPtrStmt *stmt) {
...
+  // Flatten nested tensor types to scalar, e.g. vec3 to f32
+  if (auto nested = dt->cast<TensorType>()) {
+    dt = nested->get_element_type();
+  }
   if (stmt->offset_used_as_index()) {
+    // Origin is a local or shared array allocation - use OpAccessChain
     if (stmt->origin->is<AllocaStmt>()) {
-      spirv::SType ptr_type = ir_->get_pointer_type(
-          ir_->get_primitive_type(dt), origin_val.stype.storage_class);
+      auto scalar_stype = maybe_retype_derived_ptr(
```
hughperkins: lots more new code. lets mov to the uint backed shared array file please.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -324,14 +357,16 @@ void TaskCodegen::visit(MatrixPtrStmt *stmt) {
 void TaskCodegen::visit(LocalLoadStmt *stmt) {
   auto ptr = stmt->src;
   spirv::Value ptr_val = ir_->query_value(ptr->raw_name());
-  spirv::Value val = ir_->load_variable(
-      ptr_val, ir_->get_primitive_type(stmt->element_type()));
+  spirv::Value val = load_shared_float(*ir_, ptr_val, ptr, stmt->element_type(),
```
hughperkins: I would expect tis to fork somehow. Like we should have:
1. path for the original code, where we dont have to think about uint backed stuff, plus
2. pah for new code, which should be abstracted tou to the new file, and we dont have to tink about when there arent any uint backed floats.

hughperkins: (like, my concern is: why are we deleting `  spirv::Value val = ir_->load_variable(
      ptr_val, ir_->get_primitive_type(stmt->element_type()));` ?

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -324,14 +357,16 @@ void TaskCodegen::visit(MatrixPtrStmt *stmt) {
...
+                                       uint_backed_shared_float_ptr_stmts_);
   ir_->register_value(stmt->raw_name(), val);
 }
 
 void TaskCodegen::visit(LocalStoreStmt *stmt) {
   spirv::Value ptr_val = ir_->query_value(stmt->dest->raw_name());
   spirv::Value val = ir_->query_value(stmt->val->raw_name());
+  val = store_shared_float(*ir_, val, stmt->dest, stmt->val->element_type(),
```
hughperkins: either make this a conditional fork please, or make ti very clear from the name that this only applies to uint backed floats on sharedd arrays. maybe something like:

if(uint_backed_val = store_uint_backed_shared_float(...)) {
    val = uint_backed_val;
}

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1524,10 +1559,12 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
   // Shared arrays have already created an accesschain, use it directly.
   const bool dest_is_ptr = dest_val.stype.flag == TypeKind::kPtr;
 
+  // The dest_is_ptr ternaries guard against calling at_buffer() on workgroup
+  // pointers, which would fail (at_buffer only handles StorageBuffer).
```
hughperkins: I feel it would be good to say what at_buffer does:

"at_buffer gets the address in device memory, which only applies to StorageBuffer. For shared memory, dest_val is already a pointer to the shared memory." Actually, I don't remember exactly why at_buffer doesn't apply to shared memory. Can you check with Opus, and summarize into a sentence or two, to save my having to chekc with Opus instead please.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1524,10 +1559,12 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
...
   if (dt->is_primitive(PrimitiveTypeID::f64)) {
     if (caps_->get(DeviceCapability::spirv_has_atomic_float64_add) &&
         stmt->op_type == AtomicOpType::add) {
-      addr_ptr = at_buffer(stmt->dest, dt);
+      addr_ptr = dest_is_ptr ? dest_val : at_buffer(stmt->dest, dt);
```
hughperkins, line 1562: why does the uint backed shared mean we have to change this line?

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1536,7 +1573,7 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
   } else if (dt->is_primitive(PrimitiveTypeID::f32)) {
     if (caps_->get(DeviceCapability::spirv_has_atomic_float_add) &&
         stmt->op_type == AtomicOpType::add) {
-      addr_ptr = at_buffer(stmt->dest, dt);
+      addr_ptr = dest_is_ptr ? dest_val : at_buffer(stmt->dest, dt);
```
hughperkins, line 1571: why does the uint backed shared mean we have to change this line?

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1577,6 +1614,8 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
       val = ir_->make_value(atomic_fp_op, ir_->get_primitive_type(dt), addr_ptr,
                             /*scope=*/ir_->const_i32_one_,
                             /*semantics=*/ir_->const_i32_zero_, data);
+    } else if (dest_is_ptr) {
+      val = shared_float_atomic(*ir_, stmt->op_type, addr_ptr, data, dt);
```
hughperkins, line 1640: why is dest_is_ptr not true for shared int atomics?

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2096,6 +2135,8 @@ void TaskCodegen::generate_struct_for_kernel(OffloadedStmt *stmt) {
   task_attribs_.buffer_binds = get_buffer_binds();
 }
 
+// Return the SPIR-V address for a global/storage-buffer access. Only works for
+// device-buffer-backed pointers (via ptr_to_buffers_), not workgroup arrays.
```
hughperkins: oh nice 🎉

hughperkins: wait, what do you mean 'spirv address'? It is simply the 'device address' right? or ... 🤔  It is an address that is used inside spirv to represent global/storage buffers, but doesn't have any direct hardware meaning at this point? Or ... 🤔 ?

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -1088,17 +1088,20 @@ Value IRBuilder::float_atomic(AtomicOpType op_type,
         addr_ptr, data, [&](Value lhs, Value rhs) { return mul(lhs, rhs); },
         dt);
   } else if (op_type == AtomicOpType::min) {
+    // Use dt-derived type instead of t_fp32_ so FMin/FMax work for f16/f64.
+    auto float_type = get_primitive_type(dt);
```
hughperkins: ✅

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -1088,17 +1088,20 @@ Value IRBuilder::float_atomic(AtomicOpType op_type,
...
     return atomic_operation(
         addr_ptr, data,
         [&](Value lhs, Value rhs) {
-          return call_glsl450(t_fp32_, /*FMin*/ 37, lhs, rhs);
+          return call_glsl450(float_type, /*FMin*/ 37, lhs, rhs);
         },
         dt);
   } else if (op_type == AtomicOpType::max) {
+    auto float_type = get_primitive_type(dt);
```
hughperkins: ✅

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -1088,17 +1088,20 @@ Value IRBuilder::float_atomic(AtomicOpType op_type,
...
     return atomic_operation(
         addr_ptr, data,
         [&](Value lhs, Value rhs) {
-          return call_glsl450(t_fp32_, /*FMax*/ 40, lhs, rhs);
+          return call_glsl450(float_type, /*FMax*/ 40, lhs, rhs);
```
hughperkins, line 1110: ✅

quadrants/codegen/spirv/spirv_ir_builder.cpp
```diff
@@ -1157,7 +1160,7 @@ Value IRBuilder::atomic_operation(Value addr_ptr,
     // int loaded = atomicCompSwap(vals[0], old, new);
     /*
     * Don't need this part, theoretically
-    auto semantics = uint_imm ediate_number(
+    auto semantics = uint_immediate_number(
```
hughperkins: how did this compile before? Are we using this bit?
- if we arent, lets not chane it ? (but add a FIXME t oremove it)
- if we are now, what changes so taht we are using it now?
- if we always were, how did it compile before?

quadrants/codegen/spirv/spirv_codegen.cpp
hughperkins: (like, a "type" is specific to a framework, like a spirv type, an llvm type etc.

But an address is just an address, not framework specific, I think? )

quadrants/codegen/spirv/spirv_ir_builder.cpp
duburcqa: This is just a comment. I just fixed a typo.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: Before this PR, shared arrays never reached the float-atomic path; now they can, so all branches guard at_buffer() with dest_is_ptr ternaries.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: added comment.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: added comment.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: Added comment.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: Added comment. It is simply the device address indeed.

quadrants/codegen/spirv/spirv_ir_builder.cpp
duburcqa: Reverted anyway.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: It should be better now.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: It is not possible to reduce this any further from what I understand.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: Fixed.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: done.

quadrants/codegen/spirv/spirv_codegen.cpp
duburcqa: updated.

v01dXYZ:
@duburcqa `atomic_operation` seems to have the same emulation (it seems it is enabled except for `add`). I try to look at the SPIRV emitted for `min`/`max` and the differences between `origin/main`, but I don't see great differences.

hughperkins:
I asked Opus to minimize the diff between this branch and main, excluding tests and the new files you added. it came up with https://github.com/Genesis-Embodied-AI/quadrants/pull/452 . Thougths?

duburcqa:
I was assisted by Claude Opus to write this PR. I have read and review every changes in this PR. I take full responsibility for the lines added and removed in this PR. I won't blame any issue on Claude Opus.

hughperkins (approved):
Approving, as a codeowner, on the basis that Alexis has reviewed and approved this PR."

PR #436, opened 2026-03-30
[Build] Fix flaky test_clock_accuracy

hughperkins:
The test was flaky (~10% failure rate) due to two issues:
- block_dim=1 caused each thread to run on a different SM with slightly different clock rates, breaking the proportionality check at high i
- Cold instruction cache on first kernel execution inflated a[0]

Replace the sin-based workload with an LCG, use parallel execution within a single warp, capture timing inside the loop via a data-dependent store to prevent compiler hoisting, and warm up the instruction cache by delaying the start timestamp to j=10.

Example failure without this PR:

Screenshot 2026-04-11 at 05 27 10

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_intrinsics.py
```diff
@@ -50,19 +50,22 @@ def foo():
...
-                a[i] = qd.clock_counter() - start
-
-    foo()
+            x = state[i]
+            start = qd.i64(0)
+            for j in range((i + 1) * 50000):
+                x = (1664527 * x + 1013904223) % 2147483647
+                if j == 10:
+                    start = qd.clock_counter()
+                if x > 10:
```
hughperkins, line 83: needs a comment explaining why we are doing this x comparison

tests/python/test_intrinsics.py
```diff
@@ -50,19 +50,22 @@ def foo():
...
+                    a[i] = qd.clock_counter() - start
```
hughperkins, line 84: needs a comment explaining why we write inside the loop, not after it.

tests/python/test_intrinsics.py
```diff
@@ -50,19 +50,22 @@ def foo():
...
-                x = qd.sin(x * 1.0001 + j * 1e-6) + 1.2345
-            if x < 10.0:
-                a[i] = qd.clock_counter() - start
-
-    foo()
+            x = state[i]
+            start = qd.i64(0)
+            for j in range((i + 1) * 50000):
+                x = (1664527 * x + 1013904223) % 2147483647
+                if j == 10:
```
hughperkins, line 75: nees a comment explaining why we dont sart the loop till j=1, and why we do it inside the loop

tests/python/test_intrinsics.py
```diff
@@ -50,19 +50,22 @@ def foo():
...
-            for j in range((i + 1) * 2000):
-                x = qd.sin(x * 1.0001 + j * 1e-6) + 1.2345
-            if x < 10.0:
-                a[i] = qd.clock_counter() - start
-
-    foo()
+            x = state[i]
+            start = qd.i64(0)
+            for j in range((i + 1) * 50000):
+                x = (1664527 * x + 1013904223) % 2147483647
```
hughperkins, line 72: needs a comment saying that this is an LCG, and why we use it

tests/python/test_intrinsics.py
```diff
@@ -50,19 +50,22 @@ def foo():
...
+                if j == 10:
+                    start = qd.clock_counter()
+                if x > 10:
+                    a[i] = qd.clock_counter() - start
+            state[i] = x
```
hughperkins, line 86: needs a commment saying why we write the result back to state

tests/python/test_intrinsics.py
```diff
@@ -50,19 +50,22 @@ def foo():
...
         for i in range(32):
-            start = qd.clock_counter()
-            x = qd.random() * 0.5 + 0.5
-            for j in range((i + 1) * 2000):
-                x = qd.sin(x * 1.0001 + j * 1e-6) + 1.2345
-            if x < 10.0:
-                a[i] = qd.clock_counter() - start
-
-    foo()
+            x = state[i]
```
hughperkins, line 66: needs a comment saying why we read x from state initially

tests/python/test_intrinsics.py
```diff
@@ -50,19 +50,22 @@ def foo():
 @test_utils.test(arch=qd.cuda)
 def test_clock_accuracy():
     a = qd.field(dtype=qd.i64, shape=32)
+    state = qd.field(dtype=qd.i32, shape=32)
 
     @qd.kernel
-    def foo():
-        qd.loop_config(block_dim=1)
+    def measure_sequence_timings():
         for i in range(32):
```
hughperkins, line 64: needs a comment saying the significant of shape 32

hughperkins:
> I don't understand your PR:
> 
> > block_dim=1 caused each thread to run on a different SM with slightly different clock rates
> 
> Can you elaborate?
> 
> > Replace the sin-based workload with an LCG
> 
> This seems irrelevant. Unless you can explain why this is necessary, I would recommend avoid unless modifications.

please could ou place your commetns on the relevant lines of the pr. Otherwise we dont have threading ,which is hard to follow I feel.

hughperkins:
> sin won't actually work, because the if x < 10 line assumes a uniform distribution in the output values, which sin won't give.

what are you talking about? x is the output of 1 + sin. It cannot be larger than 10.

Please comment onto the line, so the comments go into a thread. This is clearly related to a pseicfic line, ie the choice of LCG vs sin.

PR #437, opened 2026-03-30
[Perf] Skip per-call overhead in perf_dispatch when 1 compatible impl.

hughperkins:
…mpatible

When only a single registered implementation passes is_compatible(), cache it in _sole_compatible_impl and bypass geometry hash computation, dict lookups, and time.time() checks on every subsequent call.

This eliminates ~10 us/call of Python overhead that caused ~13% FPS regression on CPU benchmarks (e.g. franka_random) where decomposed is incompatible and monolith is the sole option.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/quadrants/lang/_perf_dispatch.py
+++ b/python/quadrants/lang/_perf_dispatch.py
@@ -280,10 +280,10 @@ class PerformanceDispatcher(Generic[P, R]):
             raise QuadrantsRuntimeError("No suitable functions were found.")
 
         elif len(compatible_set) == 1:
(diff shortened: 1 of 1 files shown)


python/quadrants/lang/_perf_dispatch.py
```diff
@@ -252,6 +253,9 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs):
         if self._forced_impl is not None:
             return self._forced_impl(*args, **kwargs)
 
+        if self._sole_compatible_impl is not None:
```
hughperkins: why add this? why not just use _forced_impl, which is designed specifically for this?

PR #441, opened 2026-03-31
[Misc] QD_DUMP_IR dumps to files with the task_id added to the filename

hughperkins:
… tasks

When a kernel contains multiple top-level for loops, each loop is compiled as a separate offloaded task. Previously all tasks wrote their LLVM IR to the same file (kernel->name + "_llvm.ll"), so only the last task to finish was visible. Use the per-task name instead so each gets its own file.

Made-with: Cursor

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
@claude review

hughperkins:
@claude review

hughperkins:
@claude review

PR #13, opened 2025-06-04
[Mac] Update cocoa for mac os 15

hughperkins:
Issue: #

See https://genesis-ai-company.slack.com/archives/C08UKDR6KGR/p1749077734526669?thread_ts=1749076147.261559&cid=C08UKDR6KGR for context

Pre-requisite for merging https://github.com/genesis-company/taichi/pull/5

### Brief Summary

Update cocoa for mac os 15

copilot:summary

### Walkthrough

copilot:walkthrough

PR #450, opened 2026-04-05
[Perf] Enable zero-copy in to_torch() and to_numpy()

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/docs/source/user_guide/interop.md
+++ b/docs/source/user_guide/interop.md
@@ -149,18 +149,16 @@ assert v1.data_ptr() == v2.data_ptr()   # same underlying memory
 
 ### Apple Metal: synchronisation
 
-On Apple Metal, Quadrants and PyTorch MPS use separate Metal command queues. Quadrants kernel writes are made visible to the MPS-backed view via an automatic `qd.sync()` on every `to_torch()` / `to_numpy()` call. When you ask Quadrants for an independent buffer with `copy=True`, Quadrants additionally calls `torch.mps.synchronize()` after cloning so the returned tensor reflects the latest device writes:
+On Apple Metal, Quadrants and PyTorch MPS use separate Metal command queues. Quadrants kernel writes are made visible via an automatic `qd.sync()` on every `to_torch()` / `to_numpy()` call (both `copy=True` and `copy=False`). For `copy=False`, `qd.sync()` flushes the Quadrants queue so the zero-copy view sees the latest writes:
 
 ```python
 qd.init(arch=qd.metal)
 f = qd.field(qd.f32, shape=(64,))
 
 run_kernel(f)                       # queues writes on the Quadrants Metal stream
-view = f.to_torch()                 # qd.sync() runs internally; view sees the writes
-copy = f.to_torch(copy=True)        # qd.sync() + torch.mps.synchronize() run internally
+view = f.to_torch(copy=False)       # qd.sync() runs internally; view sees the writes
+copy = f.to_torch(copy=True)        # kernel copy via qd.sync()
 ```
-
-`view.clone()`, called by you on a tensor you already hold, is a plain PyTorch op and does **not** go through Quadrants -- it neither calls `qd.sync()` nor `torch.mps.synchronize()`. If you need that, either go through `f.to_torch(copy=True)` (which does both internally) or call `torch.mps.synchronize()` yourself before / after the clone.
 
 The reverse direction (PyTorch writes to a zero-copy view, then a Quadrants kernel reads from the same field) is **not** automatically synchronised. Because Quadrants and PyTorch MPS submit work to separate Metal command queues, a kernel launched immediately after a torch write may execute before the torch write has actually committed to memory:
 
--- a/docs/source/user_guide/interop.md
+++ b/docs/source/user_guide/interop.md
@@ -149,7 +149,7 @@ assert v1.data_ptr() == v2.data_ptr()   # same underlying memory
 
 ### Apple Metal: synchronisation
 
-On Apple Metal, Quadrants and PyTorch MPS use separate Metal command queues. Every `to_torch()` / `to_numpy()` call runs `qd.sync()` internally to flush the Quadrants queue. Additionally, `copy=True` (the default) calls `torch.mps.synchronize()` after the kernel copy so the returned tensor is immediately safe to use in torch ops. `copy=False` does **not** call `torch.mps.synchronize()`:
+On Apple Metal, Quadrants and PyTorch MPS use separate Metal command queues. Every `to_torch()` / `to_numpy()` call runs `qd.sync()` internally to flush the Quadrants queue. Additionally, `copy=True` (the default) calls `torch.mps.synchronize()` after the kernel copy. This is necessary because, on Metal, Quadrants and Torch do not share the same compute streams. `copy=False` does **not** call `torch.mps.synchronize()`:
 
 ```python
 qd.init(arch=qd.metal)
--- a/docs/source/user_guide/interop.md
+++ b/docs/source/user_guide/interop.md
@@ -84,7 +84,9 @@ clone = f.to_torch(copy=True)   # independent copy (default)
 auto  = f.to_torch()            # same as copy=True
 ```
 
-When using `copy=False`, modifications via the view are visible to subsequent Quadrants kernel reads, and vice-versa. The view stays valid until the underlying storage is reallocated -- typically on `qd.init()` or `qd.reset()`, after which a fresh call to `to_torch(copy=False)` / `to_numpy(copy=False)` returns a new view.
+When using `to_torch(copy=False)`, the returned tensor shares memory with the field. Modifications via the tensor are visible to subsequent Quadrants kernel reads, and vice-versa. The view stays valid until the underlying storage is reallocated -- typically on `qd.init()` or `qd.reset()`, after which a fresh call to `to_torch(copy=False)` returns a new view.
+
+`to_numpy(copy=False)` also shares memory, but the returned array is **read-only** on NumPy >= 2.0 (NumPy flags arrays imported via DLPack v0 capsules as non-writable). Use `to_torch(copy=False)` if you need a writable zero-copy view.
 
 ### When zero-copy is available
 
@@ -124,7 +126,7 @@ qd.sync()             # not strictly required; safe pattern
 print(f[0])           # 2.0
 ```
 
-Round-trip with NumPy on a CPU backend:
+Read-only view with NumPy on a CPU backend:
 
 ```python
 qd.init(arch=qd.cpu)
@@ -132,9 +134,9 @@ qd.init(arch=qd.cpu)
 a = qd.ndarray(qd.i32, shape=(8,))
 a.from_numpy(np.arange(8, dtype=np.int32))
 
-view = a.to_numpy(copy=False)
-view[0] = 100
-print(a.to_numpy()[0])   # 100
+view = a.to_numpy(copy=False)   # read-only zero-copy view
+print(view[0])                  # 0 -- reads directly from a's memory
+print(view.flags.writeable)     # False (NumPy >= 2.0, DLPack v0 capsule)
 ```
 
 ### Caching
--- a/docs/source/user_guide/interop.md
+++ b/docs/source/user_guide/interop.md
@@ -1,7 +1,8 @@
 # Numpy and Torch interop
 
-Quadrants provides interop with both numpy and PyTorch. There are two mechanisms:
+Quadrants provides interop with both numpy and PyTorch. There are three mechanisms:
 - **Copy-based**: convert data between quadrants fields/ndarrays and numpy arrays or torch tensors
+- **Zero-copy via DLPack**: obtain a torch tensor or numpy array that aliases the underlying Quadrants memory
 - **Direct pass-through**: pass torch tensors directly into kernels as ndarray arguments (zero-copy)
 
 ## Copy-based interop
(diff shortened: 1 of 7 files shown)


python/quadrants/lang/_interop.py
```diff
@@ -0,0 +1,91 @@
+"""Zero-copy tensor interop between Quadrants and PyTorch/NumPy via DLPack.
+
+Provides cached, zero-copy conversion of Fields and Ndarrays to PyTorch tensors using the DLPack protocol.
+The returned tensors directly alias the Quadrants device memory -- modifications to either side are visible
+in the other (after ``qd.sync()`` on Apple Metal).
```
duburcqa: ` (after `qd.sync()` on Apple Metal)` is inaccurate `torch.mps.synchronize()` is also required in case of copy.

python/quadrants/lang/_interop.py
```diff
@@ -0,0 +1,91 @@
...
+
+_ARCH_METAL = _qd_core.Arch.metal
+_ARCH_VULKAN = _qd_core.Arch.vulkan
+
+_DLPACK_SUPPORTED_DTYPES = frozenset(
+    {primitive_types.i32, primitive_types.i64, primitive_types.f32, primitive_types.f64, primitive_types.u1}
+)
+
+
+@functools.lru_cache(maxsize=1)
```
duburcqa: LRU cache is slow. It should be profiled but I would not be surprised if systematic recomputation is faster. I would recommend defined a global constant  instead of a function.

python/quadrants/lang/_interop.py
```diff
@@ -0,0 +1,91 @@
...
+    except (ImportError, ValueError):
+        return False
+
+
+def can_zerocopy(is_field: bool, dtype=None, is_scalar_field: bool = False, shape: tuple[int, ...] = ()) -> bool:
+    """Check whether zero-copy DLPack export is available for the current backend and data type.
+
+    Args:
+        is_field: ``True`` for SNode-backed Fields, ``False`` for Ndarrays.
+        dtype: The Quadrants dtype (e.g. ``qd.f32``).  Types not in the C++ DLPack whitelist (f16, u8, …) return False.
```
duburcqa: Avoid non-ascii char '…'.

python/quadrants/lang/_interop.py
```diff
@@ -0,0 +1,91 @@
...
+    """``True`` when the installed PyTorch supports DLPack ``bytes_offset`` on MPS.
+
+    Required for zero-copy export of Fields on Apple Metal.
+    Available since torch > 2.9.1 (see pytorch/pytorch#168193).
+    """
+    try:
+        import torch  # pylint: disable=C0415
+
+        parts = torch.__version__.replace("+", ".").split(".")[:3]
+        return tuple(map(int, parts)) > (2, 9, 1)
```
duburcqa: If you are already relying on package `packing` somewhere else, using it here would be more robust. If not, just keep this.

python/quadrants/lang/_interop.py
```diff
@@ -0,0 +1,91 @@
...
+            impl.get_runtime().sync()
+        return tc
+
+
+def invalidate_zerocopy_cache(obj) -> None:
+    """Remove the cached DLPack torch tensor, if any."""
+    try:
+        del obj._qd_dlpack_tc
+    except AttributeError:
+        pass
```
duburcqa: This caching mechanism is missing transpose. This is important for efficiency.

duburcqa: This caching mechanism is missing numpy view. This is important for efficiency.
NumPy>=1.22 (genesis requires >=1.26) supports np.from_dlpack() natively. Using it would skip importing torch and the torch tensor intermediary for numpy-only workloads. Rather than caching a .numpy() derived from a torch tensor, a np.from_dlpack() view could be cached independently. It also avoids the torch import when the user never uses torch at all.

python/quadrants/lang/_ndarray.py
```diff
@@ -65,10 +66,45 @@ def to_dlpack(self):
...
+
+        Uses zero-copy when possible (all backends except Vulkan, and when the dtype is supported by DLPack).
+
+        Args:
+            copy: ``None``/``False`` return a zero-copy view, ``True`` returns an independent copy.
+        """
+        from quadrants.lang._interop import (  # pylint: disable=C0415
+            can_zerocopy,
+            dlpack_to_torch,
+        )
```
duburcqa: I don't see any good reason to keep this import local. No obvious circular deps or direct dependency to torch. Maybe I'm missing something.

python/quadrants/lang/_ndarray.py
```diff
@@ -65,10 +66,45 @@ def to_dlpack(self):
...
         return impl.get_runtime().prog.ndarray_to_dlpack(self, self.arr)
 
+    @python_scope
+    def to_torch(self, *, copy=None):
+        """Converts this ndarray to a ``torch.Tensor`` via DLPack.
+
+        Uses zero-copy when possible (all backends except Vulkan, and when the dtype is supported by DLPack).
+
+        Args:
+            copy: ``None``/``False`` return a zero-copy view, ``True`` returns an independent copy.
```
duburcqa: That is not what `None` does.

python/quadrants/lang/_ndarray.py
```diff
@@ -65,10 +66,45 @@ def to_dlpack(self):
...
+        )
+
+        if copy is not True and can_zerocopy(is_field=False, dtype=self.dtype):
+            tc = dlpack_to_torch(self)
+            return tc.clone() if copy is True else tc
+
+        if copy is False:
+            raise ValueError("Zero-copy not available for this backend/dtype combination")
+
+        import torch  # pylint: disable=C0415
```
duburcqa: Personally I prefer to move local imports in top of its local scope.

python/quadrants/lang/_ndarray.py
```diff
@@ -132,12 +168,27 @@ def fill(self, val):
...
         Returns:
             numpy.ndarray: The result numpy array.
         """
+        if copy is not True:
+            from quadrants.lang._interop import dlpack_to_torch  # pylint: disable=C0415
+
+            try:
+                tc = dlpack_to_torch(self)
+                if tc.device.type == "cpu":
+                    return tc.numpy()
```
duburcqa: This is very efficient. Not acceptable.

python/quadrants/lang/_ndarray.py
```diff
@@ -132,12 +168,27 @@ def fill(self, val):
...
+            except (ImportError, RuntimeError):
+                pass
```
duburcqa: I'm not in favour of using `try-except` if the condition can be checked. What about storing a flag directly at instance level to keep track of whether zero-copy is supported by this tensor?

python/quadrants/lang/field.py
```diff
@@ -236,12 +248,35 @@ def fill(self, val):
...
+                np_arr = tc.numpy()
+                if dtype is not None and np_arr.dtype != dtype:
+                    if copy is False:
+                        raise ValueError("copy=False is incompatible with dtype conversion")
+                    return np_arr.astype(dtype)
+                return np_arr
+            if copy is False:
+                raise ValueError("Zero-copy to numpy requires a CPU backend")
+        elif copy is False:
+            raise ValueError("Zero-copy not available for this backend/type combination")
```
duburcqa: It would be nice to be more specific.

python/quadrants/lang/matrix.py
```diff
@@ -1327,17 +1352,43 @@ def to_numpy(self, keep_dims=False, dtype=None):
...
+            dlpack_to_torch,
+        )
+
+        if copy is not True and can_zerocopy(is_field=True, dtype=self.dtype, shape=self.shape):
+            import torch  # pylint: disable=C0415
+
+            tc = dlpack_to_torch(self)
+            as_vector = self.m == 1 and not keep_dims
+            shape_ext = (self.n,) if as_vector else (self.n, self.m)
+            expected = self.shape + shape_ext
```
duburcqa: Duplicated logics.

python/quadrants/lang/matrix.py
```diff
@@ -1312,10 +1312,35 @@ def to_numpy(self, keep_dims=False, dtype=None):
...
+            can_zerocopy,
+            dlpack_to_torch,
+        )
+
+        if copy is not True and can_zerocopy(is_field=True, dtype=self.dtype, shape=self.shape):
+            tc = dlpack_to_torch(self)
+            if tc.device.type == "cpu":
+                as_vector = self.m == 1 and not keep_dims
+                shape_ext = (self.n,) if as_vector else (self.n, self.m)
+                expected = self.shape + shape_ext
```
duburcqa: Duplicated logics.

python/quadrants/lang/matrix.py
```diff
@@ -1312,10 +1312,35 @@ def to_numpy(self, keep_dims=False, dtype=None):
...
 
         Returns:
             numpy.ndarray: The result NumPy array.
         """
+        from quadrants.lang._interop import (  # pylint: disable=C0415
+            can_zerocopy,
+            dlpack_to_torch,
+        )
+
+        if copy is not True and can_zerocopy(is_field=True, dtype=self.dtype, shape=self.shape):
```
duburcqa: This condition is wrong. You should prefer zero-copy even if a copy is requested. This is orthogonal. Zero-copy is always faster even if you need to copy in the end.

python/quadrants/lang/_ndarray.py
```diff
@@ -65,10 +66,45 @@ def to_dlpack(self):
...
+            copy: ``None``/``False`` return a zero-copy view, ``True`` returns an independent copy.
+        """
+        from quadrants.lang._interop import (  # pylint: disable=C0415
+            can_zerocopy,
+            dlpack_to_torch,
+        )
+
+        if copy is not True and can_zerocopy(is_field=False, dtype=self.dtype):
+            tc = dlpack_to_torch(self)
+            return tc.clone() if copy is True else tc
```
duburcqa: `if copy is True` is a dead branch. But it should not be. You should just not guard against `copy is not True`.

python/quadrants/lang/_ndarray.py
```diff
@@ -65,10 +66,45 @@ def to_dlpack(self):
...
+
+        Args:
+            copy: ``None``/``False`` return a zero-copy view, ``True`` returns an independent copy.
+        """
+        from quadrants.lang._interop import (  # pylint: disable=C0415
+            can_zerocopy,
+            dlpack_to_torch,
+        )
+
+        if copy is not True and can_zerocopy(is_field=False, dtype=self.dtype):
```
duburcqa: `can_zerocopy(is_field=False, dtype=self.dtype)` should be stored as a constant attribute of the instance for efficiency.

python/quadrants/lang/_interop.py
```diff
@@ -0,0 +1,91 @@
...
+        import torch  # pylint: disable=C0415
+
+        tc = torch.from_dlpack(obj.to_dlpack())
+        obj._qd_dlpack_tc = tc
+        if impl.current_cfg().arch == _ARCH_METAL:
+            impl.get_runtime().sync()
+        return tc
+
+
+def invalidate_zerocopy_cache(obj) -> None:
```
duburcqa: What about cache invalidation for `ScalarField` and `MatrixField` have no corresponding `_reset` hook?

python/quadrants/lang/struct.py
```diff
@@ -549,31 +549,40 @@ def from_torch(self, array_dict):
...
         The dictionary may be nested when converting nested structs.
 
+        Args:
+            copy: ``None`` (default) prefers zero-copy, ``True`` forces a copy, ``False`` requires zero-copy or raises.
+
         Returns:
             Dict[str, Union[numpy.ndarray, Dict]]: The result NumPy array.
         """
-        return {k: v.to_numpy() for k, v in self._items}
+        # Struct children have interleaved SNode memory; DLPack cannot represent this, so always use kernel copy.
```
duburcqa: Really?

duburcqa: struct members are separate SNode leaves with contiguous storage each, individual members could be zero-copied via DLPack from what I understand.

python/quadrants/lang/matrix.py
duburcqa: You should use zerocopy systematically, and call `torch.Tensor.numpy` on cpu backend.

python/quadrants/lang/_ndarray.py
duburcqa: You are missing call to `torch.mps.synchronize()` before clone on metal backend.

python/quadrants/lang/field.py
```diff
@@ -255,8 +290,32 @@ def to_numpy(self, dtype=None):
...
+        if copy is not True and can_zerocopy(is_field=True, dtype=self.dtype, is_scalar_field=True, shape=self.shape):
+            import torch  # pylint: disable=C0415
+
+            tc = dlpack_to_torch(self)
+            if device is not None and tc.device != torch.device(device):
+                if copy is False:
+                    raise ValueError(
+                        f"copy=False is incompatible with device transfer (data on {tc.device}, requested {device})"
+                    )
+                return tc.to(device)
```
duburcqa: Missing call to `torch.mps.synchronize()` on metal backend.

hughperkins:
@duburcqa conversation with Opus about this AI. Thoughts?

<details>
<summary><b>User:</b> this code was migrated from genesis, by an AI. Can you compare this code with what is currently in genesis, what is currently in quadrants main, and figure out why the
  code isnt ssimply automatically correct, from being ported from Genesis?</summary>

**Assistant:**

## Why it isn't automatically correct

Three structural reasons, in order of importance:

**1. The original lives in the *consumer*, not the *library*.** `qd_to_python` is in Genesis. Genesis owns the entire call path: it controls `use_zerocopy`, `gs.backend`, *and* it owns when `qd.reset()` is called relative to cache lifetime. When you push the same logic *down* into Quadrants' own classes, you inherit a different threat model: arbitrary user code can call `qd.reset()` between the cache fill and the next read, so cached views become use-after-free. The PR's commit history shows the AI hitting exactly this â€” segfaults on AMD GPU and Metal â€” and patching around it by disabling the numpy cache and changing the default of `to_numpy` mid-PR. The reviewer's comment #18 (no `_reset` hook on `ScalarField`/`MatrixField`) is the still-unfixed version of the same problem.

**2. The AI re-decided design choices instead of transcribing them.** Genesis's design has clear invariants that are easy to miss when refactoring: "always try zerocopy first, then clone if requested" (not "guard zerocopy behind copy preference"); "sync MPS twice â€” once before reading, once after cloning"; "cache transpose and numpy variants too". The AI re-derived the algorithm from intent and got the boundaries wrong in `if copy is not True and can_zerocopy(...)`, dropped the second MPS sync, and shrunk the cache to a single attribute. These map directly to comments #15/#16/#21, #1/#22/#23, and #5/#6.

**3. The Quadrants API is a different shape than the Genesis wrapper.** Genesis exposes one entry point, so it gets one cache, one set of conditions, one sync sequence. Quadrants has four to six entry points (`Ndarray.to_torch`, `Ndarray._ndarray_to_numpy`, `Ndarray._ndarray_matrix_to_numpy`, `ScalarField.to_torch`/`to_numpy`, `MatrixField.to_torch`/`to_numpy`, `StructField.to_torch`/`to_numpy`). Each got a copy-pasted version of the same boilerplate, drifting slightly between sites â€” comments #13 and #14 are just "you have the same logic twice in matrix.py", and the situation is the same across files. The Genesis design can't be ported one-method-at-a-time without first factoring a single helper that all the methods call.

A faithful port would have:
- Put the *whole* Genesis algorithm into `_interop.py` as one helper (`qd_to_python`-equivalent), kept the four cache attributes (`_tc`/`_T_tc`/`_np`/`_T_np`), replaced `gs.use_zerocopy`/`gs.backend` with `can_zerocopy(...)` evaluated *inside* that helper, and made every `to_torch`/`to_numpy` a one-liner that calls it.
- Added `_reset` hooks to **every** Field/Ndarray subclass before exposing the cache, since Quadrants exposes `qd.reset()` to users that Genesis never had to model.
- Carried over both `qd.sync()` *and* `torch.mps.synchronize()` verbatim.
- Kept the Genesis decision rule literally: "if zerocopy supported, take that path; clone iff `copy=True`."

The current PR is roughly an *interpretation* of Genesis rather than a port, and most of duburcqa's review is asking for the original Genesis behaviour back.

</details>

<details>
<summary><b>User:</b> it sounds like you are recommending we do the actions under 'a faithful port', is that a fair impression?</summary>

**Assistant:**

Yes, that's a fair impression â€” but with one caveat I should make explicit.

I was describing what a faithful port *would have looked like*, and that aligns closely with what duburcqa is asking for in the review. So if you accept the reviewer's framing, then yes, "a faithful port" is roughly the recommended path.

The caveat: there's one place where Genesis itself may not be the right target. The use-after-free risk on `qd.reset()` is real and **not present in Genesis** (because Genesis owns the reset call). Inside Quadrants, you can't just copy Genesis verbatim â€” you also need invalidation hooks on every cache holder. So it's "faithful port + reset hooks", not pure faithful port.

Two other things worth flagging before you commit to that direction:

1. **Genesis's design isn't universally praised in the review either.** The reviewer's comment #11 ("don't use try/except, use a flag") is actually a critique of a pattern Genesis uses too. So a fully faithful port would still draw that comment. Minor, but worth knowing.

2. **The `to_numpy` lifetime question is a genuine design call, not just a port mistake.** Your fix commit on 2026-04-20 ("Make to_numpy() always copy unless copy=False is explicitly requested") was reacting to a real segfault. Genesis defaults to zero-copy because Genesis users know about the model. Quadrants has a wider audience and `numpy` arrays are conventionally expected to outlive their source. Reverting to "default zero-copy for `to_numpy`" to match Genesis would re-introduce that hazard unless you also fix invalidation. It's worth deciding deliberately rather than just matching Genesis.

So my honest recommendation is: yes, do the faithful-port refactor as the structural backbone (one central helper, full cache shape, both syncs, correct `copy=True` â‡’ zerocopy-then-clone), AND add reset hooks on all field types, AND have an explicit conversation with duburcqa about the `to_numpy` default before changing it back.

</details>

docs/source/user_guide/interop.md
```diff
@@ -69,6 +70,128 @@ arr = np.zeros((4, 3, 2), dtype=np.float32)
...
+
+```python
+f = qd.field(qd.f32, shape=(1024,))
+
+view  = f.to_torch(copy=False)  # zero-copy view: aliases f's memory
+clone = f.to_torch(copy=True)   # independent copy
+auto  = f.to_torch()            # default: copy=None (see below)
+```
+
+Modifications via the view are visible to subsequent Quadrants kernel reads, and vice-versa. The view stays valid until the underlying storage is reallocated -- typically on `qd.init()` or `qd.reset()`, after which a fresh call to `to_torch()` / `to_numpy()` returns a new view.
```
duburcqa: You should mention somewhere in the doc that one should be careful it is asynchronous on Apple Metal because torch and quadrants do not share the same compute "stream" (not sure if it is the correct terminology). Which means that doing:
```python
a_tc = a.to_torch(copy=False)
a_tc.zero_()
my_kernel(a)
```
Does not guarantee that my_kernel with be computed after reset to zero on Apple Metal, until `torch.mps.synchronise()` is called explicitly after `a_tc.zero_()`. Currently it is not the case on purpose. This is managed manually in Genesis, only after performing all the desired operations on all fields involved. This is more efficient because it keeps filling the compute queue will emptying it, partially masking Python overhead.

docs/source/user_guide/interop.md
```diff
@@ -69,6 +70,128 @@ arr = np.zeros((4, 3, 2), dtype=np.float32)
...
+
+### When zero-copy is available
+
+Zero-copy uses [DLPack](https://github.com/dmlc/dlpack) and requires:
+
+- a backend with DLPack support: `cpu` (`x64`/`arm64`), `cuda`, `amdgpu`, or `metal` (Vulkan is not supported);
```
duburcqa: It is not supported, because it is mostly irrelevant for now. Vulkan-backed DLPack tensors cannot be processed by any existing well-known scientific computing library such as torch, numpy...etc

docs/source/user_guide/interop.md
```diff
@@ -69,6 +70,128 @@ arr = np.zeros((4, 3, 2), dtype=np.float32)
...
+- a DLPack-supported dtype: `i32`, `i64`, `f32`, `f64`, `u1` (other dtypes such as `f16`, `u8`, `u16` fall back to the kernel-copy path);
+- on Apple Metal, `torch >= 2.9.2` for fields (required for DLPack `bytes_offset` on MPS; see [pytorch/pytorch#168193](https://github.com/pytorch/pytorch/pull/168193));
+- 0-dim `ScalarField` instances are not zero-copyable on any backend (PyTorch DLPack `bytes_offset` limitation);
```
duburcqa, line 98: It should be clarified that Quadrants' implementation has been designed with torch as only first-class user interface in mind, otherwise bringing torch limitation to justify Quadrants' limitation is weird.

docs/source/user_guide/interop.md
```diff
@@ -69,6 +70,128 @@ arr = np.zeros((4, 3, 2), dtype=np.float32)
...
+- `StructField` members are not zero-copyable yet (see [Struct fields](#struct-fields) below); `copy=False` raises on `StructField`.
```
duburcqa: Only StructField with AOS should not be zero-copyable. If the layout is SAO, it should be fine as far as I know.

docs/source/user_guide/interop.md
```diff
@@ -69,6 +70,128 @@ arr = np.zeros((4, 3, 2), dtype=np.float32)
...
+
+Zero-copy `to_numpy()` additionally requires a CPU backend, because numpy arrays cannot reference GPU memory.
+
+### Semantics of `copy`
+
+| Value | Behaviour |
+|---|---|
+| `None` (default) | Zero-copy view if supported, otherwise an independent copy. Never raises for support reasons. |
```
duburcqa: I think it is a bad idea to use `None` by default. It should rather be `False`.

In Genesis we started with `copy=None` by default, before realising it was causing troubles for users that were expecting a copy and doing in-place modification on the output of forgetting to copy if before storing in a buffer.

If people care about efficiency, they should know about this `copy` option. If they don't, then they probably care more about reliability and user-friendliness. And I don't think it is very user-friendly that something converting a tensor to a completely different format from another library returns a shared view instead of a copy. Moreover, I think that having a consistent, simple, behaviour by default is just easier to support.

docs/source/user_guide/interop.md
```diff
@@ -69,6 +70,128 @@ arr = np.zeros((4, 3, 2), dtype=np.float32)
...
+
+qd.reset()
+qd.init(arch=qd.cuda)
+f2 = qd.field(qd.f32, shape=(1024,))
+v3 = f2.to_torch(copy=False)            # fresh view; v1/v2 must not be used
+```
+
+### Apple Metal: synchronisation
+
+On Apple Metal, Quadrants and PyTorch MPS use separate Metal command queues. Quadrants kernel writes are made visible to the MPS-backed view via an automatic `qd.sync()` on every `to_torch()` / `to_numpy()` call. Cloning a view also synchronises MPS so that the clone sees the latest writes:
```
duburcqa: `via an automatic `qd.sync()` on every `to_torch()` / `to_numpy()` call`

I don't think this is a good idea (see my earlier comment), at least it should be optional.

docs/source/user_guide/interop.md
duburcqa: Ok I just found out the section where you mention the behaviour on Metal. I'm not convinced by the behaviour that was chosen though.

docs/source/user_guide/interop.md
```diff
@@ -69,6 +70,128 @@ arr = np.zeros((4, 3, 2), dtype=np.float32)
...
+
+```python
+qd.init(arch=qd.metal)
+f = qd.field(qd.f32, shape=(64,))
+
+run_kernel(f)            # queues writes on the Quadrants Metal stream
+view = f.to_torch()      # qd.sync() runs internally; view sees the writes
+copy = view.clone()      # torch.mps.synchronize() runs internally; copy is up-to-date
```
duburcqa: I'm not convinced this is true:
https://github.com/Genesis-Embodied-AI/Genesis/blob/0a7760fddf955956dded652796487b55821c1fdd/genesis/utils/misc.py#L476

docs/source/user_guide/interop.md
hughperkins: do you mean, it should default to `copy=True`?

docs/source/user_guide/interop.md
hughperkins: so what do you think should happen?

python/quadrants/lang/_interop.py
hughperkins: Opus here.

`packaging` isn't currently a runtime dependency of Quadrants — it only appears in three CI workflow YAMLs (`win.yml`, `manylinux_wheel.yml`, `macosx.yml`), not in `pyproject.toml`, `setup.py`, or `requirements*.txt`. Per the second branch of your suggestion ("if not, just keep this"), I'll leave the ad-hoc parser in place. It handles `+cpu`/`+cu121` local-version suffixes via `replace("+", ".")`, and falls through to `False` via `except ValueError` on dev/rc/nightly tags (e.g. `2.10.0a0`) — which is the safe default since it just gates off the Metal-DLPack-bytes-offset optimisation rather than crashing.

python/quadrants/lang/_ndarray.py
hughperkins: Complicated change. Won't do.

hughperkins: (added FIXME)

hughperkins:
@claude review

duburcqa:
ok to merge

hughperkins:
Thanks! 🙌

hughperkins:
@claude review

hughperkins:
@claude review

hughperkins:
@claude review

hughperkins:
@claude review

hughperkins:
@claude review

python/quadrants/lang/field.py
```diff
@@ -385,8 +566,20 @@ def fill(self, val):
...
+        Args:
+            dtype: Optional target numpy dtype.
+            copy: ``True`` (default) returns an independent copy, ``False`` requires zero-copy or raises.
+        """
+        if copy is False:
+            arr = _try_zerocopy_numpy(self, copy=False, is_scalar=True)
+            if arr is not None:
+                if dtype is not None and arr.dtype != dtype:
+                    raise ValueError(f"copy=False is incompatible with dtype conversion ({arr.dtype} -> {dtype})")
+                return arr
```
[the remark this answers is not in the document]
hughperkins: self-classified by ai as nit

tests/python/test_to_torch_copy.py
```diff
@@ -0,0 +1,778 @@
...
+    """Skip the current test if the active backend doesn't support DLPack zero-copy."""
+    if qd.cfg.arch in _NO_ZEROCOPY_ARCHS:
+        pytest.skip(f"DLPack zero-copy not supported on {qd.cfg.arch.name}")
+    if qd.cfg.arch == qd.metal:
+        from quadrants.lang.field import (
+            _TORCH_MPS_SUPPORTS_DLPACK_BYTES_OFFSET,  # pylint: disable=C0415
+        )
+
+        if not _TORCH_MPS_SUPPORTS_DLPACK_BYTES_OFFSET:
+            pytest.skip("Metal DLPack zero-copy requires torch >= 2.9.2")
```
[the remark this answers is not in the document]
hughperkins: nit

PR #462, opened 2026-04-09
[Bug] Warn and fall back instead of erroring when CUDA toolkit missing

hughperkins:
When the conditional module for graph_do_while cannot be compiled on sm90+ GPUs (because cudadevrt is not found), emit a warning about degraded mode and fall back to host-side conditional loop instead of raising a fatal error.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/runtime/cuda/graph_manager.cpp
+++ b/quadrants/runtime/cuda/graph_manager.cpp
@@ -215,8 +215,9 @@ void GraphManager::resolve_ctx_ndarray_ptrs(
 // then loads the _qd_graph_do_while_cond function for use in conditional
 // while nodes. Only called once; subsequent calls are no-ops.
 void GraphManager::ensure_condition_kernel_loaded() {
-  if (cond_kernel_func_)
+  if (cond_kernel_load_attempted_)
     return;
+  cond_kernel_load_attempted_ = true;
 
   int cc = CUDAContext::get_instance().get_compute_capability();
   if (cc < 90) {
@@ -273,34 +274,56 @@ void GraphManager::ensure_condition_kernel_loaded() {
       break;
     }
   }
-  QD_ERROR_IF(cudadevrt_path.empty(),
-              "Cannot find cudadevrt library — required for graph_do_while. "
-              "Install the CUDA toolkit and set CUDA_HOME or CUDA_PATH.");
+  if (cudadevrt_path.empty()) {
+    QD_WARN(
+        "Could not find CUDA toolkit (cudadevrt library), so Quadrants "
+        "will fall back to degraded mode — the graph do-while loop will "
+        "run on the host instead of the GPU, and a major performance hit "
+        "is to be expected. Make sure the CUDA toolkit is installed at the "
+        "default location, otherwise either CUDA_HOME or CUDA_PATH must "
+        "be set to enable native graph conditionals.");
+    return;
+  }
 
-  // CUlinkState handle for the JIT linker session that combines our PTX
-  // with libcudadevrt.a to resolve the cudaGraphSetConditional extern.
+  // JIT-link our PTX with libcudadevrt.a to resolve the
+  // cudaGraphSetConditional extern, then load the resulting cubin.
+  // If anything goes wrong (e.g. incompatible toolkit version, corrupt
+  // library), warn and leave cond_kernel_func_ null so the caller falls
+  // back to host-side do-while.
   void *link_state = nullptr;
-  driver.link_create(0, nullptr, nullptr, &link_state);
+  try {
+    driver.link_create(0, nullptr, nullptr, &link_state);
 
-  std::size_t ptx_len = std::strlen(kConditionKernelPTX) + 1;
-  driver.link_add_data(link_state, /*CU_JIT_INPUT_PTX=*/1,
-                       const_cast<char *>(kConditionKernelPTX), ptx_len,
-                       /*name=*/"qd_cond", 0, nullptr, nullptr);
+    std::size_t ptx_len = std::strlen(kConditionKernelPTX) + 1;
+    driver.link_add_data(link_state, /*CU_JIT_INPUT_PTX=*/1,
+                         const_cast<char *>(kConditionKernelPTX), ptx_len,
+                         /*name=*/"qd_cond", 0, nullptr, nullptr);
 
-  driver.link_add_file(link_state, /*CU_JIT_INPUT_LIBRARY=*/4,
-                       cudadevrt_path.c_str(), 0, nullptr, nullptr);
+    driver.link_add_file(link_state, /*CU_JIT_INPUT_LIBRARY=*/4,
+                         cudadevrt_path.c_str(), 0, nullptr, nullptr);
 
-  void *cubin = nullptr;
-  std::size_t cubin_size = 0;
-  driver.link_complete(link_state, &cubin, &cubin_size);
+    void *cubin = nullptr;
+    std::size_t cubin_size = 0;
+    driver.link_complete(link_state, &cubin, &cubin_size);
 
-  driver.module_load_data(&cond_kernel_module_, cubin);
-  driver.module_get_function(&cond_kernel_func_, cond_kernel_module_,
-                             "_qd_graph_do_while_cond");
-  driver.link_destroy(link_state);
+    driver.module_load_data(&cond_kernel_module_, cubin);
+    driver.module_get_function(&cond_kernel_func_, cond_kernel_module_,
+                               "_qd_graph_do_while_cond");
+    driver.link_destroy(link_state);
 
-  QD_TRACE("Loaded graph_do_while condition kernel ({} bytes cubin)",
-           cubin_size);
+    QD_TRACE("Loaded graph_do_while condition kernel ({} bytes cubin)",
+             cubin_size);
+  } catch (const std::exception &e) {
+    if (link_state) {
+      driver.link_destroy.call_with_warning(link_state);
+    }
+    QD_WARN(
+        "Failed to compile graph conditional kernel: {}. Quadrants will "
+        "fall back to degraded mode — the graph do-while loop will run on "
+        "the host instead of the GPU, and a major performance hit is to be "
+        "expected.",
+        e.what());
+  }
 }
 
 void *GraphManager::add_kernel_node(void *graph,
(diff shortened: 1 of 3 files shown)


quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -273,34 +273,56 @@ void GraphManager::ensure_condition_kernel_loaded() {
...
-
-  driver.module_load_data(&cond_kernel_module_, cubin);
-  driver.module_get_function(&cond_kernel_func_, cond_kernel_module_,
-                             "_qd_graph_do_while_cond");
-  driver.link_destroy(link_state);
-
-  QD_TRACE("Loaded graph_do_while condition kernel ({} bytes cubin)",
-           cubin_size);
+  try {
+    driver.link_create(0, nullptr, nullptr, &link_state);
```
hughperkins, line 295: I didnt read these try... catch body lines in detail, but I copied and pasted the before and after lines into external files, and used Beyond Compare to check that only the indentaion changed.

hughperkins: Image

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -273,34 +273,56 @@ void GraphManager::ensure_condition_kernel_loaded() {
...
+  } catch (const std::exception &e) {
+    if (link_state) {
+      driver.link_destroy.call_with_warning(link_state);
+    }
+    QD_WARN(
+        "Failed to compile graph conditional kernel: {}. Quadrants will "
+        "fall back to degraded mode — the graph do-while loop will run on "
+        "the host instead of the GPU, and a major performance hit is to be "
+        "expected. Install a compatible CUDA toolkit and set CUDA_HOME or "
+        "CUDA_PATH to enable native graph conditionals.",
```
hughperkins: the last sentence doesnt apply here

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -456,14 +477,13 @@ bool GraphManager::try_launch(
     if (!cond_kernel_func_) {
       int cc = CUDAContext::get_instance().get_compute_capability();
       if (cc >= 90) {
-        // SM 9.0+ should always be able to load the condition kernel.
-        // Failing here means prerequisites are missing.
-        QD_ERROR(
-            "Condition kernel not available on SM {}; "
-            "cannot build graph_do_while",
+        QD_WARN(
+            "Condition kernel not available on SM {}. Quadrants will fall "
```
hughperkins: That warning is strange. It makes it osunds like having SM120 or similar is to blame. Really the messsage needs to say something like the user has a GPU that is
  compataible with GPU-side conditionals, which give faster performance, but for some reason software prerequisites appear to be missing, so we are falling back to degraded
  mode.

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -456,14 +477,14 @@ bool GraphManager::try_launch(
...
-            "cannot build graph_do_while",
+        QD_WARN(
+            "This GPU (SM {}) supports native graph conditionals for "
+            "faster performance, but the required software prerequisites "
+            "appear to be missing. Quadrants will fall back to degraded "
+            "mode — the graph do-while loop will run on the host instead "
+            "of the GPU, and a major performance hit is to be expected.",
             cc);
       }
-      // Pre-SM 9.0: fall back to host-side do-while loop.
```
hughperkins, line 466: put back these cmmeetns

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -456,14 +477,14 @@ bool GraphManager::try_launch(
     if (!cond_kernel_func_) {
       int cc = CUDAContext::get_instance().get_compute_capability();
       if (cc >= 90) {
-        // SM 9.0+ should always be able to load the condition kernel.
```
hughperkins, line 459: put back these cmmeetns

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -273,34 +273,55 @@ void GraphManager::ensure_condition_kernel_loaded() {
...
   }
-  QD_ERROR_IF(cudadevrt_path.empty(),
-              "Cannot find cudadevrt library — required for graph_do_while. "
-              "Install the CUDA toolkit and set CUDA_HOME or CUDA_PATH.");
+  if (cudadevrt_path.empty()) {
+    QD_WARN(
+        "Could not find CUDA toolkit (cudadevrt library), so Quadrants "
+        "will fall back to degraded mode — the graph do-while loop will "
+        "run on the host instead of the GPU, and a major performance hit "
+        "is to be expected. Install the CUDA toolkit and set CUDA_HOME or "
```
erizmr: Do we want to add the instructions of this optional dependency in readme or doc?

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -273,34 +273,55 @@ void GraphManager::ensure_condition_kernel_loaded() {
...
+    driver.link_complete(link_state, &cubin, &cubin_size);
+
+    driver.module_load_data(&cond_kernel_module_, cubin);
+    driver.module_get_function(&cond_kernel_func_, cond_kernel_module_,
+                               "_qd_graph_do_while_cond");
+    driver.link_destroy(link_state);
+
+    QD_TRACE("Loaded graph_do_while condition kernel ({} bytes cubin)",
+             cubin_size);
+  } catch (const std::exception &e) {
```
duburcqa, line 316: I think catching any possible exception if worst than catching nothing. It should be more selective because allowing fallback systematically is undesirable. We only want to do so if the issue is clearly identified, because warnings have high chances to be ignored by users, so it should be used only when necessary to get around known limitations.

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -458,9 +479,12 @@ bool GraphManager::try_launch(
...
         // Failing here means prerequisites are missing.
-        QD_ERROR(
-            "Condition kernel not available on SM {}; "
-            "cannot build graph_do_while",
+        QD_WARN(
+            "This GPU (SM {}) supports native graph conditionals for "
+            "faster performance, but the required software prerequisites "
+            "appear to be missing. Quadrants will fall back to degraded "
+            "mode — the graph do-while loop will run on the host instead "
+            "of the GPU, and a major performance hit is to be expected.",
```
duburcqa: Same. This should be an error because there is no legitimate reason to expect this failure mode to ever happen. If it does, I don't think a warning is sufficient. There is something fundamentally broken with the GPU of the user and I would rather maximise the chance to get a bug report on Quadrants' Github repo.

quadrants/runtime/cuda/graph_manager.cpp
hughperkins: what failures are you imagining we might hit here where we would prefer to crash rather than fall back?

quadrants/runtime/cuda/graph_manager.cpp
hughperkins: This is an outer warning. The cuda toolkit warning will have already fired at this point, if that is the reason, or the other try catch.

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -273,34 +274,55 @@ void GraphManager::ensure_condition_kernel_loaded() {
...
-  QD_ERROR_IF(cudadevrt_path.empty(),
-              "Cannot find cudadevrt library — required for graph_do_while. "
-              "Install the CUDA toolkit and set CUDA_HOME or CUDA_PATH.");
+  if (cudadevrt_path.empty()) {
+    QD_WARN(
+        "Could not find CUDA toolkit (cudadevrt library), so Quadrants "
+        "will fall back to degraded mode — the graph do-while loop will "
+        "run on the host instead of the GPU, and a major performance hit "
+        "is to be expected. Install the CUDA toolkit and set CUDA_HOME or "
+        "CUDA_PATH to enable native graph conditionals.");
```
duburcqa: > Install the CUDA toolkit and set CUDA_HOME or CUDA_PATH to enable native graph conditionals.

Make sure the CUDA toolkit is installed at default location, otherwise either CUDA_HOME or CUDA_PATH must be set ro enable native graph conditionals.

quadrants/runtime/cuda/graph_manager.cpp
duburcqa: Everything but 'cudadevrt' not found I would say. Is there anything else we would expect to happen during "normal" operation, ie something we want to support?

hughperkins: what is our criteria for deciding here?
- we want to make sure cuda toolkit is optional?
- but anything else we want to force the user to file a bug report and wait for a new version of genesis and quadrants to come out?

hughperkins: (like, bear in mind, if they swapped to an older GPU, they would still be able to run genesis in this case; just not on an SM90+ gpu: we would force them to wait for a newer vesrion of quadrants that addresses whatever-the-error-is)

quadrants/runtime/cuda/graph_manager.cpp
duburcqa: You are right. My bad.

quadrants/runtime/cuda/graph_manager.cpp
duburcqa: * we want to make sure cuda toolkit is optional?
* but anything else we want to force the user to file a bug report and wait for a new version of genesis and quadrants to come out?

Yes. This is my opinion. Feel free to think different.

duburcqa: > like, bear in mind, if they swapped to an older GPU, they would still be able to run genesis in this case; just not on an SM90+ gpu: we would force them to wait for a newer vesrion of quadrants that addresses whatever-the-error-is

Yes. But at least we make steady progress. I have been enduring this since more than one year in Genesis now, and I think we can be quite proud of the current state after all these cumulative fixes. So I would say this approach is quite successful. As long as we fix issues promptly, people are fairly comprehensive. Well, if the tutorials are not crashing of course, otherwise it is a different story...

hughperkins: Wouldnt failure to find cuda toolkit that must already be installed, because torch, count as such a bug? 🤔

duburcqa: > Wouldnt failure to find cuda toolkit that must already be installed, because torch, count as such a bug?

Actually yes. But since you are not planning to fix this anytime soon and mingrui has already been hitting this issue, I don't see any other way than throwing a warning for now. Hence, my previous reply: "As long as we fix issues promptly, people are fairly comprehensive."

hughperkins: I dont remember saying "we are not planning to fix this anytime soon". We also havent confirmed with mingrui:
- that he has cuda toolkit installed
- where it is installed

Are you saying if we did these things, and got him to either set CUDA_HOME (if non default), or add his additional 'default location' to the pr, we could switch back from warnings to errors?

duburcqa: > that he has cuda toolkit installed

I doubt Mingrui does not have torch installed.

> Are you saying if we did these things, and got him to either set CUDA_HOME (if non default), or add his additional 'default location' to the pr, we could switch back from warnings to errors?

Yes. If we are a robust handling of default location and CUDA_HOME, it should probably be flipped back to error in my view.

PR #464, opened 2026-04-09
[Misc] CUDA graph 5 Add fatbin

hughperkins:
Issue: #

### Brief Summary

copilot:summary

Ran unit tests on 5090 (SM120):

Screenshot 2026-04-09 at 13 25 23
Screenshot 2026-04-09 at 13 25 26
Screenshot 2026-04-09 at 13 25 29
Screenshot 2026-04-09 at 13 25 34

Note: I also ran all genesis tests:

Screenshot 2026-04-09 at 14 03 57

### Walkthrough

copilot:walkthrough


docs/source/user_guide/building_cudagraph_conditional_fatbin.md
```diff
@@ -0,0 +1,71 @@
...
+This will:
+
+1. Compile `quadrants/runtime/cuda/graph_do_while_cond.cu` with relocatable device
+   code for each target SM architecture.
+2. Device-link the result with `libcudadevrt.a` to resolve the
+   `cudaGraphSetConditional` extern.
+3. Write the fatbin as a C byte array to
+   `quadrants/runtime/cuda/graph_do_while_cond_fatbin.h`.
+
+After regenerating, rebuild Quadrants and commit the updated header.
```
hughperkins: do we need to rebuild in order to commit?

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -210,97 +139,40 @@ void GraphManager::resolve_ctx_ndarray_ptrs(
...
+// subsequent calls are no-ops.
 void GraphManager::ensure_condition_kernel_loaded() {
   if (cond_kernel_func_)
     return;
 
   int cc = CUDAContext::get_instance().get_compute_capability();
   if (cc < 90) {
     QD_INFO(
-        "graph_do_while natively requires SM 9.0+, but this device is SM {}. "
+        "graph_do_while requires SM 9.0+, but this device is SM {}. "
```
hughperkins: this isnt true. the gpu-side node needs SM90, on CUDA.

quadrants/runtime/cuda/graph_manager.cpp
```diff
@@ -210,97 +139,41 @@ void GraphManager::resolve_ctx_ndarray_ptrs(
...
-  driver.link_add_data(link_state, /*CU_JIT_INPUT_PTX=*/1,
-                       const_cast<char *>(kConditionKernelPTX), ptx_len,
-                       /*name=*/"qd_cond", 0, nullptr, nullptr);
+  static_assert(kConditionKernelFatbinSize > 0,
+                "Condition kernel fatbin is empty — regenerate with "
+                "scripts/build_condition_kernel_fatbin.sh");
 
-  driver.link_add_file(link_state, /*CU_JIT_INPUT_LIBRARY=*/4,
-                       cudadevrt_path.c_str(), 0, nullptr, nullptr);
+  uint32_t ret = driver.module_load_data.call(&cond_kernel_module_,
```
hughperkins: To Opus:  what does 'river.module_load_data.call(&cond_kernel_module_,
                                                kConditionKernelFatbin);' do?

hughperkins: Opus response:

  It calls cuModuleLoadData(module_ptr, fatbin_data), which loads the fatbin into a CUDA module. The key difference is .call() vs operator():
  • `.call()` (line 63) — calls the CUDA driver function and returns the error code as uint32. The caller checks the result.
  • `operator()` (line 91) — calls the function and fatally errors (QD_ERROR_IF) if it fails.

  The fatbin code uses .call() so it can check the return value and give a specific error message mentioning the SM version and the regeneration script, rather than getting a
  generic CUDA driver error.

scripts/build_condition_kernel_fatbin.sh
```diff
@@ -0,0 +1,62 @@
...
+#
+# Usage:
+#   ./scripts/build_condition_kernel_fatbin.sh
+#
+# See docs/source/user_guide/building_cudagraph_conditional_fatbin.md
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
```
hughperkins: should we use ${SCRIPT_DIR} ?

scripts/build_condition_kernel_fatbin.sh
```diff
@@ -0,0 +1,62 @@
...
+#
+# See docs/source/user_guide/building_cudagraph_conditional_fatbin.md
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+SRC="${REPO_ROOT}/quadrants/runtime/cuda/graph_do_while_cond.cu"
+OUT_DIR="${REPO_ROOT}/quadrants/runtime/cuda"
+OBJ=$(mktemp /tmp/condition_kernel.XXXXXX.o)
```
hughperkins: questoin to Opus: will the sass work also on windows, if we built it on linux?

hughperkins: Opus repsonse:

  Yes. SASS is GPU machine code — it runs on the GPU, not the CPU. The GPU doesn't care what OS the host is running. The same sm_90 SASS runs identically on an H100 whether the
  host is Linux or Windows. cuModuleLoadData loads the fatbin into the GPU the same way on both platforms.
  This is actually another advantage of the fatbin approach over the old JIT path — the JIT path needed platform-specific logic to find libcudadevrt.a (Linux) vs cudadevrt.lib
  (Windows), which is exactly the code we deleted in the merge conflict. The fatbin just works everywhere.

scripts/build_condition_kernel_fatbin.sh
```diff
@@ -0,0 +1,62 @@
...
+FATBIN=$(mktemp /tmp/condition_kernel.XXXXXX.fatbin)
+
+cleanup() { rm -f "${OBJ}" "${FATBIN}"; }
+trap cleanup EXIT
```
hughperkins: to Opus: under what conditions does this run?

hughperkins: from Opus:

  trap cleanup EXIT runs the cleanup function whenever the shell exits, regardless of the reason:
  • Script finishes normally
  • set -e triggers an early exit on error (e.g. nvcc fails)
  • Explicit exit call
  • Signal-based termination (SIGINT/Ctrl-C, SIGTERM, etc. — because these cause the shell to exit, which triggers the EXIT trap)

  So the temp files (${OBJ}, ${FATBIN}) are cleaned up in all cases. It's the standard pattern for temp file cleanup in bash scripts.

scripts/build_condition_kernel_fatbin.sh
hughperkins, line 1: I'd prefer this script in python. How would it look?

hughperkins: from Opus:

```
#!/usr/bin/env python3
  """Builds the pre-compiled fatbin for the graph_do_while condition kernel.
  Requires the CUDA toolkit (nvcc) to be installed.
  The output is a C header (graph_do_while_cond_fatbin.h) containing the fatbin
  as a byte array, which is compiled into the quadrants binary. At runtime
  this lets us load the condition kernel without needing libcudadevrt.a on
  the user's system.
  Usage:
      python scripts/build_condition_kernel_fatbin.py
  See docs/source/user_guide/building_cudagraph_conditional_fatbin.md
  """
  import shutil
  import subprocess
  import sys
  import tempfile
  from pathlib import Path
  REPO_ROOT = Path(__file__).resolve().parent.parent
  SRC = REPO_ROOT / "quadrants" / "runtime" / "cuda" / "graph_do_while_cond.cu"
  OUT_HEADER = REPO_ROOT / "quadrants" / "runtime" / "cuda" / "graph_do_while_cond_fatbin.h"
  SM_TARGETS = [
      "-gencode=arch=compute_90,code=sm_90",
      "-gencode=arch=compute_100,code=sm_100",
      "-gencode=arch=compute_120,code=sm_120",
  ]
  def find_nvcc() -> str:
      nvcc = shutil.which("nvcc")
      if nvcc:
          return nvcc
      for candidate in ["/usr/local/cuda/bin/nvcc"]:
          if Path(candidate).exists():
              return candidate
      sys.exit("Error: nvcc not found. Install the CUDA toolkit and add it to PATH.")
  def run(cmd: list[str]) -> None:
      print(f"  {' '.join(cmd)}")
      subprocess.check_call(cmd)
  def generate_header(fatbin_path: Path, out_path: Path) -> None:
      data = fatbin_path.read_bytes()
      lines = [
          "// Auto-generated by scripts/build_condition_kernel_fatbin.py",
          "// Do not edit manually. Regenerate with that script.",
          "//",
          "// Source: quadrants/runtime/cuda/graph_do_while_cond.cu",
          f"// Size: {len(data)} bytes",
          "",
          "#pragma once",
          "",
          "#include <cstddef>",
          "",
          "static const unsigned char kConditionKernelFatbin[] = {",
      ]
      for i in range(0, len(data), 16):
          chunk = data[i : i + 16]
          lines.append("    " + ", ".join(f"0x{b:02x}" for b in chunk) + ",")
      lines.append("};")
      lines.append("")
      lines.append(f"static const std::size_t kConditionKernelFatbinSize = {len(data)};")
      lines.append("")
      out_path.write_text("\n".join(lines))
      print(f"Wrote {out_path} ({len(data)} bytes)")
  def main() -> None:
      nvcc = find_nvcc()
      with tempfile.TemporaryDirectory() as tmpdir:
          obj = Path(tmpdir) / "condition_kernel.o"
          fatbin = Path(tmpdir) / "condition_kernel.fatbin"
          print("Compiling ...")
          run([nvcc, "-dc", "-rdc=true", *SM_TARGETS, str(SRC), "-o", str(obj)])
          print("Device-linking with libcudadevrt ...")
          run([nvcc, "-dlink", *SM_TARGETS, str(obj), "-lcudadevrt", "-fatbin", "-o", str(fatbin)])
          print("Generating C header ...")
          generate_header(fatbin, OUT_HEADER)
      print("Done.")
  if __name__ == "__main__":
      main()
```

scripts/build_condition_kernel_fatbin.py
```diff
@@ -0,0 +1,95 @@
...
+import sys
+import tempfile
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+SRC = REPO_ROOT / "quadrants" / "runtime" / "cuda" / "graph_do_while_cond.cu"
+OUT_HEADER = REPO_ROOT / "quadrants" / "runtime" / "cuda" / "graph_do_while_cond_fatbin.h"
+
+SM_TARGETS = [
+    "-gencode=arch=compute_90,code=sm_90",
```
hughperkins: cant we just specify versions, rather than the entire string?

scripts/build_condition_kernel_fatbin.py
```diff
@@ -0,0 +1,95 @@
...
+
+def main() -> None:
+    nvcc = find_nvcc()
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        obj = Path(tmpdir) / "condition_kernel.o"
+        fatbin = Path(tmpdir) / "condition_kernel.fatbin"
+
+        print("Compiling ...")
+        run([nvcc, "-dc", "-rdc=true", *SM_TARGETS, str(SRC), "-o", str(obj)])
```
hughperkins: question to Opus: waht does this do?

scripts/build_condition_kernel_fatbin.py
```diff
@@ -0,0 +1,92 @@
...
+        return nvcc
+    for candidate in ["/usr/local/cuda/bin/nvcc"]:
+        if Path(candidate).exists():
+            return candidate
+    sys.exit("Error: nvcc not found. Install the CUDA toolkit and add it to PATH.")
+
+
+def run(cmd: list[str]) -> None:
+    print(f"  {' '.join(cmd)}")
+    subprocess.check_call(cmd)
```
hughperkins, line 46: what ahppens if this fails? do we see the full error messages?

hughperkins: opus response:

  Yes — check_call inherits the parent's stdout/stderr, so nvcc's error messages print in real time as the command runs. When it fails, Python then raises CalledProcessError
  which adds the exit code. So you see both the full nvcc error output and the Python traceback confirming the non-zero exit.
  If we had used capture_output=True or subprocess.run(..., capture_output=True), the error output would be hidden unless explicitly printed from the exception. The current
  approach is the right one for a build script.

scripts/build_condition_kernel_fatbin.py
```diff
@@ -0,0 +1,92 @@
...
+    out_path.write_text("\n".join(lines))
+    print(f"Wrote {out_path} ({len(data)} bytes)")
+
+
+def main() -> None:
+    nvcc = find_nvcc()
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        obj = Path(tmpdir) / "condition_kernel.o"
+        fatbin = Path(tmpdir) / "condition_kernel.fatbin"
```
hughperkins: `fatbin_path`, `obj_path`

scripts/build_condition_kernel_fatbin.py
```diff
@@ -0,0 +1,92 @@
...
+    sys.exit("Error: nvcc not found. Install the CUDA toolkit and add it to PATH.")
+
+
+def run(cmd: list[str]) -> None:
+    print(f"  {' '.join(cmd)}")
+    subprocess.check_call(cmd)
+
+
+def generate_header(fatbin_path: Path, out_path: Path) -> None:
+    data = fatbin_path.read_bytes()
```
hughperkins: `fatbin_data`

scripts/build_condition_kernel_fatbin.py
```diff
@@ -0,0 +1,92 @@
...
+    lines = [
```
hughperkins: `header_lines`

scripts/build_condition_kernel_fatbin.py
```diff
@@ -0,0 +1,92 @@
...
+        "// Source: quadrants/runtime/cuda/graph_do_while_cond.cu",
+        f"// Size: {len(data)} bytes",
+        "",
+        "#pragma once",
+        "",
+        "#include <cstddef>",
+        "",
+        "static const unsigned char kConditionKernelFatbin[] = {",
+    ]
+    for i in range(0, len(data), 16):
```
hughperkins: add comment '# loops over chunks of 16 bytes at a time'

hughperkins: opus said :   That's the kind of comment the code already conveys — the range(0, len(fatbin_data), 16) with chunk = fatbin_data[i : i + 16] is self-explanatory. Adding a comment would be
  restating what the code says. I'd skip it unless you feel strongly?

hughperkins: fine... 😅

scripts/build_condition_kernel_fatbin.py
hughperkins: opus response:

  This is the device linking step. It resolves the cudaGraphSetConditional extern in the compiled object.
  The flow is two stages:
  1. nvcc -dc -rdc=true — compiles graph_do_while_cond.cu into a relocatable device object. The call to cudaGraphSetConditional is left as an unresolved extern (because -rdc=true
      means "I'll link later").
  2. nvcc -dlink — takes that object and links it against -lcudadevrt (which provides the cudaGraphSetConditional implementation), producing a self-contained fatbin with the
     extern fully resolved. The -fatbin flag says to output a fat binary containing SASS for all the SM targets.

  After this step, the fatbin has no remaining external dependencies — cudaGraphSetConditional is baked in. That's why we don't need libcudadevrt.a at runtime.

hughperkins: what is the meaning of each commandeline option?

hughperkins: Opus esponse:

  For the -dlink command:
  • `-dlink` — perform device linking (resolve cross-module device function references)
  • `*SM_TARGETS` — the -gencode flags specifying which SM architectures to produce SASS for (sm_90, sm_100, sm_120)
  • `str(obj_path)` — input: the relocatable device object from the compile step
  • `-lcudadevrt` — link against NVIDIA's device runtime library (provides cudaGraphSetConditional)
  • `-fatbin` — output a fatbin (fat binary containing SASS for multiple architectures) rather than a single cubin
  • `-o str(fatbin_path)` — output file path

  And for the -dc compile command:
  • `-dc` — compile to device-only relocatable object (no host code, externs left unresolved)
  • `-rdc=true` — enable relocatable device code (allows cross-module device function calls, required for calling cudaGraphSetConditional which lives in a separate library)
  • `*SM_TARGETS` — same gencode flags
  • `str(SRC)` — the .cu source file
  • `-o str(obj_path)` — output file path

hughperkins:
With the exception of the generated fatbin header file, I have read every line added in this PR, and reviewed the lines. I take responsibilty for the lines added and removed in this PR, and won't blame any issues on Opus.

quadrants/runtime/cuda/graph_do_while_cond_fatbin.h
```diff
@@ -0,0 +1,1030 @@
...
+    0x90, 0x00, 0x00, 0x00, 0xd0, 0x07, 0x00, 0x00, 0x4e, 0x56, 0x49, 0x44, 0x49, 0x41, 0x20, 0x43,
+    0x6f, 0x72, 0x70, 0x00, 0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00,
+    0x07, 0x00, 0x00, 0x00, 0x36, 0x00, 0x00, 0x00, 0x60, 0x00, 0x00, 0x00, 0x00, 0x70, 0x74, 0x78,
+    0x61, 0x73, 0x00, 0x43, 0x75, 0x64, 0x61, 0x20, 0x63, 0x6f, 0x6d, 0x70, 0x69, 0x6c, 0x61, 0x74,
+    0x69, 0x6f, 0x6e, 0x20, 0x74, 0x6f, 0x6f, 0x6c, 0x73, 0x2c, 0x20, 0x72, 0x65, 0x6c, 0x65, 0x61,
+    0x73, 0x65, 0x20, 0x31, 0x32, 0x2e, 0x38, 0x2c, 0x20, 0x56, 0x31, 0x32, 0x2e, 0x38, 0x2e, 0x39,
+    0x33, 0x00, 0x42, 0x75, 0x69, 0x6c, 0x64, 0x20, 0x63, 0x75, 0x64, 0x61, 0x5f, 0x31, 0x32, 0x2e,
+    0x38, 0x2e, 0x72, 0x31, 0x32, 0x2e, 0x38, 0x2f, 0x63, 0x6f, 0x6d, 0x70, 0x69, 0x6c, 0x65, 0x72,
+    0x2e, 0x33, 0x35, 0x35, 0x38, 0x33, 0x38, 0x37, 0x30, 0x5f, 0x30, 0x00, 0x2d, 0x61, 0x72, 0x63,
+    0x68, 0x20, 0x73, 0x6d, 0x5f, 0x31, 0x30, 0x30, 0x20, 0x2d, 0x6d, 0x20, 0x36, 0x34, 0x20, 0x2d,
```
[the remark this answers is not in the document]
hughperkins: curious. It parsed this into binary, and then interpreated that? 😲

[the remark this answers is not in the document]
hughperkins: The authro agent says:

  see, another agent says it is missing 90:
  "[Pasted text #1 +31 lines]
  "


  The other agent's methodology is wrong. Using strings to look for ASCII text like "sm_90" in a binary is unreliable — the SM version is encoded in binary ELF headers, not
  necessarily as an ASCII string. The strings output they found (like -arch sm_100 -m 64) are nvcc command-line fragments embedded in per-section metadata, not an exhaustive
  list of architectures.
  cuobjdump -lelf is the authoritative tool for inspecting fatbin contents — it actually parses the fatbin format. And it clearly shows all three:

  ELF file    1: check_fatbin.1.sm_90.cubin
  ELF file    2: check_fatbin.2.sm_100.cubin
  ELF file    3: check_fatbin.3.sm_120.cubin

  The fatbin is correct. sm_90 is present.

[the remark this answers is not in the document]
hughperkins: how can I verify the header file myself? what command?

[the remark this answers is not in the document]
hughperkins: Extract the bytes from the header and run cuobjdump:

  cd /path/to/quadrants
  python3 -c "
  import re
  from pathlib import Path
  text = Path('quadrants/runtime/cuda/graph_do_while_cond_fatbin.h').read_text()
  m = re.search(r'kConditionKernelFatbin\[\] = \{([^}]+)\}', text, re.DOTALL)
  data = bytes(int(x.strip(), 16) for x in m.group(1).split(',') if x.strip())
  Path('/tmp/check.fatbin').write_bytes(data)
  " && cuobjdump -lelf /tmp/check.fatbin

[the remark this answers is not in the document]
hughperkins: My output:

```
[Quadrants Build] hugh@hugh5090:~/git/quadrants$ export PATH=$PATH:/usr/local/cuda/bin
[Quadrants Build] hugh@hugh5090:~/git/quadrants$ python3 -c "
import re
from pathlib import Path
text = Path('quadrants/runtime/cuda/graph_do_while_cond_fatbin.h').read_text()
m = re.search(r'kConditionKernelFatbin\[\] = \{([^}]+)\}', text, re.DOTALL)
data = bytes(int(x.strip(), 16) for x in m.group(1).split(',') if x.strip())
Path('/tmp/check.fatbin').write_bytes(data)
" && cuobjdump -lelf /tmp/check.fatbin
ELF file    1: check.1.sm_90.cubin
ELF file    2: check.2.sm_100.cubin
ELF file    3: check.3.sm_120.cubin
```

erizmr (approved):

hughperkins:
Thanks!

PR #472, opened 2026-04-12
[Perf] Tiles 2a: add cholesky and ger

hughperkins:
Issue: #

### Brief Summary

  PR 2a: Rank-1 update and Cholesky factorization for Tile16x16

  What it adds to `_tile16.py`:
  • _get_col(k) / _set_col(k, val) — register access helpers that encapsulate the 16-way if k == N dispatch, eliminating repetition in all downstream methods
  • _ger_sub(a, b) — in-place rank-1 subtract (self -= a @ b^T) via subgroup shuffles
  • cholesky_(eps) — in-place 16x16 Cholesky factorization (A = L @ L^T), diagonal clamped to sqrt(max(value, eps)) for numerical stability

  Tests added:
  • test_tile16_ger_sub — parametrized over f32/f64
  • test_tile16_cholesky — parametrized over f32/f64, 3 source offsets (0, 5, 32), and 3 src-dst deltas (0, 3, 16); verifies both correctness against scipy.linalg.cholesky and
    that untouched dst regions are preserved

  Good points:
  • _get_col/_set_col helpers make _ger_sub (4 lines) and cholesky_ (22 lines) very readable compared to the raw 16-way unrolled form
  • These helpers pay forward — PR 2b's _trsm and any future tile operations can reuse them
  • Cholesky test has strong coverage: multiple offsets exercise the load/store boundary logic, f64 tests catch precision regressions, and the untouched-region check catches OOB
     writes
  • All internal — no public API surface, nothing to break for users

  Bad points:
  • _get_col/_set_col are themselves still 16-line boilerplate each; this is inherent to the struct-field approach but could theoretically be code-generated
  • The existing PR 1 methods (_eye_, _load, _store, _load3d, _store3d) still use raw self.rN access — they'd benefit from the same helpers but refactoring them would change PR
    1's diff
  • cholesky_ epsilon must be explicitly typed (qd.f32(1e-6) or qd.f64(1e-12)) to match the tile dtype; a mismatched type would likely cause a compile error, but there's no
    explicit validation or dtype-inference convenience

PR 2a walkthrough (relative to PR 1)

  The challenge

  PR 1 gave us a register-resident 16x16 tile that can load, store, and initialize to zero or identity. But a tile that can only shuttle data to and from memory isn't useful yet
   — we need it to compute. The immediate goal is blocked Cholesky factorization, which requires two operations: rank-1 updates and the factorization itself.

  What this PR adds

  Two new computational methods on the tile, plus two helper methods that make them readable.

  Register access helpers: `_get_col` and `_set_col`

  The tile's 16 columns live in named fields r0–r15. You can't index them by variable — self.r[k] doesn't work because they're struct fields, not an array. So any operation that
   needs "read/write column k" requires a 16-way if k == 0: ... if k == 1: ... dispatch.
  PR 1's _load, _store, and _eye_ each hardcode their own version of this pattern. Rather than repeat it again for every new method, this PR extracts it into two @qd.func
  helpers:
  • _get_col(self, k) — returns the value of register k
  • _set_col(self, k, val) — sets register k to val

  Each is 16 if branches — the only place this boilerplate lives now. Every subsequent method uses these instead of touching r0–r15 directly.

  Rank-1 subtract: `_ger_sub`

  _ger_sub(self, a, b) computes self -= a @ b^T in-place. Each thread provides its element of vectors a and b. The outer product is computed column by column: for each column j,
   broadcast b[j] from thread j to all threads via subgroup.shuffle, then subtract a * b[j] from column j.
  With the helpers, this is 4 lines:

  for j in range(_TILE):
      bc = qd.simt.subgroup.shuffle(b, qd.u32(j))
      self._set_col(j, self._get_col(j) - a * bc)

  This is used in blocked Cholesky for diagonal block updates (A_kk -= L_k @ L_k^T) and off-diagonal block updates (A_ik -= L_i @ L_k^T).

  Cholesky factorization: `cholesky_`

  cholesky_(self, eps) factorizes the tile in-place: replaces the lower triangle with L such that A ≈ L @ L^T. It processes one column k at a time, for k = 0..15:
  1. Diagonal element (only thread k executes this): compute s = sum(L[k,j]^2 for j < k) using _get_col, then set L[k,k] = sqrt(max(A[k,k] - s, eps)). The eps clamp prevents
     sqrt of negative for near-singular matrices.
  2. Broadcast diagonal: shuffle(diag_val, k) sends L[k,k] from thread k to all threads.
  3. Cross-thread dot product (all threads execute this, required for uniform shuffle): compute dot = sum(L[i,j] * L[k,j] for j < k). Each thread reads its own L[i,j] via
     _get_col(j), and gets L[k,j] by shuffling from thread k.
  4. Off-diagonal elements (threads with tid > k): set L[i,k] = (A[i,k] - dot) / L[k,k].

  With the helpers, this is 22 lines instead of the ~224 it would take with raw register access.

  Tests

  `test_tile16_ger_sub` — loads a matrix and two vectors, runs _ger_sub, verifies against numpy.outer. Parametrized over f32 (atol=1e-5) and f64 (atol=1e-10).
  `test_tile16_cholesky` — generates a well-conditioned 16x16 SPD matrix (B @ B^T + 16*I), runs cholesky_, compares the lower triangle against scipy.linalg.cholesky.
  Parametrized over:
  • dtype: f32 and f64 (with appropriate epsilon: 1e-6 vs 1e-12, and tolerance: 1e-5 vs 1e-10)
  • src_offset: 0, 5, 32 — where in a 64x64 array the SPD matrix sits, exercising the load/store offset logic
  • dst_delta: 0, 3, 16 — difference between load and store positions, verifying that store writes only the target region

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_tile16.py
+++ b/tests/python/test_tile16.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pytest
+import scipy.linalg
 
 import quadrants as qd
 from quadrants.lang.simt._tile16 import _TILE, _make_tile16x16
@@ -296,3 +297,71 @@ def test_tile16_make_caching():
     assert a is not c
     d = _make_tile16x16(qd.f64)
     assert c is d
+
+
+def _make_spd(seed: int = 42):
+    """Return a well-conditioned 16x16 symmetric positive-definite matrix."""
+    rng = np.random.RandomState(seed)
+    B = rng.randn(_TILE, _TILE).astype(np.float64)
+    return (B @ B.T + _TILE * np.eye(_TILE)).astype(np.float32)
+
+
+@test_utils.test(arch=qd.gpu)
+def test_tile16_ger_sub():
+    Tile = _make_tile16x16(qd.f32)
+    mat = qd.ndarray(qd.f32, (_TILE, _TILE))
+    vec_a = qd.ndarray(qd.f32, (_TILE,))
+    vec_b = qd.ndarray(qd.f32, (_TILE,))
+    out = qd.ndarray(qd.f32, (_TILE, _TILE))
+
+    @qd.kernel
+    def k1(
+        mat_arr: qd.types.NDArray[qd.f32, 2],
+        a_arr: qd.types.NDArray[qd.f32, 1],
+        b_arr: qd.types.NDArray[qd.f32, 1],
+        out_arr: qd.types.NDArray[qd.f32, 2],
+    ):
+        qd.loop_config(block_dim=_TILE)
+        for _ in range(_TILE):
+            t = Tile()
+            t._load(mat_arr, 0, _TILE, 0, _TILE)
+            tid = qd.simt.subgroup.invocation_id()
+            a_val = a_arr[tid]
+            b_val = b_arr[tid]
+            t._ger_sub(a_val, b_val)
+            t._store(out_arr, 0, _TILE, 0, _TILE)
+
+    M = np.arange(_TILE * _TILE, dtype=np.float32).reshape(_TILE, _TILE)
+    a = np.arange(_TILE, dtype=np.float32) + 1.0
+    b = np.arange(_TILE, dtype=np.float32) + 2.0
+    mat.from_numpy(M)
+    vec_a.from_numpy(a)
+    vec_b.from_numpy(b)
+    k1(mat, vec_a, vec_b, out)
+
+    expected = M - np.outer(a, b)
+    np.testing.assert_allclose(out.to_numpy(), expected, atol=1e-5)
+
+
+@test_utils.test(arch=qd.gpu)
+def test_tile16_cholesky():
+    Tile = _make_tile16x16(qd.f32)
+    src = qd.ndarray(qd.f32, (_TILE, _TILE))
+    dst = qd.ndarray(qd.f32, (_TILE, _TILE))
+
+    @qd.kernel
+    def k1(src_arr: qd.types.NDArray[qd.f32, 2], dst_arr: qd.types.NDArray[qd.f32, 2]):
+        qd.loop_config(block_dim=_TILE)
+        for _ in range(_TILE):
+            t = Tile()
+            t._load(src_arr, 0, _TILE, 0, _TILE)
+            t.cholesky_(qd.f32(1e-6))
+            t._store(dst_arr, 0, _TILE, 0, _TILE)
+
+    A = _make_spd()
+    src.from_numpy(A)
+    k1(src, dst)
+
+    L_gpu = np.tril(dst.to_numpy())
+    L_ref = scipy.linalg.cholesky(A.astype(np.float64), lower=True).astype(np.float32)
+    np.testing.assert_allclose(L_gpu, L_ref, atol=1e-5)
--- a/tests/python/test_tile16.py
+++ b/tests/python/test_tile16.py
@@ -299,27 +299,30 @@ def test_tile16_make_caching():
     assert c is d
 
 
-def _make_spd(seed: int = 42):
+def _make_spd(np_dtype=np.float32, seed: int = 42):
     """Return a well-conditioned 16x16 symmetric positive-definite matrix."""
     rng = np.random.RandomState(seed)
     B = rng.randn(_TILE, _TILE).astype(np.float64)
-    return (B @ B.T + _TILE * np.eye(_TILE)).astype(np.float32)
+    return (B @ B.T + _TILE * np.eye(_TILE)).astype(np_dtype)
 
 
+@pytest.mark.parametrize("qd_dtype", _QD_DTYPES)
 @test_utils.test(arch=qd.gpu)
-def test_tile16_ger_sub():
-    Tile = _make_tile16x16(qd.f32)
-    mat = qd.ndarray(qd.f32, (_TILE, _TILE))
-    vec_a = qd.ndarray(qd.f32, (_TILE,))
-    vec_b = qd.ndarray(qd.f32, (_TILE,))
-    out = qd.ndarray(qd.f32, (_TILE, _TILE))
+def test_tile16_ger_sub(qd_dtype):
+    test_utils.skip_if_f64_unsupported(qd_dtype)
+    np_dtype = _NP_DTYPES[qd_dtype]
+    Tile = _make_tile16x16(qd_dtype)
+    mat = qd.ndarray(qd_dtype, (_TILE, _TILE))
+    vec_a = qd.ndarray(qd_dtype, (_TILE,))
+    vec_b = qd.ndarray(qd_dtype, (_TILE,))
+    out = qd.ndarray(qd_dtype, (_TILE, _TILE))
 
     @qd.kernel
     def k1(
-        mat_arr: qd.types.NDArray[qd.f32, 2],
-        a_arr: qd.types.NDArray[qd.f32, 1],
-        b_arr: qd.types.NDArray[qd.f32, 1],
-        out_arr: qd.types.NDArray[qd.f32, 2],
+        mat_arr: qd.types.NDArray[qd_dtype, 2],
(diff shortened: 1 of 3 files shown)


tests/python/test_tile16.py
```diff
@@ -296,3 +297,71 @@ def test_tile16_make_caching():
...
+
+
+def _make_spd(seed: int = 42):
+    """Return a well-conditioned 16x16 symmetric positive-definite matrix."""
+    rng = np.random.RandomState(seed)
+    B = rng.randn(_TILE, _TILE).astype(np.float64)
+    return (B @ B.T + _TILE * np.eye(_TILE)).astype(np.float32)
+
+
+@test_utils.test(arch=qd.gpu)
```
hughperkins, line 310: these should be on f64 too, and tolerance adjusted appropariately

tests/python/test_tile16.py
```diff
@@ -296,3 +297,71 @@ def test_tile16_make_caching():
...
+    vec_a.from_numpy(a)
+    vec_b.from_numpy(b)
+    k1(mat, vec_a, vec_b, out)
+
+    expected = M - np.outer(a, b)
+    np.testing.assert_allclose(out.to_numpy(), expected, atol=1e-5)
+
+
+@test_utils.test(arch=qd.gpu)
+def test_tile16_cholesky():
```
hughperkins: I think I'd like to parametrize over different src and destination offsets, to check also that src and destination offset store/load works ok (previous tests were subject to shortcut, in the absence of cholesky).

hughperkins:
Tests look good to me.

hughperkins:
quesiton to opus:

Opus respones:

```
  The main concern is cholesky_ — it's 268 lines of extremely repetitive code with the same if k == 0: ... if k == 1: ... if k == 15: ... pattern repeated 6 times. The same
  "read register k" / "write register k" / "accumulate registers 0..k-1" patterns appear over and over.
  This is inherent to the struct-field approach (you can't index self.r0..self.r15 by variable), but it could be factored into reusable @qd.func helpers. For example:
  • _get_col(self, k) — the 16-way if k == N: cur = self.rN pattern
  • _set_col(self, k, val) — the 16-way if k == N: self.rN = val pattern
  • _dot_prefix(self, other, k) — the if k > 0: dot += self.r0 * other.r0; ... accumulation

  That would reduce cholesky_ from ~260 lines to maybe ~30, and _ger_sub would benefit too. It would also make any future methods (like solve_triangular_ in PR 2b) much shorter.
  The rest looks fine — tests are clean, no secrets, no hacks.
  Want me to refactor it?
```

### Yes

Opus response:

```
  Done. Here's what changed:
  Added `_get_col(k)` and `_set_col(k, val)` helpers — each is a 16-way if k == N dispatch over the r0..r15 registers. These are the only place the register-name boilerplate
  lives now.
  Refactored using helpers:
  • _ger_sub: 34 lines → 4 lines
  • cholesky_: 224 lines → 22 lines
  • _trsm: 158 lines → 12 lines

  All 104 tests pass on CUDA and Vulkan. All downstream branches (2b through 5) updated and pushed.
```

Observation from me:

Code went down from +358 lines to +209 lines 🙌

hughperkins:
### Are there any layout inconsistencies in this code relative to itself and to other existing repo code?

Opus response:

```
  No layout inconsistencies. The code is clean:
  • Indentation is consistent (4 spaces throughout)
  • Blank line spacing between methods matches the existing PR 1 patterns (one blank line between @qd.func methods inside the class)
  • Docstring style matches existing methods (triple-quoted, imperative mood)
  • Test function spacing matches (two blank lines between top-level test functions)
  • Decorator ordering is consistent (@pytest.mark.parametrize → @test_utils.test → def)
  • No mixed banners or separators
  • Import ordering follows the existing convention (stdlib → third-party → quadrants → tests)
```

### Are there any coding inconsistencies in this code relative to itself and to other existing repo code?

```
  I see one inconsistency:
  Tolerance handling — PR 1 tests pass no atol at all (relying on the default), while the new PR 2a tests use inline ternaries (atol = 1e-10 if qd_dtype == qd.f64 else 1e-5).
  This is fine for now since PR 1 tests are exact (zeros, identity, load/store roundtrips) and don't need tolerance. No actual inconsistency.
  Everything else looks consistent:
  • Variable naming (np_dtype, qd_dtype, Tile, src, dst, result, expected) matches PR 1 patterns
  • Test structure (skip_if_f64 → setup → kernel → host verify) is uniform
  • _make_spd parameter name np_dtype matches the existing _NP_DTYPES convention
  • cholesky_ naming follows the trailing-underscore-for-in-place convention used by _eye_()
  • _ger_sub is private (underscore prefix) consistent with other internal methods

  No coding inconsistencies to fix.
```

### Is there anything in this PR that could be implemented in a way that will run faster at runtime?

```
  Looking at the cholesky_ implementation:

   quadrants/python/quadrants/lang/simt/_tile16.py lines 395-419

          def cholesky_(self, eps):
              ...
              for k in range(_TILE):
                  diag_val = qd.cast(0.0, dtype)
                  if tid == k:
                      s = qd.cast(0.0, dtype)
                      for j in range(_TILE):
                          if k > j:
                              c = self._get_col(j)
                              s += c * c
                      diag_val = qd.sqrt(qd.max(self._get_col(k) - s, eps))
                      self._set_col(k, diag_val)
                  diag_k = qd.simt.subgroup.shuffle(diag_val, qd.u32(k))
                  dot = qd.cast(0.0, dtype)
                  for j in range(_TILE):
                      if k > j:
                          my_col = self._get_col(j)
                          Lkj = qd.simt.subgroup.shuffle(my_col, qd.u32(k))
                          dot += Lkj * my_col
                  if tid > k:
                      new_val = (self._get_col(k) - dot) / diag_k
                      self._set_col(k, new_val)

  The diagonal computation (lines 400-406, inside if tid == k) does redundant work: only one thread executes it per iteration, but it computes s by looping over all previous
  columns. Since for k in range(_TILE) unrolls at compile time, the inner for j in range(_TILE): if k > j: also unrolls — this part is already efficient (the comparisons are
  compile-time constants).
  The cross-thread dot product (lines 413-417) runs on all threads, including tid <= k which then throws away dot. But this is likely unavoidable since the shuffles must be
  executed uniformly across the subgroup.
  So no — there's nothing here that would meaningfully run faster with a different implementation. The algorithm is already O(n^3) with minimal overhead, and the shuffle-based
  approach avoids shared memory entirely. The _get_col/_set_col calls should inline to direct register accesses after compilation.
```

### Is there anything in this PR which might exhibit incorrect behavior in some way?

```
  Looking at the code carefully for correctness issues:
  `_get_col` / `_set_col`: Use if chains (not elif), so all 16 comparisons run. Only one fires since k has a single value. Correct — but worth noting that _get_col initializes
  val = 0.0 first, so if some bizarre k > 15 were passed, it silently returns 0 rather than erroring. In practice k always comes from range(_TILE) so this is a non-issue.
  `_ger_sub`: Each iteration reads column j, computes the result, then writes it back. Different iterations touch different columns. No read-after-write hazard. Correct.
  `cholesky_`: The shuffle-based dot product loop runs on ALL threads (including tid <= k), which is required — divergent shuffles are undefined behavior on some GPUs. Thread k
  modifies only column k (via _set_col(k, diag_val)), so the subsequent dot product correctly reads unmodified columns j < k. Correct.
  Cholesky test: When src_offset=32 and dst_delta=16, dst_offset=48 and dst_row_end=64, which equals GRID=64 — so the store region just barely fits. Correct.
  I don't see any incorrect behavior. The algorithm is a faithful implementation of column-by-column Cholesky, the refactored helpers are semantically identical to the originals
   (verified by the 104 passing tests), and the test coverage exercises meaningful edge cases (offsets, partial columns, dtype combinations).
```

erizmr (approved):

hughperkins:
Thanks!

PR #474, opened 2026-04-13
[Perf] Tiles 2b: add triangular solve

hughperkins:
Issue: #

### Brief Summary

  This PR adds triangular solve to Tile16x16, building on the Cholesky factorization from PR 2a. Together, cholesky_ + solve_triangular_ give users the two primitives needed for
  blocked Cholesky-based linear solves entirely in registers.
  Two methods are added to _tile16.py:
  • `_trsm(self, L)` -- the internal workhorse. Solves X @ L^T = B in-place (self starts as B, ends as X) by forward-substituting column-by-column. Each column c computes a dot
    product of previously-solved columns against row c of L (fetched via subgroup.shuffle), subtracts from the current column, and divides by the diagonal. Uses the
    _get_col/_set_col helpers from PR 2a.
  • `solve_triangular_(self, B, lower=True)` -- the public API wrapper. Validates lower=True at compile time, then delegates to B._trsm(self). The argument order follows the
    convention "self is the factored matrix, B is modified in-place."

  The test loads a pre-computed Cholesky factor L and a random matrix B, runs L.solve_triangular_(B), and verifies X @ L^T ≈ B against scipy. Parametrized over f32/f64.
  Good points:
  • Very small and focused: 68 lines total (28 impl + 40 test), easy to review.
  • Reuses _get_col/_set_col from PR 2a, so the algorithm reads cleanly as a loop rather than 16-way unrolled register access.
  • The algorithm structure mirrors cholesky_ (column-by-column forward substitution with shuffles), making it easy to verify correctness by analogy.
  • Test validates against scipy with tight f64 tolerance.

  Bad points:
  • Only lower=True is supported; lower=False raises at compile time rather than implementing upper-triangular solve.
  • Test only exercises the full 16x16 case (offset 0, full tile). No partial-tile or offset-based parametrization like the Cholesky test in PR 2a has.
  • No test for the lower=False error path.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/quadrants/lang/simt/_tile16.py
+++ b/python/quadrants/lang/simt/_tile16.py
@@ -306,6 +306,268 @@ def _make_tile16x16_class(dtype):
             if tid == 15:
                 self.r15 = 1.0
 
+        @qd.func
+        def _ger_sub(self, a, b):
+            """General rank-1 subtract in-place: self -= a @ b^T."""
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(0))
+            self.r0 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(1))
+            self.r1 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(2))
+            self.r2 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(3))
+            self.r3 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(4))
+            self.r4 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(5))
+            self.r5 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(6))
+            self.r6 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(7))
+            self.r7 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(8))
+            self.r8 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(9))
+            self.r9 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(10))
+            self.r10 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(11))
+            self.r11 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(12))
+            self.r12 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(13))
+            self.r13 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(14))
+            self.r14 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(15))
+            self.r15 -= a * bc
+
+        @qd.func
+        def cholesky_(self, eps):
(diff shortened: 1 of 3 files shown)


hughperkins:
Questions to opus:

```
  Yes, two issues:
  1. The test reference computation is wrong. The _trsm solves X @ L^T = B, but the scipy reference computes (L^{-T} @ B)^T = B^T @ L^{-1} instead of B @ L^{-T}. This test
     actually fails (we saw this earlier). The fix is B.T not trans="T":


  # Wrong (current):
  X_ref = scipy.linalg.solve_triangular(L_ref, B, lower=True, trans="T").T
  # Correct:
  X_ref = scipy.linalg.solve_triangular(L_ref, B.T, lower=True).T

  2. Missing `# type: ignore[reportOperatorIssue]` in _trsm -- the same DSL operator expressions (+=, *, /) that needed suppression in cholesky_ appear here without it. Pyright
     CI will fail on this branch too.

  Let me also check if the protocol stubs were updated.

  ⬢ Grepped 4 greps
    Grepped "_trsm|solve_triangular_" in quadrants/python/quadrants/lang/simt/_tile16.py
    Grepped "def _trsm|def solve_triangular_" in quadrants/python/quadrants/lang/simt/_tile16.py
    Grepped "class _Tile16x16Proto" in quadrants/python/quadrants/lang/simt/_tile16.py
    Grepped "_Tile16x16Proto" in quadrants/python/quadrants/lang/simt/_tile16.py

  3. `solve_triangular_` is missing `@qd.func` -- cholesky_ and _ger_sub both have it, but solve_triangular_ does not. It'll fail when called from a kernel.

  Want me to fix all three issues?
```

### why dont the tests catch this?

```
  Good question -- let me reconsider each:
  Issue 1 (wrong reference): The test does fail -- we saw it earlier when running on hp/tiles-2b. So this IS caught. It's a test bug (wrong expected value), not a code bug.
  Issue 2 (missing pyright suppression): This would be caught by pyright CI, same as the hp/tiles-2a failure you just linked.
  Issue 3 (missing `@qd.func`): Let me check whether this actually matters.

  ⬢ Grepped "..._eye_|_load|_store|_get_col|_set_col)" in quadrants/python/quadrants/lang/simt/_tile16.py
    Found 398 matches

  I was wrong about issue #3. The tests on tiles-4 call L.solve_triangular_(B) inside a kernel and all 785 tests pass. So solve_triangular_ works without @qd.func -- Quadrants'
  AST transformer inlines dataclass method bodies regardless of the decorator. The if not lower: raise TypeError(...) acts as a compile-time guard, not runtime, which is why it
  works.
  So to correct my answer: there are really two issues, not three:
  1. Wrong test reference -- the test does fail (we saw it earlier), so it IS caught. It's just a test bug that was shipped. Fix: B.T instead of trans="T".
  2. Missing pyright suppression in `_trsm` -- pyright CI will catch this, same as it caught the cholesky_ issue on 2a.

  Want me to fix both on hp/tiles-2b and propagate?
```

### yes

```
  Done. Both fixes applied to hp/tiles-2b and propagated to all downstream branches:
  • Test fix: solve_triangular(L, B.T, lower=True).T instead of solve_triangular(L, B, lower=True, trans="T").T
  • Pyright fix: # type: ignore[reportOperatorIssue] on the two operator lines in _trsm
```

hughperkins:
### add test for the lower-False error path

```
Done. Added test_tile16_solve_triangular_upper_raises which verifies that calling solve_triangular_(..., lower=False) raises TypeError with the expected message.
```

erizmr (approved):

hughperkins:
Thanks! 🙌

PR #475, opened 2026-04-13
[Misc] Tiles 2c: Refactor: use _get_col/_set_col in tiles load/store/init

hughperkins:
Replace 16-way explicit register access with loop-based _get_col/_set_col helpers, eliminating ~145 lines of boilerplate.

Issue: #

### Brief Summary

  PR 2c: Refactor load/store/eye to use _get_col/_set_col helpers
  Purpose: Pure refactoring -- replaces 16-way explicit register access in _load, _store, _load3d, _store3d, and _eye_ with loops over _get_col/_set_col, the same helpers
  introduced in PR 2a for the linear algebra methods.
  Changes (all in `_tile16.py`, net -176 lines):
  • _load / _load3d: The 16 guarded assignments if col_start + N < col_end: self.rN = arr[row, col_start + N] become a 3-line loop: for j in range(16): if col_start + j <
    col_end: self._set_col(j, arr[row, col_start + j]). The row-clamping preamble is unchanged.
  • _store / _store3d: Same pattern in reverse -- arr[row, col_start + j] = self._get_col(j).
  • _eye_: 16 zero assignments + 16 conditional self.rN = 1.0 become two small loops: zero all registers, then set the diagonal.

  No test changes -- existing load/store/eye tests from PR 1 already exercise all these code paths.
  Good points:
  • Large net line reduction (~160 lines removed) with zero behavior change.
  • Every method body now fits on screen and reads as a clear loop rather than a wall of near-identical lines.
  • Eliminates the last direct self.rN access outside of _get_col/_set_col themselves, so future register layout changes only need to touch those two helpers.
  • Consistent with the same refactoring already applied to _ger_sub, cholesky_, and _trsm in PRs 2a/2b.

  Bad points:
  • No new tests added; relies entirely on existing coverage. A reviewer must trust that the mechanical transformation is correct.
  • Compile-time expansion: each _set_col/_get_col call in a loop unrolls to 16 if-branches, so _load now generates 16x16 = 256 branches in the IR vs the original 16. The
    compiler should optimize these away (each inner branch is on a loop-constant), but it's a theoretical compile-time regression.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/quadrants/lang/simt/_tile16.py
+++ b/python/quadrants/lang/simt/_tile16.py
@@ -306,6 +306,268 @@ def _make_tile16x16_class(dtype):
             if tid == 15:
                 self.r15 = 1.0
 
+        @qd.func
+        def _ger_sub(self, a, b):
+            """General rank-1 subtract in-place: self -= a @ b^T."""
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(0))
+            self.r0 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(1))
+            self.r1 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(2))
+            self.r2 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(3))
+            self.r3 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(4))
+            self.r4 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(5))
+            self.r5 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(6))
+            self.r6 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(7))
+            self.r7 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(8))
+            self.r8 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(9))
+            self.r9 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(10))
+            self.r10 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(11))
+            self.r11 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(12))
+            self.r12 -= a * bc
+            bc = qd.simt.subgroup.shuffle(b, qd.u32(13))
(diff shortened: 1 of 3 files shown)


hughperkins:

Opus response:

```
  Looking at the 2c-specific diff (from 2b to 2c), it's a clean mechanical refactoring. I don't see anything embarrassing -- the transformations are straightforward and
  consistent across all five methods.
  The only thing I'd flag is minor: the _eye_ two-loop pattern could be a single loop (self._set_col(j, 1.0 if tid == j else 0.0)) but that's a style preference, not
  embarrassing. And in Quadrants DSL, a ternary might not compile the same way, so the two-loop version is the safer choice.
  This one looks clean.

```

hughperkins:
###   talk me through '• Compile-time expansion: each _set_col/_get_col call in a loop unrolls to 16 if-branches, so _load now generates 16x16 = 256 branches in the IR vs the  original 16. The  compiler should optimize these away (each inner branch is on a loop-constant), but it's a theoretical compile-time regression.'

```
  The original _load had 16 explicit lines:

  if col_start < col_end:
      self.r0 = arr[row, col_start]
  if col_start + 1 < col_end:
      self.r1 = arr[row, col_start + 1]
  # ... 14 more

  That's 16 branches in the generated IR -- one per column.
  The new code is:

  for j in range(16):
      if col_start + j < col_end:
          self._set_col(j, arr[row, col_start + j])

  Quadrants unrolls range(16) at compile time, so this becomes 16 iterations. But each _set_col(j, val) is itself:

  if k == 0: self.r0 = val
  if k == 1: self.r1 = val
  # ... 14 more

  So iteration j=0 expands to: bounds check + 16 if-branches (only k==0 actually fires). Iteration j=1: bounds check + 16 if-branches (only k==1 fires). And so on. Total: 16
  iterations x 16 branches = 256 branches in the unrolled IR.
  In practice, since the loop is unrolled at compile time, j is a constant in each iteration. So the compiler sees if 0 == 0: self.r0 = val; if 0 == 1: self.r1 = val; ... and can
  trivially dead-code-eliminate 15 of the 16 branches per iteration, getting back to 16 effective assignments. But that optimization happens later in the pipeline -- the IR is
  temporarily 16x larger.
  The question is whether this matters. For _load/_store it's 256 branches, which is fine. For cholesky_ which has multiple _get_col/_set_col calls inside nested loops, the
  unrolled IR could be larger, but that was already accepted in PR 2a. The practical impact is likely negligible -- Quadrants already unrolls aggressively and the dead-code pass is
   cheap.
```

erizmr (approved):

hughperkins:
Thanks! 🙌

PR #483, opened 2026-04-14
[Perf] Tiles 5: add Cholesky benchmark demo

hughperkins:
Issue: #

## Summary
  Adds a complete blocked Cholesky factorization demo (`misc/demos/cholesky_blocked.py`) that benchmarks register-resident Tile16x16 against two scalar-Crout baselines, and
  links it from the Tile16x16 user guide.
  ### What's in this PR
  **`misc/demos/cholesky_blocked.py`** (323 lines) — a self-contained benchmark script that implements and compares three batched 92×92 Cholesky kernels across 4096
  environments:
  1. **Baseline**: scalar Cholesky-Crout in shared memory, 64 threads, 2N+1 sequential `block.sync()` calls.
  2. **Blocked**: same 6×6 tile blocking structure as Tile16x16 but using shared memory and scalar Crout for diagonal blocks, 16 threads.
  3. **Tile16x16**: fully register-resident via `Tile16x16`, no shared memory, zero syncs, 16 threads.
  The script loads SPD test matrices, runs warmup + timed iterations, verifies reconstruction error, and prints a results table.
  **`docs/source/user_guide/tile16.md`** — adds a "Example: blocked Cholesky" section at the end with a link to the demo and a results table (5.19× speedup on Blackwell).
  ### Good points
  - **Pedagogically valuable**: shows three progressively better implementations side-by-side, making it easy to see *why* register-resident tiles win (elimination of shared
  memory and sync barriers).
  - **Self-contained**: runs with a single `python misc/demos/cholesky_blocked.py` — generates its own test data, verifies correctness, and benchmarks.
  - **Concrete performance evidence**: documents real numbers on production hardware (RTX PRO 6000 Blackwell), giving users calibrated expectations.
  - **Small diff**: only 2 files, 335 lines added, no changes to library code.
  ### Bad points
  - **CUDA-only**: the demo hardcodes `qd.init(arch=qd.cuda)` and won't run on Vulkan/Metal.
  - **Hardcoded problem size**: N=92 and 4096 environments are baked in as module-level constants rather than being CLI-configurable.
  - **Not a unit test**: the demo lives in `misc/demos/` and isn't exercised by CI, so it could silently break if the Tile16x16 API changes.
  - **Benchmark numbers will drift**: the results table in the docs is a static snapshot from one specific GPU; future hardware or compiler changes won't be reflected.

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -13,13 +13,14 @@ import quadrants as qd
 from quadrants.lang.simt._tile16 import _make_tile16x16
 
 Tile = _make_tile16x16(qd.f32)
+N = Tile.SIZE  # 16
 
 @qd.func
 def my_blocked_op(A, row0, col0, eps):
     t = Tile.zeros()
-    t[:] = A[row0:row0+16, col0:col0+16]
+    t[:] = A[row0:row0+N, col0:col0+N]
     t.cholesky_(eps)
-    A[row0:row0+16, col0:col0+16] = t
+    A[row0:row0+N, col0:col0+N] = t
 ```
 
 ## Creating a tile
@@ -49,10 +50,9 @@ arr[batch, row0:row1, col0:col1] = t       # store to 3D array
 
 ### Slice value rules
 
-- **Start indices are required.** `arr[:16, :16]` is not allowed; write `arr[0:16, 0:16]`.
-- **Stop indices are optional.** If omitted, they default to `start + 16`. So `arr[row0:, col0:]` is equivalent to `arr[row0:row0+16, col0:col0+16]`.
-- **Row range** `[row0, row1)`: thread `tid` accesses row `row0 + tid`. Threads where `row0 + tid >= row1` are skipped. Additionally, rows beyond the array's shape are skipped. Typically `row1 = row0 + 16`, but smaller ranges work for partial tiles.
-- **Column range** `[col0, col1)`: each active thread loads/stores columns `col0` through `min(col1, arr.shape[-1]) - 1`. Tile columns beyond this range are left as zero (load) or skipped (store). At most 16 columns are accessed (tile registers `r0`–`r15` map to `col0`, `col0+1`, …, `col0+15`).
+- **Both start and stop indices are required.** `arr[:N, :N]` and `arr[0:, 0:]` are not allowed; write `arr[0:N, 0:N]`.
+- **Row range** `[row0, row1)`: thread `tid` accesses row `row0 + tid`. Threads where `row0 + tid >= row1` are skipped. Additionally, rows beyond the array's shape are skipped. Typically `row1 = row0 + Tile.SIZE`, but smaller ranges work for partial tiles.
+- **Column range** `[col0, col1)`: each active thread loads/stores columns `col0` through `min(col1, arr.shape[-1]) - 1`. Tile columns beyond this range are left as zero (load) or skipped (store). At most `Tile.SIZE` columns are accessed (tile registers `r0`–`r15` map to `col0`, `col0+1`, …, `col0+15`).
 - **Batch index** (3D only): a scalar integer indexing the leading dimension.
 
 ### Notes
@@ -89,7 +89,7 @@ v = arr[batch, row0:row1, col]   # 3D: same, with batch index
 t -= qd.outer(v, v)
 ```
 
-The row slice follows the same rules as tile slices: start is required, stop defaults to `start + 16` if omitted. `col` is a scalar column index. Each thread loads `arr[row0 + tid, col]`; threads where `row0 + tid >= row1` get zero.
+The row slice follows the same rules as tile slices: both start and stop are required. `col` is a scalar column index. Each thread loads `arr[row0 + tid, col]`; threads where `row0 + tid >= row1` get zero.
 
 ## Cholesky factorization
 
@@ -109,17 +109,19 @@ Solves `X @ L^T = B` in-place, replacing `B` with `X`. `L` must be a lower-trian
 
 ## Kernel structure
 
-Tile16x16 requires exactly 16 threads per subgroup. Use `qd.loop_config(block_dim=16)` before the parallel loop:
+Each tile operation uses 16 lanes of a subgroup. Set `block_dim=Tile.SIZE` so that each thread block is one 16-thread group:
 
 ```python
+N = Tile.SIZE
+
 @qd.kernel
 def my_kernel(A: qd.types.NDArray[qd.f32, 3]):
-    qd.loop_config(block_dim=16)
+    qd.loop_config(block_dim=N)
     for i in range(A.shape[0]):
         t = Tile.zeros()
-        t[:] = A[i, 0:16, 0:16]
+        t[:] = A[i, 0:N, 0:N]
         t.cholesky_(1e-6)
-        A[i, 0:16, 0:16] = t
+        A[i, 0:N, 0:N] = t
 ```
 
 ## f64 support
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -160,3 +160,7 @@ Not all GPU backends support f64. Use `test_utils.skip_if_f64_unsupported()` in 
 | `t -= qd.outer(a, b)` | General rank-1 subtract |
 | `t.cholesky_(eps)` | In-place Cholesky factorization |
 | `L.solve_triangular_(B)` | Triangular solve (in-place on B) |
+
+## Example: blocked Cholesky
+
+See [`misc/demos/cholesky_blocked.py`](../../../misc/demos/cholesky_blocked.py) for a complete blocked Cholesky factorization using Tile16x16, with benchmarks against NumPy.
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -156,4 +156,12 @@ t = qd.simt.Tile16x16.zeros(dtype=qd.f64)
 
 ## Example: blocked Cholesky
 
-See [`misc/demos/cholesky_blocked.py`](../../../misc/demos/cholesky_blocked.py) for a complete blocked Cholesky factorization using Tile16x16, with benchmarks against NumPy.
+See [`misc/demos/cholesky_blocked.py`](../../../misc/demos/cholesky_blocked.py) for a complete blocked Cholesky factorization using Tile16x16, benchmarked against scalar-Crout baselines (shared memory with 64 threads, and blocked shared memory with 16 threads).
+
+Results on RTX PRO 6000 Blackwell, 4096 environments, N=92, f32:
+
+| Kernel | Threads | Time (us) | vs baseline |
+|--------|--------:|----------:|------------:|
+| baseline (scalar Crout, shared mem) | 64 | 2766 | 1.00x |
+| blocked (scalar Crout, shared mem) | 16 | 2556 | 1.08x |
+| **tile16 (Tile16x16, no shared memory)** | 16 | 533 | **5.19x** |
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -105,7 +105,7 @@ Factorizes the tile in-place: replaces the lower triangle with `L` such that `L 
 L.solve_triangular_(B)
 ```
 
-Solves `X @ L^T = B` in-place, replacing `B` with `X`. `L` must be a lower-triangular tile (e.g. from `cholesky_()`). Only `lower=True` is supported; passing `lower=False` raises `TypeError`.
+Solves `X @ L^T = B` in-place, replacing `B` with `X`. `L` must be a lower-triangular, non-singular tile (all diagonal elements non-zero, e.g. from `cholesky_()`). Only `lower=True` is supported; passing `lower=False` raises `TypeError`.
 
 ## Kernel structure
 
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -164,3 +164,11 @@ Not all GPU backends support f64. Use `test_utils.skip_if_f64_unsupported()` in 
 ## Example: blocked Cholesky
 
 See [`misc/demos/cholesky_blocked.py`](../../../misc/demos/cholesky_blocked.py) for a complete blocked Cholesky factorization using Tile16x16, benchmarked against scalar-Crout baselines (shared memory with 64 threads, and blocked shared memory with 16 threads).
+
+Results on RTX PRO 6000 Blackwell, 4096 environments, N=92, f32:
+
+| Kernel | Threads | Time (us) | vs baseline |
+|--------|--------:|----------:|------------:|
+| baseline (scalar Crout, shared mem) | 64 | 2766 | 1.00x |
+| blocked (scalar Crout, shared mem) | 16 | 2556 | 1.08x |
+| **tile16 (Tile16x16, no shared memory)** | 16 | 533 | **5.19x** |
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -36,7 +36,7 @@ The `dtype` argument is optional — if omitted it defaults to the runtime's `de
 
 ## Loading and storing
 
-Load/store transfers data between a tile and device memory arrays using slice syntax. Both `qd.ndarray` and `qd.field` are supported (use `qd.Template` as the annotation for field arguments). Each thread accesses row `row0 + tid`, where `tid` is the thread's subgroup lane index (obtained internally via `subgroup.invocation_id()`).
+Load/store transfers data between a tile and device memory arrays using slice syntax. Both `qd.ndarray` and `qd.field` are supported. Each thread accesses row `row0 + tid`, where `tid` is the thread's subgroup lane index (obtained internally via `subgroup.invocation_id()`).
 
 ### Slice syntax
 
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -107,6 +107,22 @@ L.solve_triangular_(B)
 
 Solves `X @ L^T = B` in-place, replacing `B` with `X`. `L` must be a lower-triangular, non-singular tile (all diagonal elements non-zero, e.g. from `cholesky_()`). Only `lower=True` is supported; passing `lower=False` raises `TypeError`.
 
+## SharedArray support
+
+Tiles can load from and store to `qd.simt.block.SharedArray` using the same slice syntax as device arrays:
+
+```python
+sh = qd.simt.block.SharedArray((qd.simt.Tile16x16.SIZE, qd.simt.Tile16x16.SIZE), qd.f32)
+t = qd.simt.Tile16x16.zeros(dtype=qd.f32)
+t[:] = src[0:N, 0:N]
+sh[0:N, 0:N] = t          # store tile to shared memory
+qd.simt.block.sync()
+t2 = qd.simt.Tile16x16.zeros(dtype=qd.f32)
+t2[:] = sh[0:N, 0:N]      # load tile from shared memory
+```
+
+Column clamping applies the same way as for device arrays — columns beyond the SharedArray width are left as zero on load or skipped on store. Column vector slices (`v = sh[K0:K1, col]`) also work with SharedArray.
+
 ## Kernel structure
 
 ### Block size
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -1,0 +1,152 @@
+# Tile16x16: register-resident 16x16 tiles
+
+`Tile16x16` provides a 16x16 matrix tile that lives entirely in registers, distributed across 16 threads in a subgroup (warp). Each thread holds one row as 16 scalar registers. Cross-thread communication uses subgroup shuffles — no shared memory needed.
+
+This is useful for implementing blocked linear algebra kernels (Cholesky, triangular solve, etc.) where you want to keep working data in registers for maximum throughput.
+
+Tile16x16 runs on all GPU backends supported by Quadrants: CUDA, AMD, Metal, and Vulkan. It builds on `qd.simt.subgroup.shuffle`, which is cross-platform — no vendor-specific libraries required.
+
+## Quick start
+
+```python
+import quadrants as qd
+from quadrants.lang.simt._tile16 import _make_tile16x16
+
+Tile = _make_tile16x16(qd.f32)
+
+@qd.func
+def my_blocked_op(A, row0, col0, eps):
+    t = Tile.zeros()
+    t[:] = A[row0:row0+16, col0:col0+16]
+    t.cholesky_(eps)
+    A[row0:row0+16, col0:col0+16] = t
+```
+
+## Creating a tile
+
+```python
+Tile = _make_tile16x16(qd.f32)   # or qd.f64
+
+t = Tile.zeros()       # all zeros
+t = Tile.eye()         # 16x16 identity
+```
+
+`_make_tile16x16` returns a `qd.dataclass` type whose 16 fields (`r0`–`r15`) are the scalar registers for one row. The result is cached per dtype.
+
+## Loading and storing
+
+Load/store transfers data between a tile and device memory arrays using slice syntax. Each thread accesses row `row0 + tid`, where `tid` is the thread's subgroup lane index (obtained internally via `subgroup.invocation_id()`).
+
+### 2D arrays
+
+```python
+t = Tile.zeros()
+t[:] = arr[row0:row0+16, col0:col0+16]    # load
+arr[row0:row0+16, col0:col0+16] = t       # store
+```
+
+### 3D arrays
+
+For arrays with a leading batch dimension (e.g. `H[batch, row, col]`):
+
+```python
+t = Tile.zeros()
+t[:] = arr[batch, row0:row0+16, col0:col0+16]    # load
+arr[batch, row0:row0+16, col0:col0+16] = t       # store
+```
+
+The load/store automatically clamps to the array's shape, so out-of-bounds columns are left as zero (load) or skipped (store).
+
+The `[:]` on the load LHS is required — it distinguishes an in-place tile load from a variable rebinding. The store side does not need `[:]` because the array subscript on the LHS already triggers the correct assignment path.
+
+## Identity initialization
+
+For padding partial tiles in blocked algorithms:
+
+```python
+t = Tile.eye()    # create a new identity tile
+t._eye_()         # or reset an existing tile to identity in-place
+```
+
+Each thread sets its diagonal element to 1.0 and all others to 0.0.
+
+## Rank-1 updates
+
+```python
+t -= qd.outer(v, v)    # t -= v @ v^T  (symmetric)
+t -= qd.outer(a, b)    # t -= a @ b^T  (general)
+```
+
+Each thread provides its element(s) of the vector(s). The outer product is computed via subgroup shuffles and subtracted from the tile in-place. `qd.outer(a, b)` returns a deferred proxy — it is only valid as the RHS of `-=` on a Tile16x16. Composition like `qd.outer(a, b) + qd.outer(c, d)` raises `TypeError`.
+
+### Loading column vectors for outer products
+
+Column vectors can be loaded from arrays using slice syntax:
+
+```python
+t -= qd.outer(arr[row0:row0+16, col], arr[row0:row0+16, col])
+```
+
+Each thread loads one element from the column. Out-of-range threads get zero. This also works with 3D arrays:
+
+```python
+t -= qd.outer(arr[batch, row0:row0+16, col], arr[batch, row0:row0+16, col])
+```
+
+## Cholesky factorization
+
+```python
+t.cholesky_(eps)
+```
+
+Factorizes the tile in-place: replaces the lower triangle with `L` such that `L @ L^T ≈ A`. The `eps` parameter clamps the diagonal to avoid numerical issues with near-singular matrices. After this call, the lower triangle of `t` contains `L`.
+
+## Triangular solve
+
+```python
+L.solve_triangular_(B)
+```
+
+Solves `X @ L^T = B` in-place, replacing `B` with `X`. `L` must be a lower-triangular tile (e.g. from `cholesky_()`). Only `lower=True` is supported; passing `lower=False` raises `TypeError`.
+
+## Kernel structure
+
+Tile16x16 requires exactly 16 threads per subgroup. Use `qd.loop_config(block_dim=16)` before the parallel loop:
+
+```python
+@qd.kernel
+def my_kernel(A: qd.types.NDArray[qd.f32, 3]):
+    qd.loop_config(block_dim=16)
+    for i in range(A.shape[0]):
+        t = Tile.zeros()
+        t[:] = A[i, 0:16, 0:16]
+        t.cholesky_(1e-6)
+        A[i, 0:16, 0:16] = t
+```
+
+## f64 support
+
+Pass `qd.f64` to the factory for double precision:
+
+```python
+Tile64 = _make_tile16x16(qd.f64)
+```
+
+Not all GPU backends support f64. Use `test_utils.skip_if_f64_unsupported()` in tests.
+
+## Method reference
+
+| Operation | Description |
+|-----------|-------------|
+| `Tile.zeros()` | Create a zero-initialized tile |
+| `Tile.eye()` | Create an identity tile |
+| `Tile.SIZE` | Tile dimension constant (16) |
+| `t[:] = arr[r0:r1, c0:c1]` | Load from 2D array |
+| `t[:] = arr[i, r0:r1, c0:c1]` | Load from 3D array |
+| `arr[r0:r1, c0:c1] = t` | Store to 2D array |
+| `arr[i, r0:r1, c0:c1] = t` | Store to 3D array |
+| `t._eye_()` | Set to identity matrix (in-place) |
+| `t -= qd.outer(v, v)` | Symmetric rank-1 subtract |
+| `t -= qd.outer(a, b)` | General rank-1 subtract |
+| `t.cholesky_(eps)` | In-place Cholesky factorization |
+| `L.solve_triangular_(B)` | Triangular solve (in-place on B) |
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -35,7 +35,7 @@ t = qd.simt.Tile16x16.eye(dtype=qd.f32)      # 16x16 identity, f32
 t = qd.simt.Tile16x16.eye(dtype=qd.f64)      # 16x16 identity, f64
 ```
 
-The `dtype` argument is required. The underlying tile dataclass has 16 fields (`r0`–`r15`) — the scalar registers for one row. Tile types are cached per dtype internally.
+The `dtype` argument is required. The underlying tile dataclass has 16 fields (`r0`–`r15`) — the scalar registers for one row.
 
 ## Loading and storing
 
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -4,7 +4,7 @@
 
 This is useful for implementing blocked linear algebra kernels (Cholesky, triangular solve, etc.) where you want to keep working data in registers for maximum throughput.
 
-Tile16x16 runs on all GPU backends supported by Quadrants: CUDA, AMD, Metal, and Vulkan. It builds on `qd.simt.subgroup.shuffle`, which is cross-platform — no vendor-specific libraries required.
+Tile16x16 runs on all GPU backends supported by Quadrants: CUDA, AMD, Metal, and Vulkan. It builds on `qd.simt.subgroup.shuffle`, which is cross-platform — no vendor-specific libraries required. Using Tile16x16 on a CPU backend raises `QuadrantsSyntaxError`.
 
 ## Quick start
 
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -84,13 +84,15 @@ Each thread provides its element(s) of the vector(s). The outer product is compu
 Column vectors can be loaded from arrays using slice syntax:
 
 ```python
-t -= qd.outer(arr[row0:row0+16, col], arr[row0:row0+16, col])
+v = arr[row0:row0+16, col]
+t -= qd.outer(v, v)
 ```
 
 Each thread loads one element from the column. Out-of-range threads get zero. This also works with 3D arrays:
 
 ```python
-t -= qd.outer(arr[batch, row0:row0+16, col], arr[batch, row0:row0+16, col])
+v = arr[batch, row0:row0+16, col]
+t -= qd.outer(v, v)
 ```
 
 ## Cholesky factorization
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -67,7 +67,7 @@ For padding partial tiles in blocked algorithms:
 
 ```python
 t = Tile.eye()    # create a new identity tile
-t._eye_()         # or reset an existing tile to identity in-place
+t.eye_()         # or reset an existing tile to identity in-place
 ```
 
 Each thread sets its diagonal element to 1.0 and all others to 0.0.
@@ -155,7 +155,7 @@ Not all GPU backends support f64. Use `test_utils.skip_if_f64_unsupported()` in 
 | `t[:] = arr[i, r0:r1, c0:c1]` | Load from 3D array |
 | `arr[r0:r1, c0:c1] = t` | Store to 2D array |
 | `arr[i, r0:r1, c0:c1] = t` | Store to 3D array |
-| `t._eye_()` | Set to identity matrix (in-place) |
+| `t.eye_()` | Set to identity matrix (in-place) |
 | `t -= qd.outer(v, v)` | Symmetric rank-1 subtract |
 | `t -= qd.outer(a, b)` | General rank-1 subtract |
 | `t.cholesky_(eps)` | In-place Cholesky factorization |
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -163,4 +163,4 @@ Not all GPU backends support f64. Use `test_utils.skip_if_f64_unsupported()` in 
 
 ## Example: blocked Cholesky
 
-See [`misc/demos/cholesky_blocked.py`](../../../misc/demos/cholesky_blocked.py) for a complete blocked Cholesky factorization using Tile16x16, with benchmarks against NumPy.
+See [`misc/demos/cholesky_blocked.py`](../../../misc/demos/cholesky_blocked.py) for a complete blocked Cholesky factorization using Tile16x16, benchmarked against scalar-Crout baselines (shared memory with 64 threads, and blocked shared memory with 16 threads).
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -4,7 +4,7 @@
 
 This is useful for implementing blocked linear algebra kernels (Cholesky, triangular solve, etc.) where you want to keep working data in registers for maximum throughput.
 
-Tile16x16 runs on all GPU backends supported by Quadrants: CUDA, AMD, Metal, and Vulkan. It builds on `qd.simt.subgroup.shuffle`, which is cross-platform — no vendor-specific libraries required.
+Tile16x16 runs on all GPU backends supported by Quadrants: CUDA, AMD, Metal, and Vulkan. It builds on `qd.simt.subgroup.shuffle`, which is cross-platform — no vendor-specific libraries required. Using Tile16x16 on a CPU backend raises `QuadrantsSyntaxError`.
 
 ## Quick start
 
@@ -32,11 +32,11 @@ t = qd.simt.Tile16x16.eye(dtype=qd.f32)      # 16x16 identity, f32
 t = qd.simt.Tile16x16.eye(dtype=qd.f64)      # 16x16 identity, f64
 ```
 
-The `dtype` argument is required. The underlying tile dataclass has 16 fields (`r0`–`r15`) — the scalar registers for one row.
+The `dtype` argument is optional — if omitted it defaults to the runtime's `default_fp` (usually `qd.f32`). The underlying tile dataclass has 16 fields (`r0`–`r15`) — the scalar registers for one row.
 
 ## Loading and storing
 
-Load/store transfers data between a tile and device memory arrays using slice syntax. Each thread accesses row `row0 + tid`, where `tid` is the thread's subgroup lane index (obtained internally via `subgroup.invocation_id()`).
+Load/store transfers data between a tile and device memory arrays using slice syntax. Both `qd.ndarray` and `qd.field` are supported. Each thread accesses row `row0 + tid`, where `tid` is the thread's subgroup lane index (obtained internally via `subgroup.invocation_id()`).
 
 ### Slice syntax
 
@@ -106,6 +106,36 @@ L.solve_triangular_(B)
 ```
 
 Solves `X @ L^T = B` in-place, replacing `B` with `X`. `L` must be a lower-triangular, non-singular tile (all diagonal elements non-zero, e.g. from `cholesky_()`). Only `lower=True` is supported; passing `lower=False` raises `TypeError`.
+
+### Combined Cholesky + triangular solve
+
+A common pattern is to factorize a tile and immediately solve against it:
+
+```python
+L = qd.simt.Tile16x16.zeros(dtype=qd.f32)
+L[:] = A[0:N, 0:N]
+L.cholesky_(eps)
+B = qd.simt.Tile16x16.zeros(dtype=qd.f32)
+B[:] = rhs[0:N, 0:N]
+L.solve_triangular_(B)
+rhs[0:N, 0:N] = B
+```
+
+## SharedArray support
+
+Tiles can load from and store to `qd.simt.block.SharedArray` using the same slice syntax as device arrays:
+
+```python
+sh = qd.simt.block.SharedArray((qd.simt.Tile16x16.SIZE, qd.simt.Tile16x16.SIZE), qd.f32)
+t = qd.simt.Tile16x16.zeros(dtype=qd.f32)
+t[:] = src[0:N, 0:N]
+sh[0:N, 0:N] = t          # store tile to shared memory
+qd.simt.block.sync()
+t2 = qd.simt.Tile16x16.zeros(dtype=qd.f32)
+t2[:] = sh[0:N, 0:N]      # load tile from shared memory
+```
+
+Column clamping applies the same way as for device arrays — columns beyond the SharedArray width are left as zero on load or skipped on store. Column vector slices (`v = sh[K0:K1, col]`) also work with SharedArray.
 
 ## Kernel structure
 
--- a/docs/source/user_guide/tile16.md
+++ b/docs/source/user_guide/tile16.md
@@ -109,7 +109,9 @@ Solves `X @ L^T = B` in-place, replacing `B` with `X`. `L` must be a lower-trian
 
 ## Kernel structure
 
-Each tile operation uses 16 lanes of a subgroup. Set `block_dim=Tile.SIZE` so that each thread block is one 16-thread group:
+### Block size
+
+Set `block_dim=Tile.SIZE` so that each thread block contains exactly 16 threads — one per tile row:
 
 ```python
 N = Tile.SIZE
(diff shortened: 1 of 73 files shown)


docs/source/user_guide/tile16.md
```diff
@@ -160,3 +160,7 @@ Not all GPU backends support f64. Use `test_utils.skip_if_f64_unsupported()` in
 | `t -= qd.outer(a, b)` | General rank-1 subtract |
 | `t.cholesky_(eps)` | In-place Cholesky factorization |
 | `L.solve_triangular_(B)` | Triangular solve (in-place on B) |
+
+## Example: blocked Cholesky
+
+See [`misc/demos/cholesky_blocked.py`](../../../misc/demos/cholesky_blocked.py) for a complete blocked Cholesky factorization using Tile16x16, with benchmarks against NumPy.
```
hughperkins: "against numpy" 🤔

hughperkins: what? 😮

hughperkins:
<details>
<summary>1. To what extent do the tests match what is described in the doc?</summary>

# PR #483 review: do the tests match the doc?

## Summary

The doc (`docs/source/user_guide/tile16.md`) describes the full Tile16x16 API. The tests (`tests/python/test_tile16.py`) provide strong coverage of every individual API operation described in the doc. The one gap is the newly-added blocked Cholesky example â€” the composition of tile operations into a multi-block algorithm is only exercised in a benchmark demo, not in the test suite.

## What the doc describes and what is tested

### Well-covered (doc feature â†’ test)

| Doc section | Test(s) | Notes |
|---|---|---|
| `Tile16x16.zeros()` | `test_tile16_zeros` (line 33) | Both `Tile()` and `.zeros()` alias, f32/f64, ndarray/field |
| `Tile16x16.eye()` / `eye_()` | `test_tile16_eye` (line 61) | Constructor and in-place, both dtypes |
| `Tile16x16.SIZE` | `test_tile16_size_constant` (line 264), `test_tile16_size_constant_in_kernel` (line 270) | Constant outside and inside kernels |
| Slice load/store 2D | `test_tile16_slice_load_store_roundtrip` (line 516), `test_tile16_load_store` (line 105) | Multiple offsets, partial cols/rows |
| Slice load/store 3D (batch) | `test_tile16_slice_3d_batch` (line 567), `test_tile16_load3d_store3d` (line 229) | Multiple batch indices, partial tiles |
| Row clamping | `test_tile16_clamp_to_array_rows` (line 147), `test_tile16_3d_clamp_to_array_rows` (line 183), `test_tile16_load_store_partial_rows` (line 1073) | Both load-side (zeros for extra rows) and store-side (no OOB) |
| Column clamping | `test_tile16_load_clamp_to_array_cols` (line 288), `test_tile16_store_partial_cols_untouched` (line 318), `test_tile16_load_store_partial` (line 1045) | Load-side zero padding and store-side truncation |
| Start/stop indices required | `test_tile16_load_missing_start_raises` (line 814), `test_tile16_load_missing_stop_raises` (line 831), `test_tile16_store_missing_stop_raises` (line 849) | Plus vec variants (lines 996, 1016) |
| Negative indices rejected | `test_tile16_load_negative_row_raises` (line 778), `test_tile16_load_negative_col_raises` (line 796), `test_tile16_store_negative_row_raises` (line 1621), `test_tile16_store_negative_col_raises` (line 1639) | Load and store sides |
| `[:]` required on load LHS | `test_tile16_load_without_slice_rebinds` (line 938) | Verifies omitting `[:]` rebinds to proxy, not tile |
| `t -= qd.outer(a, b)` | `test_tile16_ger_sub` (line 367), `test_tile16_slice_ger_sub_via_outer` (line 593) | General rank-1 subtract, both internal and slice API |
| `t -= qd.outer(v, v)` | `test_tile16_outer_symmetric_same_variable` (line 668), `test_tile16_syr_sub` (line 1103) | Symmetric variant |
| Column vector loading 2D | `test_tile16_vec_proxy_ger_sub_2d` (line 634), `test_tile16_vec_proxy_syr_sub_2d` (line 1405) | Non-zero row offsets |
| Column vector loading 3D | `test_tile16_vec_proxy_ger_sub_3d` (line 700), `test_tile16_vec_proxy_syr_sub_3d` (line 1439) | Batch dimension |
| `qd.outer` composition error | `test_outer_composition_raises` (line 742) | Verifies TypeError on `p1 + p2` |
| `t += qd.outer(...)` rejected | `test_tile16_augassign_add_outer_raises` (line 959) | Only `-=` supported |
| `t.cholesky_(eps)` | `test_tile16_cholesky` (line 414), `test_tile16_potrf` (line 1134) | Both internal and public API, multiple offsets, f32/f64 |
| `L.solve_triangular_(B)` | `test_tile16_trsm` (line 462), `test_tile16_potrf_then_trsm` (line 1164) | Standalone and chained with cholesky |
| `lower=False` raises TypeError | `test_tile16_solve_triangular_upper_raises` (line 503) | |
| f64 support | Parametrized across most tests via `_QD_DTYPES = [qd.f32, qd.f64]` | |
| `block_dim=Tile16x16.SIZE` | Every kernel test uses `block_dim=_TILE` | Implicitly covered |

### Additional test coverage beyond the doc

The tests also cover several things not explicitly documented:

- **SharedArray â†” Tile16x16 interop** (lines 1204â€“1397): roundtrip, partial cols, Cholesky through shared memory, clamping. The doc doesn't have a SharedArray integration section, but these tests validate the infrastructure used by the blocked demo.
- **Vec proxy from SharedArray** (`test_tile16_vec_proxy_shared_array`, line 1473): column vector load from SharedArray for outer products.
- **Multi-column accumulation** (`test_tile16_vec_proxy_multi_column_accumulate`, line 1546): the Cholesky "lookback" pattern of accumulating multiple rank-1 updates.
- **Cross-dtype roundtrip** (`test_tile16_f64_roundtrip_into_f32_array`, line 1583): f32 data through f64 tiles.
- **CPU backend error** (`test_tile16_raises_on_cpu`, line 1674): clear error on unsupported backend.
- **Proxy default dtype and reinit** (lines 1700â€“1795): default_fp inference, survives init/reset cycles.

## Gap: blocked Cholesky composition

The doc's new "Example: blocked Cholesky" section (lines 157â€“167 of `tile16.md`) points to `misc/demos/cholesky_blocked.py`. This demo combines `cholesky_()`, `solve_triangular_()`, `qd.outer()`, and tile load/store into a multi-block algorithm over a 92Ã—92 matrix decomposed as a 6Ã—6 grid of 16Ã—16 tiles. The demo has its own `verify()` function (line 296 of the demo) that checks reconstruction error, but it is a manual benchmark script â€” not part of the automated test suite.

Each individual operation is well-tested in isolation. However, **there is no automated test for the multi-block composition** â€” i.e., a test that runs a blocked Cholesky on a matrix larger than 16Ã—16 using multiple tiles. The closest is `test_tile16_potrf_then_trsm` (line 1164), which chains cholesky + trsm on a single 16Ã—16 tile, and `test_tile16_vec_proxy_multi_column_accumulate` (line 1546), which tests the rank-1 lookback loop. Neither exercises the full blocked loop with multiple diagonal/off-diagonal tile iterations.

This is a moderate gap: correctness of the composition depends on tile stores from one iteration being visible to tile loads in the next (the "self-read-after-self-write" pattern described in the demo's comments, line 240). This memory ordering property is not tested anywhere else.

## Verdict

The tests match the doc well for the Tile16x16 API primitives. The main gap is an integration test for the blocked Cholesky algorithm â€” the key motivating example for the feature. Adding a small parameterized test (e.g., 32Ã—32 with 2Ã—2 blocks) that runs the tile16 blocked Cholesky kernel and verifies `L @ L^T â‰ˆ A` would close this gap without duplicating the full benchmark.

</details>

<details>
<summary>2. Is there any code that is duplicate by or in this PR that we could refactorize better somehow?</summary>

# PR #483 â€” Duplication Review

## Summary

The new file is a self-contained demo/benchmark, so some repetition is expected and arguably desirable for readability. That said, there are a few repeated patterns within the PR worth noting, and one minor overlap with existing test code.

## 1. Prior-block contribution loops in `cholesky_blocked()` (within PR)

The inner loop that subtracts prior-block contributions appears twice with nearly identical structure:

**Diagonal block** (lines 131â€“139):
```python
for r in range(TILE):
    c = tid
    if c <= r:
        s = qd.f32(0.0)
        for jb in range(kb):
            j0 = jb * TILE
            for t in range(TILE):
                s += H[k0 + r, j0 + t] * H[k0 + c, j0 + t]
        H[k0 + r, k0 + c] -= s
```

**Off-diagonal block** (lines 165â€“172):
```python
for r in range(TILE):
    c = tid
    s = qd.f32(0.0)
    for jb in range(kb):
        j0 = jb * TILE
        for t in range(TILE):
            s += H[i0 + r, j0 + t] * H[k0 + c, j0 + t]
    H[i0 + r, k0 + c] -= s
```

The only difference is the row base (`k0` vs `i0`) and the `if c <= r` guard (present for the symmetric diagonal case, absent for the general off-diagonal case). A `@qd.func` helper parameterized on row-base could unify these, but given this is a scalar baseline kernel meant to contrast with Tile16x16, the repetition is tolerable.

## 2. Prior-column rank-1 subtraction in `cholesky_tile16()` (within PR)

The rank-1 subtraction loop also appears twice:

**Diagonal block** (lines 216â€“220):
```python
for jb in range(kb):
    j0 = jb * TILE
    for t in range(TILE):
        v = L_tile16_field[env, k0:k1, j0 + t]
        L_kk -= qd.outer(v, v)
```

**Off-diagonal block** (lines 236â€“241):
```python
for jb in range(kb):
    j0 = jb * TILE
    for t in range(TILE):
        v_own = L_tile16_field[env, i0:i1, j0 + t]
        v_diag = L_tile16_field[env, k0:k1, j0 + t]
        L_ik -= qd.outer(v_own, v_diag)
```

The diagonal case is a specialization where both vectors come from the same row range (`k0:k1`). These could potentially share a `@qd.func` helper, but the symmetric `outer(v, v)` vs general `outer(v_own, v_diag)` distinction may make that awkward. Low priority.

## 3. Compile-verify pattern in `main()` (within PR)

Lines 294â€“307 repeat the same compile â†’ sync â†’ verify sequence three times:

```python
print("Compiling baseline ...")
cholesky_baseline()
qd.sync()
verify("baseline", L_baseline_field, A_np)
# ... same pattern for blocked and tile16 ...
```

This could be a loop over a list of `(name, kernel_fn, L_field)` tuples, similar to how the benchmark results are already printed via a loop (lines 315â€“322). Minor cleanup.

## 4. SPD matrix generation vs test utility (PR vs existing repo)

The demo's `make_spd_matrices()` (`misc/demos/cholesky_blocked.py:50â€“53`) generates batched NxN SPD matrices. The test suite has `_make_spd()` (`tests/python/test_tile16.py:357â€“361`) which generates a single 16x16 SPD matrix. These serve different purposes (batched benchmark vs single-tile unit test) and use different APIs (`np.einsum` vs `B @ B.T`), so extracting a shared utility isn't worthwhile.

## 5. No duplication with library code

The three kernels in the demo are all new code with no overlap with `_tile16.py` or other library internals. The demo consumes the Tile16x16 public API (`zeros`, `eye`, `cholesky_`, `solve_triangular_`, `qd.outer`) without reimplementing any of it.

## Verdict

The most actionable item is **#3** (loop over the compile-verify steps in `main()`), which is a trivial cleanup. Items **#1** and **#2** are inherent to the blocked Cholesky algorithm structure â€” the diagonal and off-diagonal cases are intentionally shown separately for pedagogical clarity in a demo file. No significant duplication with the existing codebase.

</details>

<details>
<summary>3. Is there any code or code change in this PR that might be embarrassing in some way if published in a public github repo?</summary>

# PR #483 â€” Embarrassment Check

**Overall: Clean.** No credentials, secrets, profanity, debug leftovers, or licensing issues found. Two minor items worth considering:

## 1. Internal project reference: `"dex_hand dimensions"`

`cholesky_blocked.py` line 287:

```python
print(f"Blocked Cholesky {N}x{N} benchmark (dex_hand dimensions)")
```

This names a specific internal application as the reason `N=92` was chosen. If "dex_hand" is an unreleased or internal project, this leaks its existence and the size of its state matrices. Consider replacing with something generic like `"N=92"` or `"robotics-scale matrices"`, or omit the parenthetical entirely.

## 2. Specific GPU model in docs and docstring

Both `tile16.md` and the script docstring cite **"RTX PRO 6000 Blackwell"**. This is standard practice for published benchmarks and not embarrassing per se, but it does pin the results to specific hardware. Fine to keep â€” just confirm you're comfortable with it being the permanent public reference point.

## Nothing else found

- No hardcoded credentials, API keys, or tokens.
- No TODO/HACK/FIXME comments.
- No profanity or embarrassing variable names.
- No personal information (emails, paths, usernames).
- No stray debug prints â€” all `print()` calls are intentional benchmark output.
- No misleading comments; code comments are accurate and useful.
- No licensing headers missing (none are used elsewhere in `misc/demos/`).

</details>

<details>
<summary>4. Are there any layout inconsistencies in this code relative to itself and to other existing repo code?</summary>

# PR #483 â€” Layout Inconsistencies

## Internal consistency

The new file (`misc/demos/cholesky_blocked.py`) is internally consistent: 4-space indentation, two blank lines between top-level definitions, blank lines separating logical blocks within functions, and uniform comment style throughout. No issues here.

## vs. existing repo code

Several layout choices diverge from the conventions established by the ~25 other scripts in `misc/`:

1. **Subdirectory placement.** Every other misc script lives flat in `misc/` (`benchmark_bls.py`, `benchmark_reduction.py`, `spMv_linear_solve.py`, etc.). This is the sole file in the new `misc/demos/` subdirectory. If `demos/` is intended as a new category, it's fine, but right now it's a one-file directory that breaks the flat convention.

2. **Shebang line.** `#!/usr/bin/env python3` â€” no other file in `misc/` has one.

3. **`if __name__ == "__main__": main()` guard.** Every existing `misc/` benchmark executes directly at module level (e.g., `benchmark_reduction.py`, `benchmark_reduction_tmps.py`, `benchmark_tensor_access.py`). The new file wraps everything in `main()`. This is arguably better practice, but inconsistent with every peer file.

4. **Module-level docstring.** The new file has a 25-line docstring. Most `misc/` files have no docstring at all (a few have a bare `# TODO` comment). This is a positive deviation but worth noting as a style delta.

5. **`qd.field(dtype=qd.f32, ...)`** uses the keyword form. All existing `misc/` files use the positional form `qd.field(qd.f32, ...)`:

   ```
   # Existing convention (positional):
   a = qd.field(qd.i32, shape=N)          # benchmark_reduction.py
   a = qd.field(qd.f32, shape=N)          # benchmark_reduction_tmps.py
   b = qd.field(qd.f32, shape=n)          # spMv_linear_solve.py

   # New file (keyword):
   A_field = qd.field(dtype=qd.f32, shape=(N_ENVS, N, N))
   ```

6. **Section-separator comments** (`# ----------- ... -----------`). Not used anywhere else in `misc/`. The library source files (e.g., `_tile16.py`) don't use them either.

## Severity

None of these are bugs. Items 1, 2, 5, and 6 are minor style deltas that could be normalized for consistency. Items 3 and 4 are arguably improvements over the existing convention â€” if adopted here, they could be adopted repo-wide later, but for now they make this file the odd one out.

</details>

<details>
<summary>5. Are there any coding inconsistencies in this code relative to itself and to other existing repo code?</summary>

# PR #483 â€” Coding Inconsistencies Review

## Internal Consistency (within the new file)

The demo file `misc/demos/cholesky_blocked.py` is internally consistent. Naming conventions, spacing, and structure are uniform across all three kernels.

## Inconsistencies with Existing Repo Code

### 1. `np.random.default_rng` vs `np.random.RandomState`

The demo uses `np.random.default_rng(42)` in `make_spd_matrices()`. Every other file in the repo that generates random numbers uses `np.random.RandomState`:

- `test_tile16.py` uses `np.random.RandomState` in all 9+ call sites (`_make_spd`, test helpers)
- No other file in the repo uses `default_rng`

`default_rng` is the modern NumPy API and arguably better, but it's inconsistent with the established codebase convention. Low severity â€” fine either way, but worth noting.

### 2. `qd.cuda` vs `qd.gpu` for arch init

The demo hardcodes `qd.init(arch=qd.cuda)`. Existing patterns are split:

- `qd.cuda`: `misc/minimal_timed.py`, `misc/benchmark_scatter_bls.py`, `misc/test_gpu_async.py`, `misc/test_memory_pool.py`
- `qd.gpu`: `misc/benchmark_reduction.py`, `misc/benchmark_bls.py`
- `test_tile16.py` uses `qd.gpu` exclusively

`qd.cuda` limits the demo to NVIDIA. Given the demo is specifically benchmarked on an RTX PRO 6000 and is a showcase for Tile16x16, hardcoding `qd.cuda` is defensible. Not a real issue, just a factual note.

### 3. `qd.field()` call syntax: keyword vs positional

The demo uses keyword-style `qd.field(dtype=qd.f32, shape=(...))`. Most existing `misc/` files use positional style:

```python
# Existing misc/ pattern (positional):
a = qd.field(qd.f32, shape=N)
b = qd.field(qd.i32, shape=())
x, y = qd.field(qd.f32), qd.field(qd.f32)

# Demo (keyword):
A_field = qd.field(dtype=qd.f32, shape=(N_ENVS, N, N))
```

The keyword style is clearer for the 3D shapes used here, and `test_tile16.py` also uses keyword form for some fields (e.g. `eps_field = qd.field(dtype=qd.f32, shape=())`). Not a real problem.

### 4. `qd.loop_config(name=...)` â€” new pattern unique to this file

The demo passes `name=` to `qd.loop_config`:

```python
qd.loop_config(name="chol_baseline", block_dim=64)
```

This is a valid API (tested in `test_loop_config_name.py`), but no other `misc/` file or `test_tile16.py` uses it. It's appropriate here for profiling/benchmarking purposes â€” no issue.

### 5. `if __name__ == "__main__"` guard â€” unusual for misc/

The demo wraps execution in `main()` behind an `if __name__ == "__main__"` guard. Only 4 of the ~26 `misc/` files do this (and those are utility scripts like `copyright.py`, `make_changelog.py`). The benchmark scripts (`benchmark_reduction.py`, `benchmark_tensor_access.py`, etc.) all run at module level without a guard.

This is actually *better* practice for a demo, and matches the fact that this lives in the new `misc/demos/` subdirectory rather than `misc/` root. Consistent with being a more polished file.

### 6. Module-level `qd.init()` and field allocation before `main()`

The demo calls `qd.init(arch=qd.cuda)` and allocates all four fields at module scope (lines 42â€“47), then does the actual work in `main()`. This is consistent with the pattern in the existing misc/ benchmarks (e.g. `benchmark_reduction.py` does all init and field allocation at module level). No issue.

## Summary

No significant coding inconsistencies. The file is well-structured and follows repo conventions. The two minor deviations from existing patterns are:

| Item | Severity | Verdict |
|------|----------|---------|
| `default_rng` vs `RandomState` | Trivial | Fine to keep (modern API), but inconsistent with the rest of the codebase |
| `if __name__` guard | Trivial | Actually better practice; appropriate for a demo file |

Everything else â€” kernel structure, API usage (`Tile16x16.zeros`, `Tile16x16.eye`, `cholesky_`, `solve_triangular_`, `qd.outer`, slice syntax), field/loop patterns, naming conventions â€” is consistent with `_tile16.py` and `test_tile16.py`.

</details>

<details>
<summary>6. Is there anything in this PR that could be implemented in less code?</summary>

## Question 6: Could anything be implemented in less code?

The kernels themselves are already tightly written â€” their verbosity is inherent to the algorithm. The main savings are in the benchmark/verification harness.

### 1. `main()` has a repetitive compile-verify-benchmark pattern (lines 320â€“348)

Each kernel goes through the same three steps: print â†’ call â†’ sync â†’ verify â†’ benchmark. This is repeated three times with only the name/function/field/thread-count varying. A list-of-tuples and a loop would cut ~15 lines to ~8:

```python
kernels = [
    ("baseline (scalar Crout, shared mem)", 64, cholesky_baseline, L_baseline_field),
    ("blocked  (scalar Crout, shared mem)", 16, cholesky_blocked, L_blocked_field),
    ("tile16   (Tile16x16, no shared memory)", 16, cholesky_tile16, L_tile16_field),
]

for name, threads, fn, field in kernels:
    print(f"Compiling {name}...")
    fn()
    qd.sync()
    verify(name, field, A_np)

# ... then benchmark in the same loop or a second pass
```

### 2. `verify()` loop can be replaced with vectorized numpy (lines 300â€“308)

The per-environment Python loop can be a single batched operation:

```python
def verify(name, L_field, A_np):
    n_checked = min(8, N_ENVS)
    L_np = np.tril(L_field.to_numpy()[:n_checked])
    recon = L_np @ L_np.transpose(0, 2, 1)
    max_err = np.max(np.abs(recon - A_np[:n_checked]))
    print(f"  {name} max reconstruction error (first {n_checked} envs): {max_err:.2e}")
    return max_err
```

This drops the 10-line loop+nan-check to 4 lines. (The nan/inf check is arguably unnecessary anyway â€” `np.max` on an array with nans returns nan, which would be visible in the printed output.)

### 3. Minor: four separate `qd.field` declarations could be a dict (lines 70â€“73)

The four field declarations differ only in name. A dict comprehension keeps them together and makes it easy to add/remove kernels:

```python
fields = {name: qd.field(dtype=qd.f32, shape=(N_ENVS, N, N))
          for name in ["A", "L_baseline", "L_blocked", "L_tile16"]}
```

### Not worth changing

The three kernels themselves (`cholesky_baseline`, `cholesky_blocked`, `cholesky_tile16`) look verbose but each implements a genuinely different algorithm. The "subtract prior-block contributions" loops in `cholesky_blocked` (lines 157â€“166 vs 191â€“199) look similar but operate on different index patterns (diagonal vs off-diagonal blocks), so extracting a helper would add indirection without real savings. Similarly, the tile16 kernel's two `outer` subtraction loops (lines 242â€“246 vs 262â€“267) differ in whether one or two distinct vectors are loaded. These are deliberate structural parallels, not copy-paste bloat.

**Bottom line**: The kernels are fine. The harness (`main`, `verify`, field declarations) has ~20 lines of unnecessary repetition that a loop and vectorized numpy would eliminate.

</details>

<details>
<summary>7. Is there anything in this PR that could be implemented in a way that will run faster at runtime?</summary>

# Q7: Runtime performance opportunities

Focusing on `cholesky_tile16` (Kernel 3) since it's the showcased kernel.

## 1. Redundant `v_diag` loads from global memory (significant)

In the off-diagonal update loop, `v_diag` is loaded from global memory once per `(ib, jb, t)` triple, but it only depends on `(jb, t)` â€” it's invariant across `ib`:

```python
for ib in range(kb + 1, N_BLOCKS):
    ...
    for jb in range(kb):
        j0 = jb * TILE
        for t in range(TILE):
            v_own = L_tile16_field[env, i0:i1, j0 + t]
            v_diag = L_tile16_field[env, k0:k1, j0 + t]  # same for all ib!
            L_ik -= qd.outer(v_own, v_diag)
```

Moreover, the exact same vectors were already loaded during the diagonal block update a few lines earlier (line 245: `v = L_tile16_field[env, k0:k1, j0 + t]`).

With N_BLOCKS=6, I count ~560 total `v_diag` vector loads across all `kb` iterations where only ~240 are unique â€” a 2.3x redundancy. The worst case is at `kb=1`: the same 16 vectors are loaded 5 times (once per off-diagonal block) instead of once.

**Fix**: Restructure so that all off-diagonal tiles are live simultaneously and the `(jb, t)` loop is outermost:

```python
# Pre-load all off-diagonal blocks from A
L_ik_tiles = [qd.simt.Tile16x16.zeros(dtype=qd.f32) for _ in range(kb + 1, N_BLOCKS)]
for idx, ib in enumerate(range(kb + 1, N_BLOCKS)):
    i0, i1 = ib * TILE, qd.min(ib * TILE + TILE, N)
    L_ik_tiles[idx][:] = A_field[env, i0:i1, k0:k1]

# Single pass over prior blocks â€” one v_diag load serves all tiles
for jb in range(kb):
    j0 = jb * TILE
    for t in range(TILE):
        v_diag = L_tile16_field[env, k0:k1, j0 + t]  # ONE load
        L_kk -= qd.outer(v_diag, v_diag)
        for idx, ib in enumerate(range(kb + 1, N_BLOCKS)):
            i0, i1 = ib * TILE, qd.min(ib * TILE + TILE, N)
            v_own = L_tile16_field[env, i0:i1, j0 + t]
            L_ik_tiles[idx] -= qd.outer(v_own, v_diag)
```

Register cost: at most N_BLOCKS=6 tiles live simultaneously (when `kb=0`, though that iteration has no prior-block work). Each Tile16x16 is 16 registers per thread, so 96 registers total â€” well within budget on Blackwell.

This also eliminates the separate diagonal-update loop (lines 242â€“246), since its work is folded into the unified pass.

## 2. Uncoalesced column vector loads

The field layout is `(N_ENVS, N, N)` in row-major order. Column vector loads like `L_tile16_field[env, k0:k1, j0 + t]` read elements strided by `N` (=92) in the innermost dimension. Each thread loads `field[env, k0 + tid, j0 + t]` â€” adjacent threads access addresses 92Ã—4 = 368 bytes apart, which is uncoalesced and requires multiple memory transactions.

This is inherent to doing rank-1 column-wise updates against a row-major matrix. If the framework supports it, storing the L factor transposed (or maintaining a separate column-major copy of the prior blocks) would make these loads coalesced. Whether this is worth the extra store is workload-dependent, but for 6 block-columns and up to 5 prior blocks each, the volume of strided loads is substantial.

## 3. Store diagonal tile earlier (minor)

The diagonal result is stored after all off-diagonal blocks are processed (line 276), but it could be stored immediately after `cholesky_()` (line 249). The off-diagonal processing uses `L_kk` from registers (via `solve_triangular_`), not from global memory, so this reordering is safe. Storing earlier gives the write more time to reach L2 cache before the *next* `kb` iteration reads those same addresses in the prior-block loops.

</details>

<details>
<summary>8. Is there anything in this PR which might exhibit incorrect behavior in some way?</summary>

# PR #483 Review â€” Question 8: Incorrect behavior

## 1. Stale docstring: "64x64" and "4x4 grid"

The module docstring says:

```27:28:misc/demos/cholesky_blocked.py
"""Benchmark 64x64 blocked Cholesky factorization using Tile16x16.
```

and describes a "4x4 grid of 16x16 tiles." But `N = 92`, giving `N_BLOCKS = 6` â€” a 6x6 grid of tiles and a 92x92 factorization. These descriptions are stale from an earlier version of the demo and are factually wrong for the code as shipped. The same docstring also claims "64 sequential syncs" for the baseline, but the baseline does `1 + 2*N = 185` syncs.

Not a runtime bug, but anyone reading the docstring to understand the benchmark will be misled.

## 2. Uninitialized SharedArray padding in `cholesky_blocked`

When `N % TILE != 0` (here N=92, TILE=16, last block has 12 valid rows out of 16), the blocked kernel allocates `SharedArray((N_PADDED, N_PADDED + 1), qd.f32)` (96 x 97) but only loads rows 0..91:

```146:151:misc/demos/cholesky_blocked.py
        for row in range(N):
            c = tid
            while c <= row:
                H[row, c] = A_field[env, row, c]
                c += TILE
        qd.simt.block.sync()
```

Rows 92â€“95 are never initialized, yet the Crout factorization and triangular solve process all 16 rows of the last block unconditionally (the `for r in range(TILE)` loops on lines 157 and 191 have no guard for `k0 + r < N` or `i0 + r < N`). The write-back only stores valid rows (line 212: `for row in range(N)`), so the output is correct â€” garbage in the padding rows never reaches global memory.

This works because the Crout factorization processes columns left-to-right: valid columns 0â€“11 only depend on earlier valid columns, so garbage in columns 12â€“15 cannot contaminate them. However, if shared memory happens to contain NaN or Inf from a prior kernel's use of the same shared memory, those values will propagate through the padding-row computations. While they won't corrupt the output, they waste cycles and could theoretically slow down the GPU (some architectures handle NaN/denormal math at reduced throughput). More importantly, the correctness argument is subtle and fragile â€” a future edit that changes the write-back to iterate over `N_PADDED` instead of `N` would silently produce garbage output.

## 3. No incorrect behavior found in the core algorithms

After careful analysis of all three kernels, the blocked Cholesky algorithm (right-looking: prior-block subtraction, diagonal factorization, off-diagonal triangular solve) is correctly implemented in each variant:

- **Baseline**: Scalar Crout with correct sync placement â€” sync after thread 0 writes the diagonal, sync after all threads write off-diagonal entries. The stride-by-64 parallelization over off-diagonal rows is correct.

- **Blocked**: The prior-block subtraction correctly computes the Schur complement; each thread handles a single column (`c = tid`) and iterates over rows serially, so no write conflicts exist within the subtraction loop. The per-column sync inside the Crout factorization and triangular solve correctly serializes column dependencies.

- **Tile16**: The "same thread reads its own prior store" property holds â€” thread `i` in the subgroup owns row `i` of each tile, so when it later reads `L_tile16_field[env, k0+i, j0+t]`, it reads back a value it wrote in a prior iteration. Per-thread sequential consistency on the GPU guarantees visibility without explicit sync. The eye-initialization for partial diagonal blocks and zero-initialization for partial off-diagonal blocks correctly pads the tiles so the Cholesky factor is `[L 0; 0 I]` for the diagonal and `[X; 0]` for off-diagonal blocks.

The store ordering in the tile16 kernel (off-diagonal blocks stored before the diagonal block) is also safe: the diagonal block `L_{kk}` is only consumed within its own iteration via the register-resident `L_kk` variable (for `solve_triangular_`), and is never read back from global memory by subsequent iterations.

</details>

<details>
<summary>9. Is there anything in this PR that might reduce runtime speed?</summary>

## Question 9: Is there anything in this PR that might reduce runtime speed?

**No.** This PR does not modify any existing library code. The changes are:

1. **A new standalone demo/benchmark script** (`misc/demos/cholesky_blocked.py`) â€” not imported by the library, not on any code path that runs unless a user explicitly executes it.
2. **Documentation additions** (`docs/source/user_guide/tile16.md`) â€” a new section with a results table and a link to the demo. No runtime impact.

Neither change touches the `quadrants` package itself, so there is zero risk of regressing runtime performance of existing functionality.

### Performance notes on the demo itself

The demo is a benchmark, not library code, so its performance doesn't affect users. That said, the kernels are cleanly written with no obvious anti-patterns:

- The Tile16x16 kernel avoids shared memory and block syncs entirely, which is the point of the demo.
- The baseline and blocked kernels use `block.sync()` where structurally required; there are no gratuitous syncs.
- The benchmark loop correctly separates warmup from measured iterations and calls `qd.sync()` before timing, so results are reliable.
- One minor note: the `verify` function copies full fields to NumPy and loops in Python, but this runs outside the timed region and only checks 8 environments, so it doesn't affect reported numbers.

</details>

<details>
<summary>10. Are there any public methods, classes, or similar added that are not listed in the doc?</summary>

## Q10: Are there any public methods, classes, or similar added that are not listed in the doc?

**No.** The doc coverage is comprehensive. Every public API item is documented.

PR #483 itself adds no new public API â€” it only adds a doc section ("Example: blocked Cholesky") and a demo file (`misc/demos/cholesky_blocked.py`).

Checking the full feature surface (across the PR stack) against `tile16.md`:

| Public API | Documented? |
|---|---|
| `qd.simt.Tile16x16.zeros(dtype=...)` | Yes â€” "Creating a tile" section + method reference |
| `qd.simt.Tile16x16.eye(dtype=...)` | Yes â€” "Creating a tile" section + method reference |
| `qd.simt.Tile16x16.SIZE` | Yes â€” method reference |
| `t.eye_()` | Yes â€” "Identity initialization" section + method reference |
| `t.cholesky_(eps)` | Yes â€” "Cholesky factorization" section + method reference |
| `L.solve_triangular_(B)` | Yes â€” "Triangular solve" section + method reference |
| `t[:] = arr[r0:r1, c0:c1]` (2D load) | Yes â€” "Loading and storing" section + method reference |
| `t[:] = arr[i, r0:r1, c0:c1]` (3D load) | Yes |
| `arr[r0:r1, c0:c1] = t` (2D/3D store) | Yes |
| `t -= qd.outer(a, b)` | Yes â€” "Rank-1 updates" section + method reference |
| `qd.outer(a, b)` (top-level function) | Yes â€” described in "Rank-1 updates" |
| `v = arr[r0:r1, col]` (column vector load) | Yes â€” "Loading column vectors for outer products" |
| `v = arr[batch, r0:r1, col]` (3D vec load) | Yes |

Internal/private items (`_load`, `_store`, `_ger_sub`, `_trsm`, `_TileSliceProxy`, `_VecSliceProxy`, `_TileRefProxy`, `_OuterProduct`, `try_tile_ref`, `try_tile_slice`, etc.) are all underscore-prefixed and correctly omitted from the doc.

</details>

<details>
<summary>11. Check the test coverage of the tests. Is there any code that is being added or modified in this PR, that is not being exercised by the tests? Is there any functionality that is implied by consistency or otherwise that is not being tested by these tests?</summary>

# Question 11: Test coverage analysis

The demo (`misc/demos/cholesky_blocked.py`) is not itself library code, so it doesn't strictly need unit tests. However, it exercises several Tile16x16 usage patterns that are **not covered** by the existing test suite in `tests/python/test_tile16.py`. If any of these patterns regressed, the tests would not catch it.

## What IS covered

The individual Tile16x16 primitives used by the demo are all unit-tested in isolation:

- `Tile16x16.zeros()`, `Tile16x16.eye()`, load/store (2D and 3D), `cholesky_()`, `solve_triangular_()`, `t -= qd.outer(a, b)`, vec proxy from 2D/3D arrays, partial row/col clamping, same-variable `qd.outer(v, v)`, multi-column accumulate loop, SharedArray interop.

## What is NOT covered

### 1. Multi-tile blocked composition (the whole point of the demo)

No test performs a blocked factorization on a matrix larger than 16x16. The demo chains `cholesky_()` â†’ `solve_triangular_()` â†’ `outer()` across a 4x4 grid of tile blocks in nested loops (`cholesky_blocked.py` lines 204-250). An integration test that factors e.g. a 32x32 or 48x48 matrix using multiple Tile16x16 blocks and verifies the result would be the single most impactful addition.

### 2. Store-then-load from the same field within one kernel

The demo's `cholesky_tile16` kernel stores tiles to `L_tile16_field` (line 247, 250) and reads them back in later loop iterations (lines 219, 239-240). This self-read-after-self-write through global memory is central to the blocked algorithm's correctness. No test ever writes to a field and reads it back within the same kernel invocation â€” tests always use separate src/dst arrays.

### 3. Asymmetric `qd.outer()` with two 3D vec proxies from different row ranges

At line 241: `L_ik -= qd.outer(v_own, v_diag)` where `v_own` reads rows `[i0:i1]` and `v_diag` reads rows `[k0:k1]` from the same 3D field. The existing test `test_tile16_vec_proxy_ger_sub_3d` (line 700) uses two 3D vec proxies with the **same** row range but different columns (`vecs_arr[1, 0:_TILE, 0]` and `vecs_arr[1, 0:_TILE, 1]`). No test uses two 3D vec proxies with **different** row ranges.

### 4. `eye()` initialization followed by partial-row overwrite

At lines 209-210: `L_kk = qd.simt.Tile16x16.eye(...)` then `L_kk[:] = A_field[env, k0:k1, k0:k1]` where `k1 - k0 < 16` for the last block (N=92, so block 5 has rows 80-91 = 12 rows). The identity in rows 12-15 must survive the partial load. `test_tile16_eye` (line 61) tests `eye` in isolation, and partial-row loads are tested separately, but the **combination** (eye then partial overwrite, verifying untouched rows remain identity) is not tested.

### 5. Simultaneous partial rows AND partial cols on a 3D load/store

For the last diagonal block (k0=80, k1=92), the load at line 210 reads a 12Ã—12 sub-block from a 3D field. Tests exercise partial rows (e.g. `test_tile16_load_store_partial_rows`) or partial cols in 2D, and 3D full-tile loads, but no test does a 3D load where **both** row and column ranges are shorter than 16 simultaneously.

## Recommendation

The most valuable single addition would be a test that performs a 2-block (32x32) or 3-block (48x48) blocked Cholesky using the same pattern as the demo's `cholesky_tile16` kernel: multiple tiles composed through global-memory store/load, with `cholesky_()` + `solve_triangular_()` + `outer()` chained across blocks. This would cover gaps 1, 2, and 3 simultaneously and verify the end-to-end algorithm that the demo showcases.

</details>

<details>
<summary>12. Check the factorization of the tests. Are there some tests that could be merged, e.g. making use of parametrize? Reminder: you can use if qd.static(...) in quadrants kernels to switch between different compiled code.</summary>

# Review Q12: Test factorization opportunities

## 1. Negative-index error tests (load & store)

Four separate tests each define a near-identical kernel that differs only in which slice dimension gets the `-1`:

- `test_tile16_load_negative_row_raises` (line 778) â€” `s[-1:_TILE, 0:_TILE]`
- `test_tile16_load_negative_col_raises` (line 796) â€” `s[0:_TILE, -1:_TILE]`
- `test_tile16_store_negative_row_raises` (line 1621) â€” `d[-1:_TILE, 0:_TILE] = t`
- `test_tile16_store_negative_col_raises` (line 1638) â€” `d[0:_TILE, -1:_TILE] = t`

These could be collapsed into two tests (one for load, one for store) parametrized over `"row"/"col"`, using `if qd.static(dim == "row")` to select which slice is negative inside the kernel.

## 2. Missing-bound error tests (load, store, vec)

Five tests follow the same pattern â€” identical boilerplate differing only in which slice bound is omitted:

- `test_tile16_load_missing_start_raises` (line 814) vs `test_tile16_load_missing_stop_raises` (line 831)
- `test_tile16_store_missing_stop_raises` (line 849) vs `test_tile16_store_missing_start_raises` (line 1657)
- `test_tile16_vec_slice_missing_stop_raises` (line 996) vs `test_tile16_vec_slice_missing_start_raises` (line 1016)

Each pair could become a single function parametrized over `"start"/"stop"`, using `qd.static()` to select the omitted bound. That reduces five functions to three.

## 3. 2D vs 3D row-clamping tests

`test_tile16_clamp_to_array_rows` (line 147) and `test_tile16_3d_clamp_to_array_rows` (line 183) are nearly identical. The only differences are:

- Tensor shape: `(src_rows, _TILE)` vs `(1, src_rows, _TILE)`
- Kernel calls: `_load`/`_store` vs `_load3d`/`_store3d`
- Assertion indexing: `result[:NROWS, :]` vs `result[0, :NROWS, :]`

These could merge into one test with a parametrized `ndim` (2 or 3) and `if qd.static(ndim == 3)` inside the kernel to call the 3D variants.

## 4. SharedArray clamp-store vs clamp-load

`test_tile16_shared_array_clamp_store` (line 1346) and `test_tile16_shared_array_clamp_load` (line 1373) both test auto-clamping when the SharedArray is narrower than 16 columns. One tests the store direction, the other the load direction. They share the same structure and could be merged with a `clamp_side` parameter (mirroring the existing pattern in `test_tile16_clamp_to_array_rows`), using `qd.static()` to select which SharedArray dimension is narrow.

## 5. `test_tile16_f64_roundtrip_into_f32_array` has two inline kernels

At line 1583, this test defines two nearly identical kernels `roundtrip_f32` and `roundtrip_f64` that differ only in the tile dtype argument to `Tile16x16.zeros()`. These could be a single kernel parametrized over tile dtype, using `if qd.static(tile_dtype == qd.f64)` to select, or simply restructured as a `@pytest.mark.parametrize("tile_dtype", [qd.f32, qd.f64])` test with a single kernel that uses `qd.static()`.

## Summary

| Opportunity | Tests involved | Saves |
|---|---|---|
| Negative-index errors | 4 â†’ 2 | 2 functions |
| Missing-bound errors | 5 â†’ 3 | 2 functions |
| 2D/3D row clamping | 2 â†’ 1 | 1 function |
| SharedArray clamp direction | 2 â†’ 1 | 1 function |
| f64 roundtrip inline kernels | 2 kernels â†’ 1 | internal dedup |

Total: roughly 6 fewer test functions, with no loss of coverage, all leveraging `qd.static()` for compile-time branching inside the merged kernels.

</details>

<details>
<summary>13. Are there any type: ignore at the head of new or modified files that could be removed?</summary>

## Q13: Are there any `type: ignore` at the head of new or modified files that could be removed?

**Yes â€” one file has a blanket `# type: ignore` that deserves scrutiny.**

### `python/quadrants/lang/ast/ast_transformer.py` â€” line 1

```
# type: ignore
```

This file is modified by the broader feature branch (new imports added from `quadrants.lang.util`). The blanket `# type: ignore` at line 1 suppresses **all** type errors for the entire file. This is a pre-existing suppression, not introduced by this PR, but since the file is being actively modified it's worth asking whether it's still needed or could be narrowed to specific lines.

### `python/quadrants/lang/simt/_tile16.py` â€” line 1

```
# pyright: reportInvalidTypeForm=false
```

This is a **targeted** Pyright suppression (not a blanket `type: ignore`) and is justified: the file uses `TYPE_CHECKING`-gated protocol stubs with type forms that Pyright rejects. This one is fine to keep.

### Other files

The remaining modified files (`tile_slicing.py`, `__init__.py`, `impl.py`, `util.py`, `test_tile16.py`, `tile16.md`, `cholesky_blocked.py`) have **no** `type: ignore` or type-suppression comments at their heads.

### Recommendation

The `# type: ignore` in `ast_transformer.py` is the only candidate for removal or narrowing. It predates this PR, so fixing it is out of scope for this change, but it would be worth filing a follow-up issue to audit whether the blanket suppression is still necessary.

</details>

<details>
<summary>14. Are there any local imports we can move to the top of their module?</summary>

# Q14: Local imports that could be moved to module level

## Summary

Across the 7 Python files checked, there are **19 local imports** (imports inside functions/methods). Most are intentionally local due to circular import risks or optional dependencies. Two can safely be moved to the top of their module.

## Can be moved to module level

1. **`python/quadrants/lang/util.py` line 371** â€” `import types` inside `is_from_quadrants_module()`. This is a standard library module with zero circular-import risk. Every other stdlib import in this file (`functools`, `os`, `traceback`, `warnings`) is already at the top.

2. **`python/quadrants/lang/simt/_tile16.py` lines 530â€“532** â€” `from quadrants.lang.exception import QuadrantsSyntaxError` inside `_Tile16x16Proxy._resolve()`. The `exception` module does not depend on `_tile16`, so there is no circular-import risk. (The companion `from quadrants.lang import impl` on line 529 *must* stay local â€” see below â€” but the exception import can be separated out.)

## Borderline â€” could be moved but low value

3. **`tests/python/test_tile16.py` line 1758** â€” `from quadrants.lang import impl` inside `test_proxy_default_dtype_survives_reinit()`. Test files have no circular-import concerns, and `impl` is already indirectly available via `qd`. Moving it to the top would be fine but is low-stakes in a test file.

## Correctly kept local (should NOT move)

| File | Line(s) | Import | Reason |
|------|---------|--------|--------|
| `_tile16.py` | 529 | `from quadrants.lang import impl` | Circular: `impl.py` â†’ `tile_slicing` â†’ `_tile16` |
| `impl.py` | 877, 917 | `from . import _py_tensor` | Guarded by `is_python_backend()`; depends on optional PyTorch |
| `impl.py` | 878, 918 | `from .util import dtype_to_torch_dtype` | Same â€” only needed in PyTorch backend path |
| `ast_transformer.py` | 624 | `from quadrants import math as qd_math` | Circular: `quadrants.math` depends on `quadrants.lang` |
| `ast_transformer.py` | 664â€“666 | `from quadrants.lang import matrix_ops` | Circular: `matrix_ops` depends on other `lang` modules |
| `ast_transformer.py` | 699 | `from quadrants.lang.matrix_ops import matmul` | Same circular risk as above |
| `util.py` | 44, 136, 216, 283 | `import torch` | Optional dependency; guarded by availability checks |
| `util.py` | 53 | `from distutils.spawn import find_executable` | Deprecated module; lazy import avoids warnings |
| `util.py` | 359 | `import quadrants as _qd_pkg` | Circular: `util.py` is inside the `quadrants` package |

## Verdict

Items 1 and 2 above are safe, low-risk moves. Neither is introduced by this PR (they are pre-existing), so they're nits rather than blockers.

</details>

<details>
<summary>15. Check the diff between this branch and origin/main: are there any comments that have been deleted or modified?</summary>

# PR #483 â€” Deleted or modified comments

All changes are in `python/quadrants/lang/simt/_tile16.py` unless otherwise noted.

## Deleted comments

1. **`python/quadrants/lang/ast/ast_transformer.py`** â€” Inline comment `# ignore this built-in module` deleted. The line containing it was replaced with `violation = False` (the comment's behavior was preserved, the explanatory comment was dropped):

   ```
   -                        # ignore this built-in module
   +                        violation = False
   +                    if violation and _is_from_quadrants_module(node.value.ptr):
                            violation = False
   ```

2. **`_tile16.py`** â€” A 3-line standalone comment block was deleted from its original location and moved (reformatted to 2 lines) ~50 lines later in the same function:

   ```
   -    # StructType.__call__ already defaults missing args to 0, so Tile()
   -    # produces a zero-initialized tile without needing default values in the
   -    # class definition (which @qd.dataclass doesn't support).
   ```

   Replaced by a condensed 2-line version at a new position (after the class body now includes more methods):

   ```
   +    # StructType.__call__ already defaults missing args to 0, so Tile() produces a zero-initialized tile
   +    # without needing default values in the class definition (which @qd.dataclass doesn't support).
   ```

## Modified comments (docstrings)

All in `_tile16.py`. These are rewording/reformatting changes, not deletions of meaning:

3. **Class docstring** reformatted from multi-line to two-line (no content change):
   - `"""A 16x16 tile distributed one row per subgroup thread...` â€” newline between sentences removed.

4. **`_load` / `_load3d` / `_store` / `_store3d` docstrings** â€” `row_end`/`col_end` renamed to `row_stop`/`col_stop` throughout (matching the parameter rename). Formatting also condensed from multi-line to fewer lines.

5. **`eye_` docstring** (previously `_eye_`) â€” condensed from 3-line to 2-line format. No content change.

6. **`cholesky_` docstring** â€” reformatted from 2 separate lines to a wrapped 2-line paragraph. No content change.

7. **`_trsm` docstring** â€” same reformatting (2 separate lines merged into wrapped paragraph). No content change.

8. **`solve_triangular_` docstring** â€” **content expanded**: added "and non-singular (all diagonal elements non-zero). Passing a singular matrix causes division by zero, producing inf/NaN without warning." This is a meaningful addition, not just reformatting.

## Added type-suppression comments

9. Two `# pyright: ignore[reportReturnType]` comments added on the `return` statements in `_get_tile16x16_class()`. These are new â€” the original code had no suppression comments on those lines.

## Summary

- **1 comment fully deleted** (the `# ignore this built-in module` inline comment in `ast_transformer.py`)
- **1 comment block relocated and reformatted** (the `StructType.__call__` explanation in `_tile16.py`)
- **6 docstrings reformatted** (condensed line wrapping, `row_end`â†’`row_stop` renames)
- **1 docstring substantively expanded** (`solve_triangular_` now documents singular-matrix behavior)
- **2 type-suppression comments added**

No comments were deleted without replacement or reason. The deletions are either intentional cleanup (removing a now-redundant comment) or reformatting.

</details>

<details>
<summary>16. Check the diff between this branch and origin/main: are there any new/modified comments that are being wrapped at 80 columns, instead of 120 columns?</summary>

# Q16: Comments wrapped at 80 columns instead of 120?

**Yes â€” `misc/demos/cholesky_blocked.py` wraps comments and docstrings at ~80 columns.**

The rest of the PR (Python source in `_tile16.py`, `tile_slicing.py`, etc. and the markdown doc `tile16.md`) uses lines well beyond 80 â€” docstrings wrap near 110â€“120 chars, and `tile16.md` has lines up to 321 chars. So those files are fine.

The demo file is the outlier:

## Module docstring (lines 6â€“15)

The numbered-list entries in the docstring wrap at 72â€“79 characters:

```
1. Baseline: scalar Cholesky-Crout, 64 threads, shared memory, 64 sequential   â† 76 chars
   syncs. Thread 0 computes each diagonal, remaining threads parallelize        â† 72 chars
   off-diagonal updates.

2. Blocked: 4x4 grid of 16x16 tiles, 16 threads, shared memory, scalar Crout   â† 76 chars
   for diagonal blocks. Same blocking structure as Tile16x16 but all data lives â† 79 chars
   in shared memory with block.sync() between every step.

3. Tile16x16: same blocked structure but fully register-resident via Tile16x16. â† 79 chars
   No shared memory, zero syncs. Prior tiles read from global memory (L2).      â† 74 chars
```

At 120 columns each numbered item would fit on two lines instead of three.

## Inline comment block (lines 212â€“215)

```python
            # Subtract rank-1 contributions from prior column-blocks.
            # Each vector load reads from global memory, but only data this       â† 75 chars
            # subgroup itself wrote in an earlier iteration (same thread reads     â† 78 chars
            # its own prior store), so no cross-thread sync is needed.            â† 70 chars
```

At 120 columns this three-line continuation would fit in two lines, e.g.:

```python
            # Each vector load reads from global memory, but only data this subgroup itself wrote in an earlier
            # iteration (same thread reads its own prior store), so no cross-thread sync is needed.
```

## Verdict

Only `misc/demos/cholesky_blocked.py` has this issue. The docstring and one multi-line inline comment block are wrapped at ~80 columns. Everything else in the PR follows the wider convention.

</details>

hughperkins:
Opus evaluation of above reveiw:

```
Evaluation of the review

  The review is thorough and well-structured. Here are my assessments and recommendations, grouped by priority.


  High-value findings to act on

  Stale docstring (Q8). The module docstring says "64x64" and "4x4 grid" but the code uses N=92 with 6 blocks. This is a clear factual error that should be fixed before merge.
  Easy fix.
  `"dex_hand dimensions"` reference (Q3). If dex_hand is an internal project, this leaks it. Worth genericizing to something like "robotics-scale matrices".
  80-column wrapping (Q16). The docstring and one inline comment block in cholesky_blocked.py wrap at ~80 columns instead of the repo's 120-column convention. Easy to reflow.


  Medium-value findings worth discussing

  No integration test for blocked Cholesky (Q1, Q11). Multiple questions converge on the same gap: the demo's multi-tile composition (store-then-reload, chained
  cholesky+trsm+outer across blocks) isn't exercised by the test suite. A small 32x32 blocked Cholesky test would close several gaps at once. However, since this PR only adds a
  demo file (not library code), this is a "nice to have" rather than a blocker.
  Redundant `v_diag` loads in `cholesky_tile16` (Q7). The same vectors are loaded from global memory multiple times across off-diagonal blocks. The suggested restructuring
  (hoist v_diag loads outside the ib loop) would reduce global memory traffic by ~57% for this kernel. Since this is a demo/benchmark, the performance win would strengthen the
  showcase.


  Low-value findings — skip or defer

  Compile-verify boilerplate in `main()` (Q2, Q6). Three agents flagged that the compile/sync/verify sequence repeats 3 times. True, but it's a 326-line demo script, not library
   code. A loop would save ~10 lines. Not worth blocking a PR.
  `default_rng` vs `RandomState` (Q5). The demo uses the modern NumPy API while the rest of the repo uses RandomState. This is a positive deviation — no reason to regress to the
   legacy API for consistency.
  Layout deltas: shebang, `__main__` guard, section separators (Q4). The demo file is in a new misc/demos/ subdirectory and intentionally more polished. These are style
  improvements, not problems.
  Test factorization (Q12). Six test functions could be merged via parametrize + qd.static(). Valid but out of scope for this PR (the tests aren't changed in this diff).
  Local imports, type: ignore (Q13, Q14). Pre-existing nits in files not touched by this PR. Worth a separate cleanup PR, not this one.
  Deleted/modified comments (Q15). All changes are intentional reformatting or improvements. No information was lost. No action needed.


  Answers that were not useful

  Q9 (runtime speed reduction) correctly concluded "no risk" but that was obvious from the diff — the PR only adds files. The answer padded itself with performance observations
  about the demo. Not harmful, but zero signal.
  Q10 (undocumented public APIs) also correctly concluded "no gaps" but the PR adds no API. The table of all Tile16x16 API items is thorough but wasn't really needed to answer
  the question for this specific diff.


  Recommendation

  Three items to fix before merge:
  1. Fix the stale "64x64" / "4x4 grid" / "64 sequential syncs" in the docstring
  2. Replace "dex_hand dimensions" with something generic
  3. Reflow the 80-column-wrapped comments to 120

  One item to consider:
  4. Hoist v_diag loads outside the ib loop to reduce redundant global memory reads (strengthens the benchmark)

  Everything else is either out of scope for this PR or too minor to act on.
```

hughperkins:
Opus actions to address 'bad points' in the description:

<details>
<summary><b>User:</b> modify the demo to work on qd.gpu</summary>

**Assistant:**

Done. Changed `qd.cuda` to `qd.gpu` and pushed.

</details>

<details>
<summary><b>User:</b> add a unit test to qaudrants unit tests that smoke tests that cholesksy_blocked.md runs to completion without crashing â€¦</summary>

add a unit test to qaudrants unit tests that smoke tests that cholesksy_blocked.md runs to completion without crashing (exit code 0). This should run on qd.gpu

---

**Assistant:**

Let me find existing test patterns and see if there are any subprocess-based smoke tests already.

</details>

<details>
<summary><b>User:</b> sorry, cholesky_blocked.py</summary>

**Assistant:**

Let me check how the existing test file for tile16 is structured and see test_utils.test() usage:

Let me check how `test_utils.test()` works with GPU requirements:

Now let me check how `subprocess` is used in existing tests for a pattern to follow:

The demo uses module-level `qd.init()` and `qd.field()`, so it needs to run as a subprocess. Let me add a simple smoke test:

Now add the smoke test at the end of the file:

Done. Added `test_tile16_cholesky_blocked_demo` at the end of `test_tile16.py` â€” it runs `misc/demos/cholesky_blocked.py` as a subprocess, requires `qd.gpu`, and fails the test if the exit code is non-zero (with stderr in the failure message). 5-minute timeout to allow for compilation.

</details>

<details>
<summary><b>User:</b> run hte test, hceck it passes</summary>

**Assistant:**

Passes on both CUDA and Vulkan. 15 seconds total.

</details>

erizmr (approved):

PR #485, opened 2026-04-15
[AMDGPU] Enable HIP memory pool and surface pool-exhaustion errors.

### LINKED ISSUE #2434: [Bug]: invalid argument while calling memset (hipMemset) with >3000 num_envs on AMD
### Bug Description

First of all, thanks for your impressive work !
I'd glad to try the latest v0.4.0 on my computer but got some problem: 
RuntimeError: [amdgpu_driver.h:operator()@81] AMDGPU Error hipErrorInvalidValue: invalid argument while calling memset (hipMemset)

### Steps to Reproduce

The  trigger is pretty simple: 
```python
import torch
import genesis as gs

gs.init(backend=gs.gpu, performance_mode=False, logging_level="debug")

scene = gs.Scene(
    show_viewer   = False,
    rigid_options = gs.options.RigidOptions(
        dt                = 0.01,
    ),
)

plane = scene.add_entity(
    gs.morphs.Plane(),
)

franka = scene.add_entity(
    gs.morphs.MJCF(file='xml/franka_emika_panda/panda.xml'),
)

scene.build(n_envs=3000)

# control all the robots
franka.control_dofs_position(
    torch.tile(
        torch.tensor([0, 0, 0, -1.0, 0, 1.0, 0, 0.02, 0.02], device=gs.device), (3000, 1)
    ),
)
for i in range(1000):
    scene.step()
```
With num_envs = 2048, everything works fine and hit a great performance. But as I increase the num_envs, that problem trigger. 

But I am not sure whether this problem is triggered by my own environment setup. I've checked the merge log and found rocm 7.2.0 so I use my system rocm instead of running from the docker (which is rocm 6.4.1 I think). 

### Expected Behavior

Got impressive performance with 2048:
```base
[Genesis] [15:03:53] [INFO] Running at 1,029,874.19 FPS (502.87 FPS per env, 2048 envs).
[Genesis] [15:03:53] [INFO] Running at 1,029,742.79 FPS (502.80 FPS per env, 2048 envs).
```
Expecting more envs working. 

### Screenshots/Videos

_No response_

### Relevant log output

```shell
RuntimeError: [amdgpu_driver.h:operator()@81] AMDGPU Error hipErrorInvalidValue: invalid argument while calling memset (hipMemset)
> /home/luoshaoyinn/workspace/gs-soccer/.venv/lib/python3.12/site-packages/quadrants/lang/_ndarray.py(283)__init__()
    281         super().__init__()
    282         self.dtype = cook_dtype(dtype)
--> 283         self.arr = impl.get_runtime().prog.create_ndarray(
    284             self.dtype, arr_shape, layout=Layout.NULL, zero_fill=True, dbg_info=_ti_core.DebugInfo(get_traceback())
    285         )
```

Full log at: 
[traceback.txt](https://github.com/user-attachments/files/25408978/traceback.txt)

Environment managed by uv: 
[pyproject.toml.txt](https://github.com/user-attachments/files/25409066/pyproject.toml.txt)

### Environment

- OS: arch rolling
- GPU/CPU : RX7900xtx (gfx1100)
- GPU-driver version: amdgpu from pacman lastest
- ROCm 7.2 from pacman

### Release version or Commit ID

v0.4.0 from pypi 

### Additional Context

Upd:
The problem seems to be quadrant. I built wheel from the latest main on quadrant and test this script:
```python
import quadrants as qd
import numpy as np

# Initialize with AMDGPU backend
qd.init(arch=qd.amdgpu)

def test_allocation(size_gb):
    # Calculate number of float32 elements
    # 1 GB = 1024^3 bytes. float32 = 4 bytes.
    num_elements = int((size_gb * 1024**3) // 4)
    print(f"Attempting to allocate {size_gb} GB ({num_elements} elements)...")
    
    try:
        # Create a large field
        field = qd.field(dtype=qd.f32, shape=(num_elements,))
        # Trigger the memset/initialization (where your error occurred)
        field.fill(0) 
        print("Success!")
    except Exception as e:
        print(f"Failed at {size_gb} GB!")
        print(e)

# Test 2GB (should pass) and 4GB+ (likely to fail)
test_allocation(1)
test_allocation(1.000000005)
```

And got surprisingly:
```shell
[Quadrants] version 0.4.1, llvm 20.1.0, commit ea7d24f1, linux, python 3.12.12
[Quadrants] Starting on arch=amdgpu
Attempting to allocate 1 GB (268435456 elements)...
Success!
Attempting to allocate 1.000000005 GB (268435457 elements)...
[E 02/19/26 16:35:32.549 65280] [amdgpu_driver.h:operator()@81] AMDGPU Error hipErrorInvalidValue: invalid argument while calling memset (hipMemset)


Failed at 1.000000005 GB!
[amdgpu_driver.h:operator()@81] AMDGPU Error hipErrorInvalidValue: invalid argument while calling memset (hipMemset)
```

Memset for a array larger than 1GB cause hipMemset broken

duburcqa:
## Summary

Brings the AMDGPU backend in line with CUDA so that `device_memory_GB` is no longer required for allocations past the default 1 GiB preallocation, and converts the cryptic `hipErrorInvalidValue while calling memset` symptom from
[Genesis-Embodied-AI/Genesis#2434](https://github.com/Genesis-Embodied-AI/Genesis/issues/2434) into a clear pool-exhaustion error.

## Motivation

The reported bug looked like a 1 GiB size cliff in `hipMemset` / `hipMemsetD32`, but a direct HIP probe (`hipMalloc` + `hipMemset` past 1 GiB on a fresh block) shows the runtime itself is fine. The real cause is on the AMDGPU allocation path:

- `AmdgpuDevice::allocate_memory_runtime` always routes through the fixed-size preallocated pool, unlike CUDA which uses `cuMemAllocAsync` to grow on demand when the device supports memory pools.
- When the pool exhausts, the device-side `allocate_from_reserved_memory` fires `quadrants_assert_runtime`; the CUDA branch hard-halts via `__assertfail`, but AMDGPU has no equivalent. The kernel returns without writing `*result_buffer`, the host reads a stale pointer,
and the subsequent `hipMemset` trips `hipErrorInvalidValue` — masking pool exhaustion as a memset bug.

## Changes

- Wire `hipMallocAsync` and related APIs into the AMDGPU backend, matching the CUDA memory-pool path; the fixed-size `device_memory_GB` preallocation is skipped on modern ROCm devices and kept as a fallback.
- In `allocate_llvm_runtime_memory_jit` (AMDGPU + CUDA), surface pool exhaustion as a clear host-side `Out of pre-allocated memory...` error instead of letting a downstream memset fail on a stale pointer.
- Document the ROCm 5.2+ requirement in the README.
- Add a regression test that allocates an ndarray past the default 1 GiB preallocation across all backends.

## Good points

- AMDGPU users no longer need to tune `device_memory_GB` by hand; workloads that outgrow the old 1 GiB default now just work. Directly addresses Genesis-Embodied-AI/Genesis#2434.
- The AMDGPU backend is now structurally symmetric with CUDA (same `use_memory_pool` branching in `allocate_memory_runtime` / `dealloc_memory`, same sparse-SNode preallocation guard), which cuts the divergence maintainers have to reason about.
- The cryptic HIP driver error on pool exhaustion is replaced with an actionable "Out of pre-allocated memory. Consider using ti.init(device_memory_fraction=0.9) or ti.init(device_memory_GB=N)..." message on both backends. Saves the next person a multi-hour debugging
session.

## Bad points

- HIP enum values (`hipDeviceAttributeMemoryPoolsSupported = 88`, `hipMemPoolAttrReleaseThreshold = 4`) are hardcoded in `amdgpu_driver.h` rather than pulled from a HIP header. A future ROCm ABI reshuffle could break the probe. Mitigated by using the non-throwing
`.call()` variant so a wrong value degrades to "no pool" instead of aborting `qd.init`, but this is a latent fragility.
- The CUDA device-side `__assertfail` hard-halt on pool exhaustion is kept alongside the new host-side propagation for defense in depth, so there are now two mechanisms handling the same failure mode. Arguably the device-side one could be removed later, but doing it in this PR would widen the scope into CUDA behavior.

## Verification

- AMDGPU (ROCm 7, gfx1100): `qd.ndarray(qd.i32, shape=(2**28 + 1,))` succeeds with default `qd.init(arch=qd.amdgpu)`, no `device_memory_GB` override.
- With `device_memory_GB=1` forced, the former `hipErrorInvalidValue` becomes a clean `Out of AMDGPU pre-allocated memory...` error.
- Test passes locally on `arch=arm64`; AMDGPU / CUDA / Vulkan / Metal coverage via CI.

## Notes

ROCm < 5.2 is no longer supported (`hipMallocAsync` is a hard prerequisite). Stated in the README prerequisites. It was released in May 2022 so I think it is reasonable.

Resolves https://github.com/Genesis-Embodied-AI/Genesis/issues/2434

### DIFF
--- a/quadrants/rhi/amdgpu/amdgpu_device.h
+++ b/quadrants/rhi/amdgpu/amdgpu_device.h
@@ -53,6 +53,7 @@ class AmdgpuDevice : public LlvmDevice {
     bool is_imported{false};
     bool use_preallocated{true};
     bool use_cached{false};
+    bool use_memory_pool{false};
     void *mapped{nullptr};
   };
 
--- a/quadrants/runtime/llvm/llvm_runtime_executor.cpp
+++ b/quadrants/runtime/llvm/llvm_runtime_executor.cpp
@@ -64,6 +64,7 @@ LlvmRuntimeExecutor::LlvmRuntimeExecutor(CompileConfig &config, KernelProfilerBa
       config.arch = host_arch();
     } else {
       // AMDGPU runtime created successfully
+      use_device_memory_pool_ = AMDGPUContext::get_instance().supports_mem_pool();
     }
 #else
     QD_WARN("Quadrants is not compiled with AMDGPU.");
--- a/quadrants/runtime/llvm/llvm_runtime_executor.cpp
+++ b/quadrants/runtime/llvm/llvm_runtime_executor.cpp
@@ -357,7 +357,10 @@ void LlvmRuntimeExecutor::initialize_llvm_runtime_snodes(const LlvmOfflineCache:
     }
   }
 
-  if (config_.arch == Arch::cuda && use_device_memory_pool() && !all_dense) {
+  if ((config_.arch == Arch::cuda || config_.arch == Arch::amdgpu) && use_device_memory_pool() && !all_dense) {
+    // Sparse SNode trees allocate runtime state via runtime_memory_allocate_aligned during snode_initialize.
+    // When the device memory pool is active, the eager preallocate_runtime_memory() in materialize_runtime is
+    // skipped, so the bump allocator is only wired up lazily here when a sparse tree actually needs it.
     preallocate_runtime_memory();
   }
 
--- a/tests/python/test_memory.py
+++ b/tests/python/test_memory.py
@@ -19,6 +19,20 @@ def test_memory_allocate():
         x[i] = i
 
 
+@pytest.mark.run_in_serial
+@test_utils.test()
+def test_ndarray_exceeds_default_device_memory_gb():
+    # Allocate an ndarray larger than the default device_memory_GB (1 GiB) to exercise that on-demand growth works
+    # without user-side pool tuning. On AMDGPU this path goes through hipMallocAsync since ROCm 5.2+; CUDA uses
+    # cuMemAllocAsync; CPU/Vulkan/Metal are unaffected by device_memory_GB. The check is that the allocation and its
+    # zero-fill both succeed.
+    n = (1 << 30) // 4 + 1
+    arr = qd.ndarray(qd.i32, shape=(n,))
+    assert arr[0] == 0
+    assert arr[n // 2] == 0
+    assert arr[n - 1] == 0
+
+
 @test_utils.test(arch=get_host_arch_list())
 def test_oop_memory_leak():
     @qd.data_oriented
--- a/tests/python/test_memory.py
+++ b/tests/python/test_memory.py
@@ -22,15 +22,32 @@ def test_memory_allocate():
 @pytest.mark.run_in_serial
 @test_utils.test()
 def test_ndarray_exceeds_default_device_memory_gb():
-    # Allocate an ndarray larger than the default device_memory_GB (1 GiB) to exercise that on-demand growth works
-    # without user-side pool tuning. On AMDGPU this path goes through hipMallocAsync since ROCm 5.2+; CUDA uses
-    # cuMemAllocAsync; CPU/Vulkan/Metal are unaffected by device_memory_GB. The check is that the allocation and its
-    # zero-fill both succeed.
+    # device_memory_GB defaults to 1, so allocate an ndarray whose byte size exceeds 1 GiB to exercise that
+    # on-demand growth works without user-side pool tuning. On AMDGPU this path goes through hipMallocAsync since
+    # ROCm 5.2+; CUDA uses cuMemAllocAsync; CPU/Vulkan/Metal are unaffected by device_memory_GB.
     n = (1 << 30) // 4 + 1
     arr = qd.ndarray(qd.i32, shape=(n,))
     assert arr[0] == 0
     assert arr[n // 2] == 0
     assert arr[n - 1] == 0
+
+
+@test_utils.test(require=qd.extension.sparse)
+def test_sparse_field_with_device_memory_pool():
+    # With the device memory pool active on CUDA / AMDGPU, the eager preallocate_runtime_memory() in
+    # materialize_runtime is skipped. Sparse (non-dense) SNode trees depend on a lazy fallback in
+    # initialize_llvm_runtime_snodes to wire up the device-side bump allocator; exercise that path.
+    x = qd.field(qd.i32)
+    qd.root.pointer(qd.i, 32).dense(qd.i, 8).place(x)
+
+    @qd.kernel
+    def write():
+        x[0] = 42
+        x[255] = 7
+
+    write()
+    assert x[0] == 42
+    assert x[255] == 7
 
 
 @test_utils.test(arch=get_host_arch_list())
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ On Genesis simulator (Linux + NVIDIA 5090):
 ## Prerequisites
 - Python 3.10-3.13
 - Mac OS 14, 15, Windows, or Ubuntu 22.04-24.04 or compatible
+- ROCm 5.2 or newer for AMD GPU support
 
 ## Procedure
 ```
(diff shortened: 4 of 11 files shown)


hughperkins:
@claude review

tests/python/test_memory.py
```diff
@@ -19,6 +19,20 @@ def test_memory_allocate():
         x[i] = i
 
 
+@pytest.mark.run_in_serial
+@test_utils.test()
+def test_ndarray_exceeds_default_device_memory_gb():
+    # Allocate an ndarray larger than the default device_memory_GB (1 GiB) to exercise that on-demand growth works
```
hughperkins: I parse this as 'default device memory gb is 1GB', but I think you intend to mean that 1GB is greater than the default device memory gb?

hughperkins:
can you state why we enable hip memory pool please?

By the way, I quite like @erizmr 's idea that the description of each pr should contain description ,good points, bad points. I make this for my branches using the opus 4.6 prompt "for the purposes of pr description, describe what is in this PR, and the good and bad points about this PR", which works pretty well I find.

quadrants/rhi/amdgpu/amdgpu_device.h
```diff
@@ -53,6 +53,7 @@ class AmdgpuDevice : public LlvmDevice {
     bool is_imported{false};
     bool use_preallocated{true};
     bool use_cached{false};
+    bool use_memory_pool{false};
```
hughperkins, line 56: why does this default to false? who/what/when/why enables it to true?

duburcqa: The entire implementation is a copy-paste of CUDA backend.

duburcqa:
> can you state why we enable hip memory pool please?

Well, without it, batched Genesis  simulation is unusable. Sounds like a big problem? x)

quadrants/rhi/amdgpu/amdgpu_device.h
duburcqa: More specifically, the implementation mirrors CUDA's AllocInfo (where use_memory_pool{false} also defaults to false; it's set from params.use_memory_pool in allocate_memory_runtime).

tests/python/test_memory.py
duburcqa: fixed.

quadrants/runtime/llvm/llvm_runtime_executor.cpp
```diff
@@ -64,6 +64,7 @@ LlvmRuntimeExecutor::LlvmRuntimeExecutor(CompileConfig &config, KernelProfilerBa
       config.arch = host_arch();
     } else {
       // AMDGPU runtime created successfully
+      use_device_memory_pool_ = AMDGPUContext::get_instance().supports_mem_pool();
     }
 #else
     QD_WARN("Quadrants is not compiled with AMDGPU.");
```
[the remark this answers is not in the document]
duburcqa: fixed.

quadrants/runtime/llvm/llvm_runtime_executor.cpp
```diff
@@ -64,6 +64,7 @@ LlvmRuntimeExecutor::LlvmRuntimeExecutor(CompileConfig &config, KernelProfilerBa
       config.arch = host_arch();
     } else {
       // AMDGPU runtime created successfully
+      use_device_memory_pool_ = AMDGPUContext::get_instance().supports_mem_pool();
```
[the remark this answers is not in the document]
duburcqa: fixed.

duburcqa:
> By the way, I quite like @erizmr 's idea that the description of each pr should contain description ,good points, bad points. I make this for my branches using the opus 4.6 prompt "for the purposes of pr description, describe what is in this PR, and the good and bad points about this PR", which works pretty well I find.

Updated with good and bad points.

duburcqa:
@claude review

hughperkins (approved):
Thanks!

PR #504, opened 2026-04-17
[Perf] Tiles 4a: Add tensor_type parametrization to tile16 tests

hughperkins:
Parametrize 16 existing tile16 tests with [qd.ndarray, qd.field] to verify both tensor backends produce identical results. Uses an _ann() helper to select the correct kernel annotation per tensor type.

  ## Summary
  Parametrize 16 existing Tile16x16 tests with `[qd.ndarray, qd.field]` so that every core tile operation is verified against both tensor backends.
  ### What's in this PR
  A single file change to `tests/python/test_tile16.py` (+138 / -79 lines):
  - Adds an `_ann(tensor_type, dtype, ndim)` helper that returns `qd.types.NDArray[dtype, ndim]` for ndarrays or `qd.Template` for fields, so kernel signatures work with both
  backends.
  - Adds `_ATOLS` and `_EPS_VALS` convenience dicts keyed by `qd_dtype`.
  - Adds `@pytest.mark.parametrize("tensor_type", [qd.ndarray, qd.field])` to 16 tests covering zeros, eye, load/store (2D and 3D), row/column clamping, ger_sub, Cholesky,
  triangular solve, slice syntax, outer products, and vec-proxy loads.
  - Replaces hardcoded `qd.ndarray(...)` allocations with `tensor_type(...)` in those tests.
  - Replaces `qd.types.NDArray[qd_dtype, N]` kernel annotations with the `Ann` variable from `_ann()`.
  - Leaves error/raises tests and f32-only tests unchanged (they continue to use `qd.ndarray` directly).
  ### Good points
  - **Doubles coverage cheaply**: every parametrized test now runs on both `qd.field` and `qd.ndarray`, catching backend-specific regressions with no new test logic.
  - **Minimal, mechanical change**: no kernel body modifications, no new dependencies, no API changes — easy to review.
  - **All 278 tests pass** (80 skipped: f64 on Vulkan), verified locally on CUDA + Vulkan.
  - **Clean separation**: error-path and f32-hardcoded tests are intentionally left un-parametrized, avoiding unnecessary test matrix expansion.
  ### Bad points
  - **No `fastcache=True`**: kernels still use the default caching mode and capture variables from the enclosing scope, so they don't exercise the stricter purity-checked code
  path.
  - **Annotation indirection**: the `_ann()` helper adds a level of indirection that makes kernel signatures slightly harder to read at a glance.
  - **Test runtime roughly doubles**: the 16 parametrized tests now run 2× as many variants (ndarray + field × dtype × arch), adding ~4-5 minutes to the full suite.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_intrinsics.py
+++ b/tests/python/test_intrinsics.py
@@ -89,7 +89,7 @@ def test_clock_accuracy():
 
     for i in range(1, 31):
         assert a[i - 1] < a[i] < a[i + 1]
-        assert -1 < a[i] / a[0] - (i + 1) < 1
+        assert -2 < a[i] / a[0] - (i + 1) < 2
 
 
 @test_utils.test(arch=clock_freq_supported_archs)
--- a/tests/python/test_intrinsics.py
+++ b/tests/python/test_intrinsics.py
@@ -89,7 +89,7 @@ def test_clock_accuracy():
 
     for i in range(1, 31):
         assert a[i - 1] < a[i] < a[i + 1]
-        assert -2 < a[i] / a[0] - (i + 1) < 2
+        assert -1 < a[i] / a[0] - (i + 1) < 1
 
 
 @test_utils.test(arch=clock_freq_supported_archs)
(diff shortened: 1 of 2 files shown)


erizmr (approved):

hughperkins:
Thanks! 🙌

PR #505, opened 2026-04-17
[Perf] Tiles 4b: Migrate tiles16 tests to enable fastcache

hughperkins:
Issue: #

### Brief Summary

copilot:summary

  ## Summary
  Migrate all 31 tile16 test kernels to `fastcache=True` by passing the tile size as a `qd.Template` parameter and explicitly parameterizing all closure-captured variables.
  **Base branch:** `hp/tiles-4a`
  ### What's in this PR
  A single file change to `tests/python/test_tile16.py` (+273 / -222 lines):
  - All 31 `@qd.kernel` decorators changed to `@qd.kernel(fastcache=True)`.
  - Adds `tile_size: qd.Template` as the last parameter to every kernel, passed as `_TILE` at each call site. Inside kernels, `_TILE` references are replaced with `tile_size`
  (for `block_dim`, `range()`, and `_load`/`_store` args).
  - Captured integer variables (e.g. `src_offset`, `src_row_end`, `batch`, `NCOLS`, `NBATCH`) are added as `qd.i32` kernel parameters to satisfy purity requirements.
  - Removes `test_tile16_size_constant_in_kernel` (it tests `Tile.SIZE` access inside a kernel, which requires the public `Tile16x16` proxy not yet available on this branch).
  ### Good points
  - **No library code changes**: the `tile_size: qd.Template` approach avoids needing the `Tile16x16Proxy`, `simt/__init__.py` changes, or purity checker modifications. The
  entire PR is test-only.
  - **Exercises the purity checker**: all kernels now run under `fastcache=True`, catching any accidental closure captures that would break caching.
  - **Clean layering**: builds directly on tiles-4a (tensor_type parametrization), keeping each PR focused on one concern.
  - **All 276 tests pass** with `-W error::UserWarning` (80 skipped: f64 on Vulkan).
  ### Bad points
  - **Verbose call sites**: kernels like `test_tile16_load_store` now have 11 arguments at the call site, which is noisy.
  - **Removes a test**: `test_tile16_size_constant_in_kernel` is dropped rather than adapted, reducing coverage for the `Tile.SIZE` access path.
  - **`tile_size` is redundant**: it's always `_TILE` (16) — every kernel receives the same constant. This is a testing-only workaround; production code would use the
  `Tile16x16` proxy directly.

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_intrinsics.py
+++ b/tests/python/test_intrinsics.py
@@ -89,7 +89,7 @@ def test_clock_accuracy():
 
     for i in range(1, 31):
         assert a[i - 1] < a[i] < a[i + 1]
-        assert -1 < a[i] / a[0] - (i + 1) < 1
+        assert -2 < a[i] / a[0] - (i + 1) < 2
 
 
 @test_utils.test(arch=clock_freq_supported_archs)
--- a/tests/python/test_intrinsics.py
+++ b/tests/python/test_intrinsics.py
@@ -89,7 +89,7 @@ def test_clock_accuracy():
 
     for i in range(1, 31):
         assert a[i - 1] < a[i] < a[i + 1]
-        assert -2 < a[i] / a[0] - (i + 1) < 2
+        assert -1 < a[i] / a[0] - (i + 1) < 1
 
 
 @test_utils.test(arch=clock_freq_supported_archs)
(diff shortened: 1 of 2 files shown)


erizmr (approved):

hughperkins:
Thank you 🙌

PR #508, opened 2026-04-17
[Perf] Tiles 4d: Consolidate slice error tests using parametrize

hughperkins:
Issue: #

## Summary
  Consolidates 7 individual slice-error tests into 3 parametrized tests, reducing boilerplate while adding store-side error coverage that was previously missing.
  ### What's in this PR
  - **`test_tile16_load_slice_errors`** — replaces `test_tile16_load_negative_row_raises`, `test_tile16_load_negative_col_raises`, `test_tile16_load_missing_start_raises`,
  `test_tile16_load_missing_stop_raises`. Parametrized over 4 cases (`neg_row`, `neg_col`, `no_start`, `no_stop`) using `qd.static()` branching and `bad_slice: qd.Template`.
  - **`test_tile16_store_slice_errors`** — new parametrized test covering negative-index and missing-start/stop errors on the *store* side (`d[-1:tile_size, ...]`,
  `d[:tile_size, ...]`, `d[0:, ...]`). Previously only `test_tile16_store_missing_stop_raises` existed; this adds 3 new error cases.
  - **`test_tile16_vec_slice_errors`** — replaces `test_tile16_vec_slice_missing_stop_raises` and `test_tile16_vec_slice_missing_start_raises`. Parametrized over 2 cases.
  Net change: 89 lines removed, 50 lines added.
  ### Good
  - Reduces 7 nearly-identical test functions to 3, eliminating duplicated kernel boilerplate
  - Adds store-side error coverage for negative indices and missing start that didn't exist before
  - Uses `bad_slice: qd.Template` as an explicit kernel parameter (consistent with `fastcache=True` purity)
  ### Bad
  - The `qd.static()` branching pattern in the error tests is less immediately readable than separate test functions — you have to trace the parametrize values to understand
  what each case tests
  - All 4 load-error and 4 store-error cases compile into a single kernel with dead branches, which is slightly more work for the compiler (negligible in practice since these
  are error tests)

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ On Genesis simulator (Linux + NVIDIA 5090):
 ## Prerequisites
 - Python 3.10-3.13
 - Mac OS 14, 15, Windows, or Ubuntu 22.04-24.04 or compatible
+- ROCm 5.2 or newer for AMD GPU support
 
 ## Procedure
 ```
--- a/docs/source/user_guide/index.md
+++ b/docs/source/user_guide/index.md
@@ -32,6 +32,14 @@ interop
 ```
 
 ```{toctree}
+:caption: SIMT primitives
+:maxdepth: 1
+:titlesonly:
+
+tile16
+```
+
+```{toctree}
 :caption: Performance
 :maxdepth: 1
 :titlesonly:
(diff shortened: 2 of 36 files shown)


erizmr (approved):

PR #527, opened 2026-04-20
[Type] Tensor 7: Test layout= + needs_grad= grad-buffer round-…

hughperkins:
…trip

Test-only PR closing the gap left by PRs 5 and 6: the combination of layout= and needs_grad= was checked at allocation time, but not yet exercised through a kernel write/read on the .grad buffer with a non- identity layout.

This pins down the pre-impl POC Q3b finding (grad SNode inherits the primal's axis_seq) so any upstream Quadrants regression in this area surfaces immediately.

Adds 8 tests:
- rank-2 transposed-storage primal+grad kernel round-trip
- rank-3 (2, 0, 1) primal+grad kernel round-trip
- all 6 rank-3 layout permutations with primal+grad written and read

All tests pass; no production-code or doc changes.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
Would prefer to add the ndarray layout first really, but ...

PR #530, opened 2026-04-20
[Type] Tensor 10: Layout aliasing across supported Quadrants p…

hughperkins:
…atterns

Test-only PR. Investigation found that in-kernel rebinding (``y = x``) is **not** supported by Quadrants for any ndarray — that's an upstream limitation that raises ``QuadrantsTypeError: Invalid constant scalar data type: AnyArray`` regardless of layout. The test file's docstring spells this out explicitly.

This PR pins down the aliasing patterns that Quadrants does support and that flexible-tensors layout metadata must propagate through:

1. Same Ndarray passed twice to the same kernel: both AnyArrays get the same layout via the runtime feature tuple.
2. Same Ndarray reused across two consecutive kernel calls: layout persists.
3. Repeated ``.grad`` access inside a single kernel: each access inherits layout from the parent AnyArray.
4. Same Ndarray bound to two kernels with different parameter names but compatible annotations: layout travels with the value, not the annotation.
5. Tagged + untagged ndarrays in the same kernel: layout isolation per argument.
6. Two separately-allocated tagged ndarrays: independence.

6 tests, all pass.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_tensor_ndarray_layout_aliasing.py
```diff
@@ -0,0 +1,195 @@
+"""Aliasing of layout metadata across the supported Quadrants aliasing
```
hughperkins, line 1: lets not mention pr numbers

hughperkins: also lets put all these tests into a single file probably, across all the prs.

PR #533, opened 2026-04-20
[Type] Tensor 13: qd.tensor accepts layout= for NDARRAY

hughperkins:
Removes the PR-6-era NotImplementedError gate. With the AST-rewrite plumbing landed in PR 8 and the cache-key contract pinned in PR 12, the public factory now wires non-identity layouts straight through to ndarray-backed tensors.

Behaviour:
- shape= is the **canonical** shape the user indexes inside kernels.
- The factory allocates the underlying ndarray at the *physical* (permuted) shape: physical[k] = canonical[layout[k]].
- The instance is auto-tagged with _qd_layout, so kernel subscripts x[i, j, ...] are translated to physical access by build_Subscript.
- Identity layouts (None or range(ndim)) collapse to no tag — same as the FIELD path — so untagged + identity-tagged stay byte-identical.
- order= still forbidden as a kwarg (single source of truth: layout=).
- ValueError up front for wrong-length / non-permutation layouts.

Tests (test_flexible_factory_layout_ndarray.py, 18 cases):
- No-tag / identity collapse
- Physical shape and tag presence after rank-2 and rank-3 calls
- Validation (length, permutation, order kwarg)
- Rank-2 transpose matches the no-layout reference under transpose
- Rank-2 explicit value spot-checks
- Rank-3 every permutation parametrized
- AugAssign through factory-allocated tensors
- needs_grad layout inheritance
- Cache-key distinction (factory-tagged path)

Drive-by: PR-6 NotImplementedError tests in test_flexible_layout.py flipped to assert the factory now succeeds, with depth coverage delegated to the new file. All 180 flexible-tensors tests pass.

Doc: flexible_tensors.md "Controlling physical layout" gains an "`layout=` on the ndarray backend" subsection with a worked example and a note about the current physical-shape `tensor.shape` quirk.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
yay 🙌

PR #537, opened 2026-04-21
[AutoDiff] Autodiff 12: Heap-backed adstack on LLVM backends (CPU/CUDA/AMDGPU)

duburcqa:
# Heap-backed adstack on LLVM backends (CPU / CUDA / AMDGPU)

> Replaces the per-thread worker-stack allocation of `AdStackAllocaStmt` with a host-grown shared heap, lifting the 256 KB CPU thread-stack budget that the prior `create_entry_block_alloca` path imposed on deep reverse-mode kernels. SPIR-V side is the following PR.

## TL;DR

Prior behaviour on CPU / CUDA / AMDGPU: every `AdStackAllocaStmt` lowered to a function-scope `alloca` at the task's entry block, so every adstack lived on the LLVM stack frame (= worker-thread stack on CPU, per-thread local memory on GPU). A kernel with many loop-carried values at `default_ad_stack_size=256` crossed the worker-thread limit and silently corrupted adjacent stack memory; the previous PR in the stack added a 256 KB codegen-time guard that hard-aborted those kernels.

This PR moves the storage off the stack:

- **Codegen**: a pre-scan of each offloaded task body computes per-task `{ad_stack_offsets_, ad_stack_per_thread_stride_}`, and `visit(AdStackAllocaStmt)` emits `base = runtime->adstack_heap_buffer + linear_thread_idx * stride + offset` instead of an alloca. Base is loaded once in `entry_block` and reused.
- **Runtime**: `LlvmRuntimeExecutor::ensure_adstack_heap(needed_bytes)` grows the per-runtime slab via amortised doubling, publishes the new pointer/size into `runtime->{adstack_heap_buffer, adstack_heap_size}` by caching the two device field-pointer addresses on first grow and writing through them on every subsequent grow.
- **Launchers**: CPU / CUDA / AMDGPU kernel launchers call `ensure_adstack_heap(per_thread_stride * num_threads)` before each task launch. Dynamic-bound range-for tasks resolve `num_threads` by reading `begin` / `end` from `runtime->temporaries` via a host-side DtoH memcpy.
- **CUDA graphs**: rejected at launch when any task has `per_thread_stride > 0`, because graph baking precludes the host-side `ensure_adstack_heap` step between dispatches.
- **`default_ad_stack_size` exposed** via `qd.init()`; raised from 32 → 256 now that the per-thread on-chip / worker-stack budget no longer caps it.
- **The codegen budget guard from the previous PR is removed**; the stack frame no longer carries adstack storage, so the 256 KB ceiling is obsolete.

Nothing changes for kernels that don't enable the adstack extension.

## Why

Prior to this PR, a kernel like a reverse-mode articulated-body dynamics step in Genesis hit the 256 KB CPU-stack budget at modest capacities (4 loop-carried f64 variables × 4096 entries × 16 bytes each already crosses it). The two alternatives — ship with `default_ad_stack_size` capped at a value small enough to fit on every worker stack, or ask users to lower `ad_stack_size` per-kernel — either regress correctness on large kernels or force tuning noise on the user. Moving the storage off-stack removes the constraint entirely: per-thread slice size is bounded only by `num_threads * per_thread_stride` and the driver's allocator.

### Codegen (`quadrants/codegen/llvm/codegen_llvm.{h,cpp}`, `llvm_compiled_data.h`)

`TaskCodeGenLLVM` grows three new per-task fields:

- `ad_stack_per_thread_stride_` — sum of `AdStackAllocaStmt::size_in_bytes()` (aligned up to 8) for every adstack in the task.
- `ad_stack_offsets_` — map from each alloca stmt to its offset within the per-thread slice.
- `ad_stack_heap_base_llvm_` — cached SSA value of the heap base pointer, emitted once in `entry_block`.

`init_offloaded_task_function` pre-scans the task body before any codegen runs and populates the first two, so that later sibling allocas never shift an earlier alloca's offset out from under a cached SSA pointer.

`visit(AdStackAllocaStmt)` now emits:

```cpp
base   = LLVMRuntime_get_adstack_heap_buffer(runtime)  // cached in entry_block
tid_64 = zext(linear_thread_idx(context))              // i32 → i64
slice  = tid_64 * stride                               // widened mul to avoid i32 overflow
ptr    = base + slice + offset
```

`linear_thread_idx` is the arch-appropriate invocation id (`RuntimeContext::cpu_thread_id` on CPU; `block_idx * block_dim + thread_idx` on CUDA / AMDGPU), matching how `rand_states` is indexed.

The old 256 KB function-scope budget guard (introduced in the previous PR) is deleted; its `ad_stack_fn_scope_bytes_` accumulator is gone too. Heap-backed storage makes the ceiling irrelevant.

`OffloadedTask` gains an `AdStackSizingInfo ad_stack` sub-struct that propagates sizing to the host launcher: `per_thread_stride`, `static_num_threads`, `dynamic_gpu_range_for`, plus const values and gtmps byte offsets for range-for `begin` / `end`.

### Per-arch codegen tweaks

- **`codegen_cpu.cpp`** — fills `current_task->ad_stack` with the pre-scanned stride and sets `static_num_threads = cpu_thread_id_range` (CPU thread count is known at compile time).
- **`codegen_cuda.cpp`** — fills `current_task->ad_stack.static_num_threads = grid_dim * block_dim` for const-bound tasks, and marks `dynamic_gpu_range_for = true` + records `begin_offset_bytes` / `end_offset_bytes` / `begin_const_value` / `end_const_value` for dynamic range-for tasks so the launcher can resolve the actual iteration count at launch time.
- **`codegen_amdgpu.cpp`** — same as CUDA.

### Runtime (`llvm_runtime_executor.{h,cpp}`, `runtime.cpp`)

`LLVMRuntime` gains two new fields: `Ptr adstack_heap_buffer = nullptr; u64 adstack_heap_size = 0;`. These are read by every adstack-backed task on the device side; the host writes to them through the cached field-pointer addresses.

`LlvmRuntimeExecutor::ensure_adstack_heap(needed_bytes)`:

1. No-op if `needed_bytes == 0 || needed_bytes <= adstack_heap_size_`.
2. Otherwise `new_size = max(needed_bytes, 2 * adstack_heap_size_)` (amortised doubling).
3. Allocates through the per-arch driver (`llvm_device()->allocate_memory`), wraps in a `DeviceAllocationGuard`.
4. On first grow, calls a one-shot runtime query to fetch the device addresses of `runtime->adstack_heap_buffer` and `runtime->adstack_heap_size`, caches them.
5. Publishes via `memcpy_host_to_device` (CUDA / AMDGPU) or plain pointer stores (CPU) against the cached addresses — no per-grow kernel launch.
6. Releases the previous `DeviceAllocationGuard` via move-assignment. Safety of the release (see the detailed block comment in the `.cpp` and the matching field comment in the `.h`): CPU uses `std::free` (trivially safe); CUDA `cuMemFree_v2` synchronises before returning; AMDGPU `dealloc_memory` pools through `CachingAllocator::release` without sync, and cross-launch safety on AMDGPU is provided by the synchronous `hipFree(context_pointer)` at the tail of `amdgpu::KernelLauncher::launch_llvm_kernel` (the latent-fix in #536).

`get_runtime_temporaries_device_ptr()` — cached lookup of `runtime->temporaries`, used by the GPU launchers to read back dynamic range-for bounds.

### Per-arch launchers

- **`runtime/cpu/kernel_launcher.{h,cpp}`** — `Context` gains a parallel `ad_stack_needed_bytes` vector, precomputed at register time (CPU sizing is static). `launch_offloaded_tasks` calls `ensure_adstack_heap` per task.
- **`runtime/cuda/kernel_launcher.cpp`** — adds `resolve_num_threads(task)` which DtoH-memcpys `begin` / `end` from `runtime->temporaries` for dynamic range-for tasks; calls `ensure_adstack_heap` per task.
- **`runtime/amdgpu/kernel_launcher.cpp`** — same as CUDA.
- **`runtime/cuda/graph_manager.cpp`** — hard-errors `graph=True` on kernels where any task has `per_thread_stride > 0`. Graph baking precludes host-side intervention between dispatches.

### Capacity knob (`compile_config.h`, `python/export_lang.cpp`)

- `default_ad_stack_size` raised from 32 to 256.
- Exposed as a `qd.init()` kwarg. The comment block is rewritten to reflect the new heap-backing reality.
- `ad_stack_size` (per-stack explicit capacity) is unchanged.

### Docs (`docs/source/user_guide/autodiff.md`)

Drops the "SPIR-V on-chip cap" limitation from the known-limitations list (that was about the prior Function-scope SPIR-V path; with the SPIR-V heap landing in the next PR, it's gone too). Adds a "Tuning the capacity" section explaining `default_ad_stack_size` vs `ad_stack_size` and the K+2-pushes-per-iteration rule for picking N.

### Tests (`tests/python/test_adstack.py`)

Heap-specific additions:

- `test_adstack_heap_grow_on_demand` — two launches at increasing capacity pinpoint that the amortised-doubling grow path fires and the second launch reuses the bigger slab.
- `test_adstack_heap_backed_exceeds_old_threadstack_budget` — a kernel whose per-thread adstack bytes exceed the pre-PR 256 KB ceiling now compiles and runs correctly.
- `test_adstack_cuda_graph_rejected_with_adstack` — `graph=True` on an adstack kernel raises.

## Side-effect audit

| Concern                                  | Where checked                                                            | Verdict                                                                  |
| ---------------------------------------- | ------------------------------------------------------------------------ | ------------------------------------------------------------------------ |
| Offline cache key                        | `AdStackSizingInfo` fields serialised via the existing `OffloadedTask` key hash | Auto-covered, kernels without adstack get zero stride and hash identically |
| Stmt clone / serialization               | No new `Stmt` fields (all sizing lives in codegen / `OffloadedTask`)     | N/A                                                                      |
| CUDA graphs                              | Hard rejection with clear error message when any task has `per_thread_stride > 0` | Fails loudly, not silently                                               |
| AMDGPU release lifecycle                 | Relies on #536's synchronous `hipFree(context_pointer)` tail             | Cross-launch invariant spelled out in both `.cpp` and `.h` comments      |
| Non-adstack kernels                      | `per_thread_stride == 0` path short-circuits the launcher-side `ensure_adstack_heap` | Zero-cost for kernels that don't enable the extension                    |

## Stack

Split 2/3 of the former "heap-backed adstack" PR. Based on #536 (latent fixes). Followed by #493 (SPIR-V heap).


docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+Each adstack has a capacity that is fixed at compile time and cannot be modified at runtime. When the compiler can prove the worst-case loop trip count, that value is used; otherwise it falls back to a conservative default.
 
-If a kernel overflows its adstack at runtime, Quadrants raises a Python `RuntimeError` naming the overflow at the next `qd.sync()`; if the default is already too large for the target driver, pipeline creation itself fails with a similar exception at kernel-launch time. Heap-backed SPIR-V adstack, which would lift the per-thread ceiling, is left for future work.
+If a kernel overflows its adstack at runtime, Quadrants raises a Python `RuntimeError` naming the overflow at the next `qd.sync()`.
+
+**Tuning the capacity.** Two `qd.init()` knobs control adstack sizing:
+
+- `default_ad_stack_size=N` (default `256`): the fallback capacity for loops whose trip count the compiler cannot prove statically. Every adstack whose max_size was not deducible shares this value. Prefer tuning this knob, since it only affects the branch where the compiler needed to guess.
```
hughperkins: lets have the units please
`default_ad_stack_size_mb`

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+- `ad_stack_size=N` (default `0 = adaptive`): a hard override that forces every adstack in the program to exactly `N` slots, regardless of what the compiler proved. Prefer this knob only when a targeted experiment needs uniform sizing (e.g. stress-testing the runtime heap path).
```
hughperkins: units `ad_stack_size_mb`

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+**How to pick `default_ad_stack_size`.** The reverse pass of a `K`-iteration dynamic loop emits `K + 2` pushes per adstack (the trip count plus two setup pushes: one for the initial adjoint slot and one for the primal's starting value). Size the default at the flat trip count of the deepest unprovable dynamic loop in the program, plus that headroom. Common shapes:
```
hughperkins: `default_ad_stack_size_mb`

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+- A single `qd.ndrange(n, m)` whose bounds come from a field: worst case is `n_max * m_max` iterations. Pick `N >= n_max * m_max + 2`. At `max_n_dofs_per_entity = 16`, 16 x 16 = 256 hits the default exactly.
```
hughperkins: dont we need to multiply by 4?

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+- Nested `for i in range(a[None]): for j in range(b[None]):`: worst case is `a_max * b_max`, same rule.
```
hughperkins: x 4?

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+- A single dynamic `for i in range(a[None])`: `N >= a_max + 2`.
```
hughperkins: * 4?

docs/source/user_guide/autodiff.md
hughperkins: by the way, why + 2?

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+**Memory cost.** The adstack pipeline allocates one small scratch buffer per loop-carried variable that the reverse pass has to remember. For example, a kernel whose dynamic loop reads and updates one float accumulator needs 1 adstack; a kernel whose loop updates four different floats needs 4. Integer counters and boolean branch flags used by the reverse pass also count (typically one each per dynamic `if` or nested loop). The total memory Quadrants allocates across all those buffers is roughly
```
hughperkins: why do we need to qualify with "that the reverse pass has to remember."? Are there loop-carried variables tha the reverse pass does not have to remember?

docs/source/user_guide/autodiff.md
hughperkins: if this is in units, or int32s, or something then maybe something like `default_ad_stack_size_count` or `default_ad_stack_size_units` or default_ad_stack_size_i32s`?

hughperkins: what happens if there is an i64 in the loop?

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+```
+num_threads * stack_size * bytes_per_element * num_loop_carried_variables
```
hughperkins: can you clarify where num_threads suddenly springs from? I'm guessing it's from the top level for loop, but you don't introduce t his I think. Or at least, I dont remember your introducing this.

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+```
+
+where `bytes_per_element` depends on the element type and the backend. On the LLVM backends (CPU / CUDA / AMDGPU) each adstack slot stores both a primal and an adjoint value, so f32 costs 8, i32 costs 8, and bool costs 2 bytes per slot. On the SPIR-V backends (Metal / Vulkan) integer adstacks only store the primal (the reverse pass does not accumulate integer adjoints), and bool is widened to i32 at storage time because SPIR-V has no defined layout for `OpTypeBool`, so f32 costs 8, i32 costs 4, and bool costs 4 bytes per slot. The buffer lives on the device on GPU and in host RAM on CPU. `num_threads` is the number of threads the kernel actually dispatches, not a worst-case grid: on CPU this is the thread pool size (tens of threads), so the memory footprint stays small; on GPU it is the dispatched ndrange. The buffer grows on demand to match the largest size any launch has needed so far and is then reused across subsequent launches, so you do not need to reserve memory up front.
```
hughperkins: lets ditch the `where`, otherwsie no room to breathe. Seems lik a bunch new concepts here, so lets give the reader time to breathe. its a new paragraph.

hughperkins: "the LLVM backends" => "LLVM backends"

hughperkins: make a table rather than filling the text with examples.

hughperkins: this all seesm like way too much detail. Do we really need to know this to use autodiff? Move it to an 'advanced' or 'under the hood' section.

hughperkins: keep 'The buffer grows on demand to match the largest size any launch has needed so far and is then reused across subsequent launches, so you do not need to reserve memory up front.'

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+Two shapes blow up the cost:
```
hughperkins: not sure how 'two shapes' relates to the bullet points
- the second bullet point appears to be a palliative, not a shape
- the first bullet point seems to be just one shape

I really think this is all very advanced stuff though. If a user has to know all this, they're not going to start using autodiff.

MOve it to an 'advanced' or 'gotchas' section or similar.

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+- A big `ndrange` on GPU. A kernel over `ndrange(1024, 1024)` with a `default_ad_stack_size` of `256` and four f32 loop-carried variables allocates roughly `1024 * 1024 * 256 * 8 * 4 bytes = 8 GB` - easy to overrun a consumer GPU's memory budget.
+- Doubling `default_ad_stack_size` doubles the backing buffer linearly, so it is the easiest knob to reach for when you hit an out-of-memory error.
+
+If the GPU driver returns an out-of-memory error, the order of remedies is: drop `default_ad_stack_size` toward the real worst-case trip count of your dynamic loops, reduce the number of loop-carried variables the reverse pass has to remember (split a kernel, checkpoint manually, or fold two accumulators into one), then raise `device_memory_fraction` or `device_memory_GB` at `qd.init()` if your GPU has headroom.
```
hughperkins: 'tip 'n tricks' seciton, or similar

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,37 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+**When the default is too small.** The runtime surfaces overflow as a `QuadrantsAssertionError` (LLVM backends) or `RuntimeError` (SPIR-V backends) on the next `qd.sync()`, and the message recommends bumping `default_ad_stack_size`. Pick `N` using the rules above and retry.
```
hughperkins: 'advanced' or 'reference' or smilar. User dosent need to know this. They'll find out soon enough...

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+Each adstack has a capacity that is fixed at compile time. When the compiler can prove the worst-case loop trip count, that value is used; otherwise it falls back to a conservative default.
 
-If a kernel overflows its adstack at runtime, Quadrants raises a Python `RuntimeError` naming the overflow at the next `qd.sync()`; if the default is already too large for the target driver, pipeline creation itself fails with a similar exception at kernel-launch time. Heap-backed SPIR-V adstack, which would lift the per-thread ceiling, is left for future work.
+If a kernel overflows its adstack at runtime, Quadrants raises a Python exception (`QuadrantsAssertionError` on LLVM backends, `RuntimeError` on SPIR-V backends) on the next `qd.sync()`. The message points at `default_ad_stack_size`: bump that and retry.
```
hughperkins: I dont think the user needs to know this. It's expected to be the case. If it's not then they'll be annoyed. But not part of the user doc please.

(You could move it to an 'advanced' or 'reference' section if you wish, I feel, but I think it's really just expected).

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+Two `qd.init()` knobs control adstack sizing. Both are measured in **slots per stack** (not bytes; see the advanced section below for per-slot sizes).
+
+- `default_ad_stack_size=N` (default `256`): the fallback capacity for loops whose trip count the compiler cannot prove. Every adstack whose size was not deducible uses this value. Prefer this knob — it only kicks in where the compiler had to guess.
```
hughperkins: preference for `default_ad_stack_size_slots` (and similar for the other config options below)

hughperkins: (or `default_ad_stack_slots_count` or `default_ad_stack_num_slots`)

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+- `ad_stack_size=N` (default `0 = adaptive`): a hard override that forces every adstack in the program to exactly `N` slots, regardless of what the compiler proved. Use only for targeted experiments (e.g. stress-testing the runtime heap path).
```
hughperkins: `ad_stack_num_slots`

docs/source/user_guide/autodiff.md
duburcqa: `ad_stack_size` already exists. I would prefer to be consistent since they both come together.

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
 
 ### Adstack overflow
```
hughperkins: I feel like this wholesection si not needed by the user to know, since it's automatic now right?

Lets just move it to an 'under the hood'section?

docs/source/user_guide/autodiff.md
duburcqa: This was pre-existing.

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+If a kernel overflows its adstack at runtime, Quadrants raises a Python exception (`QuadrantsAssertionError` on LLVM backends, `RuntimeError` on SPIR-V backends) on the next `qd.sync()`. The message points at `default_ad_stack_size`: bump that and retry.
+
+Two `qd.init()` knobs control adstack sizing. Both are measured in **slots per stack** (not bytes; see the advanced section below for per-slot sizes).
+
+- `default_ad_stack_size=N` (default `256`): the fallback capacity for loops whose trip count the compiler cannot prove. Every adstack whose size was not deducible uses this value. Prefer this knob — it only kicks in where the compiler had to guess.
+- `ad_stack_size=N` (default `0 = adaptive`): a hard override that forces every adstack in the program to exactly `N` slots, regardless of what the compiler proved. Use only for targeted experiments (e.g. stress-testing the runtime heap path).
+
+The backing buffer grows on demand to match the largest size any launch has needed so far and is then reused across subsequent launches, so you do not need to reserve memory up front.
```
hughperkins: this is the paragraph that suggests to me that this whole section is not needed by the user, and can move to an advanced seciton?

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+### Advanced: adstack memory tuning
+
+*You only need this section if you hit an adstack OOM on GPU or want to cap the memory footprint explicitly.*
```
hughperkins: Good. I like this preample.

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+**Sizing rule.** The reverse pass of a `K`-iteration dynamic loop emits `K + 2` pushes per adstack: the per-iteration push, one setup push for the initial adjoint slot, and one for the primal's starting value. Size `default_ad_stack_size` at the worst-case trip count of the deepest unprovable dynamic loop in the program, plus that headroom of 2:
```
hughperkins: `per adstack slot`

hughperkins: (or ... is it per slot? or per ... stack? 🤔  per loop? please clarify)

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+At `max_n_dofs_per_entity = 16`, a 16 x 16 ndrange hits the default exactly (`256`).
+
+**Memory footprint.** The pipeline allocates one scratch buffer per loop-carried variable. Total memory across all buffers is approximately:
+
+```
+num_threads * default_ad_stack_size * bytes_per_slot * num_loop_carried_variables
+```
+
+`num_threads` is the number of threads the kernel actually dispatches (tens on CPU via the thread pool; up to the full ndrange on GPU). `num_loop_carried_variables` counts the reverse-pass state: each f32 accumulator, each integer counter, and each boolean branch flag is one buffer. `bytes_per_slot` depends on the element type and the backend:
```
hughperkins: Not a fan of things in parentheses. If it's not necessary, then lets move it to advanced. oteherwise lets take it out of parentheses please, make it into a new sentence.

hughperkins: I'm not sure how f32, integer, boolean are fitting together here. Does this mean an f32 accumulator is one buffer? Or some kind of union, or a struct containing an f32 + an int + a bool? Not clear to me.

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+| Backend | f32 | i32 | i64 | bool |
```
hughperkins: why arent we consdiering other types, such as u8, f64, etc?

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+| --- | --- | --- | --- | --- |
+| LLVM (CPU / CUDA / AMDGPU) | 8 | 8 | 16 | 2 |
+| SPIR-V (Metal / Vulkan) | 8 | 4 | 8 | 4 |
+
+LLVM backends store both a primal and an adjoint per slot, doubling the size. SPIR-V integer adstacks only store the primal (the reverse pass does not accumulate integer adjoints); `bool` is widened to `i32` at storage time because SPIR-V has no defined layout for `OpTypeBool`. Adstack buffers live on the device on GPU and in host RAM on CPU.
```
hughperkins: I think stuff abou spir-v should move to a separate paragraph. Lets have one paragraph for the overall cross-paltform veiw, and then seprate paragraph(s) for platform-sepcfici quirks.

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+**Avoiding OOM on GPU.** A big `ndrange` combined with several loop-carried f32 variables adds up fast: `ndrange(1024, 1024)` × `default_ad_stack_size=256` × 4 f32 variables is already ~8 GB, enough to exhaust a consumer GPU. Remedies, in order:
```
hughperkins: Good. I like.

docs/source/user_guide/autodiff.md
```diff
@@ -234,9 +234,51 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+1. Lower `default_ad_stack_size` toward the real worst-case trip count of your loops.
+2. Reduce the number of loop-carried variables the reverse pass has to remember (split the kernel, checkpoint manually, or fold two accumulators into one).
```
hughperkins: not sure if you've defined 'loop-carried variables'.  You might have. Could you confirm whether there is a clear definition of this concept earlier on?

docs/source/user_guide/autodiff.md
```diff
@@ -213,13 +213,13 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+**How Quadrants does it.** An opt-in compiler pipeline called the *autodiff stack* (*adstack*) allocates a per-variable stack alongside each primal that is updated inside the loop and therefore changes from one iteration to the next. The forward pass pushes an entry each iteration; the reverse pass pops them back off in reverse order to recover the correct primal for every chain-rule step. It is opt-in because it costs extra per-thread memory and compile time, and because most kernels do not need it. Running with adstack enabled when it is not strictly needed is safe. Running without it when it is needed raises a `QuadrantsCompilationError` in most cases (the autodiff pass rejects a non-static range that would otherwise lose its primal); in the narrow cases where the kernel compiles anyway, the reverse pass reads a stale value for every iteration and the gradients come out wrong but non-zero.
```
hughperkins: break the first sentence into two. so each setnence just states a single concept.

hughperkins: " the autodiff stack (adstack) " => "adstack". We only ever refer to it as adstack, so let's say adstack is its name. We can put the long form in brakcets i youf want "called the *adstack* (short for "(a)uto(d)iff (stack)")"

hughperkins: I think remove "and therefore changes from one iteration to the next"

hughperkins: " It is opt-in" => "adstack is opt-in"

hughperkins: "Running without it when it is needed raises a `QuadrantsCompilationError` in most cases (the autodiff pass rejects a non-static range that would otherwise lose its primal);" => nice 🙌

hughperkins: "in the narrow cases where the kernel compiles anyway, the reverse pass reads a stale value for every iteration and the gradients come out wrong but non-zero." => would be nice to get rid of such exceptional cases. Do we know what they are? Can we document them?

docs/source/user_guide/autodiff.md
hughperkins: I know it's not part of your changes in this pr, but it's part of your set of prs overall, and anyway, I think it would be good to address:

I couldnt understand this sentence. I mean, I probably could if I thought about it a long time, bu I think it would be good to explain step by step, so people can just read fluidly, and undersatnd it, without grinding to a halt, and having to work through stuff in their head.

docs/source/user_guide/autodiff.md
```diff
@@ -213,13 +213,13 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
 
 **Workflow.** Enable the pipeline at init time and keep using the normal reverse-mode workflow: `qd.init(..., ad_stack_experimental_enabled=True)`. The flag is compile-time, so it must be set before the offending kernel compiles.
 
 ### Examples of dynamic loops that need it
 
-- A loop-carried dependency (a variable read, written, and read again across iterations, e.g. `v = v * 0.95 + 0.01`).
+- A loop-carried variable - one whose value is carried forward from each iteration into the next, e.g. `v = v * 0.95 + 0.01`.
 - A local variable used as an index into a global field.
 - Non-linear ops (`sin`, `cos`, `exp`, `sqrt`, `tanh`, `pow`, ...) whose derivative depends on the primal value at that iteration.
```
hughperkins: can we have a counter-example of a non-linear op that doesnt need adstack, in the 'do not need it' section below please.

docs/source/user_guide/autodiff.md
hughperkins, line 228: Might be worth stating here explicilty "Note the definiton of a 'loop carried variable' above. We will be using this definition throughout this document.

docs/source/user_guide/autodiff.md
```diff
@@ -232,11 +232,60 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
 
-If a kernel overflows its adstack at runtime, Quadrants raises a Python `RuntimeError` naming the overflow at the next `qd.sync()`; if the default is already too large for the target driver, pipeline creation itself fails with a similar exception at kernel-launch time. Heap-backed SPIR-V adstack, which would lift the per-thread ceiling, is left for future work.
+**Tuning.** Two `qd.init()` knobs control adstack sizing, both measured in slots per adstack (not bytes):
+
+- `default_ad_stack_size=N` (default `256`): the fallback capacity for loops whose trip count the compiler cannot prove statically. Every adstack whose size was not deducible shares this value. Prefer this knob — it only affects the branch where the compiler had to guess.
+- `ad_stack_size=N` (default `0 = adaptive`): hard override forcing every adstack in the program to exactly `N` slots regardless of what the compiler proved. Use only for targeted experiments, for example stress-testing the runtime heap path.
```
hughperkins, line 243: terrible name. Not asking you to change it. But it sounds like a very reasonable parameter to modify by the user, whereas in fact it's basically debug only. Should have a `debug_` prefix or something I feel. (or even be private).

docs/source/user_guide/autodiff.md
```diff
@@ -232,11 +232,60 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+**Sizing rule.** A `K`-iteration dynamic loop consumes `K + 2` slots in each of its adstacks: one slot per forward iteration, plus two setup slots (one for the initial adjoint, one for the primal's starting value). `default_ad_stack_size` is a per-stack slot count, so size it at the worst-case trip count of the deepest unprovable dynamic loop in the program, plus that `+ 2`.
```
hughperkins: so far we haven't discussed the concept that a loop can have multiple adstacks. Can you have a pargraph to present this concept first please. (I guess each loop-carried variable gets its own adstack?)

docs/source/user_guide/autodiff.md
duburcqa: Yes. Let's address this in a follow-up.

docs/source/user_guide/autodiff.md
hughperkins: "plus that `+ 2`." redundant redundancy.

docs/source/user_guide/autodiff.md
```diff
@@ -232,11 +232,60 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+| Loop shape | Required `default_ad_stack_size` |
+| --- | --- |
+| single dynamic `for i in range(a[None])` | `a_max + 2` |
+| nested `for i in range(a[None]): for j in range(b[None])` | `a_max * b_max + 2` |
+| `qd.ndrange(n, m)` with field-derived `n`, `m` | `n_max * m_max + 2` |
+
+At `max_n_dofs_per_entity = 16`, a 16 x 16 ndrange hits the default exactly (`256`).
```
hughperkins: mixtue of fonts is visually distracting. either put backticks around `16 x 16` too, or remove from `= 16`

docs/source/user_guide/autodiff.md
```diff
@@ -232,11 +232,60 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+**Memory footprint.** The pipeline allocates one scratch buffer per piece of reverse-pass state. That count includes every loop-carried variable the reverse pass has to replay, plus any integer counter and any boolean branch flag it has to read back. Total memory across all buffers is approximately
```
hughperkins: whats a 'piece'? per stack carried variable?

hughperkins: oh you define it next 🤔

hughperkins: Maybe we just avoid the issue by avoiding having to use this noun at all? for example

"The pipeline allocates a scratch buffer for each loop-carried variable, and also for any loop counter, and any boolean branch flags." ?

Still, I'm unclear about this 'integer counter' and 'boolean branch flag'. You havent defined them before. Could you define these in a previous paragraph please.

docs/source/user_guide/autodiff.md
```diff
@@ -232,11 +232,60 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+
+```
+num_threads * default_ad_stack_size * bytes_per_slot * num_buffers
```
hughperkins, line 260: lets define num_threads before we use it, rather than after, somehow.

"In order to calculate the memory used, we will also need the number of threads the kernel dispatches, which we denote `num_threads`. On CPU this is the thread-pool size, typically tens. On GPU it is the full ndrange."

(and then after this, you can put:

Total memory across all buffers is approximately
(etc)

)

docs/source/user_guide/autodiff.md
```diff
@@ -232,11 +232,60 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+```
+
+`num_threads` is the number of threads the kernel actually dispatches. On CPU that is the thread-pool size, typically tens. On GPU it is the full ndrange. `bytes_per_slot` scales with the element's storage size and the backend; see the two tables below.
+
```
hughperkins, line 264: "On all backends, an adstack always stores a primal value."

"In addition, on LLVM backends, each adstack slot stores also an adjoint value because [explain why]"

"On SPIR-V backends (Metal / Vulkan), integer adstacks only store the primal because the reverse pass does not accumulate integer adjoints ..."

docs/source/user_guide/autodiff.md
```diff
@@ -232,11 +232,60 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+On LLVM backends (CPU / CUDA / AMDGPU), each adstack slot stores both a primal and an adjoint value, so `bytes_per_slot = 2 * sizeof(T)` for every element type `T`. Common cases:
```
hughperkins: Nice 🙌

docs/source/user_guide/autodiff.md
```diff
@@ -232,11 +232,60 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
+| --- | --- |
+| f32 | 8 |
+| f64 | 16 |
+| i32 / u32 | 8 |
+| i64 / u64 | 16 |
+| bool | 2 |
+
+On SPIR-V backends (Metal / Vulkan), integer adstacks only store the primal because the reverse pass does not accumulate integer adjoints, so `bytes_per_slot = sizeof(T)` for integer `T` and `bytes_per_slot = 2 * sizeof(T)` for real `T`. SPIR-V has no defined layout for `OpTypeBool`, so booleans are widened to i32 at storage time:
+
+| T | SPIR-V bytes/slot |
```
hughperkins, line 277: Nice 🙌

docs/source/user_guide/autodiff.md
```diff
@@ -213,22 +214,22 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
 
 **Workflow.** Enable the pipeline at init time and keep using the normal reverse-mode workflow: `qd.init(..., ad_stack_experimental_enabled=True)`. The flag is compile-time, so it must be set before the offending kernel compiles.
 
 ### Examples of dynamic loops that need it
 
-- A loop-carried variable - one whose value is carried forward from each iteration into the next, e.g. `v = v * 0.95 + 0.01`.
+- A loop-carried variable - one whose value is carried forward from each iteration into the next, e.g. `v = v * 0.95 + 0.01`. The rest of this document uses "loop-carried variable" in this sense.
 - A local variable used as an index into a global field.
-- Non-linear ops (`sin`, `cos`, `exp`, `sqrt`, `tanh`, `pow`, ...) whose derivative depends on the primal value at that iteration.
+- Non-linear ops (`sin`, `cos`, `exp`, `sqrt`, `tanh`, `pow`, ...) whose input changes from one iteration to the next. Example: `for i in range(n): total += qd.sin(v); v = v * 0.95 + 0.01` - `qd.sin(v)` is evaluated against a different `v` each iteration, and the reverse pass needs every one of those `v` values to compute the derivative at each step.
```
hughperkins: why does it matter that they are non-linear ops? how is this different from the fact that v is a loop-carried variable? what about other ops such as addition or subtractio netc?

docs/source/user_guide/autodiff.md
```diff
@@ -213,41 +214,85 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
 - An `if` whose condition depends on a variable that mutates across iterations.
 
 ### Examples of dynamic loops that do not need it
 
-- Read-only streaming over a field with a linear accumulator: `for i in x: total += x[i]`. The per-iteration gradient contribution is a constant, no per-iteration primal replay is needed.
-- A linear reduction over a dynamic range: `for i in range(n): total += a * x[i] + b`. Same rationale.
-- A dynamic range whose body does not carry state across iterations and whose only non-linear op uses a loop-invariant value (the primal is the same every iteration, so replay is trivially correct).
+- A loop that only reads from a field and accumulates into a running total, e.g. `for i in x: total += x[i]`. In each iteration the body adds `x[i]` to `total`. The gradient of that addition with respect to `x[i]` is simply `1` - a constant that does not depend on `x[i]` itself. The reverse pass can therefore compute it without recovering any iteration-specific primal, so the adstack is unnecessary.
```
hughperkins: it seems that the more general case then is any expression whose derivative is a constant? though derivative wrt what? 🤔  How do we know that we are taking the gradient wrt x[i]?

docs/source/user_guide/autodiff.md
```diff
@@ -238,9 +239,11 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
 - `ad_stack_size=N` (default `0 = adaptive`): hard override forcing every adstack in the program to exactly `N` slots regardless of what the compiler proved. Use only for targeted experiments, for example stress-testing the runtime heap path.
 
+**One adstack per variable.** A dynamic loop does not have a single adstack. The compiler allocates one for each [loop-carried variable](#examples-of-dynamic-loops-that-need-it) the reverse pass has to replay. It also allocates one for each *integer counter*: a loop-carried integer variable, for example a running `count += 1` or an index `idx += step` used downstream. And one for each *boolean branch flag*: a per-iteration boolean the compiler emits internally whenever an `if` inside the loop body depends on a loop-carried variable - it records which branch ran on each iteration so the reverse pass walks the matching one. A kernel with four f32 loop-carried variables and one integer counter therefore allocates five separate adstacks, each sized independently by the rule below.
```
hughperkins: How is the integer counter different than a loop-carried variable?

docs/source/user_guide/autodiff.md
```diff
@@ -249,17 +252,17 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
 | nested `for i in range(a[None]): for j in range(b[None])` | `a_max * b_max + 2` |
 | `qd.ndrange(n, m)` with field-derived `n`, `m` | `n_max * m_max + 2` |
 
-At `max_n_dofs_per_entity = 16`, a 16 x 16 ndrange hits the default exactly (`256`).
+At `max_n_dofs_per_entity = 16`, a `16 x 16` ndrange hits the default exactly (`256`).
 
-**Memory footprint.** The pipeline allocates one scratch buffer per piece of reverse-pass state. That count includes every loop-carried variable the reverse pass has to replay, plus any integer counter and any boolean branch flag it has to read back. Total memory across all buffers is approximately
+**Memory footprint.** With one scratch buffer per adstack (see above), the total memory cost depends on two further quantities. The first is the number of threads the kernel actually dispatches, which we call `num_threads`. On CPU that is the thread-pool size, typically tens. On GPU it is the full ndrange. The second is `bytes_per_slot`, which scales with the element's storage size and the backend; the two tables below work through its concrete values. Total memory across all buffers is then approximately:
```
hughperkins: have we used the term 'element' thus far? Have we defined it?

docs/source/user_guide/autodiff.md
```diff
@@ -249,17 +252,17 @@ The pattern often hides inside in-place time-stepping updates like `x[i] = x[i]
...
 
 ```
 num_threads * default_ad_stack_size * bytes_per_slot * num_buffers
 ```
 
-`num_threads` is the number of threads the kernel actually dispatches. On CPU that is the thread-pool size, typically tens. On GPU it is the full ndrange. `bytes_per_slot` scales with the element's storage size and the backend; see the two tables below.
+On all backends, every adstack slot stores a *primal* value - that is what the reverse pass pops to recover the forward-pass value at each chain-rule step.
 
-On LLVM backends (CPU / CUDA / AMDGPU), each adstack slot stores both a primal and an adjoint value, so `bytes_per_slot = 2 * sizeof(T)` for every element type `T`. Common cases:
+In addition, LLVM backends (CPU / CUDA / AMDGPU) store an *adjoint* slot of the same type alongside the primal for every element type. The adjoint slot is where the reverse pass accumulates chain-rule contributions; LLVM carries it even for integer and boolean types because the codegen uses a uniform two-element slot layout, which keeps push/pop branch-free at the cost of the unused adjoint slot for non-real types. So `bytes_per_slot = 2 * sizeof(T)` on LLVM for every element type `T`:
```
hughperkins: Nice. thanks!

hughperkins:
Checklist:
- Doc updated, and looks good to me (readable, makes sense, a few variable names that seem questionable, but they are pre-existing, and should be addressed seprately from this PR, to avoid inflating the PR)

=> ok to merge

PR #538, opened 2026-04-21
[AutoDiff] Autodiff 14: Bound GfxRuntime::ctx_buffers_ retirement queue across flush() calls

duburcqa:
# Bound `GfxRuntime::ctx_buffers_` retirement queue across `flush()` calls

> Follow-up to #536. Caps the deferred-free memory that can accumulate across `flush()` calls without an intervening `synchronize()`. Bounded-wait approximation of the Ideal semaphore-keyed retirement pattern (non-blocking poll). The truly non-blocking variant is deferred to a per-backend RHI change.

## TL;DR

#536 fixed a use-after-free on the SPIR-V side by leaving `ctx_buffers_` alone in `flush()` (only clearing it in `synchronize()` after `wait_idle()` drains the stream). codex flagged the resulting trade-off: in async workloads that call `flush()` repeatedly without syncing — e.g. a pipeline of submits that forward-chain semaphores to the next stage — `ctx_buffers_` can grow unbounded because nothing reclaims the deferred-free entries.

```cpp
// quadrants/runtime/gfx/runtime.h
static constexpr std::size_t kPendingRetirementsDepth = 3;
std::deque<std::pair<StreamSemaphore, std::vector<std::unique_ptr<DeviceAllocationGuard>>>> pending_retirements_;

// quadrants/runtime/gfx/runtime.cpp — flush()
if (pending_retirements_.size() >= kPendingRetirementsDepth) {
  device_->get_compute_stream()->command_sync();
  pending_retirements_.clear();
}
if (!ctx_buffers_.empty()) {
  pending_retirements_.emplace_back(sema, std::move(ctx_buffers_));
  ctx_buffers_.clear();
}
```

Each `flush()` snapshots the current `ctx_buffers_`, pairs it with the submission semaphore, and pushes the pair onto a FIFO capped at 3 entries. On the 4th flush without a `synchronize()`, `command_sync()` drains the compute stream and the FIFO is cleared. Never reclaims an in-flight allocation; never grows beyond `3 * ctx_buffers_per_flush` of live-but-deferred memory.

## Why the bounded-wait approximation and not the Ideal polling variant

Codex's suggestion was "per-submission retirement tied to the returned semaphore" — i.e. non-blocking poll of each semaphore's signaled status and retirement of exactly the batches that are done. That requires a `bool is_signaled() const` method on `StreamSemaphoreObject` plus per-backend implementations:

- **Vulkan**: needs a `VkFence` associated with each submit and `vkGetFenceStatus`. Quadrants RHI's Vulkan backend (`quadrants/rhi/vulkan/vulkan_device.{h,cpp}`) today returns only a binary `VkSemaphore` from submit; adding fence support means either threading a parallel fence through the submit API or binding a fence to the `VulkanStreamSemaphoreObject` on construction. Binary-semaphore polling is not defined by the spec.
- **Metal**: `MTLSharedEvent.signaledValue >= targetValue` polls cleanly. The current `MetalStreamSemaphoreObject` (`quadrants/rhi/metal/metal_device.{h,mm}`) wraps an `id<MTLEvent>`; swapping or extending to `MTLSharedEvent` is straightforward but still an RHI API change.
- **CPU**: trivial — the submit is synchronous, so "signaled" is always true.

None of those changes are conceptually hard, but collectively they touch the RHI public surface (`quadrants/rhi/public_device.h`), all three backend implementations, and every call site that threads a semaphore through. That was too large to bundle with #536 or with the heap-backed adstack PRs that actually drive ctx_buffers_ growth. This PR ships the minimum correctness-safe, bounded-growth approximation that does not require any RHI API changes. A follow-up PR can add `is_signaled()` and swap the `command_sync()` retirement line for per-entry polling.

### `quadrants/runtime/gfx/runtime.h`

- New `std::deque<std::pair<StreamSemaphore, std::vector<DeviceAllocationUnique>>> pending_retirements_` member.
- `static constexpr std::size_t kPendingRetirementsDepth = 3`.
- Field-level comment block documenting the bounded-wait rationale and the Ideal polling path as follow-up.
- Add `#include <deque>` and `#include <utility>` to the header.

### `quadrants/runtime/gfx/runtime.cpp`

- `flush()` — after the submit, if `pending_retirements_.size() >= kPendingRetirementsDepth`, call `command_sync()` and clear the FIFO. Then move the current `ctx_buffers_` into a new `pending_retirements_` entry paired with the just-returned semaphore.
- `synchronize()` — `wait_idle()` already guarantees every cmdlist that ever referenced a buffer in either collection has completed, so both `pending_retirements_` and `ctx_buffers_` are cleared unconditionally.

## Side-effect audit

| Concern                                   | Verdict |
| ----------------------------------------- | ------- |
| Use-after-free                            | Impossible: `pending_retirements_` entries only release their buffers after `command_sync()` or `wait_idle()`. |
| Unbounded growth in async workloads       | Bounded at `kPendingRetirementsDepth * ctx_buffers_per_flush` entries. |
| Latency                                   | `command_sync()` fires every `kPendingRetirementsDepth + 1` flushes without a user-sync. In steady-state this is a no-op; in sustained-async workloads it introduces a stall once per N flushes. Tolerable as the correctness-safe default; the polling follow-up removes the stall entirely. |
| Non-SPIR-V backends                       | Unaffected — `GfxRuntime` is SPIR-V only. LLVM backends use `LlvmRuntimeExecutor` which has its own retirement story (synchronous `hipFree` on AMDGPU via Autodiff 11, `cuMemFree_v2` auto-sync on CUDA). |

## Stack

Autodiff 14 of 14. Based on #493 (SPIR-V heap-backed adstack). End of the chain.

### DIFF
--- a/quadrants/runtime/gfx/runtime.cpp
+++ b/quadrants/runtime/gfx/runtime.cpp
@@ -734,6 +734,9 @@ void GfxRuntime::synchronize() {
       profiler_->insert_record(record.first, record.second);
     }
   }
+  // `wait_idle()` above guarantees every cmdlist that ever referenced a buffer in `pending_retirements_` or
+  // `ctx_buffers_` has completed on the GPU, so both can be dropped here unconditionally.
+  pending_retirements_.clear();
   ctx_buffers_.clear();
   ndarrays_in_use_.clear();
   // Async adstack-overflow report: every launch in this sync window that overflowed wrote a non-zero sentinel into
@@ -763,10 +766,24 @@ StreamSemaphore GfxRuntime::flush() {
   if (current_cmdlist_) {
     sema = device_->get_compute_stream()->submit(current_cmdlist_.get());
     current_cmdlist_ = nullptr;
-    // Do NOT clear ctx_buffers_ here: submit() returns as soon as the cmdlist is queued, not when the GPU has
-    // finished executing. The deferred-free buffers in ctx_buffers_ (e.g. the old adstack heap buffer left over
-    // after a grow-on-demand resize) may still be referenced by commands in flight. Only `synchronize()` clears
-    // the vector, after `wait_idle()` has drained the stream.
+    // Pair the just-queued semaphore with the current batch of deferred-free buffers and push onto the
+    // retirement FIFO. `ctx_buffers_` is then ready to accept the next submission's deferred-frees.
+    //
+    // Bound the FIFO at `kPendingRetirementsDepth` entries: when the queue is full we block on the oldest
+    // entry's semaphore, drain it, and pop it before pushing the new one. The wait is almost always a no-op
+    // in steady state (the oldest submission has long since completed by the time `kPendingRetirementsDepth`
+    // more flushes have piled on top of it), but it's a hard upper bound on how much deferred-free memory
+    // can be queued up in async workloads that never call `synchronize()` - which is precisely the path that
+    // allowed unbounded growth before this PR. See the field-level comment on `pending_retirements_` for the
+    // Ideal non-blocking variant and why it is deferred to follow-up work.
+    if (pending_retirements_.size() >= kPendingRetirementsDepth) {
+      device_->get_compute_stream()->command_sync();
+      pending_retirements_.clear();
+    }
+    if (!ctx_buffers_.empty()) {
+      pending_retirements_.emplace_back(sema, std::move(ctx_buffers_));
+      ctx_buffers_.clear();
+    }
   } else {
     auto [cmdlist, res] = device_->get_compute_stream()->new_command_list_unique();
     QD_ASSERT(res == RhiResult::success);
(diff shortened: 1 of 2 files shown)


duburcqa:
Closing per discussion: the unbounded-growth path requires a workload that calls `flush()` without ever touching any Python-side observable (result fetch, `to_numpy()`, etc.), which does not exist in Quadrants today. The FIFO-depth bound in this PR trades a theoretical cost for a real, measurable `command_sync()` stall at an arbitrary depth threshold — net-negative against the pre-existing behaviour that is already correctness-safe after #536. The proper fix (non-blocking per-semaphore polling, `bool is_signaled() const` on `StreamSemaphoreObject` with per-backend implementations) should stand alone if it is ever motivated by a real workload, not ride in behind a bot comment. A rationale comment is being added to the pre-existing `ctx_buffers_` code in #536 explaining why the current simple design is kept.

PR #542, opened 2026-04-21
[Build] Autodiff 15: Replace 2022 MoltenVK pin with LunarG Vulkan SDK fetch

duburcqa:
# LunarG-sourced MoltenVK on Apple, with buffer-device-address unblocked

> One commit, no behaviour change for users who don't opt into Apple Vulkan / Metal. Replaces the 2022 MoltenVK pin shipped by Taichi with a LunarG Vulkan SDK fetch driven by `./build.py --shell`, and drops the hard-disabled `spirv_has_physical_storage_buffer` (PSB) cap on Apple now that the fresh MoltenVK supports `VK_KHR_buffer_device_address` (BDA). Prerequisite for the runtime-evaluated adstack sizer on Metal / Vulkan (Autodiff 17).

## TL;DR

```bash
# macOS / arm64, from a clean checkout:
./build.py --shell -- cmake -S . -B build -DQD_WITH_VULKAN=ON
# $VULKAN_SDK / $MOLTENVK_DIR are exported by the shell hook;
# CMake's find_file locates libMoltenVK.dylib under the SDK and stages it into build/.
./build.py
python -c "import quadrants as qd; qd.init(arch=qd.vulkan)"
# Now reports spirv_has_physical_storage_buffer=True on Apple.
```

The shell hook fetches LunarG's macOS installer once, extracts the SDK into `~/.cache/quadrants/vulkan-macos-1.4.321.0/`, and exports `VULKAN_SDK` / `MOLTENVK_DIR`. CMake then picks up `libMoltenVK.dylib` from the SDK instead of Taichi's legacy pinned dylib. Downstream, the adstack sizer compute shader (Autodiff 17) relies on BDA to walk `SizeExpr` trees on device, so enabling PSB on Apple is the gating change that makes that shader legal to dispatch.

## Why

The previous Apple Vulkan path pinned a 2022 MoltenVK dylib hosted on `taichi_assets`, predating the `VK_KHR_buffer_device_address` / physical-storage-buffer capability. Two concrete consequences:

- `vulkan_device_creator.cpp` hard-guarded `spirv_has_physical_storage_buffer` off on Apple behind `#if !defined(__APPLE__) && false`, citing [taichi-dev/taichi#6295](https://github.com/taichi-dev/taichi/issues/6295).
- The adstack sizer shader that lands in Autodiff 17 needs `OpLoad` through a `PhysicalStorageBuffer`-class pointer to read `SizeExpr` leaves; with PSB disabled on Apple, every reverse-mode kernel would hard-error at launch time on Metal.

The less-targeted workarounds are insufficient: keeping the Taichi pin and papering over the capability check would ship a MoltenVK that cannot serve BDA loads; asking every Quadrants developer to install LunarG's SDK globally breaks hermetic CI. Fetching through `build.py --shell` gives us a single, reproducible SDK path the rest of the build consumes.

This PR replaces the pin with a LunarG-SDK fetch, drops the Apple PSB kill-switch, and fails the build loudly if neither `MOLTENVK_DIR` nor `VULKAN_SDK` is set.

## Entry point

`setup_vulkan()` in `.github/workflows/scripts/ti_build/vulkan.py`. Invoked by `./build.py --shell` (and by CI's `ti_build/entry.py`) before any CMake configure runs. It:

1. For `Darwin/arm64`, downloads LunarG's `vulkansdk-macos-1.4.321.0.zip` (installer bundle) into `~/.cache/quadrants/vulkan-macos-1.4.321.0-installer/`.
2. Runs the installer CLI non-interactively (`--accept-licenses --default-answer --confirm-command install`) into `~/.cache/quadrants/vulkan-macos-1.4.321.0/`.
3. Exports `VULKAN_SDK=$prefix/macOS`, prepends `$sdk/bin` / `$sdk/lib` to `PATH` / `DYLD_LIBRARY_PATH`, sets `VK_LAYER_PATH`, and exports `MOLTENVK_DIR=$sdk/lib` (the directory where LunarG drops `libMoltenVK.dylib` flat).

`quadrants/rhi/CMakeLists.txt` (Apple branch) reads `MOLTENVK_DIR` / `VULKAN_SDK`, `find_file`s `libMoltenVK.dylib` under those paths, and `configure_file`s it into `${CMAKE_BINARY_DIR}/libMoltenVK.dylib` + installs into `${INSTALL_LIB_DIR}/runtime/`. A missing SDK is a `FATAL_ERROR` with an inline remediation message.

### 1. SDK acquisition (`vulkan.py`)

| Platform | Source | Prefix |
| --- | --- | --- |
| Linux | `vulkansdk-linux-x86_64-1.4.321.1.tar.xz` (tarball, unchanged) | `~/.cache/quadrants/vulkan-1.4.321.1/x86_64/` |
| Darwin / arm64 | `vulkansdk-macos-1.4.321.0.zip` (installer bundle) | `~/.cache/quadrants/vulkan-macos-1.4.321.0/` |
| Windows | MSI (unchanged) | `~/.cache/quadrants/vulkan-win-1.4.321.1/` |

The macOS branch is the only new one. LunarG didn't publish a `1.4.321.1` macOS asset, so the patch-level is inlined to `1.4.321.0`. `zipfile` drops the installer bundle without preserving the Unix execute bit, so the script `chmod 0755`s the installer binary before running it (idempotent, scoped to the single file). The CLI `install` command writes the SDK into the `--root` prefix.

### 2. CMake pickup (`quadrants/rhi/CMakeLists.txt`)

| Env var | Meaning | Consumer |
| --- | --- | --- |
| `MOLTENVK_DIR` | path that directly contains `libMoltenVK.dylib` | `find_file(MOLTEN_VK libMoltenVK.dylib NO_DEFAULT_PATH PATHS ${MOLTENVK_DIR})` |
| `VULKAN_SDK` | SDK prefix; `${VULKAN_SDK}/lib` is tried if `MOLTENVK_DIR` is unset | same `find_file` call, fallback path |

`configure_file` stages the located dylib into `${CMAKE_BINARY_DIR}/libMoltenVK.dylib` (copy, not symlink, so the install step can re-digest it) and `install(FILES ... DESTINATION ${INSTALL_LIB_DIR}/runtime)` ships it alongside the runtime. A missing SDK is a `FATAL_ERROR` pointing at `./build.py --shell`; there is no silent fallback to the legacy pin on purpose.

### 3. PSB capability unblocked (`vulkan_device_creator.cpp`)

Removes the `#if !defined(__APPLE__) && false` kill-switch gate around `caps.set(DeviceCapability::spirv_has_physical_storage_buffer, true)`. The surrounding `CHECK_VERSION(1, 3) || buffer_device_address_feature.bufferDeviceAddress` gate is retained, so devices that still don't advertise BDA (ancient drivers, headless CI without Vulkan) remain safe. A multi-line `//` comment points at the Taichi issue the old kill-switch referenced and at this PR's rationale for dropping it.

### 4. `build.py --shell` plumbing (`entry.py`)

`entry.py` already threaded `setup_vulkan()` through the shell entry-point. The diff here adds an `--shell` kwarg pass-through so the macOS branch's additional env var exports (`MOLTENVK_DIR`, `VK_LAYER_PATH`) survive into the child process the user lands in.

| Backend | Affected by this PR? | Verdict |
| --- | --- | --- |
| CPU (LLVM) | No | N/A - does not compile Vulkan RHI |
| CUDA (LLVM) | No | N/A - does not compile Vulkan RHI |
| AMDGPU (LLVM) | No | N/A - does not compile Vulkan RHI |
| Metal (SPIR-V) | Yes | MoltenVK is now sourced from LunarG; PSB capability is on. Validated via CI `Test on Mac (15, 3.12)` / `(26, 3.12)` + local `qd.init(arch=qd.vulkan)` smoke. |
| Vulkan on Linux (SPIR-V) | No (already LunarG-sourced) | No behaviour change. `.tar.xz` branch in `vulkan.py` untouched. |
| Vulkan on Windows (SPIR-V) | No | MSI branch in `vulkan.py` untouched. |

### CI

- `Test on Mac (15, 3.10 / 3.11 / 3.12 / 3.13)` and `Test on Mac (26, 3.10 / 3.11 / 3.12 / 3.13)` exercise the new fetch end-to-end: if the installer fails, if the SDK is missing `libMoltenVK.dylib`, or if the `FATAL_ERROR` branch triggers, the build step fails visibly.
- `Manylinux wheel Build/Test (ubuntu-22.04 / ubuntu-22.04-arm)` validates that the Linux branch of `vulkan.py` is unchanged - these jobs still pick up the same LunarG tarball.
- `Windows 2025 Build/Test (3.10-3.13)` validates that the Windows branch of `vulkan.py` is unchanged.

### Local smoke

- `./build.py --shell -- cmake -S . -B build -DQD_WITH_VULKAN=ON && ./build.py` on macOS-26 / arm64 succeeds and stages `libMoltenVK.dylib` into `build/`.
- `python -c "import quadrants as qd; qd.init(arch=qd.vulkan); print(qd.lang.impl.current_cfg().spirv_has_physical_storage_buffer)"` reports `True` after this PR; reports `False` before it.
- `CMAKE_BUILD_TYPE=Debug cmake --log-level=DEBUG` shows the `MoltenVK: using LunarG Vulkan SDK copy at ...` status line.

No unit tests are added: the change is a build-system rewire + a capability-bit flip, both of which surface as CI build failures if regressed.

## Side-effect audit

| --- | --- | --- |
| Linux / Windows Vulkan SDK fetch | `vulkan.py` `case ("Linux", "x86_64")` / `case ("Windows", "AMD64")` branches untouched | ok - no behaviour change outside Apple |
| `VULKAN_SDK` env var semantics | exported from the shell hook; consumed by `quadrants/rhi/CMakeLists.txt` (BSD find path), `quadrants/rhi/vulkan/vulkan_device_creator.cpp` (runtime loader), shader compiler `glslang` lookup | ok - existing consumers keep using `$VULKAN_SDK`; the new macOS prefix looks identical in layout |
| `MOLTENVK_DIR` env var | new. Only read by `quadrants/rhi/CMakeLists.txt`; no runtime lookup | ok - opt-in; falls back to `$VULKAN_SDK/lib` |
| CMake cache | `find_file(MOLTEN_VK ...)` is cached; after the SDK is installed the first configure populates it and subsequent configures skip | ok - deleting `~/.cache/quadrants/vulkan-macos-.../` + `rm -rf build` regenerates from scratch |
| BDA gate in `vulkan_device_creator.cpp` | preserved as `CHECK_VERSION(1, 3) \|\| buffer_device_address_feature.bufferDeviceAddress` outer check | ok - headless CI drivers that don't advertise BDA will not get PSB set |
| Non-Apple behaviour | Apple-guarded block (`if (APPLE)`) in `CMakeLists.txt`; Apple-guarded cap in `vulkan_device_creator.cpp` | ok - Linux / Windows Vulkan paths untouched |
| Download cache | `download_dep(url, installer_dir, strip=1)` uses the existing cache primitive; re-runs short-circuit on cached unzip + on the existence of `$prefix/macOS/` | ok - no network on re-run |
| Installer execute bit | `installer_bin.chmod(0o755)` before `subprocess.check_call`; idempotent | ok - `zipfile` dropped mode `0644`, handled here |
| `$VK_LAYER_PATH` | set on every platform branch including the new macOS one | ok - layer validation remains wired |
| Legacy Taichi pin removal | `CMakeLists.txt` previously `curl`-ed `libMoltenVK.dylib.zip` from `taichi_assets` - removed. `FATAL_ERROR` replaces the silent fallback. | intentional; no legacy fallback is shipped |


hughperkins:
Questions:
1. do we need any doc changes?
2. what expectations should a user have had previously, that change now?

duburcqa:
> the earlier comment in the code in question suggests installing moltenvk sdk using brew what do you see as the good and bad points of the two posible approaches (ie downloading from lunarg, vs using hte homebrew version?)

I'm not a huge fan of allowing system dependencies. By forcing our own version of MoltenVK, we can guarantee that it works. We do not support any other version than the one shipping with Quadrants and I don't think we want to explore such opportunity. If a dev wants to try some specific version, it is very easy to change it in the CMake file and delete the cache to force re-downloading the updated version. So I don't see any limitation for not supporting system-wise homebrew version. Except that, well, it forces downloading something, so it takes more time to bootstrap build env for the first time.

duburcqa:
Closing to re-open on the correctly-named head branch `duburcqa/moltenvk_sdk_source`; see replacement PR.

PR #543, opened 2026-04-22
[AutoDiff] Autodiff 16: Resolve reverse-mode adstack depths per-launch via runtime-evaluated SizeExpr

duburcqa:
# Runtime-evaluated adstack sizing on LLVM backends

> Makes reverse-mode adstack capacity automatic on CPU / CUDA / AMDGPU: the compiler captures a symbolic upper-bound expression per adstack alloca and the host launcher resizes the per-thread heap just before each dispatch. No user-facing API change; `default_ad_stack_size` / `ad_stack_size` remain accepted in `qd.init()` and the runtime falls back to `default_ad_stack_size` (with a `QD_WARN`) when an adstack's enclosing loop shape is outside the `SizeExpr` grammar. Shapes inside the grammar are sized exactly at launch time.

## TL;DR

```python
import quadrants as qd

qd.init(arch=qd.cpu, ad_stack_experimental_enabled=True)

n = qd.field(qd.i32, shape=())
x = qd.field(qd.f32, shape=16, needs_grad=True)
y = qd.field(qd.f32, shape=(), needs_grad=True)

@qd.kernel
def compute():
    for i in x:
        v = x[i]
        for _ in range(n[None]):           # trip count known only at launch time
            v = v * 0.95 + 0.01
        y[None] += v

n[None] = 512
compute()
y.grad[None] = 1.0
compute.grad()                              # no QuadrantsAssertionError: adstack sized to 512 + 2
```

Before this PR, the inner `range(n[None])` forced the CFG analyzer into its `default_ad_stack_size=256` fallback, so `n[None] = 512` overflowed at the next `qd.sync()`. After this PR the compiler captures the loop bound as a `SizeExpr`, the launcher evaluates it against `n[None]` at launch time, and the backing heap is grown to exactly `512 + 2` slots.

## Why

Reverse-mode AD through a dynamic loop requires an adstack whose depth equals the forward trip count plus two setup pushes. Until now Quadrants had to resolve that depth at compile time, so every loop bound that involved a field load (`n_bodies`, `n_dofs`, an ndarray shape, an enclosing parallel-for index) fell back to the user-configured `default_ad_stack_size`. For realistic reverse-mode kernels this is both:

- **Wrong-too-small**: a 14-dof mechanism stepping through an `ndrange(n_bodies, n_dofs)` easily needs >256 slots, triggering a runtime `QuadrantsAssertionError: Adstack overflow`.
- **Wrong-too-big**: bumping the default to cover worst case inflates every adstack in the program, including the ones the compiler *could* prove were tiny, and scales linearly with the dispatched ndrange on GPU.

The traditional workaround - tune `default_ad_stack_size` per program - is not tunable enough: you cannot set a different default for every adstack, and bumping it globally multiplies GPU memory by the number of threads. This PR replaces the static fallback with a symbolic upper-bound expression captured at compile time and reduced per launch against the live field / ndarray state, so each adstack gets exactly the depth that dispatch needs.

## Entry point

`LlvmRuntimeExecutor::publish_adstack_metadata(const AdStackSizingInfo &ad_stack, std::size_t num_threads, LaunchContextBuilder *ctx, void *device_runtime_context_ptr = nullptr)` at `quadrants/runtime/llvm/llvm_runtime_executor.cpp`. Called from `cpu/cuda/amdgpu/kernel_launcher.cpp` right before every task's dispatch. It:

1. Walks the task's `size_exprs` vector (post-order `SerializedSizeExpr` trees, one per `AdStackAllocaStmt`).
2. Evaluates each tree via `adstack_size_expr_eval.cpp`, reading scalar field values through `SNodeRwAccessorsBank` and launch-context ndarray shapes / scalar reads through the `LaunchContextBuilder`. The AMDGPU path threads the launcher's device-side `RuntimeContext` staging copy (`kernel_launcher.cpp:185-186`) through `device_runtime_context_ptr` so HIP kernels that resolve `ExternalTensorRead` leaves against ndarray pointers do not fault on a host pointer; CUDA relies on UVA and passes `nullptr`; CPU ignores the argument.
3. Computes per-thread stride, per-alloca offsets, and per-alloca max-sizes.
4. Writes these into `LLVMRuntime::adstack_{per_thread_stride,offsets,max_sizes}` via `runtime_get_adstack_metadata_field_ptrs` so generated code reads them at every push / load-top site.
5. Grows the per-thread heap backing buffer via `ensure_adstack_heap(per_thread_stride * dispatched_threads)`.

Shapes the grammar covers are sized exactly. Shapes outside the grammar (e.g. a `range(field[i])` / `range(ndarray[i])` indexed by a loop-carried variable that the structural walker does not fold) fall back to `default_ad_stack_size` and emit a `QD_WARN` naming the source location so the user can either restructure the loop or extend the grammar.

### 1. Compile-time `SizeExpr` capture

`quadrants/transforms/determine_ad_stack_size.cpp` runs a Bellman-Ford-first, structural-walker-second pre-pass that attempts to express each `AdStackAllocaStmt`'s max depth as a `SizeExpr` tree. Grammar covered:

| Leaf | Source | Lowering |
| --- | --- | --- |
| Const | integer literal | `SizeExpr::make_const` |
| FieldLoad | `GlobalLoadStmt` of a scalar `i32` / `i64` SNode | `make_field_load` with the SNode pointer + constant index path |
| NdarrayShape | `ExternalTensorShapeAlongAxisStmt` | `make_ndarray_shape(arg_id, axis)` |
| NdarrayRead | scalar read from an integer ndarray | `make_ndarray_read(arg_id, index_path)` |
| LoopIndexMax | `LoopIndexStmt` of an enclosing bounded range-for / parallel-for | `make_loop_index_max` (bounded by the loop's `end` or a `max` over every push site) |
| StackLoadTopMax | `AdStackLoadTopStmt` of a stash adstack | `make_stack_load_top_max` (bounded by a `max` over every push site) |

Inner operators: `add`, `sub`, `mul` (const * anything via `Add`-replication), `max`. `expr_sub`'s MaxOverRange fusion caps the fused iteration range at `min(shape_a, shape_b)` (expressed in-grammar as `a + b - max(a, b)`) so the fused body cannot read the shorter ndarray past its buffer end when a caller passes mismatched-length ndarrays. Anything else falls back to `default_ad_stack_size` and emits the `QD_WARN` noted above. `SizeExpr` objects are immutable after the pre-pass writes them and are serialised as post-order `SerializedSizeExpr::nodes` so offline-cache hits evaluate the same way as fresh compiles.

### 2. Codegen reads stride / offset / max-size from runtime metadata

| File | Change |
| --- | --- |
| `codegen_llvm.cpp` `AdStackAllocaStmt` | base addr = `heap + linear_tid * runtime->adstack_per_thread_stride + runtime->adstack_offsets[stack_id]` (all three loaded at runtime) |
| `codegen_llvm.cpp` `AdStackPushStmt` | `stack_push(runtime, stack, runtime->adstack_max_sizes[stack_id], element_size)` (cap is a runtime load, not an IR immediate) |
| `init_offloaded_task_function` pre-scan | assigns `AdStackAllocaStmt::stack_id` in declaration order; `ad_stack_offsets_` became `std::vector<std::size_t>` indexed by `stack_id` |

The heap base address is unchanged; only the per-thread stride and per-alloca offset math was moved from codegen-time immediates to runtime loads, so kernels compiled with the old flow will recompile but do not break semantics.

> Note: `default_ad_stack_size` and `ad_stack_size` are ignored on LLVM for adstacks whose bound is inside the grammar. They remain in `qd.init()` for SPIR-V backwards compatibility and for the warn-and-fallback path where the pre-pass cannot resolve the loop shape symbolically.

### 3. Host evaluator

`adstack_size_expr_eval.{h,cpp}` walks the flat post-order tree:

- Field loads go through the existing `SNodeRwAccessorsBank` to read the current scalar value on the host side.
- Ndarray shape / scalar reads go through `LaunchContextBuilder` which captures both the pointer and the shape on kernel launch.
- `LoopIndexMax` and `StackLoadTopMax` fold to whichever launch-time bound they were captured against.
- Arithmetic is saturating `int64_t`.

The evaluator runs strictly on the host, before the kernel dispatches, so there is no device-side evaluator to keep in sync.

### 4. Per-kernel metadata bridge

`runtime_get_adstack_metadata_field_ptrs(LLVMRuntime *)` in `runtime.cpp` returns pointers to `adstack_{per_thread_stride,offsets,max_sizes}`. The host writes into these fields once per dispatch. On CPU the fields are read directly from `LLVMRuntime`; on CUDA / AMDGPU the bridge triggers a `memcpy_host_to_device` so the generated code's subsequent loads see the right values. The memcpy is gated on the task actually carrying an adstack: kernels with no reverse-mode adstacks skip `publish_adstack_metadata` entirely and pay no per-launch cost.

| Backend | Pre-pass | Metadata bridge | Launcher hook | Status |
| --- | --- | --- | --- | --- |
| CPU (LLVM) | ok | direct field access | `cpu/kernel_launcher.cpp` | auto-sized inside the grammar; `default_ad_stack_size` fallback + `QD_WARN` outside |
| CUDA (LLVM) | ok | `memcpy_host_to_device` | `cuda/kernel_launcher.cpp` | same |
| AMDGPU (LLVM) | ok | `memcpy_host_to_device` (host pointer routed through the launcher's device-side `RuntimeContext` staging copy) | `amdgpu/kernel_launcher.cpp` | same |
| Metal (SPIR-V) | N/A here | N/A here | N/A here | untouched - still compile-time, handled by the stacked SPIR-V sizer PR |
| Vulkan (SPIR-V) | N/A here | N/A here | N/A here | untouched - still compile-time, handled by the stacked SPIR-V sizer PR |

### `tests/python/test_adstack.py`

- `test_adstack_field_load_bounded_loop_evaluated_per_launch` - pins the core guarantee: a kernel with `for _ in range(n[None])` bound by an i32 scalar field. Launches at `n_iter in (1, 20, 50)` and asserts the heap resized correctly (no overflow at 50, no over-allocation at 1), with the decorator set to `default_ad_stack_size=2` to prove the knob is bypassed when the grammar resolves the bound symbolically.
- `test_adstack_near_capacity` - rewritten to exercise both sides of the old `default_ad_stack_size=32` boundary at `n_iter in (30, 100)` without any knob override. Arch-restricted to `qd.cpu` until the SPIR-V sizer PR lands.
- `test_adstack_structural_pre_pass_fuses_sub_of_max_over_range_with_matching_shape_ends` / `..._with_mismatched_shape_ends` - pin the `expr_sub` MaxOverRange fusion for both the strict-equality and the cross-ndarray same-axis paths.
- `test_adstack_fuses_sub_of_max_over_range_with_mismatched_lengths_is_safe` - pins the `min(shape_a, shape_b)` end cap so the cross-ndarray fusion stays in-bounds when the two ndarrays have different lengths at launch time (pre-fix the fused body read the shorter ndarray past its buffer end).
- `test_adstack_inner_range_bounded_by_ndarray_read_at_outer_index` / `test_adstack_inner_range_bounded_by_multidim_ndarray_read` / `test_adstack_ext_tensor_read_indexed_by_stashed_outer_loop_var` - cover the `ExternalTensorRead` grammar leaves, including multi-axis stride handling and the stash-then-reload pattern.

### `tests/cpp/transforms/determine_ad_stack_size_test.cpp`

- `SizeExprBuildFromRangeFor` - const bounds fold to a `Const` leaf.
- `SizeExprBuildFromFieldLoad` - scalar i32 field load builds a `FieldLoad` leaf with the right SNode + index path.
- `SizeExprBuildFromLoopIndex` - inner `for j in range(i)` under an enclosing bounded loop folds to `LoopIndexMax`.
- `SizeExprRejectsUnsupportedOp` - negative control that an unsupported operator routes through the warn-and-fallback path.

Deliberate coverage gap: no SPIR-V cross-check here - the SPIR-V runtime metadata path lives in the stacked SPIR-V sizer PR and its tests cover the shader-side reading.

## Side-effect audit

| --- | --- | --- |
| Offline cache key | `quadrants/analysis/offline_cache_util.cpp` serialises `SerializedSizeExpr::nodes` alongside the existing `size_in_bytes` per `AdStackAllocaStmt` | ok - cache-hit reloads the same tree; evaluator runs the same on fresh + cache-hit. Empty-tree caches (pre-PR offline caches) fall back to `default_ad_stack_size`. |
| Stmt clone / serialization | `statements.h`: `AdStackAllocaStmt::size_expr` is a `std::shared_ptr<SizeExpr>` cloned via the default `QD_DEFINE_ACCEPT_AND_CLONE` | ok - `SizeExpr` is immutable post pre-pass; shared pointer is safe to share across clones |
| IR printer | `ir/ir.cpp` `AdStackAllocaStmt::repr` prints `stack_id` and a short `SizeExpr` summary | ok - IR dump readable |
| Codegen non-feature path | `AdStackAllocaStmt::stack_id == -1` falls through to the old `default_ad_stack_size`-backed path | ok - kernels compiled without the pre-pass writing a `SizeExpr` still behave as before |
| Whole-kernel CSE | `SizeExpr` is stored on `AdStackAllocaStmt`, not emitted as IR operands, so it does not participate in CSE | ok |
| LaunchContextBuilder | extended to capture scalar + shape for every ndarray arg; no change to the existing argument-pack layout - new entries appended | ok - old kernels still build their context the same way |
| Runtime initialization | `adstack_{per_thread_stride,offsets,max_sizes}` initialised to zero / null in `runtime_initialize`; kernels with no adstacks read nothing from them and skip `publish_adstack_metadata` entirely | ok - zero-adstack kernels pay nothing |
| Multi-threaded CPU launch | metadata is written once per dispatch before the thread pool starts; no concurrent writers | ok |
| CUDA / AMDGPU `memcpy_host_to_device` | one memcpy per adstack-bearing task per launch; measured overhead ~microseconds. Forward-only kernels and reverse kernels whose pre-pass resolves all bounds at compile time are exempt. | acceptable; would be visible only on workloads that launch thousands of tiny autodiff kernels per second |
| Mismatched-length ndarray fusion | `expr_sub`'s cross-ndarray fusion caps the fused end at `min(shape_a, shape_b)` | ok - pre-fix UB on mismatched shapes; post-fix both reads stay in-bounds |

-----------------------------------------------

# Runtime-evaluated SizeExpr for SPIR-V adstack sizing

> Stacked on #550. No behaviour change for kernels that do not opt into the experimental adstack path (`ad_stack_experimental_enabled=True`). Mirrors #550's LLVM runtime-metadata plumbing onto SPIR-V so Metal and Vulkan kernels size their per-dispatch adstack heaps from launch-time-evaluated `SizeExpr` trees, tightening the shader-baked per-thread stride / offset / max_size from compile-time immediates to runtime loads out of a new `AdStackMetadata` buffer. Also tightens the #550-layer warn-and-fallback into a `QD_ERROR` on unresolved grammar shapes so both backends converge on a single "inside-grammar is sized exactly, outside-grammar is a compile error" contract.

## TL;DR

```python
@qd.kernel
def compute(a: qd.types.ndarray(dtype=qd.i32, ndim=1)):
    for i in x:                                    # outer struct-for
        v = x[i]
        n = a[i]                                    # ndarray-read-at-loop-index bound
        for _ in range(n):                          # dynamic inner range
            v = v * 0.95 + 0.01
        y[None] += v
```

On LLVM backends (#550) the reverse-mode adstack for `v` is already sized per-launch from the evaluated `SizeExpr` tree. This PR extends the same behaviour to the Metal and Vulkan backends: `compute.grad(a)` now sizes the per-thread heap to exactly `n = a[i]` slots per outer iteration on SPIR-V too, where before it either baked the compile-time fallback into the shader or hit a `max_size > 0` codegen assert on dynamic bounds.

## Why

#550 landed `SizeExpr`-at-launch-time sizing for LLVM and seeded a `default_ad_stack_size` compile-time fallback into the SPIR-V immediates so the existing SPIR-V codegen kept asserting cleanly. That seed is safe but wasteful: the SPIR-V shader's per-thread heap slice still has to be provisioned at the worst-case `default_ad_stack_size` for every alloca that was not statically const-foldable, even when the live field values would give a much tighter bound at launch. Kernels like the one above allocate `n_outer_iters * default_ad_stack_size * 8 bytes` per loop-carried variable on the device; at `default_ad_stack_size=256` with a few dozen allocas that is tens of MB per thread on Metal - crossing the driver's `maxBufferLength` on anything larger than a toy workload.

The LLVM infrastructure is a complete blueprint: `AdStackAllocaStmt::size_expr` is captured during `determine_ad_stack_size`, `OffloadedTask::ad_stack.allocas[i].size_expr` persists through the offline cache, and `LlvmRuntimeExecutor::publish_adstack_metadata` evaluates each tree right before each dispatch. What was missing on SPIR-V was the final per-launch metadata hand-off to the shader. This PR adds it: one new `BufferType::AdStackMetadata` SSBO, one new `TaskAttributes::ad_stack` struct carrying per-alloca metadata plus serialized `SizeExpr`, one `GfxRuntime::launch_kernel` prelude that evaluates the trees and writes the buffer, and codegen rewrites every `AdStackAllocaStmt` / `Push` / `LoadTop` / `AccAdjoint` site to read its stride / offset / max_size from that buffer instead of emitting compile-time immediates.

This PR also tightens the grammar-unresolved path on both backends: where #550 falls back to `default_ad_stack_size` with a `QD_WARN`, this PR flips the fallback to a `QD_ERROR` so any adstack outside the grammar is a compile-time error naming the source location. Shapes inside the grammar are sized exactly per launch; shapes outside the grammar are rejected; the `ad_stack_size=N` escape hatch in `qd.init()` is retained for stress tests and for bypassing the sizer when debugging.

## Entry point

The adstack pre-pass `determine_ad_stack_size` already runs for all backends (gated on `autodiff_mode == kReverse && ad_use_stack`) and populates `AdStackAllocaStmt::size_expr`; this PR only wires SPIR-V codegen and the `GfxRuntime` launcher to honour that metadata. The relevant files:

- `quadrants/codegen/spirv/kernel_utils.h` - adds `TaskAttributes::AdStackSizingAttribs` / `AdStackAllocaAttribs` (replaces the scalar `ad_stack_heap_per_thread_stride_{float,int}` fields) and `BufferType::AdStackMetadata`.
- `quadrants/codegen/spirv/spirv_codegen.cpp` - serialises each alloca's `size_expr` into `task_attribs_.ad_stack.allocas`; emits runtime loads for stride / offset / max_size via `ensure_ad_stack_metadata_loaded`.
- `quadrants/codegen/spirv/adstack_sizer_shader.cpp` / `.h` - the on-device compute shader that evaluates each task's `SizeExpr` bytecode tree and writes per-task stride / offset / max_size into the `AdStackMetadata` buffer.
- `quadrants/runtime/gfx/runtime.cpp::GfxRuntime::launch_kernel` - evaluates every task's `SizeExpr` tree upfront (before the cmdlist opens, to serialise against any reentrant `SNodeRwAccessorsBank` reader-kernel launches), then uploads per-task metadata to the `AdStackMetadata` buffer and binds it alongside the existing heap buffers.
- `quadrants/runtime/gfx/runtime.h::GfxRuntime::Params::program_impl` - propagates the `ProgramImpl` back-reference so `evaluate_adstack_size_expr(...)` can reach `ProgramImpl::program` (same pattern the LLVM side uses).

### 1. Codegen: serialise per-alloca metadata into `TaskAttributes`

| Layer | File | What's new |
|---|---|---|
| Per-alloca attribs | `codegen/spirv/kernel_utils.h` | `AdStackAllocaAttribs { heap_kind, offset_in_elems_compile_time, max_size_compile_time, SerializedSizeExpr size_expr }` |
| Per-task attribs | `codegen/spirv/kernel_utils.h` | `AdStackSizingAttribs { per_thread_stride_{float,int}_compile_time, allocas[] }` (replaces the two scalar `ad_stack_heap_per_thread_stride_*` fields) |
| Capture at alloca site | `codegen/spirv/spirv_codegen.cpp::visit(AdStackAllocaStmt)` | `attribs.size_expr = stmt->size_expr->serialize()` when the pre-pass populated it |

The pre-scan that sums per-type strides stays put; it just seeds the compile-time fallback values that the host uses when `size_expr` is empty (offline-cache-hit path without a serialised tree, i.e. pre-PR offline caches).

### 2. Codegen: runtime loads at every adstack site

| Site | Before | After |
|---|---|---|
| `invoc_id * stride` at `get_ad_stack_heap_thread_base_{float,int}` | `OpIMul(invoc_id, uint_immediate(stride))` | `OpIMul(invoc_id, OpLoad(metadata_buffer[0 or 1]))` |
| Per-alloca `offset` at `ad_stack_heap_{float,int}_ptr` | `uint_immediate(heap_primal_offset)` | `OpLoad(metadata_buffer[2 + 2*stack_id])` (cached per alloca on `info.offset_val`) |
| `max_size` at `AdStackPushStmt` and `ad_stack_top_index` | `uint_immediate(stmt->max_size)` / `uint_immediate(max_size - 1)` | `OpLoad(metadata_buffer[2 + 2*stack_id + 1])` (cached on `info.max_size_val`) + one `OpISub` for the `max - 1` cap |
| Adjoint offset for f32 adstacks | `uint_immediate(primal + max_size)` | `OpIAdd(info.offset_val, info.max_size_val)` - one derived add instead of a second buffer load |

Each `AdStackSpirv::info` caches its SSA ids for `offset_val` / `max_size_val` / `adjoint_offset_val` on first use and reuses them at every push / load / load_top_adj / acc_adjoint site, so each alloca pays one metadata read regardless of how many push/pop sites it has in the shader body.

> Note: the `invoc_id * stride` OpIMul stays eagerly emitted at the first alloca site (not lazily at the first push) so it dominates every sibling inner-loop body - the same SPIR-V dominance rule #493's original heap-backing work already enforced. The only change is the stride now comes from `OpLoad(metadata[0])` instead of `uint_immediate(...)`.

### 3. Sizer shader capability gate + scope-state safety

`build_adstack_sizer_spirv` hard-errors through the empty-return path unless the device advertises `spirv_has_physical_storage_buffer`, `spirv_has_int64`, `spirv_has_int8`, and `spirv_has_int16`. The `emit_psb_load_i64` per-dtype switch calls `ir.i{8,16}_type()` unconditionally - without those capabilities the accessors return default-constructed `SType(id=0)` and `spirv-val` rejects the binary - so gating on all four surfaces a clear "legacy device missing a required hardware feature" diagnostic at launcher time instead of an invalid-bytecode rejection at pipeline creation. `GfxRuntime::launch_kernel` mirrors the gate so the launcher's error message and the shader builder's return agree.

The shader body also clears `scope[pending_var_id]` to zero on every `MaxOverRange` pop (`done_lbl`). `scope_arr` is a single function-storage alloca zero-initialised once at `main()` entry, but `var_id_counter` resets per alloca, so different stacks in the same task reuse `scope[0]`, `scope[1]`, and so on. Without the pop-path clear, the NEXT stack's outer linear pre-order walk reads the previous stack's terminal bound value as an `ExternalTensorRead` index into a potentially-smaller ndarray - out-of-bounds on Metal (hung command buffer) or silently-zero on Vulkan with `robustBufferAccess`. The clear preserves the "index 0 is always valid for any non-empty ndarray" invariant the outer walk relies on.

### 4. Runtime: upfront metadata evaluation + buffer upload

`GfxRuntime::launch_kernel` evaluates every task's `SizeExpr` trees **before** `ensure_current_cmdlist()` opens the per-launch cmdlist. This is the load-bearing ordering change: the evaluator can call `SNodeRwAccessorsBank::read_int(...)` to resolve a field-load-bounded expression, which itself launches a reader kernel through `GfxRuntime::launch_kernel` - recursing into this same function with the outer cmdlist half-built would stomp `current_cmdlist_`. Upfront evaluation keeps the reader kernels serialised against any in-flight cmdlist through the normal submit/synchronize path.

The metadata buffer is allocated **per task** (not pooled across the launch) so a sibling task's host memcpy cannot race-overwrite this task's values between cmdlist record and cmdlist submit. Each per-task buffer is retired into `ctx_buffers_` so it stays alive for the full submit/sync window. Layout:

```
u32[0]                 = stride_float
u32[1]                 = stride_int
u32[2 + 2*i + 0]       = alloca i's offset in its heap (elements)
u32[2 + 2*i + 1]       = alloca i's max_size
```

Host write happens through a `device_->map_range(...)` memcpy; unmap before bind. The shader reads via the same `BufferType::AdStackMetadata` access-chain plumbing as every other StorageBuffer in the task's bind set.

### 5. Program <-> runtime plumbing

`GfxRuntime::Params` gains a `ProgramImpl *program_impl` field, set by `MetalProgramImpl::materialize_runtime` and `VulkanProgramImpl::materialize_runtime` to `this`. The runtime uses it to reach `program_impl_->program` for `evaluate_adstack_size_expr(...)`. `Program::Program` already sets `program_impl_->program = this` post-`materialize_runtime`, so the back-reference is non-null by the time any `launch_kernel` runs.

| Backend | Pre-pass (from #550) | Codegen runtime loads | Per-launch evaluation | Heap grow on stride change |
|---|---|---|---|---|
| CPU (LLVM) | yes | yes (from #550) | yes (from #550) | yes (from #550) |
| CUDA (LLVM) | yes | yes (from #550) | yes (from #550) | yes (from #550) |
| AMDGPU (LLVM) | yes | yes (from #550) | yes (from #550) | yes (from #550) |
| Metal (SPIR-V) | yes | yes (this PR) | yes (this PR) | reused from #493's heap-grow path |
| Vulkan (SPIR-V) | yes | yes (this PR) | yes (this PR) | reused from #493's heap-grow path |

The SPIR-V backends also continue to honour the `max_size_compile_time` fallback on the offline-cache-hit path where the symbolic `SizeExpr` tree was not serialised (empty `size_expr.nodes`); the runtime falls back to it verbatim, clamping to `>= 1` like the LLVM counterpart does.

### New tests

- `test_adstack_bounded_inner_loop_sized_by_structural_prepass` (`tests/python/test_adstack.py`): pins the SPIR-V structural pre-pass correctness - a kernel whose inner range-for trip count is a constant product that the pre-pass must fold while the CFG analyzer would fall back to a pessimistic default. Arch-restricted to Metal / Vulkan because the LLVM backend routes the same kernel through the symbolic-tree path instead.
- `test_adstack_spirv_metadata_per_task_buffer` (`tests/python/test_adstack.py`, `cfg_optimization=False`): pins the per-task `AdStackMetadata` buffer fix - with a shared buffer the second offload's host memcpy overwrites the first offload's metadata before submit, causing an `Adstack overflow (offending stack_id=0)` at `qd.sync()` even though each stack's sizer-computed bound is correct. `cfg_optimization=False` is load-bearing; it keeps the two sibling offloads separate so the record-then-execute race surfaces.
- `AdStackSizerShader.GateReturnsEmptyWhenRequiredCapIsMissing` (`tests/cpp/codegen/adstack_sizer_shader_test.cpp`): pins the capability gate - dropping any one of PSB, Int64, Int8, Int16 from a synthetic caps config flips `build_adstack_sizer_spirv` to the empty-return path.
- `AdStackSizerShader.DumpBinary` (`tests/cpp/codegen/adstack_sizer_shader_test.cpp`): sanity-only smoke test that dumps the compiled binary to `/tmp/adstack_sizer.spv` for local `spirv-val` / `spirv-dis` use.

### Widened arch coverage

Existing tests that exercised the runtime-evaluated adstack on LLVM only now run on SPIR-V too (`test_adstack_field_load_bounded_loop_evaluated_per_launch`, `test_adstack_inner_range_bounded_by_ndarray_read_at_outer_index`, `test_adstack_ext_tensor_read_indexed_by_stashed_outer_loop_var`, `test_adstack_structural_pre_pass_fuses_sub_of_max_over_range_with_matching_shape_ends`, `test_adstack_structural_pre_pass_fuses_sub_of_max_over_range_with_mismatched_shape_ends`) by dropping the `arch=[qd.cpu, ...]` restriction that scoped them to the LLVM sizer path.

### Known coverage gaps

No dedicated Python regression test for the `scope[pending_var_id]` pop-path clear. The failure mode is hardware-dependent: on Metal / MoltenVK and Vulkan-with-`robustBufferAccess` the spurious pre-pass out-of-bounds read is either caught silently or returns zero, both of which `pending_max_accum` (which starts at 0 at every MOR push) absorbs without surfacing. A test that fails deterministically would need a Vulkan device without robust buffer access or a simulated evaluator. The fix is mechanism-verified through code review and covered indirectly by the full adstack suite running through the sizer.

Full suite: `pytest tests/python/test_adstack.py -n 8` -> 712 passed, 7 xfailed on macOS arm64 (CPU + Metal + Vulkan / SwiftShader); wider backend coverage runs through the Linux AMDGPU box.

## Side-effect audit

|---|---|---|
| Offline cache key | `TaskAttributes::QD_IO_DEF(..., ad_stack)` via the new `AdStackSizingAttribs::QD_IO_DEF(per_thread_stride_{float,int}_compile_time, allocas)` and `AdStackAllocaAttribs::QD_IO_DEF(heap_kind, offset_in_elems_compile_time, max_size_compile_time, size_expr)` | Safe - any schema change (field rename / reordering) invalidates the cache at the serializer level, same as every other `TaskAttributes` change. Pre-PR caches without the new fields are invalidated on first load and recompiled. |
| Buffer binding numbering | `TaskCodegen::binding_head_` unchanged; `AdStackMetadata` is appended lazily via `get_buffer_value(...)` like every other buffer kind, so existing bindings (Args, Rets, Root, ExtArr, AdStackOverflow, AdStackHeap*) keep their bindings | Safe - the metadata buffer lands at a fresh binding slot per task. |
| Tasks with no adstack | `attribs.ad_stack.allocas.empty()` short-circuits both the runtime evaluation loop and the metadata buffer allocation; codegen never calls `get_ad_stack_metadata_buffer()` in that case, so no `BufferType::AdStackMetadata` entry is added to `buffer_binds` | Safe - forward-only kernels and reverse kernels with only const-resolved allocas are unaffected. |
| Mid-cmdlist recursion | Metadata evaluation moved to before `ensure_current_cmdlist()` in `launch_kernel` | Safe - `SNodeRwAccessorsBank::read_int` reader-kernel launches now go through the normal submit/sync path; no partially-built cmdlist to stomp. |
| Legacy `max_size` immediates | The pre-scan still computes `ad_stack_heap_per_thread_stride_{float,int}_` as a compile-time sum and carries them into `AdStackSizingAttribs::per_thread_stride_{float,int}_compile_time`; the runtime uses those when every alloca's `size_expr.nodes.empty()` (pre-PR offline cache hit) | Safe - the cache-hit path reproduces the pre-PR shader layout. |
| Dominance of `invoc_id * stride` | `get_ad_stack_heap_thread_base_{float,int}` still eagerly emits at the first alloca site, just with `OpLoad` of the metadata stride instead of `uint_immediate` | Safe - the SSA id is defined in the same block that dominates every downstream push/load-top body. |
| Per-task metadata buffer lifetime | Each task record allocates a fresh `AdStackMetadata` buffer and retires the previous one into `ctx_buffers_` | Safe - the buffer stays alive through the submit/sync window; no host memcpy can overwrite a sibling task's values before its dispatch runs. |
| Cross-stack `scope_arr` reuse | MOR `done_lbl` clears `scope[pending_var_id]` to zero | Safe - the next stack's outer pre-pass reads land on index 0, which is valid for any non-empty ndarray (a grammar invariant). |
| Non-ASCII characters | `adstack_sizer_shader.cpp` comments use ASCII arrows (`->`) only | Safe - no smart quotes / em-dashes / Unicode in committed artifacts. |


duburcqa:
@claude review

quadrants/runtime/gfx/runtime.cpp
```diff
@@ -651,6 +719,42 @@ void GfxRuntime::launch_kernel(KernelHandle handle, LaunchContextBuilder &host_c
...
+        // the cmdlist is submitted, so a host-side write is visible to the device without an explicit barrier
+        // on Vulkan (host-visible memory has implicit HOST -> DEVICE availability at submit time) and on
+        // Metal (the shared buffer's contents are observable after `commit`).
+        void *mapped = nullptr;
+        RhiResult map_res = device_->map_range(adstack_metadata_buffer_->get_ptr(0), required, &mapped);
+        QD_ASSERT_INFO(map_res == RhiResult::success, "Failed to map adstack metadata buffer for host upload (size={})",
+                       required);
+        std::memcpy(mapped, ad_stack_metadata.data(), required);
+        device_->unmap(*adstack_metadata_buffer_);
+        bindings->rw_buffer(bind.binding, *adstack_metadata_buffer_);
```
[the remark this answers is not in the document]
duburcqa: Addressed by 8121ea5c2. `GfxRuntime::launch_kernel` now allocates a fresh `AdStackMetadata` device buffer per task and retires the previous allocation into `ctx_buffers_` so each dispatch's recorded binding points at its own data for the full submit/sync window. The shared-member aliasing is gone. Regression test `test_adstack_spirv_metadata_per_task_buffer` (with `cfg_optimization=False` to keep the two sibling offloads separate) pins the record-then-execute race.

quadrants/runtime/gfx/runtime.cpp
```diff
@@ -505,11 +509,65 @@ void GfxRuntime::launch_kernel(KernelHandle handle, LaunchContextBuilder &host_c
...
+    for (const auto &alloca_attribs : attribs.ad_stack.allocas) {
+      int64_t evaluated = -1;
+      if (!alloca_attribs.size_expr.nodes.empty()) {
+        QD_ASSERT_INFO(program_impl_ != nullptr && program_impl_->program != nullptr,
+                       "GfxRuntime::launch_kernel: `ProgramImpl::program` back-reference not set; cannot "
+                       "evaluate AdStack SizeExpr. Ensure GfxProgramImpl passes `program_impl = this` into "
+                       "`GfxRuntime::Params`.");
+        evaluated = evaluate_adstack_size_expr(alloca_attribs.size_expr, program_impl_->program, &host_ctx);
+      }
+      uint32_t max_size = evaluated > 0 ? uint32_t(evaluated) : alloca_attribs.max_size_compile_time;
```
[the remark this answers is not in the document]
duburcqa: Addressed. Callers now use `v < 0` (not `v <= 0`) to detect the `-1` empty-tree sentinel, so a legitimately-zero SizeExpr evaluation stays at 0 and is clamped to 1 by the subsequent `std::max<int64_t>(v, 1)` floor rather than being swapped for `max_size_compile_time`. See `llvm_runtime_executor.cpp:671` for the CPU path and `runtime.cpp:960-968` (runtime_module) for the device interpreter - both use the strict-negative check.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -2182,6 +2298,390 @@ std::vector<BufferBind> TaskCodegen::get_buffer_binds() {
...
+  }
+  spirv::Value buf = get_ad_stack_metadata_buffer();
+  uint32_t header = 2;  // slots 0, 1 are the two strides
+  spirv::Value off_idx = ir_->uint_immediate_number(ir_->i32_type(), header + 2u * info.stack_id);
+  spirv::Value max_idx = ir_->uint_immediate_number(ir_->i32_type(), header + 2u * info.stack_id + 1);
+  spirv::Value off_ptr = ir_->struct_array_access(ir_->u32_type(), buf, off_idx);
+  spirv::Value max_ptr = ir_->struct_array_access(ir_->u32_type(), buf, max_idx);
+  info.offset_val = ir_->load_variable(off_ptr, ir_->u32_type());
+  info.max_size_val = ir_->load_variable(max_ptr, ir_->u32_type());
+  if (info.heap_kind == AdStackHeapKind::heap_float) {
```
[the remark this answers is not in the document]
duburcqa: Addressed by 66bc60b8d. `visit(AdStackAllocaStmt)` now calls `ensure_ad_stack_metadata_loaded(info)` eagerly right after the heap-thread-base emission, so the `OpLoad`s for `info.offset_val` / `info.max_size_val` / `info.adjoint_offset_val` land in the alloca enclosing block and strictly dominate every downstream push / load-top / load-top-adj / acc-adjoint site - including the forward / backward sibling-loop-body pattern the comment called out. Mirrors the dominance discipline already enforced for `get_ad_stack_heap_thread_base_{float,int}()` at the same site.

quadrants/runtime/gfx/runtime.cpp
```diff
@@ -505,11 +514,216 @@ void GfxRuntime::launch_kernel(KernelHandle handle, LaunchContextBuilder &host_c
     ctx_blitter->host_to_device(any_arrays, ext_array_grads, ext_array_size);
   }
 
-  ensure_current_cmdlist();
-
   // Record commands
   const auto &task_attribs = ti_kernel->ti_kernel_attribs().tasks_attribs;
 
+  // Device-side adstack SizeExpr evaluation: every task with adstack allocas has its per-alloca `max_size` /
```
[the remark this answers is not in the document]
duburcqa: Confirmed. Fixed in a39d07ba5: `args_buffer` now allocated with `AllocUsage::Uniform | AllocUsage::Storage` at `runtime.cpp` so the storage_buffer bind on slot 2 satisfies VUID-VkDescriptorBufferInfo-buffer-02999. Also updated the neighbouring binding-comment block to point at the allocation site for the `Uniform | Storage` rationale.

quadrants/program/adstack_size_expr_eval.cpp
```diff
@@ -502,4 +641,148 @@ std::vector<uint8_t> encode_adstack_size_expr_device_bytecode(const AdStackSizin
...
+    return false;
+  }
+  // Scalar fields (`qd.field(dt, shape=())`) have `num_active_indices == 0`; the pre-pass emits a `FieldLoad`
+  // with an empty `indices` vector and the shader should just load `*base_psb` without any index computation.
+  // Return an empty strides vector - `compute_field_load_elem_index`'s loop iterates zero times and produces
+  // `elem_idx = 0`, which `psb_load_scalar` resolves to the exact place address.
+  std::vector<int> shape(n, 0);
+  for (int a = 0; a < n; ++a) {
+    int s = leaf->shape_along_axis(a);
+    if (s <= 0) {
```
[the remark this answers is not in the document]
duburcqa: Confirmed. Fixed in a39d07ba5 via the conservative path: `compute_dense_snode_strides` now walks the leaf's parent chain and returns `false` when any dense ancestor has more than one child. The caller already raises a clear "dense-only chain" error on false return, so an AoS layout now fails at compile time with an attributable message instead of silently reading a sibling field. Cell-size-aware stride math would need `SNodeDescriptor` memory-offset metadata threaded into the sizer, which is out of scope here.

quadrants/codegen/spirv/adstack_sizer_shader.cpp
```diff
@@ -0,0 +1,1036 @@
...
+// Helper: assemble an i64 from two adjacent little-endian u32 words in a buffer. Assumes the buffer is
+// logically uint32[] - correct for both the bytecode buffer (encoder memcpys POD structs in their native
+// layout, which is little-endian on every platform Quadrants supports) and the kernel arg buffer (same).
+Value load_buf_i64(IRBuilder &ir, Value buffer, Value base_word_idx) {
+  Value lo = load_buf_u32(ir, buffer, base_word_idx);
+  Value hi_idx = ir.add(base_word_idx, ir.uint_immediate_number(ir.u32_type(), 1u));
+  Value hi = load_buf_u32(ir, buffer, hi_idx);
+  // Zero-extend both halves to u64, shift hi, OR them, reinterpret as i64.
+  Value lo64 = ir.cast(ir.u64_type(), lo);
+  Value hi64 = ir.cast(ir.u64_type(), hi);
```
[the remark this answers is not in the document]
duburcqa: Confirmed. Fixed upstream in a39d07ba5: `IRBuilder::cast` now skips the trailing `OpBitcast(T, T)` when the intermediate `SConvert`/`UConvert` already produces the destination type (same-signedness widening / narrowing). The three sizer-shader sites (`adstack_sizer_shader.cpp:99, 100, 301`) now emit clean `OpUConvert` / `OpSConvert` instructions with no invalid trailing bitcast, and every other caller of `ir.cast` benefits from the fix.

quadrants/runtime/gfx/runtime.cpp
```diff
@@ -84,16 +89,19 @@ class HostDeviceContextBlitter {
...
+          auto grad_access_it = std::find_if(ctx_attribs_->grad_arr_access.begin(), ctx_attribs_->grad_arr_access.end(),
+                                             [indices](const auto &pair) -> bool { return pair.first == indices; });
+          uint32_t grad_access =
+              (grad_access_it != ctx_attribs_->grad_arr_access.end()) ? uint32_t(grad_access_it->second) : 0;
           auto grad_it = ext_array_grads.find(arg_id);
-          if (grad_it != ext_array_grads.end()) {
+          if (grad_it != ext_array_grads.end() && (grad_access & uint32_t(irpass::ExternalPtrAccess::READ))) {
             DeviceAllocation grad_buffer = grad_it->second;
             void *device_grad_ptr{nullptr};
             QD_ERROR_IF(device_->map(grad_buffer, &device_grad_ptr) != RhiResult::success,
```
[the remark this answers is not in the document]
duburcqa: Confirmed. Fixed in a39d07ba5: the d2h grad readback block now reads `grad_arr_access` via the same `find_if` pattern h2d uses and gates the grad readback on `grad_access & WRITE`, symmetric to h2d's `grad_access & READ` gate. A forward-only kernel whose `arr_access.WRITE=1` but `grad_arr_access=0` no longer clobbers the user's host `.grad` with uninitialised device memory on a backward follow-on.

docs/source/user_guide/autodiff.md
```diff
@@ -235,52 +235,47 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
+Adstack sizing is automatic on every backend: the compiler derives a per-launch capacity for each adstack alloca, and the host launcher (on LLVM backends) or a small on-device sizer shader (on SPIR-V backends) resizes the backing heap before each dispatch. You do not need to tune it, and there is no user-facing size knob anymore. If a loop shape falls outside the supported grammar, the compiler raises an error naming the offending source location. See [Under the hood: adstack capacity and memory](#under-the-hood-adstack-capacity-and-memory) for the supported shapes, the per-launch hand-off, and memory-footprint details.
```
hughperkins: alloca seems a bit low level I feel. Better to stick to the python equivalent I feel, unless absolutely unavoidable.

duburcqa: Yes I tried to do this but apparently on of them slipped through.

docs/source/user_guide/autodiff.md
```diff
@@ -235,52 +235,47 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
 
-**Automatic sizing on LLVM.** The compiler builds a symbolic upper-bound expression (`SizeExpr`) for every adstack alloca at compile time, and the host launcher evaluates that expression against the live field state plus the launch context's ndarray args right before each kernel dispatch. The per-thread heap grows on demand to match the largest size any launch has needed so far, and is reused across subsequent launches. There is no static capacity knob to tune and no runtime overflow path to hit for shapes the grammar covers.
+**Supported loop shapes.** The compiler builds a symbolic upper-bound expression for every adstack at compile time. Supported building blocks are:
```
hughperkins: "a symbolic upper-bound expression " seems complex. Is it fair to say that the adstack is calculated at runtime, based on [whatever it is based on] ?

quadrants/runtime/gfx/adstack_sizer_launch.cpp
```diff
@@ -0,0 +1,239 @@
...
+    // a storage SSBO in the SPIR-V IR, not a uniform). Vulkan distinguishes uniform and storage buffers
+    // via distinct `VkDescriptorType` values - binding these slots via `buffer()` (uniform) produces a
+    // descriptor set layout that doesn't match the pipeline's, and `bind_shader_resources` returns
+    // `invalid_usage` with "Layout mismatch". Use `rw_buffer` across the board so the descriptor types
+    // match; the shader is disciplined about not writing slots 0 and 2, so granting write-capable
+    // descriptors there is harmless. The backing `args_buffer` is allocated with
+    // `Uniform | Storage` at `runtime.cpp::launch_kernel`'s allocation site so the storage-buffer bind on
+    // slot 2 satisfies VUID-VkDescriptorBufferInfo-buffer-02999.
+    bindings->rw_buffer(0, adstack_sizer_bytecode_buffer_->get_ptr(per_task_bytecode_offsets[k]),
+                        per_task_bytecodes[k].size());
```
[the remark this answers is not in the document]
duburcqa: Confirmed. Fixed in 8814b68bb: per-task bytecode offsets are now rounded up to `kDescriptorOffsetAlignment = 256` (conservative cover for NVIDIA/Intel/MoltenVK 16B, Adreno 64B, older NVIDIA 256B). A small `align_up` helper lives alongside the accumulation loop; RHI does not currently expose the queried minimum, so we pick a safe fixed value rather than threading a new device-cap query for this alone.

quadrants/codegen/spirv/adstack_sizer_shader.cpp
```diff
@@ -0,0 +1,1035 @@
...
+// as a clear compile-time diagnostic rather than a mysterious overflow at the next `stack_push`.
+// Per-stack node-count cap for the eval state. Each invocation gets its own private `values_arr` of this
+// size indexed by the *local* offset within the stack's subtree (not the global bytecode-wide node index),
+// so this cap applies only per-stack, not per-kernel. Observed reverse-mode kernels have a handful of stacks
+// with ~1k-node symbolic trees (`MaxOverRange` over ndarray reads with a nested `Max` of per-substep trip
+// counts), so the cap needs to comfortably exceed that. 4096 * 8 B = 32 KiB of i64 private memory per
+// invocation; the sizer runs as a single-thread dispatch so this is not multiplied by workgroup size.
+constexpr int kMaxNodes = kAdStackSizerMaxNodesPerStack;
+constexpr int kMaxVars = 16;
+constexpr int kMaxPending = 16;
```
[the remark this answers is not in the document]
duburcqa: Confirmed. Fixed in 8814b68bb: `kMaxVars` now aliases `kAdStackSizeExprDeviceMaxBoundVars` so the shader's `scope_arr` sizes to whatever the host dense-remap caps at. For the MOR nesting axis, `kAdStackSizerMaxPendingFrames = 16` is exposed through `adstack_sizer_shader.h` and the host encoder's `encode_bytecode_common` walks each tree via a new `compute_max_mor_nesting` helper; it hard-errors before emitting bytecode past the cap, matching the existing `kMaxNodes` pattern.

quadrants/codegen/spirv/adstack_sizer_shader.cpp
```diff
@@ -0,0 +1,1035 @@
...
+    ir.make_inst(spv::OpBranchConditional, empty, empty_lbl, nonempty_lbl);
+
+    ir.start_label(empty_lbl);
+    {
+      Value zero_i64 = ir.int_immediate_number(ir.i64_type(), 0);
+      store_values_at(ir, st, current_now, zero_i64);
+      Value next_cur = ir.add(current_now, ir.int_immediate_number(ir.i32_type(), 1));
+      ir.store_variable(st.current_var, next_cur);
+      ir.make_inst(spv::OpBranch, mor_merge);
+    }
```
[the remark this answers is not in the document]
duburcqa: Fixed in 8814b68bb. At `case_mor` push time the shader now clamps `pending_end_arr[sp]` to `min(end, begin + (1 << 24))`, so the existing `next_i < end` advance check in the continuation path silently terminates after 16M iterations - mirrors the LLVM interpreter's `break` and the host evaluator's `QD_ERROR` at the same threshold, keeping the three paths converged.

quadrants/runtime/gfx/runtime.cpp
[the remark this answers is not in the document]
duburcqa: Confirmed. Fixed in 2e20856be: h2d grad mirror now gates on `grad_arr_access & (READ | WRITE)`. A kernel that only non-atomic-writes `.grad` (e.g. `x.grad[i] = val` on a torch/numpy tensor, grad_access = WRITE) now uploads the current host values before the kernel runs, so the indices it doesn't touch stay at those values through the symmetric d2h readback instead of getting overwritten with allocator garbage. Kernels with both bits clear still skip the mirror entirely.

quadrants/ir/adstack_size_expr_device.h
```diff
@@ -73,19 +86,29 @@ struct AdStackSizeExprDeviceNode {
...
+  // `StructType::get_element_offset({arg_id, DATA_PTR_POS_IN_NDARRAY})`. Unused (-1) for other kinds.
   int32_t arg_buffer_offset{-1};
-  // Offset into the global indices array + count. Indices follow the same `SerializedSizeExprNode::indices`
-  // encoding: non-negative = constant index; `-(var_id + 1)` = currently-bound loop variable captured by the
-  // nearest enclosing `MaxOverRange`.
+  // Offset into the global indices array + `indices_count` = number of axes. Layout inside the global indices
+  // array depends on the node kind:
+  //   * `kExternalTensorRead`: `indices_count` consecutive int32 index values, stride-1 summed (legacy, fine for
+  //     1-D ndarrays). Indices use the same encoding as `SerializedSizeExprNode::indices` - non-negative =
+  //     constant index, `-(var_id + 1)` = currently-bound loop variable.
```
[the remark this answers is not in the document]
duburcqa: Fixed in 2e20856be. Updated the `indices[]` layout doc in `adstack_size_expr_device.h` so `kExternalTensorRead` and `kFieldLoad` share one description: both emit `2 * indices_count` int32 entries as `(idx_a_raw, elem_stride_a)` pairs per axis, with `elem_stride_a` pre-computed on host (ndarray shape for ETR, snode dense chain for FieldLoad). Both interpreters walk the pairs in a single sequential scan.

duburcqa:
@claude review

docs/source/user_guide/autodiff.md
```diff
@@ -208,15 +208,22 @@ update_b()
...
+**How Quadrants does it.** Quadrants provides a dedicated compiler pipeline for this, called the *adstack* (short for "(a)uto(d)iff (stack)"). It allocates a per-variable stack alongside each primal that is updated inside the loop. The forward pass pushes one entry per iteration. The reverse pass walks the stack from top down: at each reverse iteration it reads the current top entry as many times as that iteration's chain-rule contributions need (one `read` per downstream use of the primal), then pops the entry once and steps to the iteration underneath. Peeking-and-reusing rather than popping per use matters because the same primal often feeds several chain-rule terms in one iteration - e.g. a `v` that appears in both a `sin(v)` and a subsequent `v * w` needs to be visible to both adjoint terms before it is discarded. Enabling adstack costs extra per-thread memory and compile time, but some kernels need it.
```
hughperkins: "from top down: at each reverse " => "from top down. At each reverse "

hughperkins: "(one read per downstream use of the primal)"

=> move to a new sentence after the current one, (so we don't need to pusht he current sentence onto the stack whilst reading, then read the parentheses, then pop the ucrrent setnecne back off the stack in our mind, which makes things harder to read I feel)

hughperkins: "Peeking-and-reusing rather than popping per use matters because the same primal often feeds several chain-rule terms in one iteration - e.g. a v that appears in both a sin(v) and a subsequent v * w needs to be visible to both adjoint terms before it is discarded. " => This seems like an implementation detail, not needed by the end-user? => "under the hood" style section

hughperkins: "The flag is compile-time, so it must be set before the offending kernel compiles. " => this seems superfluous? Remove, or move to 'under the hood' (you are justifying your impelmetnation design I feel, rather than telling the user what they need to know in order to use the feature).

docs/source/user_guide/autodiff.md
```diff
@@ -208,15 +208,22 @@ update_b()
...
 
-**Workflow.** Enable the pipeline at init time and keep using the normal reverse-mode workflow: `qd.init(..., ad_stack_experimental_enabled=True)`. The flag is compile-time, so it must be set before the offending kernel compiles.
+**Workflow.** Enable the pipeline at init time and keep using the normal reverse-mode workflow: `qd.init(..., ad_stack_experimental_enabled=True)`. The flag is compile-time, so it must be set before the offending kernel compiles. Any mainstream modern GPU works on the SPIR-V side (Apple Silicon for Metal, any recent discrete GPU for Vulkan); on the odd legacy device that is missing the hardware features the sizer relies on, the launch errors out with a clear message and you can fall back to the LLVM runtime (CPU / CUDA / AMDGPU).
```
hughperkins: "Any mainstream modern GPU works on the SPIR-V side (Apple Silicon for Metal, any recent discrete GPU for Vulkan); on the odd legacy device that is missing the hardware features the sizer relies on, the launch errors out with a clear message and you can fall back to the LLVM runtime (CPU / CUDA / AMDGPU)." => this seems like details best left for later? (maybe some 'compatibilty' section or simlar?)

docs/source/user_guide/autodiff.md
```diff
@@ -208,15 +208,22 @@ update_b()
...
+
+**Note.** Running with adstack enabled when it is not strictly needed is safe, but not the other way around. Running without it when it is needed raises a `QuadrantsCompilationError` in most cases: the autodiff pass rejects a non-static range that would otherwise lose its primal. A few edge-case loop shapes still slip past that rejection and produce silently-wrong gradients; these are tracked and fixed in the autodiff pass as they surface, so if you see wrong-but-non-zero gradients through a dynamic loop with adstack disabled, turn it on and rerun as a sanity check.
```
hughperkins: "A few edge-case loop shapes still slip past that rejection and produce silently-wrong gradients; these are tracked and fixed in the autodiff pass as they surface" => does this mean the gradients will in fact be correct? or will be silently wrong?

hughperkins: "if you see wrong-but-non-zero gradients through a dynamic loop " => how to detect this?

docs/source/user_guide/autodiff.md
```diff
@@ -208,15 +208,22 @@ update_b()
...
 
 Reverse-mode AD walks the forward kernel in reverse and applies the chain rule at every op. The chain-rule factor at each op is that op's derivative with respect to its input. For *non-linear* ops (`sin`, `cos`, `exp`, `sqrt`, `tanh`, `pow`, ...) that derivative depends on the input's primal, so the reverse pass needs the primal value that was there on the forward pass. For *linear* ops (addition, subtraction, multiplication by a constant) the derivative is itself a constant and no primal is needed. In a dynamic loop the forward pass writes a different primal at each iteration, so the reverse pass cannot simply re-read the latest value - it needs one per iteration. adstack provides exactly that: a per-iteration stash of the primal.
```
hughperkins, line 234: Nice and clear to me 🙌

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
 
 `qd.static(range(...))` loops are unrolled at compile time and never need the adstack either.
 
-### Under the hood: adstack capacity and memory
+### Supported loop shapes
 
-*You do not need to read this section to use reverse-mode AD. If a kernel exceeds its adstack capacity, Quadrants raises a Python exception at the next `qd.sync()` whose message recommends bumping `default_ad_stack_size` - that is usually enough. Read on only if you hit that overflow, want to understand why, or want to cap the memory footprint explicitly.*
+The compiler recognises a set of bound shapes:
```
hughperkins: I think we should just delete this seciton, or move it to some highly advanced seciton.

Users will just expect stuff to work I feel, and anything else is a bug.

What's an example of something that does NOT work currrnelty?

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
 
-**Tuning.** Two `qd.init()` knobs control adstack sizing, both measured in slots per adstack (not bytes):
+| Bound shape | Example |
```
hughperkins: this table seems like something to go in a reference section.

"The compiler recognises a set of bound shapes (see 'appendix 123' for authorative list). Any loop shape outside this set is rejected at compile time. The error names the offending source line; typical fixes are to restructure the loop into one of the shapes in the appendix, or to file a bug. "

(but see above comment that we should just delete this seciton altogether, or move the entire section to highly advanced seciton)

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
+### Under the hood (advanced)
 
-**Sizing rule.** A `K`-iteration dynamic loop consumes `K + 2` slots in each of its adstacks: one slot per forward iteration, plus two setup slots (one for the initial adjoint, one for the primal's starting value). `default_ad_stack_size` is a per-stack slot count, so size it at the worst-case trip count of the deepest unprovable dynamic loop in the program plus 2.
+*This section is a glimpse of how the adstack is sized and laid out in memory. You never need to read it to use reverse-mode AD - skip it unless you are curious, debugging an overflow or OOM, or planning to extend the list of supported loop shapes.*
```
hughperkins: "You never need to read it to use reverse-mode AD " => "never" seems a very strong claim. I'm pretty skeptical of this claim given the constraints on memory, loop structure etc.

I prefer the earlier pharsing. I don't remember what that was.

Oh here we go:

"*You do not need to read this section to use reverse-mode AD. Skip past it unless you hit an overflow error on SPIR-V, an out-of-memory error on GPU, or a compile error pointing at an adstack alloca.*"

so:
- remove the first sentence, superfluous
- remove 'never'
- break second sentence into two
- basically just use the original paragraph I pasted above :)

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
+| `num_buffers` | Number of adstacks the kernel allocates - one per loop-carried variable plus one per dependent branch flag (see [One adstack per variable](#one-adstack-per-variable)). |
 
-On SPIR-V backends (Metal / Vulkan) the slot layout is trimmed: adstacks whose `T` is an integer type (`i32`, `i64`, ...) only store the primal because the reverse pass does not accumulate integer adjoints, and per-thread on-chip memory is more constrained than on LLVM. So `bytes_per_slot = sizeof(T)` for integer `T` and `bytes_per_slot = 2 * sizeof(T)` for floating-point `T`. SPIR-V has no defined layout for `OpTypeBool`, so booleans are widened to i32 at storage time:
+Every adstack slot always stores a *primal* value - the forward-pass value the reverse pass pops to recover the chain-rule step. Floating-point adstacks additionally store an *adjoint* slot where the reverse pass accumulates chain-rule contributions. Integer / boolean adstacks do not need an adjoint slot, but LLVM backends still carry one for codegen uniformity. SPIR-V backends trim it. SPIR-V also widens `bool` to `i32` at storage time, because SPIR-V has no defined layout for `OpTypeBool`.
```
hughperkins: "but LLVM backends still carry one for codegen uniformity. SPIR-V backends trim it. SPIR-V also widens bool to i32 at storage time, because SPIR-V has no defined layout for OpTypeBool." => I would break this platform-specific stuff into a seprate paragraph somehow.

"Platform-specific notes:
- even though integer/boolean adstacks do not need an adjoint slot, LLVM backends still carry one for codegen uniformity
- SPIR-V stores `bool`s using 4 bytes (32 bits)"

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
```
hughperkins, line 327: "Therefore some examples of the size of adstack slots on each platform are as follows:"

docs/source/user_guide/autodiff.md
duburcqa: > What's an example of something that does NOT work currrnelty?

Actually I would prefer to keep it. We only support a small subset of all the craziness a user may want to do. Here are examples that are not supported:
```
@qd.kernel
def my_kernel(a)
    for i in range(sqrt(a)):  # on any non-linear transformation plus min/max
        [...]
```

```
@qd.kernel
def my_kernel(a, b)
    for i in range(a):
        for j in range(b):
            for k in range(a[i], b[j]):  # Non-matching running variables+indices  
                [...]
```

```
@qd.kernel
def my_kernel(a)
    for i in range(a.shape[0]):
        while a[i] < 10:  # Bound that cannot be determined without running the kernel
            a[i] = a[i] + 1
```

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
+- **When a dynamic-loop reverse kernel fails.** In normal use, enabling the adstack pipeline and running a reverse-mode kernel through a dynamic loop should just work. Two rare situations can make it fail, and it is worth knowing which one you are in:
```
hughperkins: "In normal use, enabling the adstack pipeline and running a reverse-mode kernel through a dynamic loop should just work. Two rare situations can make it fail, and it is worth knowing which one you are in:" => this si superfluous I feel.

hughperkins: Are these really limitations?  It seems like saying "a limitation of numpy tensors is that they cannot be larger than available physical memory."? 🤔

hughperkins: These should probably go in some "Gotchas" or "What can go wrong" or simliar kind of section. I'm not sure they are limtiations?

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
+    - split the inner section into its own `@qd.kernel`;
+    - pass `ad_stack_size=N` to `qd.init()` with `N` large enough to cover the real push count (bypasses the sizer).
+  - *The kernel is legitimately too deep for the hardware* - surfaces as an out-of-memory error from the allocator, before the kernel even starts running. A reverse pass through many loop-carried variables at a large ndrange can ask the runtime for more adstack memory than the device can physically back, even when the sizer's number is correct. Remedies are the ones listed under *Avoiding OOM on GPU* above: fewer loop-carried variables, a smaller ndrange, manual checkpointing, or more device-memory headroom.
+- **Compile time scales with loop nesting.** The adstack pipeline trades compile time for generality. Kernels with many loop-carried variables, nested dynamic loops, or large inner-loop bodies produce visibly slow compile times - seconds stretching into minutes, and on SPIR-V backends sometimes into the territory where the driver's shader compiler gives up. Budget compile time accordingly when migrating existing reverse-mode AD workloads.
```
hughperkins: is this a limtiation? Seems more like some kind of performance characteristic?

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
+- **Backward passes are slower than forward passes.** In particular, SPIR-V backward passes can be an order of magnitude slower than the forward pass.
```
hughperkins: this isnt really a limitation I feel. In general, in the convnet world at least, backward passes are matehtmatically twice as long as forward passes. Thtas not really a limtation. It just cmes out of the maths.

Maybe just limit to "SPIR-V backward passes can be an order of magnitude slower than the forward pass" (maybe say why?)

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
+- **Integer casts stop gradients.** Reverse-mode AD does not propagate gradients through integer casts or non-real operations: integers have no meaningful derivative, so the gradient reads as zero upstream of the cast - no error, just silently-zero gradients. Rules of thumb:
```
hughperkins: is this really a limitation? This is just a characteristc of differentiation I think? You cannot pass a gradient through a discretization step. (at least, not without using approximatinos, such as Gumbo etc).

docs/source/user_guide/autodiff.md
duburcqa: Where do you want to mention this kind of information if not in limitations?

docs/source/user_guide/autodiff.md
```diff
@@ -233,67 +240,114 @@ Reverse-mode AD walks the forward kernel in reverse and applies the chain rule a
...
+  - keep differentiable variables in `qd.f32` / `qd.f64` through the full forward chain;
+  - casting *to* a float is safe - the downstream float section remains differentiable, and the cast itself contributes a unit factor to the chain rule;
+  - casting *back to* an integer stops the gradient at that point, so only do it at integer-indexing sites, after any arithmetic whose gradient you need.
+- **Loop bounds that read an ndarray written to by the same kernel are unsafe.** If a reverse-mode kernel has a loop whose iteration count comes from `n[j]` and the same kernel also writes to `n[j]` before that loop runs (`for i in range(n[j])` after `n[j] = something`), the computed gradient may silently come out zero or wrong. Quadrants does not currently detect this. Workaround: put iteration counts in a separate ndarray (or scalar field) the kernel only reads, or split the write and the differentiable loop into two `@qd.kernel` calls. The same caveat applies across kernels - a kernel that writes an ndarray read as a loop bound by a later `.grad()` call needs the host-side contents of that ndarray to already hold the value the backward kernel will execute against.
```
hughperkins: it's not clear to me from skim-reading whether this is a limitation of our implemetation, or just a fundamental charcteristic that falls out of the maths. I'm *guessing* the former. Can you clarify in reply to this comment please.

docs/source/user_guide/autodiff.md
duburcqa: Well, the thing is that it can eat memory pretty fast, and what is being allocated and why is not immediately obvious. When you allocate a gigantic numpy array you will notice it, here it is hidden.

docs/source/user_guide/autodiff.md
duburcqa: Where would you like to put this appendix? Another markdown page?

docs/source/user_guide/autodiff.md
duburcqa: > how to detect this

There is no way to detect this I'm afraid. Just running some external finite difference check, or precising turning it on and rerun as a sanity check. We can just say: "Checking that the value of the gradient is unaffected when turning adstack on is the best way to determine whether adstack is necessary."

docs/source/user_guide/autodiff.md
hughperkins: "Performance characteristics"?

"Performance"?

docs/source/user_guide/autodiff.md
hughperkins: I would argue that a 'limitation' is some issue with our implementation, rather than a fundamental characteristic of the underlying maths etc.

As suggested, lets have a section on things that can go wrong, that users can consult.

Analogy:

doc 1:
"Limitations:
- this toaster does not work when it is not plugged into the power"

doc 2:
"Why doesn't my toaster work?
- check: is it plugged into the power?"

docs/source/user_guide/autodiff.md
hughperkins: At the end of the same markdown page.

docs/source/user_guide/autodiff.md
hughperkins: Alright, so move to a reference section. And state something like "Note that whilst we support many common loop constructs, many constructs which are allowed in the absence of adstack are not currently handled. See Appendix A for supported loop constructs."

docs/source/user_guide/autodiff.md
duburcqa: I don't know exactly. What I know for sure is that I found it very unexpected when I discovered this limitation and that it is not something uncommon. We need this in Genesis typically. I would be surprised if such limitation exists in Torch, otherwise more people would compute complete wrong gradient x) The biggest issue is that it fails silently.

docs/source/user_guide/autodiff.md
hughperkins: "Checking that the value of the gradient is unaffected when turning adstack on is the best way to determine whether adstack is necessary." Yup, if that's what we have, let's run with that. Maybe loosen "the best way" to "is a reasonable way"

hughperkins:
Nice doc! Thanks!

Checklist:
- doc has been updated, and I've reviewed
- many of the new lines of code appear to be added to new files, dedicated to the new adstack sizing functionality (rather than mixed into existing files)

=> ok to merge

PR #548, opened 2026-04-22
[spirv] Don't decorate Function-scope arrays with ArrayStride

hughperkins:
## Summary
  `IRBuilder::get_array_type()` was unconditionally emitting
  `OpDecorate <arr_type> ArrayStride <n>` on every `OpTypeArray` /
  `OpTypeRuntimeArray` it created. Per Vulkan's standalone-SPIR-V
  rules (VUID-StandaloneSpirv-None-10684), `ArrayStride` (and other
  explicit-layout decorations) are only permitted on array types that
  are used in explicit-layout storage classes — i.e. inside a
  `Block` / `BufferBlock` struct backing `StorageBuffer` / `Uniform` /
  `PushConstant`. Emitting `ArrayStride` on an array type that is
  then used as the pointee of a `Function`- or `Workgroup`-scope
  `OpVariable` produces invalid SPIR-V.
  ## Symptom
  `spirv-val --target-env vulkan1.3` rejects the SPIR-V we emit for
  any kernel that stack-allocates a `Matrix`/`Vector` temporary:
      error: line NNN: [VUID-StandaloneSpirv-None-10684]
          Invalid explicit layout decorations on type for operand
          '%_ptr_Function__arr_float_uint_9'
        %tmp284_unknown = OpVariable %_ptr_Function__arr_float_uint_9 Function
  Most Vulkan drivers silently accept it anyway, but the NVIDIA
  proprietary driver's SPIR-V → NVVM shader compiler SIGSEGVs on it
  (observed in `libnvidia-glvkspirv.so` / `libnvidia-gpucomp.so`,
  driver `580.76.05`, RTX 5090, during `vkCreateComputePipelines` —
  e.g. on any kernel calling `qd.sym_eig(A_3x3)`).
  ## Fix
  Move the `ArrayStride` decoration out of `get_array_type()` and into
  `get_struct_array_type()`, which is the only call path that wraps
  the array in a `Block` / `BufferBlock` struct and therefore
  legitimately needs an explicit layout. Direct callers of
  `get_array_type()` — the `AllocaStmt` codegen for tensor-typed
  `Function`- and `Workgroup`-scope locals (`spirv_codegen.cpp:267`) —
  now emit undecorated array types, which is what Vulkan requires for
  those storage classes.
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/codegen/spirv/spirv_ir_builder.cpp
+++ b/quadrants/codegen/spirv/spirv_ir_builder.cpp
@@ -398,6 +398,26 @@ SType IRBuilder::get_array_type(const SType &_value_type, uint32_t num_elems) {
     ib_.begin(spv::OpTypeRuntimeArray).add_seq(arr_type, value_type).commit(&global_);
   }
 
+  // NOTE: We intentionally do NOT emit OpDecorate ArrayStride here.
+  // Per the Vulkan "standalone SPIR-V" rules (VUID-StandaloneSpirv-None-10684),
+  // explicit-layout decorations like ArrayStride are only permitted on types
+  // used in explicit-layout storage classes (StorageBuffer / Uniform /
+  // PushConstant / etc. via a Block/BufferBlock struct). Emitting ArrayStride
+  // on an array type that is then used as a Function- or Workgroup-scope
+  // OpVariable produces invalid SPIR-V that spirv-val rejects and that the
+  // NVIDIA proprietary Vulkan driver's SPIR-V->NVVM compiler crashes on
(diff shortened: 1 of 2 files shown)


hughperkins:
@claude review

hughperkins:
Marking this lowpri.

Apparently this doesnt actually cause any test failures. It was caught by `spirv-val`.

hughperkins:
I'm just going to close this, because it's only a spirv-val issue. If you came across this while trying to fix spirv-val, and you want to merge it, we can reopen it, and merge it :)

PR #549, opened 2026-04-22
[AutoDiff] Autodiff 16: Resolve reverse-mode adstack depths per-launch via runtime-evaluated SizeExpr on LLVM backends (CPU/CUDA/AMDGPU)

duburcqa:
# Runtime-evaluated adstack sizing on LLVM backends

> One commit + a doc-comments follow-up. Makes reverse-mode adstack capacity automatic on CPU / CUDA / AMDGPU: the compiler captures a symbolic upper-bound expression per adstack alloca and the host launcher resizes the per-thread heap just before each dispatch. No user-facing API change; `default_ad_stack_size` / `ad_stack_size` stop being load-bearing for launch-time sizing.

## TL;DR

```python
import quadrants as qd

qd.init(arch=qd.cpu, ad_stack_experimental_enabled=True)

n = qd.field(qd.i32, shape=())
x = qd.field(qd.f32, shape=16, needs_grad=True)
y = qd.field(qd.f32, shape=(), needs_grad=True)

@qd.kernel
def compute():
    for i in x:
        v = x[i]
        for _ in range(n[None]):           # trip count known only at launch time
            v = v * 0.95 + 0.01
        y[None] += v

n[None] = 512
compute()
y.grad[None] = 1.0
compute.grad()                              # no QuadrantsAssertionError: adstack sized to 512 + 2
```

Before this PR, the inner `range(n[None])` forced the CFG analyzer into its `default_ad_stack_size=256` fallback, so `n[None] = 512` overflowed at the next `qd.sync()`. After this PR the compiler captures the loop bound as a `SizeExpr`, the launcher evaluates it against `n[None]` at launch time, and the backing heap is grown to exactly `512 + 2` slots.

## Why

Reverse-mode AD through a dynamic loop requires an adstack whose depth equals the forward trip count plus two setup pushes. Until now Quadrants had to resolve that depth at compile time, so every loop bound that involved a field load (`n_bodies`, `n_dofs`, an ndarray shape, an enclosing parallel-for index) fell back to the user-configured `default_ad_stack_size`. For realistic physics / simulation kernels this is both:

- **Wrong-too-small**: a 14-DOF robot stepping through an `ndrange(n_bodies, n_dofs)` easily needs >256 slots, triggering a runtime `QuadrantsAssertionError: Adstack overflow`.
- **Wrong-too-big**: bumping the default to cover worst case inflates every adstack in the program, including the ones the compiler *could* prove were tiny, and scales linearly with the dispatched ndrange on GPU.

The traditional workaround - tune `default_ad_stack_size` per program - is not tunable enough: you cannot set a different default for every adstack, and bumping it globally multiplies GPU memory by the number of threads. This PR replaces the static fallback with a symbolic upper-bound expression captured at compile time and reduced per launch against the live field / ndarray state, so each adstack gets exactly the depth that dispatch needs.

## Entry point

`LlvmRuntimeExecutor::publish_adstack_metadata(LaunchContextBuilder &ctx, const LlvmOffloadedTaskBody &task)` at `quadrants/runtime/llvm/llvm_runtime_executor.cpp`. Called from `cpu/cuda/amdgpu/kernel_launcher.cpp` right before every task's dispatch. It:

1. Walks the task's `size_exprs` vector (post-order `SerializedSizeExpr` trees, one per `AdStackAllocaStmt`).
2. Evaluates each tree via `adstack_size_expr_eval.cpp`, reading scalar field values through `SNodeRwAccessorsBank` and launch-context ndarray shapes / scalar reads through the `LaunchContextBuilder`.
3. Computes per-thread stride, per-alloca offsets, and per-alloca max-sizes.
4. Writes these into `LLVMRuntime::adstack_{per_thread_stride,offsets,max_sizes}` via `runtime_get_adstack_metadata_field_ptrs` so generated code reads them at every push / load-top site.
5. Grows the per-thread heap backing buffer via `ensure_adstack_heap(per_thread_stride * dispatched_threads)`.

Shapes the grammar cannot cover raise `QuadrantsCompilationError` at compile time naming the offending source location. There is no silent over-allocation and no runtime overflow path.

### 1. Compile-time `SizeExpr` capture

`quadrants/transforms/determine_ad_stack_size.cpp` replaces the old CFG Bellman-Ford + default fallback with a walker that attempts to express each `AdStackAllocaStmt`'s max depth as a `SizeExpr` tree. Grammar covered:

| Leaf | Source | Lowering |
| --- | --- | --- |
| Const | integer literal | `SizeExpr::make_const` |
| FieldLoad | `GlobalLoadStmt` of a scalar `i32` / `i64` SNode | `make_field_load` with the SNode pointer + constant index path |
| NdarrayShape | `ExternalTensorShapeAlongAxisStmt` | `make_ndarray_shape(arg_id, axis)` |
| NdarrayRead | scalar read from an integer ndarray | `make_ndarray_read(arg_id, index_path)` |
| LoopIndexMax | `LoopIndexStmt` of an enclosing bounded range-for / parallel-for | `make_loop_index_max` (bounded by the loop's `end` or a `max` over every push site) |
| StackLoadTopMax | `AdStackLoadTopStmt` of a stash adstack | `make_stack_load_top_max` (bounded by a `max` over every push site) |

Inner operators: `add`, `sub`, `mul` (const * anything via `Add`-replication), `max`. Anything else raises `QD_ERROR` naming the source statement. `SizeExpr` objects are immutable after the pre-pass writes them and are serialised as post-order `SerializedSizeExpr::nodes` so offline-cache hits evaluate the same way as fresh compiles.

### 2. Codegen reads stride / offset / max-size from runtime metadata

| File | Change |
| --- | --- |
| `codegen_llvm.cpp` `AdStackAllocaStmt` | base addr = `heap + linear_tid * runtime->adstack_per_thread_stride + runtime->adstack_offsets[stack_id]` (all three loaded at runtime) |
| `codegen_llvm.cpp` `AdStackPushStmt` | `stack_push(runtime, stack, runtime->adstack_max_sizes[stack_id], element_size)` (cap is a runtime load, not an IR immediate) |
| `init_offloaded_task_function` pre-scan | assigns `AdStackAllocaStmt::stack_id` in declaration order; `ad_stack_offsets_` became `std::vector<std::size_t>` indexed by `stack_id` |

The heap base address is unchanged; only the per-thread stride and per-alloca offset math was moved from codegen-time immediates to runtime loads, so kernels compiled with the old flow will recompile but do not break semantics.

> Note: `default_ad_stack_size` and `ad_stack_size` are ignored on LLVM. They remain in `qd.init()` for SPIR-V backwards compatibility and for the narrow offline-cache-hit corner case where `SerializedSizeExpr::nodes` is empty (the launcher falls back to `default_ad_stack_size` then).

### 3. Host evaluator

`adstack_size_expr_eval.{h,cpp}` walks the flat post-order tree:

- Field loads go through the existing `SNodeRwAccessorsBank` to read the current scalar value on the host side.
- Ndarray shape / scalar reads go through `LaunchContextBuilder` which captures both the pointer and the shape on kernel launch.
- `LoopIndexMax` and `StackLoadTopMax` fold to whichever launch-time bound they were captured against.
- Arithmetic is saturating `int64_t`.

The evaluator runs strictly on the host, before the kernel dispatches, so there is no device-side evaluator to keep in sync.

### 4. Per-kernel metadata bridge

`runtime_get_adstack_metadata_field_ptrs(LLVMRuntime *)` in `runtime.cpp` returns pointers to `adstack_{per_thread_stride,offsets,max_sizes}`. The host writes into these fields once per dispatch. On CPU the fields are read directly from `LLVMRuntime`; on CUDA / AMDGPU the bridge triggers an `memcpy_host_to_device` so the generated code's subsequent loads see the right values.

| Backend | Pre-pass | Metadata bridge | Launcher hook | Status |
| --- | --- | --- | --- | --- |
| CPU (LLVM) | ok | direct field access | `cpu/kernel_launcher.cpp` | auto-sized; no overflow path |
| CUDA (LLVM) | ok | `memcpy_host_to_device` | `cuda/kernel_launcher.cpp` | auto-sized; no overflow path |
| AMDGPU (LLVM) | ok | `memcpy_host_to_device` | `amdgpu/kernel_launcher.cpp` | auto-sized; no overflow path |
| Metal (SPIR-V) | N/A here | N/A here | N/A here | untouched - still compile-time, handled by the follow-up SPIR-V sizer PR |
| Vulkan (SPIR-V) | N/A here | N/A here | N/A here | untouched - still compile-time, handled by the follow-up SPIR-V sizer PR |

### `tests/python/test_adstack.py`

- `test_adstack_field_load_bounded_loop_evaluated_per_launch` - pins the core guarantee: a kernel with `for _ in range(n[None])` bound by an i32 scalar field. Launches at `n_iter in (1, 20, 50)` and asserts the heap resized correctly (no overflow at 50, no over-allocation at 1), with the decorator set to `default_ad_stack_size=2` to prove the knob is bypassed.
- `test_adstack_near_capacity` - rewritten to exercise both sides of the old `default_ad_stack_size=32` boundary at `n_iter in (30, 100)` without any knob override. Arch-restricted to `qd.cpu` until the SPIR-V sizer follow-up lands.
- `test_adstack_unsupported_shape_raises` - negative control: a loop bound built from an unsupported op (`ceil_div`) produces a `QuadrantsCompilationError` naming the source line.

### `tests/cpp/transforms/determine_ad_stack_size_test.cpp`

- `SizeExprBuildFromRangeFor` - const bounds fold to a `Const` leaf.
- `SizeExprBuildFromFieldLoad` - scalar i32 field load builds a `FieldLoad` leaf with the right SNode + index path.
- `SizeExprBuildFromLoopIndex` - inner `for j in range(i)` under an enclosing bounded loop folds to `LoopIndexMax`.
- `SizeExprRejectsUnsupportedOp` - negative control that an unsupported operator triggers the pre-pass error path.

Deliberate coverage gap: no SPIR-V cross-check here - the SPIR-V runtime metadata path lives in the follow-up PR and its tests cover the shader-side reading.

## Side-effect audit

| --- | --- | --- |
| Offline cache key | `quadrants/analysis/offline_cache_util.cpp` serialises `SerializedSizeExpr::nodes` alongside the existing `size_in_bytes` per `AdStackAllocaStmt` | ok - cache-hit reloads the same tree; evaluator runs the same on fresh + cache-hit |
| Stmt clone / serialization | `statements.h`: `AdStackAllocaStmt::size_expr` is a `std::shared_ptr<SizeExpr>` cloned via the default `QD_DEFINE_ACCEPT_AND_CLONE` | ok - `SizeExpr` is immutable post pre-pass; shared pointer is safe to share across clones |
| IR printer | `ir/ir.cpp` `AdStackAllocaStmt::repr` prints `stack_id` and a short `SizeExpr` summary | ok - IR dump readable |
| Codegen non-feature path | `AdStackAllocaStmt::stack_id == -1` falls through to the old `default_ad_stack_size`-backed path | ok - kernels compiled without the pre-pass writing a `SizeExpr` still behave as before |
| Whole-kernel CSE | `SizeExpr` is stored on `AdStackAllocaStmt`, not emitted as IR operands, so it does not participate in CSE | ok |
| LaunchContextBuilder | extended to capture scalar + shape for every ndarray arg; no change to the existing argument-pack layout - new entries appended | ok - old kernels still build their context the same way |
| Runtime initialization | `adstack_{per_thread_stride,offsets,max_sizes}` initialised to zero / null in `runtime_initialize`; kernels with no adstacks read nothing from them | ok - zero-adstack kernels pay nothing |
| Multi-threaded CPU launch | metadata is written once per dispatch before the thread pool starts; no concurrent writers | ok |
| CUDA / AMDGPU `memcpy_host_to_device` | one memcpy per task per launch; measured overhead ~microseconds | acceptable; would be visible only on workloads that launch thousands of tiny autodiff kernels per second |

### DIFF
--- a/quadrants/rhi/vulkan/vulkan_device_creator.cpp
+++ b/quadrants/rhi/vulkan/vulkan_device_creator.cpp
@@ -805,18 +805,19 @@ void VulkanDeviceCreator::create_logical_device(bool manual_create) {
       features2.pNext = &buffer_device_address_feature;
       vkGetPhysicalDeviceFeatures2KHR(physical_device_, &features2);
 
-      if (CHECK_VERSION(1, 3) || buffer_device_address_feature.bufferDeviceAddress) {
-        if (device_supported_features.shaderInt64) {
-          // The prior gate was `#if !defined(__APPLE__) && false`, a kill-switch referencing Taichi issue
-          // #6295 (broken BDA on the 2022-era MoltenVK pinned in `quadrants/rhi/CMakeLists.txt`). That
-          // pin has since been replaced with a LunarG-SDK-sourced MoltenVK (>= 1.3, sporting working
-          // `vkGetBufferDeviceAddress`), and the extension feature bit above already gates against devices
-          // that don't advertise BDA - so there is no reason to hard-disable PSB on Apple (or anywhere).
-          // Without this cap the adstack sizer shader hard-errors at launch time on every reverse-mode
-          // kernel because `$ENV{VULKAN_SDK}`-backed MoltenVK can't be distinguished from the legacy pin
-          // by capability alone.
-          caps.set(DeviceCapability::spirv_has_physical_storage_buffer, true);
-        }
+      // Gate strictly on the queried feature bit. Vulkan 1.3 *promotes* `VK_KHR_buffer_device_address`
+      // into core but still lets the implementation expose `bufferDeviceAddress = VK_FALSE`; the previous
+      // `CHECK_VERSION(1, 3) || feature_bit` condition therefore treated 1.3 devices as PSB-capable even
+      // when they were not, which would drive `vkGetBufferDeviceAddressKHR` / BDA usage flags on hardware
+      // that didn't advertise the feature. The `shaderInt64` guard stays because the sizer shader holds
+      // PSB pointers as `i64` SSA values.
+      if (buffer_device_address_feature.bufferDeviceAddress && device_supported_features.shaderInt64) {
+        // The prior `#if !defined(__APPLE__) && false` kill-switch referenced Taichi issue #6295 (broken
+        // BDA on the 2022-era MoltenVK pinned in `quadrants/rhi/CMakeLists.txt`). That pin has since been
+        // replaced with a LunarG-SDK-sourced MoltenVK (>= 1.3, sporting working `vkGetBufferDeviceAddress`),
+        // so there is no reason to hard-disable PSB on Apple (or anywhere) - the feature query above now
+        // correctly reflects the device's actual capability.
+        caps.set(DeviceCapability::spirv_has_physical_storage_buffer, true);
       }
       *pNextEnd = &buffer_device_address_feature;
       pNextEnd = &buffer_device_address_feature.pNext;


hughperkins:
This adds 4604 lines of code. Is the base branch correct? 🤔

hughperkins:
Question (I dont have astrong opinion on this point, but just posing the question)
- the earlier comment in the code in question suggests installing moltenvk sdk using brew
- what do you see as the good and bad points of the two posible approaches (ie downloading from lunarg, vs using hte homebrew version?)

PR #555, opened 2026-04-23
[Test] Flush xdist channel before worker exit so test failure reports are visible

hughperkins:
interactor.retire() calls os._exit(0) which kills the process before execnet's IO thread can transmit the queued test report. Replace with an explicit sendevent + sleep + os._exit so the failure details reach the controller. Also guard report.longrepr.chain access to prevent AttributeError crashes on non-standard longrepr types.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/conftest.py
+++ b/tests/python/conftest.py
@@ -1,5 +1,7 @@
 import gc
+import os
 import sys
+import time
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


hughperkins:
@claude review

hughperkins:
@claude review

PR #559, opened 2026-04-24
[Test] Fix flaky perf_dispatch tests by increasing work amounts

hughperkins:
## Summary
- The timing-sensitive `test_perf_dispatch` tests were flaky on CI (`gpu-t4-4-core` runners, especially Vulkan) because the slow/fast work amounts (10000 vs 100) produced only a ~2x wall-clock ratio — most of the time was fixed GPU dispatch overhead (~0.45ms). Under pytest-xdist contention this margin collapsed and the dispatcher picked the wrong "fastest" implementation.
- Increase work amounts so slow is always 10x medium: GPU 1M/100K, CPU 100K/10K. Work amounts are defined as module-level constants and selected per-arch using `qd.gpu`.
- Verified 40/40 passes under 4-worker xdist contention across x64, cuda, and vulkan on a 5090.

## Test plan
- [x] All 17 perf_dispatch tests pass on x64, cuda, vulkan individually
- [x] 40/40 runs pass under 4-worker xdist contention (all arches combined)
- [x] Previously reproduced ~40-50% failure rate with original work amounts under same contention


hughperkins:
@claude review

hughperkins:
@claude review

PR #561, opened 2026-04-24
[Type] Tensor 24

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
+# Tensors
+
+Quadrants offers two underlying tensor implementations, `qd.field` and `qd.ndarray`. They have different runtime/compile-time trade-offs, and different physical memory layouts can suit different kernels.
```
duburcqa: Do we have doc for them? If so, it would be nice to reference them here.

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+See [`tensor_types`](tensor_types.md), [`scalar_tensors`](scalar_tensors.md), and [`matrix_vector`](matrix_vector.md) for the underlying tensor primitives.
+
+## Choosing a backend: `qd.Backend`
+
+`qd.Backend` is an `IntEnum` with two members:
+
+| Member | Underlying type | When to prefer |
+|---|---|---|
+| `qd.Backend.FIELD` | `qd.field` | Faster at runtime; recompiles when any dimension size changes. |
+| `qd.Backend.NDARRAY` | `qd.ndarray` | Slightly slower at runtime but avoids recompilation when sizes change. |
```
duburcqa: I would remove "Slightly". Just the right amount of information to get your attention while being frustrated of any idea how much slower it is. In practice I observe up to 30% on some benchmarks, which is arguably not negligible.

hughperkins: Yes, strongly agree. Will update.

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+qd.init(arch=qd.x64)
+
+a = qd.tensor(qd.f32, shape=(4, 5))                                 # ndarray (default)
+b = qd.tensor(qd.f32, shape=(4, 5), backend=qd.Backend.FIELD)       # field
+
+assert isinstance(a, qd.Tensor)
+assert isinstance(b, qd.Tensor)
+```
+
+`qd.tensor()` (and the `qd.Vector.tensor` / `qd.Matrix.tensor` siblings) returns a `qd.Tensor` wrapper that uniformly forwards a fixed surface (`shape`, `dtype`, `layout`, `to_numpy`, `from_numpy`, `to_torch`, `from_torch`, `to_dlpack`, `fill`, `copy_from`, `grad`, host-side `__getitem__` / `__setitem__`, pickle) regardless of which backend it wraps. Drop down to the bare impl with `t._unwrap()` (returns the underlying `qd.Ndarray` or `qd.ScalarField`) only if you need a backend-specific knob.
```
duburcqa, line 36: For the record, `torch` provides `base_tensor = my_tensor.as_subclass(torch.Tensor)`. I don't know if we want to mimic this interface.

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+qd.init(arch=qd.x64)
+
+# Default (canonical) layout: same order as the canonical shape.
+a = qd.tensor(qd.f32, shape=(N, B))
+
+# Transposed storage: axis 1 (batch) becomes the outer SNode, axis 0 inner.
+b = qd.tensor(qd.f32, shape=(N, B), layout=(1, 0))
+```
+
+`layout` is a tuple of `int` listing the **canonical axis index at each successive memory-nesting level, outermost first**. It must be a permutation of `range(len(shape))`. The canonical (logical) shape that you pass and that `tensor.shape` returns is *not* affected by `layout`:
```
duburcqa, line 106: What about adding one extra statement:
"Tweaking the memory layout on a per-tensor basis is commonly used to improve runtime performance. In practice, tuning axis order is sufficient in most cases. For advanced user seeking finer-grained control over the memory layout, please see ti.root.X as reference."

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+
+```python
+b = qd.tensor(qd.f32, shape=(N, B), layout=(1, 0))
+assert b.shape == (N, B)        # canonical shape, unchanged
+b[i, j] = ...                   # canonical indexing in kernels still works
+```
+
+Any permutation is supported, up to Quadrants' `quadrants_max_num_indices` (currently 12). `layout=None` and the identity permutation (`(0, 1, ..., N-1)`) are equivalent and forward no permutation to the underlying allocator.
```
duburcqa, line 114: What about adding one extra statement:
"Tweaking the memory layout on a per-tensor basis is commonly used to improve runtime performance. In practice, tuning axis order is sufficient in most cases. For advanced user seeking finer-grained control over the memory layout, please see ti.root.X as reference."

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+
+fill(a)   # field branch
+fill(b)   # ndarray branch
+```
+
+The kernel argument is unwrapped to the bare impl before the template-mapper / AST sees it, so kernel bodies still write `x[i, j]` and pay no per-call cost for the wrapper.
+
+## Pickle
+
+`qd.Tensor` objects are picklable on **both** backends, including under non-identity layouts. Round-trip preserves the canonical data, the dtype, the shape, and the layout:
```
duburcqa: What does 'Round-trip' means?

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+a.from_numpy(np.arange(12, dtype=np.float32).reshape(3, 4))
+
+restored = pickle.loads(pickle.dumps(a))
+assert isinstance(restored, qd.Tensor)
+assert restored.shape == (3, 4)
+assert restored.layout == (1, 0)
+assert (restored.to_numpy() == a.to_numpy()).all()
+```
+
+Pickle is implemented at the wrapper layer (`Tensor.__reduce__` round-trips through `to_numpy()` plus a `qd.tensor(...)` reallocation on load). The bare `qd.field` type does not support pickle directly — `qd.Tensor` is the unit of serialization.
```
duburcqa: "`qd.Tensor` is the unit of serialization." is confusing I feel.

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+
+qd.init(arch=qd.x64)
+
+a = qd.ndarray(qd.f32, shape=(4, 5))
+t = qd.wrap(a)
+assert isinstance(t, qd.Tensor)
+assert t._unwrap() is a   # same underlying impl
+```
+
+`qd.wrap` is the only sanctioned way to construct a wrapper around a bare impl after the fact. The `qd.Tensor(impl)` constructor itself rejects double-wrapping so you can't accidentally end up with a `Tensor` containing a `Tensor`.
```
duburcqa, line 214: This is weird. Why not just support passing a `qd.ndarray` or `qd.field` to `qd.tensor` "constructor" ? This is exactly how `torch` is working for wrapping numpy arrays, so I would strongly recommend doing the same.

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+
+## Cross-backend `copy_from` is not supported
+
+`tensor.copy_from(other)` requires both tensors to share the same backend. Mixed-backend copies are deliberately unsupported because the intended deployment model is process-wide, homogeneous backend selection (e.g. via a `GS_ENABLE_NDARRAY` env var):
```
duburcqa: " (e.g. via a `GS_ENABLE_NDARRAY` env var)"  --> what is this? You are referencing Genesis internals in Quadrants' documentation??

docs/source/user_guide/tensor.md
hughperkins: Pickle then unpickle.

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+
+```python
+a = qd.tensor(qd.f32, shape=(4,), backend=qd.Backend.FIELD)
+b = qd.tensor(qd.f32, shape=(4,), backend=qd.Backend.NDARRAY)
+a.copy_from(b)   # raises: cross-backend copy unsupported
+```
+
+If you genuinely need to move data across backends, route it through NumPy (or DLPack/Torch): `a.from_numpy(b.to_numpy())`.
```
duburcqa: why numpy? torch would be better if supported, as it may leverage zero-copy.

docs/source/user_guide/tensor.md
```diff
@@ -0,0 +1,240 @@
...
+
+```python
+t_field = qd.tensor(qd.f32, shape=(4,), backend=qd.Backend.FIELD)
+t_nd    = qd.tensor(qd.f32, shape=(4,), backend=qd.Backend.NDARRAY)
+
+t_field.grad   # currently a Tensor wrapper around a zombie field
+t_nd.grad      # None
+```
+
+Use `needs_grad=True` if you intend to read `.grad`. Integer dtypes are symmetric (`grad is None` on both backends regardless of `needs_grad`). This asymmetry is tracked as a follow-up; it does not affect kernels that opt in to autograd via `needs_grad=True`.
```
duburcqa: "This asymmetry is tracked as a follow-up;" --> Are you sure about this?

docs/source/user_guide/tensor.md
hughperkins: We are using composition here. There is no inheritance relationship between Tensor, Field and NdArray. To what extent does this affect yoru analysis?

hughperkins: (Note that Opus says:

"  It would be something like:
```
  t = qd.tensor(qd.f32, shape=(4, 5))
  # Unwrap to the bare impl
  bare = t.as_subclass(qd.Ndarray)   # returns the underlying Ndarray
```
  But it's awkward because:
  • The caller has to know which backend the tensor uses to pass the right class. If it's field-backed, you'd need t.as_subclass(qd.ScalarField) instead.
  • In PyTorch it works naturally because every subclass IS-A torch.Tensor — same storage, just a different Python type. Here there's no shared base between Ndarray and
    ScalarField.

  So if you did mimic it, you'd probably end up ignoring the argument and just returning the impl regardless — at which point it's just t.unwrap() with extra steps. Not a great
  fit for the composition design."
)

docs/source/user_guide/tensor.md
hughperkins: To move data between backends, we MUST copy.

hughperkins: I suppose you mean, we only need to copy once, whereas using numpy forces copy twice?

docs/source/user_guide/tensor.md
hughperkins: Here's what Opus wrote:

"  1. Signature disambiguation. The first positional arg is currently dtype. If someone passes an Ndarray, you'd need isinstance checking to decide "is this a dtype or an impl to
   wrap?" Doable, but the function now has two completely different code paths behind one name.

  2. Copy or wrap? torch.tensor(x) always copies. torch.as_tensor(x) shares storage. Which semantics would qd.tensor(ndarray) have? Wrapping (zero-copy, shared storage) is what
  qd.wrap does. Copying would be expensive and pointless. But if it wraps, the name tensor() is misleading — torch users would expect a copy.

  3. What happens with extra kwargs? If someone writes qd.tensor(my_ndarray, backend=FIELD) — that's a cross-backend copy request, which isn't supported. Do you silently ignore
  backend=? Error? What about layout=? Applying a layout to an existing ndarray mutates its metadata — is that expected from a "constructor"?

  4. Lifetime/ownership confusion. If qd.tensor(my_ndarray) wraps without copying, the wrapper and the bare ndarray alias the same storage. Mutations through either path are
  visible to the other. Users might not expect that from something that looks like a constructor call.

  5. Double-wrapping. qd.tensor(existing_tensor) — unwrap and re-wrap? No-op? Error? Currently Tensor(Tensor) errors.

  6. The qd.wrap name is already pretty clear. It says exactly what it does. The question is whether users would look for it or just try qd.tensor(my_ndarray) first.

  The cleanest version might be: keep qd.tensor() for allocation only, but if someone passes an impl as the first arg, raise a helpful error saying "use qd.wrap(impl) instead".
  Best of both worlds — discoverable without the ambiguity."

hughperkins: I lean towards Opus suggestion of raising a helpful error saying `use qd.wrap(impl) instead`. Thoughts?

docs/source/user_guide/tensor.md
duburcqa: I'm fine with `unwrap`. It is more targeted and more natural if Quadrants uses composition.

docs/source/user_guide/tensor.md
duburcqa: I'm not ok with Opus comment. All these issues have been sorted out by Torch already. Just need to mimic what torch is doing.

duburcqa: If you want to keep 'wrap', it is up to you, but I don't think Opus is making a valid point.

docs/source/user_guide/tensor.md
duburcqa: Yes.

hughperkins: Opus says:

"  to_torch() allocates a new torch.zeros() and copies into it — it's not zero-copy via DLPack. It runs a bridge kernel to fill the output. So a.from_torch(b.to_torch()) would
  still be two copies:

  1. b.to_torch() — allocates a new torch tensor + bridge kernel copy
  2. a.from_torch(...) — another bridge kernel copy

  Same number of copies as the numpy path. The only advantage would be if to_torch(device=...) keeps it on GPU (avoiding the device→host→device round-trip), but it still
  allocates an intermediate.

  So it's not actually better in copy count — it's better in that the intermediate stays on-device instead of going through CPU. That's still a win for GPU tensors, but it's not
   the "one copy vs two" story."

Thoughts?

docs/source/user_guide/tensor.md
hughperkins: Added exception raising a helpful error saying use qd.wrap(impl) instead.

docs/source/user_guide/tensor.md
duburcqa: This is what to_torch does for now, before merging the new zero-copy support.

hughperkins: Per Opus:

"  Now I can see exactly what happens. Both to_torch() implementations (field and ndarray) do the same thing:

  1. Allocate a new torch.zeros(...) tensor
  2. Copy data into it via a bridge kernel (tensor_to_ext_arr / ndarray_to_ext_arr)

  And both from_torch() implementations:

  1. Call .contiguous() on the torch tensor (potential copy if non-contiguous)
  2. Copy data from the torch tensor into the quadrants storage via a bridge kernel (ext_arr_to_ndarray / _from_external_arr)

  So a.from_torch(b.to_torch()) is actually two copies: one out of b into a new torch tensor, one from that torch tensor into a. Same as the numpy route.

  Neither path uses DLPack zero-copy today — both to_torch() and to_numpy() allocate a fresh output and copy into it via a kernel."

This is for in the context of this PR.

hughperkins: oh, you mean , in the context of the other PR? I see.

hughperkins: Ok, so we should just switch from recommending going via ndarray to going via torch?

duburcqa:
ok to merge

hughperkins:
@claude review

hughperkins:
benchmarks for Genesis on 26b that was just fast-forward merged in:

20260428_0609_stork26b_gen2_ndarray
20260428_0609_stork26b_gen2

PR #562, opened 2026-04-24
[Vulkan] Use SDK validation layer for debugPrintf instead of apt package

hughperkins:
The apt-provided vulkan-validationlayers (1.3.x) uses a different message format than the SDK validation layer (1.4.x): the old format embedded the message ID in pMessage as "UNASSIGNED-DEBUG-PRINTF | MessageID=xxx | <msg>", while the new format puts "VVL-DEBUG-PRINTF" in pMessageIdName and the raw printf output directly in pMessage.

Update the debug callback to use pMessageIdName for identification and pMessage directly for output. Update CI to download the Vulkan SDK (1.4.321.1) instead of installing vulkan-validationlayers via apt, matching the version used during the build.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/test_gpu.yml
+++ b/.github/workflows/test_gpu.yml
@@ -137,9 +137,18 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: '3.10'
-      - name: install vulkan validation layer
+      - name: install vulkan SDK
         run: |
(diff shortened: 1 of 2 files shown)


hughperkins:
@claude review

PR #564, opened 2026-04-25
[CI] Add CI job to check line wrapping of comments and docs

hughperkins:
Checks .md files for unwanted hard-wrapping and code comments/docstrings for wrapping at 80 instead of 120 characters.

- waits 30 minutes before running, in case someone pushes many commits in a row

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


erizmr:
Approved as it is the current best solution as dicussed.

PR #579, opened 2026-04-27
[Lang] Skip IR verifier between passes unless debug=true

duburcqa:
# Skip IR verifier between passes unless debug=true

> The inter-pass IR verifier is a developer aid for catching pass-internal SSA / parent-block bugs. It does not depend on user IR, yet it ran 28 times per kernel compile in release builds and was the second-largest cost in cold GPU compile of adstack-heavy reverse-mode kernels (~21s self time). Gating it on `config.debug` (default `False`) eliminates that tax for production users while keeping full coverage for developers iterating on IR transforms.

**TL;DR**

- Adds `irpass::analysis::verify_if_debug(IRNode*, const CompileConfig&)` next to the existing bare `verify(IRNode*)` primitive.
- 28 inter-pass call sites in `compile_to_offloads.cpp` and `auto_diff.cpp` now go through the wrapper. `config` was already in scope at every site, so the substitution is mechanical.
- Default `qd.init()` (`debug=False`) skips the verifier walk; `qd.init(debug=True)` runs all 28 verify passes as before.
- Cold GPU compile profile of the Genesis Ant kernel (5min total): ~21s of own time was in `IRVerifier::basic_verify` plus its callees `Stmt::has_operand`, `BasicStmtVisitor::visit`. With the gate closed, all of that disappears.

## Why

The verifier checks two pass-invariants:
1. `stmt->parent == current_block_` after each transform - parent block consistency.
2. Every operand reachable in the dominating SSA scope stack - SSA visibility.

Neither depends on user IR. They catch bugs in the transform passes themselves, which CI tests with `debug=True` already cover. End users compiling for speed do not benefit from the per-pass walk.

## Surface API

```cpp
namespace irpass::analysis {
  void verify(IRNode *root);                                     // unchanged primitive, kept for tests
  void verify_if_debug(IRNode *root, const CompileConfig &cfg);  // new, no-op when !cfg.debug
}
```

## Mechanism

`verify_if_debug` is a thin wrapper:

```cpp
void verify_if_debug(IRNode *root, const CompileConfig &config) {
  if (!config.debug) {
    return;
  }
  verify(root);
}
```

The 28 call sites in `compile_to_offloads.cpp` (25) and `auto_diff.cpp` (3) all change from `irpass::analysis::verify(ir)` to `irpass::analysis::verify_if_debug(ir, config)`. `config` was already a local at every site.

## Per-mode matrix

| qd.init mode      | verify_if_debug behavior | inter-pass cost |
|-------------------|--------------------------|-----------------|
| default (debug=False) | early-return | 0s |
| qd.init(debug=True)   | runs verify(root) | unchanged from before |

## Tests

`tests/cpp/analysis/verify_test.cpp`
- `VerifyIfDebug.RunsBareVerifyWhenDebugEnabled` - smoke test exercising the debug=true forwarding path on a valid IR.
- `VerifyIfDebug.NoOpWhenDebugDisabled` - smoke test exercising the debug=false skip path on the same IR plus a redundant bare verify call to confirm both APIs coexist.

A stronger pin (deliberately broken IR + debug=true triggers QD_ERROR while debug=false silently passes) is not feasible without gtest death tests because `QD_ERROR` ends in `QD_UNREACHABLE` rather than throwing.

## Side-effect audit

- Bare `verify(IRNode*)` is unchanged; existing direct callers (none in non-test code per grep) still see identical behavior.
- C++ tests that use `verify` directly for IR validation continue to work.
- CI runs that exercise `debug=True` paths (Genesis differentiable_rigid suite when a regression suspect is being investigated) keep full verification coverage.

### DIFF
--- a/quadrants/analysis/verify.cpp
+++ b/quadrants/analysis/verify.cpp
@@ -6,6 +6,7 @@
 #include "quadrants/ir/statements.h"
 #include "quadrants/ir/visitors.h"
 #include "quadrants/ir/transforms.h"
+#include "quadrants/program/compile_config.h"
 #include "quadrants/system/profiler.h"
 
 namespace quadrants::lang {
@@ -138,6 +139,13 @@ void verify(IRNode *root) {
   } else {
     IRVerifier::run(root);
   }
+}
+
+void verify_if_debug(IRNode *root, const CompileConfig &config) {
+  if (!config.debug) {
+    return;
+  }
+  verify(root);
 }
 }  // namespace irpass::analysis
 
--- a/quadrants/ir/analysis.h
+++ b/quadrants/ir/analysis.h
@@ -194,6 +194,13 @@ std::unordered_set<Stmt *> constexpr_prop(Block *block, std::function<bool(Stmt 
 
 void verify(IRNode *root);
 
+// Same as verify(root) but gated on `config.debug` so release builds skip the per-pass walk. The bare verifier
+// remained the dominant secondary cost (~21s) in profiles of cold GPU compile of adstack-heavy reverse-mode kernels:
+// 28+ call sites between IR passes scale O(N * avg_operands) per call, and adstack push/pop chains over unrolled
+// iterations blow N up. Verification catches pass-internal SSA-visibility / parent-block bugs - it is a developer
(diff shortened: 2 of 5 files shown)


hughperkins:
Nice!

Ok to merge.

PR #591, opened 2026-04-29
[SPIR-V] Shrink reverse-grad kernel MSL by ~50%

duburcqa:
# SPIR-V reverse-grad kernel-size reduction: clamp + OpSelect adstack push, release-mode bounds-check elision, shared count-array, plus diagnostic-only Metal compiler-rejection logging

> Three commits. The first commit halves the cross-compiled MSL size of every reverse-grad kernel via three additive SPIR-V codegen changes (faster local Metal/Vulkan builds, lower host-side memory pressure during compile). The second commit replaces the silent `nil pipeline + nil NSError` log path in `metal_device.mm` with a non-throwing `QD_WARN` that includes the kernel name and cross-compiled MSL byte size, so when Apple's compiler service drops the connection (a known mode on some macOS / Metal-toolchain combinations) the next investigator immediately knows which kernel and how big it was, instead of bisecting from a generic `RhiResult=-1` line. The third commit drops two test cross-references from new test docstrings per the source-comment style rule.

> The PR is **NOT** a fix for the macos-15 M1 GitHub-runner failure on its own. The same g2p kernel that fails on macos-15 (cross-compiled MSL: 689,472 bytes after this PR's reduction) compiles cleanly on macOS 26 / M4 with `xcrun -sdk macosx metal -std=macos-metal2.3 -c` (218 KB AIR, exit 0). So the macos-15 failure is specific to the macos-15 Metal toolchain, not an MSL-size or MSL-content issue we can resolve from this side. This PR is the right shape regardless: the size cut is a real win, and the `metal_device.mm` warning is the right diagnostic surface for any future host that drops the XPC connection silently.

## TL;DR

```
+----------------------------------------+---------+---------+----------------------+
|                Kernel                  | Before  |  After  |       Reduction      |
+----------------------------------------+---------+---------+----------------------+
| g2p_c511 reverse-grad                  |  23,226 |  11,639 | -50.1%               |
| p2g_c509 reverse-grad                  |  37,860 |  16,318 | -56.9%               |
| mpm_grid_op_c65 reverse-grad           |  49,206 |  25,661 | -47.9%               |
| kernel_forward_velocity_c273           |  14,874 |  11,172 | -24.9%               |
| kernel_update_cartesian_space_c289     |  57,853 |   7,617 | -86.8%               |
+----------------------------------------+---------+---------+----------------------+
| Total MSL across the test              | 282,603 | 143,203 | -49.3% (-139,400 LOC)|
+----------------------------------------+---------+---------+----------------------+
```

(Numbers from a local `tests/test_grad.py::test_differentiable_push[gpu]` run with `QD_DUMP_MSL=1` `QD_OFFLINE_CACHE=0`. Test wall-clock on the same run dropped 91.9s -> 38.8s, a 58% speedup that comes from less time spent in the MSL compiler.)

## Why

Two motivations:

1. **Reduce SPIR-V codegen size waste.** The pre-PR `AdStackPushStmt` codegen emits a structured `OpSelectionMerge` / `OpBranchConditional` region per push that spirv-cross renders as ~13 MSL lines per push. For reverse-grad kernels with ~1000 pushes, that's the dominant size amplifier. Per-stack `count_var` `OpVariable Function` slots also become independent `OpPhi` mega-clusters at every enclosing loop header (700+ phis per merge). Both can be compressed without correctness change. The release-mode bounds-check (clamp + atomic-signal) on SPIR-V is currently always live; LLVM has always elided it in release; aligning the gates lets release builds skip the per-push branch entirely.

2. **Make Metal pipeline-create failures self-describing.** When Apple's MSL compiler service drops the XPC connection mid-compile, `newComputePipelineStateWithFunction:error:` returns `nil` with `error == nil`. The pre-PR path silently returned `nullptr`; the user only saw the generic `runtime.cpp:298 RhiResult=-1` line with no kernel name or byte size. The new path warn-logs the kernel name and cross-compiled MSL byte size, so any future investigator reading CI artifacts immediately sees which kernel hit the path.

### 1. AdStackPushStmt: clamp + OpSelect instead of structured if-then-else

`quadrants/codegen/spirv/spirv_codegen.cpp::TaskCodegen::visit(AdStackPushStmt*)` previously emitted a structured `OpSelectionMerge` / `OpBranchConditional` region around every push, with the then-branch doing the in-bounds store and the else-branch publishing the overflow signal. The new emit folds the entire region into:

1. `clamped_idx = GLSLstd450UMin(count, max_size - 1)`
2. unconditional store to `primal[clamped_idx]` (and `adjoint[clamped_idx] = 0` for heap_float)
3. unconditional `count++`
4. `signal = OpSelect(count >= max_size, stack_id+1, 0)`
5. unconditional `OpAtomicUMax(overflow_buffer[0], signal)`

The clamp keeps the OpAccessChain in-bounds; the atomic-max with 0 is a no-op when the stack didn't overflow, so the host-readable flag still ends up at `stack_id + 1` only when an actual overflow happened. spirv-cross emits this as straight-line MSL: ~5 lines per push instead of ~13.

### 2. Bounds-check gate switched to `check_out_of_bound`

The clamp + atomic-signal pair from above is now gated on `compile_config->check_out_of_bound || compile_config->debug` in SPIR-V codegen. Release builds elide the entire bounds check, mirroring LLVM's release-build push (which has always relied on `determine_ad_stack_size` producing a tight static bound and dropped the per-push runtime guard). LLVM's six adstack visitors switch their gate from `compile_config.debug` to `compile_config.check_out_of_bound` so the two backends key off the same flag.

| Backend | Bounds-check path | Release behaviour |
|---|---|---|
| LLVM (CPU / CUDA / AMDGPU) | `check_out_of_bound` -> `stack_init` / `stack_push` runtime calls | inline ops, no overflow flag |
| SPIR-V (Metal / Vulkan) | `check_out_of_bound \|\| debug` -> clamp + OpAtomicUMax signal | unconditional store, no overflow flag |

`CompileConfig::fit()` already promotes `debug=True` to `check_out_of_bound=True`, so existing `qd.init(debug=True)` users see no behaviour change. Users who explicitly set `qd.init(check_out_of_bound=True, debug=False)` now also get the bounds check on LLVM, which they didn't before. The OR with `debug` in the SPIR-V gate preserves the `qd.init(debug=True)` path on Metal / Vulkan, where `Program::init` force-disables `check_out_of_bound` because those arches lack `Extension::assertion`.

### 3. Shared count-array OpVariable for adstack `count_var`

Each adstack `count_var` used to be its own `OpVariable Function` of type `uint`. spirv-opt's `LocalMultiStoreElim` / `SSARewrite` promoted each into its own SSA chain, which became a separate `OpPhi` at every enclosing loop header. spirv-cross then emitted each phi as one `uint _N;` forward-decl + one `_N = _N;` alias copy per predecessor branch. Reverse-grad kernels with hundreds of adstacks crossing a single loop accumulated phi mega-clusters of 700+ entries per loop header.

This PR replaces the per-stack scalar OpVariable with a single Function-scope `uint[num_ad_stacks_]` array, allocated lazily on first `ad_stack_count_ptr(stack_id)` call and indexed by `OpAccessChain` per push / pop / load-top. spirv-opt's mem2reg passes do not promote OpAccessChain into an aggregate, so the slots stay memory-backed and never become per-stack phis. The array is sized from a pre-pass scan that counts `AdStackAllocaStmt` nodes (`num_ad_stacks_`).

This is the single biggest lever: `kernel_update_cartesian_space_c289_0_reverse_grad` drops from 57,853 MSL lines to 7,617 (-87%) entirely from this change.

### 4. Metal pipeline / library failure: `QD_WARN` + `nullptr` return (not `QD_ERROR`)

`quadrants/rhi/metal/metal_device.mm::create_compute_pipeline` and `MetalDevice::get_mtl_library` previously took the `nil pipeline + nil NSError` path silently and returned `nullptr` (or, on the `err != nil` path, called `RHI_LOG_ERROR` and returned `nullptr`). The new path logs at WARN level with `QD_WARN`, including the kernel name (where available) and cross-compiled MSL byte size:

```
[W ...] [metal_device.mm:create_compute_pipeline@206] Apple's Metal compiler service
rejected the compute-pipeline build for kernel 'g2p_c511_0_reverse_grad_0_t00'
(cross-compiled MSL size: 689472 bytes) without returning a structured error. The XPC
service drops its connection in this shape; the underlying cause is host-toolchain-
specific and is not recoverable from this side.
[E ...] [runtime.cpp:CompiledQuadrantsKernel@298] Failed to create pipeline ... RhiResult=-1
```

`QD_WARN` rather than `QD_ERROR` is critical: `QD_ERROR` ends with `throw s` (where `s` is a bare `std::string`), and `MetalDevice::create_pipeline` is declared `noexcept` and only catches `std::exception` derivatives. A throw of `std::string` here would cross the `noexcept` boundary and trip `std::terminate()`, replacing the existing clean Python `RuntimeError` translation with a fatal process abort. With `QD_WARN`, no exception is thrown inside the noexcept function; the `nullptr` return is converted by the caller to `RhiResult::error`, the `runtime.cpp:298 QD_ERROR_IF` then throws `std::string`, the existing `pybind11` translator (`quadrants/python/py_exception_translator.cpp`) catches it and raises `PyExc_RuntimeError`. Verified empirically by force-injecting the failure path locally and observing the Python-level `RuntimeError` exception.

The wording deliberately does not assert a specific cause (size, construct, driver bug, ...). The XPC connection drop is observable from this side; the actual reason in the toolchain is not.

| Backend | Adstack push shrink | Bounds-check gate | Count-array shared | Metal warn-log |
|---|---|---|---|---|
| arm64 / x64 (LLVM CPU) | N/A (LLVM emits inline) | switched to `check_out_of_bound` | N/A | N/A |
| CUDA / AMDGPU (LLVM GPU) | N/A | switched to `check_out_of_bound` | N/A | N/A |
| Vulkan (SPIR-V) | clamp + OpSelect | `check_out_of_bound \|\| debug` | yes | N/A |
| Metal (SPIR-V) | clamp + OpSelect | `check_out_of_bound \|\| debug` | yes | yes |

### `tests/python/test_adstack.py::test_adstack_overflow_raises`

Existing test, kept on `debug=True`. Verifies that an adstack push past the published `max_size` raises `RuntimeError("[Aa]dstack overflow")` on the next `qd.sync()`. `debug=True` implies `check_out_of_bound=True` via `CompileConfig::fit`, so the bounds-check codepath is live.

### `tests/python/test_adstack.py::test_adstack_overflow_raises_check_oob_explicit` (new)

Same overflow scenario as the test above but with `check_out_of_bound=True` set explicitly without `debug=True`. Pins the gating to `check_out_of_bound` rather than `debug`: a release-build user who explicitly opts into bounds-checks gets the same `RuntimeError` as a debug-mode user. Excluded on Metal / Vulkan because `Program::init` force-disables `check_out_of_bound` on arches without `Extension::assertion`, so the explicit-flag spelling alone cannot light up the bounds check there.

### `tests/python/test_adstack.py::test_adstack_overflow_flag_resets_after_catch`

Existing test, unchanged. Pins that `check_adstack_overflow()` clears the flag after raising so a subsequent `qd.sync()` returns normally.

### Local AD test status with `QD_OFFLINE_CACHE=0`

1214 tests pass on this branch (`tests/python/test_adstack.py` plus the broader `test_ad_*.py` files) across `arch=arm64`, `arch=metal-2`, `arch=vulkan-0`. Same numbers on pristine `origin/main`.

## Side-effect audit

|---|---|---|
| Offline cache key | `compile_config.h::check_out_of_bound` is already part of the config feeding the cache key; no schema change | OK |
| `task_attribs.ad_stack` serialisation | Layout unchanged: `per_thread_stride_*_compile_time` and `allocas[]` populated identically | OK |
| `info.count_var` users | All loads / stores go through `load_variable` / `store_variable` which accept `kVariablePtr` whether the underlying is a fresh `OpVariable` or an `OpAccessChain` element | OK |
| Adstack overflow signal | `OpAtomicUMax(buffer, 0)` is a no-op for the host-visible value, so the runtime still observes a clear flag iff some thread actually overflowed | OK |
| Reverse-pass count semantics | `count++` is now unconditional, so push and pop are balanced even when the in-bounds check would have skipped the increment. `LoadTop*/AccAdjoint` already clamp via UMin so an overflowed count of UINT_MAX still produces a clamped in-bounds index | OK |
| LLVM `compile_config.debug` integer-overflow checks (BinaryOpStmt / shift sites) | Untouched - still gate on `debug` (`codegen_llvm.cpp` lines 503/514/525/560); only the six adstack visitors switched to `check_out_of_bound` | OK |
| Vulkan push-constant / descriptor binding layout | Unchanged | OK |
| `metal_device.mm` raster-fallback site (`build_mtl_render_pipeline`) | This PR does NOT touch the raster site - it still uses the pre-PR `RHI_LOG_ERROR` + silent-on-`err==nil` behaviour. Out of scope for this PR; flagged here so the audit table matches the diff | Untouched (intentional) |
| Process-abort regression on the new metal `nil err` path | Verified empirically: with `QD_WARN` + `nullptr` return, force-injecting the path produces a Python `RuntimeError` (not `std::terminate`) | OK |

### DIFF
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -102,7 +102,7 @@ Per-backend support:
 |---------|------------------|
 | CPU | yes |
 | CUDA | yes |
-| AMDGPU | no (compile error) |
+| AMDGPU | no (silently dropped) |
 | Metal | no (silently dropped) |
 | Vulkan | yes (via debug-printf SPIR-V extension) |
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -36,11 +36,9 @@ x[3] = 10.0   # AssertionError in debug mode, silent corruption otherwise
 a = x[-1]     # AssertionError in debug mode
 ```
 
-The same flag also enables a deferred runtime check on the adstack used by reverse-mode autodiff: a push past the per-stack capacity (set via `qd.init(ad_stack_size=...)` or per-alloca by `determine_ad_stack_size`) raises `RuntimeError("[Aa]dstack overflow")` on the next `qd.sync()`. Without bounds-checking, an adstack overflow silently writes past the per-thread slab and produces a wrong gradient.
+### Adstack overflow check
 
-`debug=True` is a superset of `check_out_of_bound=True`. Setting `qd.init(check_out_of_bound=True)` without `debug=True` enables the field bounds check and the adstack overflow check, but skips kernel `assert` evaluation, integer overflow detection on arithmetic, and the other checks listed below. Use this when you want bounds-safety in a release-build app without paying the full debug-mode cost.
-
-On the Metal and Vulkan backends, `check_out_of_bound=True` is silently disabled at `qd.init` time because those backends lack the in-kernel assertion extension that the field bounds check relies on; passing it on its own gives you neither the field bounds check nor the adstack overflow check. Pass `debug=True` instead: that keeps the adstack overflow check live (it is gated independently and does not need the assertion extension), but the field bounds check still does not fire on these backends.
+`debug=True` also enables a deferred runtime check on the adstack used by reverse-mode autodiff. A push past the per-stack capacity (set via `qd.init(ad_stack_size=...)` or per-alloca by `determine_ad_stack_size`) raises `RuntimeError("[Aa]dstack overflow")` on the next `qd.sync()`. Without the check, an adstack overflow silently writes past the per-thread slab and produces a wrong gradient.
 
 ### Assertions in kernels
 
@@ -66,6 +64,8 @@ Debug mode adds runtime checks to every field access and assertion, which can si
 A typical workflow:
 1. Develop with `debug=True` to catch bounds errors and logic bugs
 2. Switch to `debug=False` (the default) for benchmarking and production runs
+
+For bounds safety in a release build without the rest of debug mode, set [`check_out_of_bound=True`](./init_options.md#check_out_of_bound) instead. `debug=True` always implies `check_out_of_bound=True`.
 
 ## Other debugging tools
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -36,6 +36,10 @@ x[3] = 10.0   # AssertionError in debug mode, silent corruption otherwise
 a = x[-1]     # AssertionError in debug mode
 ```
 
+The same flag also enables a deferred runtime check on the adstack used by reverse-mode autodiff: a push past the per-stack capacity (set via `qd.init(ad_stack_size=...)` or per-alloca by `determine_ad_stack_size`) raises `RuntimeError("[Aa]dstack overflow")` on the next `qd.sync()`. Without bounds-checking, an adstack overflow silently writes past the per-thread slab and produces a wrong gradient.
+
+`debug=True` is a superset of `check_out_of_bound=True`. Setting `qd.init(check_out_of_bound=True)` without `debug=True` enables the field bounds check and the adstack overflow check, but skips kernel `assert` evaluation, integer overflow detection on arithmetic, and the other checks listed below. Use this when you want bounds-safety in a release-build app without paying the full debug-mode cost.
+
 ### Assertions in kernels
 
 The `assert` statement works inside kernels when debug mode is enabled:
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -106,9 +106,9 @@ Per-backend support:
 | Metal | no (silently dropped) |
 | Vulkan | yes (via debug-printf SPIR-V extension) |
 
-Output from GPU kernels may appear out of order due to parallel execution.
+**Note.** Output from GPU kernels may appear out of order due to parallel execution.
 
-Important: remove kernel `print()` calls before benchmarking. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears at the right place in the log; this synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow makes the `print()` unreachable. The cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
+**Important.** Remove kernel `print()` calls before benchmarking. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears at the right place in the log. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow makes the `print()` unreachable; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
 
 ### Dumping compiled IR
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -108,7 +108,7 @@ Per-backend support:
 
 **Note.** Output from GPU kernels may appear out of order due to parallel execution.
 
-**Important.** Avoid kernel `print()` calls in production code if possible. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears at the right place in the log. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
+**Important.** Avoid kernel `print()` calls in production code if possible. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears as close as possible to the call site. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
 
 ### Dumping compiled IR
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -108,7 +108,7 @@ Per-backend support:
 
 **Note.** Output from GPU kernels may appear out of order due to parallel execution.
 
-**Important.** Avoid kernel `print()` calls in production code. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears at the right place in the log. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
+**Important.** Avoid kernel `print()` calls in production code if possible. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears at the right place in the log. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
 
 ### Dumping compiled IR
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -106,8 +106,6 @@ Per-backend support:
 | Metal | no (silently dropped) |
 | Vulkan | yes (via debug-printf SPIR-V extension) |
 
-**Note.** Output from GPU kernels may appear out of order due to parallel execution.
-
 **Important.** Avoid kernel `print()` calls in production code where you can. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears as close as possible to the call site. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
 
 ### Dumping compiled IR
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -87,7 +87,7 @@ See also [Troubleshooting](./troubleshooting.md).
 
 ### Printing from kernels
 
-`print()` works inside kernels for scalar values, which can be useful for debugging:
+`print()` works inside kernels for scalar values:
 
 ```python
 @qd.kernel
@@ -96,7 +96,19 @@ def debug_kernel(a: qd.Template) -> None:
         print("i =", i, "val =", a[i])
 ```
 
-Note that print output from GPU kernels may appear out of order due to parallel execution.
+Per-backend support:
+
+| Backend | Kernel `print()` |
+|---------|------------------|
+| CPU | yes |
+| CUDA | yes |
+| AMDGPU | no (compile error) |
+| Metal | no (silently dropped) |
+| Vulkan | yes (via debug-printf SPIR-V extension) |
+
+Output from GPU kernels may appear out of order due to parallel execution.
+
+Important: remove kernel `print()` calls before benchmarking. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears at the right place in the log; this synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow makes the `print()` unreachable. The cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
 
 ### Dumping compiled IR
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -108,7 +108,7 @@ Per-backend support:
 
 **Note.** Output from GPU kernels may appear out of order due to parallel execution.
 
-**Important.** Avoid kernel `print()` calls in production code. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears at the right place in the log. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow makes the `print()` unreachable; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
+**Important.** Avoid kernel `print()` calls in production code. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears at the right place in the log. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
 
 ### Dumping compiled IR
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -108,7 +108,7 @@ Per-backend support:
 
 **Note.** Output from GPU kernels may appear out of order due to parallel execution.
 
-**Important.** Avoid kernel `print()` calls in production code if possible. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears as close as possible to the call site. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
+**Important.** Avoid kernel `print()` calls in production code where you can. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears as close as possible to the call site. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
 
 ### Dumping compiled IR
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -87,7 +87,7 @@ See also [Troubleshooting](./troubleshooting.md).
 
 ### Printing from kernels
 
-`print()` works inside kernels for scalar values:
+`print()` works inside kernels for scalar values, which can be useful for debugging:
 
 ```python
 @qd.kernel
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -36,9 +36,9 @@ x[3] = 10.0   # AssertionError in debug mode, silent corruption otherwise
 a = x[-1]     # AssertionError in debug mode
 ```
 
-### Adstack overflow check
+#### Adstack overflow
 
-`debug=True` also enables a deferred runtime check on the adstack used by reverse-mode autodiff. A push past the per-stack capacity (set via `qd.init(ad_stack_size=...)` or per-alloca by `determine_ad_stack_size`) raises `RuntimeError("[Aa]dstack overflow")` on the next `qd.sync()`. Without the check, an adstack overflow silently writes past the per-thread slab and produces a wrong gradient.
+`debug=True` also enables a deferred bounds check on the adstack used by reverse-mode autodiff. A push past the per-stack capacity (set via `qd.init(ad_stack_size=...)` or per-alloca by `determine_ad_stack_size`) raises `RuntimeError("[Aa]dstack overflow")` on the next `qd.sync()`. Without the check, an adstack overflow silently writes past the per-thread slab and produces a wrong gradient.
 
 ### Assertions in kernels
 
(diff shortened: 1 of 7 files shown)


quadrants/rhi/metal/metal_device.mm
```diff
@@ -167,13 +167,34 @@
...
+                 "(cross-compiled MSL size: %zu bytes) without returning a "
+                 "structured error. This is the "
+                 "typical signature of an internal Metal-toolchain shader-size "
+                 "limit on this host: split "
+                 "the kernel into smaller pieces or reduce its body so the "
+                 "cross-compiled MSL drops below "
+                 "the host's threshold.",
+                 name.c_str(), msl.size());
       }
+      QD_ERROR("[metal_device.mm] {}", msgbuf.data());
```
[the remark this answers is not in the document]
duburcqa: Fixed in 78e94d3f1: replaced `QD_ERROR` with `QD_WARN` on both new sites; the throw no longer originates inside the noexcept `MetalDevice::create_pipeline`, so `runtime.cpp:298 QD_ERROR_IF` is the throwing site (not noexcept), pybind11 translator catches the `std::string` and raises a Python `RuntimeError`. Verified empirically by force-injecting the failure path locally - Python sees `RuntimeError` cleanly, no `std::terminate`.

quadrants/rhi/metal/metal_device.mm
```diff
@@ -167,13 +167,34 @@
                                                            error:&err];
 
     if (mtl_compute_pipeline_state == nil) {
+      std::array<char, 4096> msgbuf;
```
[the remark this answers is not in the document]
duburcqa: Fixed by updating the PR description (no code change): the side-effect audit row for the raster site now reads "Untouched (intentional)" and explicitly notes this PR does not modify `build_mtl_render_pipeline`.

duburcqa:
@claude review

docs/source/user_guide/debug.md
```diff
@@ -36,6 +36,10 @@ x[3] = 10.0   # AssertionError in debug mode, silent corruption otherwise
 a = x[-1]     # AssertionError in debug mode
 ```
 
+The same flag also enables a deferred runtime check on the adstack used by reverse-mode autodiff: a push past the per-stack capacity (set via `qd.init(ad_stack_size=...)` or per-alloca by `determine_ad_stack_size`) raises `RuntimeError("[Aa]dstack overflow")` on the next `qd.sync()`. Without bounds-checking, an adstack overflow silently writes past the per-thread slab and produces a wrong gradient.
+
+`debug=True` is a superset of `check_out_of_bound=True`. Setting `qd.init(check_out_of_bound=True)` without `debug=True` enables the field bounds check and the adstack overflow check, but skips kernel `assert` evaluation, integer overflow detection on arithmetic, and the other checks listed below. Use this when you want bounds-safety in a release-build app without paying the full debug-mode cost.
```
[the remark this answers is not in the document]
duburcqa: fixed.

docs/source/user_guide/debug.md
```diff
@@ -36,6 +36,12 @@ x[3] = 10.0   # AssertionError in debug mode, silent corruption otherwise
 a = x[-1]     # AssertionError in debug mode
 ```
 
+The same flag also enables a deferred runtime check on the adstack used by reverse-mode autodiff: a push past the per-stack capacity (set via `qd.init(ad_stack_size=...)` or per-alloca by `determine_ad_stack_size`) raises `RuntimeError("[Aa]dstack overflow")` on the next `qd.sync()`. Without bounds-checking, an adstack overflow silently writes past the per-thread slab and produces a wrong gradient.
```
hughperkins: Lets add some kind of `adstack` section/subsection header please.

hughperkins: maybe `#### adstack` ?

docs/source/user_guide/debug.md
```diff
@@ -36,6 +36,12 @@ x[3] = 10.0   # AssertionError in debug mode, silent corruption otherwise
...
+
+`debug=True` is a superset of `check_out_of_bound=True`. Setting `qd.init(check_out_of_bound=True)` without `debug=True` enables the field bounds check and the adstack overflow check, but skips kernel `assert` evaluation, integer overflow detection on arithmetic, and the other checks listed below. Use this when you want bounds-safety in a release-build app without paying the full debug-mode cost.
```
hughperkins: This seems like a mixture of general debug stuff, and adstack-speicfic stuff. Can we factorize out the general stuff to go outsdie fo the new adstack subsection, and keep just the adstack specific stuff here please.

docs/source/user_guide/debug.md
```diff
@@ -36,6 +36,12 @@ x[3] = 10.0   # AssertionError in debug mode, silent corruption otherwise
...
+
+On the Metal and Vulkan backends, `check_out_of_bound=True` is silently disabled at `qd.init` time because those backends lack the in-kernel assertion extension that the field bounds check relies on; passing it on its own gives you neither the field bounds check nor the adstack overflow check. Pass `debug=True` instead: that keeps the adstack overflow check live (it is gated independently and does not need the assertion extension), but the field bounds check still does not fire on these backends.
```
hughperkins: check_out_of_bounds tru seems alike a general thing, so lets also move it outside of the adstack section pelase.

docs/source/user_guide/debug.md
hughperkins: you havent introduced check_out_of_bounds yet. It should be a sepearate section to 'debug' I feel. But .... why introduce a separate flag? Why not just have a single debug flag, for simplicity?

duburcqa:
`========== 649 passed, 3 skipped, 2 xfailed in 1089.97s (0:18:09) ===========`

docs/source/user_guide/debug.md
hughperkins: Ok, I see you've started to provide the reasons, but I feel this could be structured more clearly, and I think it's confusing to have two flags, one of which is a subset of the other, so if we can avoid that that might be cleaner. I guess full debug is super slow?

what happens if debug is true, and check_out_of_bounds is false?

duburcqa:
|env                     |batch_size|backend|gjk_collision|constraint_solver|runtime_fps_590|runtime_fps_591|runtime_fps_delta_pct|
|------------------------|----------|-------|-------------|-----------------|---------------|---------------|---------------------|
|anymal_random           |30000     |cuda   |-            |-                |9314890        |9225767        |-0.96                |
|anymal_uniform          |30000     |cuda   |-            |-                |12361607       |12197108       |-1.33                |
|anymal_uniform_kinematic|0         |cpu    |-            |-                |2028           |2024           |-0.20                |
|anymal_uniform_kinematic|30000     |cuda   |-            |-                |10266957       |10462211       |+1.90                |
|anymal_zero             |0         |cpu    |-            |-                |7073           |7295           |+3.14                |
|anymal_zero             |30000     |cuda   |-            |-                |19096316       |18941109       |-0.81                |
|box_pyramid_3           |4096      |cuda   |-            |-                |968342         |975716         |+0.76                |
|box_pyramid_4           |4096      |cuda   |-            |-                |395824         |389584         |-1.58                |
|box_pyramid_5           |4096      |cuda   |-            |-                |141715         |139459         |-1.59                |
|box_pyramid_6           |4096      |cuda   |False        |-                |59547          |58861          |-1.15                |
|box_pyramid_6           |4096      |cuda   |True         |-                |61918          |59940          |-3.19                |
|dex_hand                |4096      |cuda   |-            |-                |17081          |17181          |+0.59                |
|duck_in_box_easy        |30000     |cuda   |False        |-                |26469341       |26766339       |+1.12                |
|duck_in_box_easy        |30000     |cuda   |True         |-                |9681964        |9660875        |-0.22                |
|duck_in_box_hard        |0         |cpu    |-            |-                |5204           |5154           |-0.96                |
|duck_in_box_hard        |30000     |cuda   |False        |-                |10077402       |10149398       |+0.71                |
|duck_in_box_hard        |30000     |cuda   |True         |-                |3493847        |3354442        |-3.99                |
|franka                  |30000     |cuda   |-            |-                |21764102       |21686917       |-0.35                |
|franka_accessors        |0         |cpu    |-            |-                |1182           |1214           |+2.71                |
|franka_accessors        |30000     |cuda   |-            |-                |15853954       |15474380       |-2.39                |
|franka_free             |30000     |cuda   |-            |-                |32626193       |32190516       |-1.34                |
|franka_random           |0         |cpu    |-            |-                |6301           |6287           |-0.22                |
|franka_random           |30000     |cuda   |-            |CG               |16793645       |16658204       |-0.81                |
|franka_random           |30000     |cuda   |-            |Newton           |16535074       |16288336       |-1.49                |
|franka_random           |30000     |cuda   |False        |-                |16786850       |16318058       |-2.79                |
|franka_random           |30000     |cuda   |True         |-                |11327811       |11463262       |+1.20                |
|g1_fall                 |4096      |cuda   |-            |Newton           |920318         |918658         |-0.18                |
|go2                     |4096      |cuda   |False        |CG               |3668476        |3653937        |-0.40                |
|go2                     |4096      |cuda   |False        |Newton           |4386696        |4383261        |-0.08                |
|go2                     |4096      |cuda   |True         |-                |3230236        |3290806        |+1.88                |
|shadow_hand_cubes       |0         |cpu    |-            |-                |41             |41             |+0.00                |
|shadow_hand_cubes_sparse|0         |cpu    |-            |-                |66             |66             |+0.00                |

[speed_test_591.txt](https://github.com/user-attachments/files/27217910/speed_test_591.txt)

docs/source/user_guide/debug.md
```diff
@@ -90,7 +96,19 @@ def debug_kernel(a: qd.Template) -> None:
...
+|---------|------------------|
+| CPU | yes |
+| CUDA | yes |
+| AMDGPU | no (compile error) |
+| Metal | no (silently dropped) |
+| Vulkan | yes (via debug-printf SPIR-V extension) |
+
+**Note.** Output from GPU kernels appears in order despite parallel execution because all kernels are queued in the same compute stream.
+
+**Important.** Avoid kernel `print()` calls in production code where you can. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears as close as possible to the call site. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
```
hughperkins, line 111: nice

docs/source/user_guide/init_options.md
```diff
@@ -45,10 +45,50 @@ Number of host threads used when compiling kernels. Default `4`. Raise on machin
...
+- Use `check_out_of_bound=True` on its own when you want bounds + adstack safety in a release build without paying for everything else `debug` adds.
+
+Per-backend support:
+
+| Backend | Field bounds check | Adstack overflow check |
+|---------|--------------------|------------------------|
+| CPU | with `check_out_of_bound=True` or `debug=True` | with `check_out_of_bound=True` or `debug=True` |
+| CUDA | with `check_out_of_bound=True` or `debug=True` | with `check_out_of_bound=True` or `debug=True` |
+| AMDGPU | with `check_out_of_bound=True` or `debug=True` | with `check_out_of_bound=True` or `debug=True` |
+| Metal | never (no in-kernel assertion mechanism) | with `debug=True` only |
```
hughperkins: wait. why the inconcistency for 'adstack overflow check' on Metal and Vulkan?

hughperkins: I think the behavior should be consistent across platforms, except for features not supported by a platform at all (so 'never' is ok for vulkan and metal for example (though not ideal of course)).

docs/source/user_guide/init_options.md
```diff
@@ -45,10 +45,50 @@ Number of host threads used when compiling kernels. Default `4`. Raise on machin
...
+| Vulkan | never (no in-kernel assertion mechanism) | with `debug=True` only |
+
+The adstack overflow check is gated independently of the assertion mechanism, so `debug=True` activates it on every backend - including Metal and Vulkan, where the field bounds check stays unavailable. On Metal and Vulkan, `check_out_of_bound` is silently reset to `False` at `qd.init` time (a warning is logged); passing it on its own gives neither check on those backends.
```
hughperkins: I think we should lave it enabled on Metal, and narrow the warning to say that only adstack overflows will be checked, not out of bounds.

But actually, now I think about it, why should 'out of bound' track 'adstack overflow'?

I think these should be two different flags.

docs/source/user_guide/init_options.md
duburcqa: I'm ok with that, but this is pre-existing in this PR, here I'm just documenting the current state. I could fix it in this PR if you want.

docs/source/user_guide/init_options.md
duburcqa: That makes sense. Do you want to use 'debug' for this or a new flag? You want to do the changes in this PR?

docs/source/user_guide/init_options.md
hughperkins: Oh, I see, I assumed that these were changes in this PR.

Yeah, ok, let's not feature-flate this PR :) Thank you for the doc :) I think it explains clearly the current situation. 🙌

docs/source/user_guide/init_options.md
hughperkins: Since you are just documenting the existing behavior, let's not change this in this doc. Thank you :)

hughperkins:
Ok, doc looks good to me. Whilst it looks like these changes just target reverse-grad autodiff, lets get genesis unit test results and genesis benchmark results please, just to be sure.

hughperkins:
oh they're already there.

checklist:
- user-facing doc filled in
- genesi benchmarks neutral
- genesis unit tests passing 

=> ok to merge

quadrants/codegen/spirv/detail/spirv_codegen.h
```diff
@@ -246,6 +249,19 @@ class TaskCodegen : public IRVisitor {
...
   uint32_t ad_stack_heap_per_thread_stride_int_{0};
+  // Total number of `AdStackAllocaStmt` the body will visit, pre-computed by the same scan that builds the heap
+  // strides. Used to size a single shared Function-scope `uint[num_ad_stacks_]` count array (see
+  // `ensure_ad_stack_count_array_var`) so that all per-stack `count` values share one OpVariable accessed via
+  // OpAccessChain. The shared array prevents spirv-opt's `LocalMultiStoreElim` / `SSARewrite` from promoting each
+  // count to a separate phi at every enclosing loop header - reverse-grad kernels that allocate hundreds of
+  // adstacks across enclosing loops would otherwise grow phi mega-clusters of hundreds of entries per loop
+  // header, which spirv-cross emits as one `uint _N;` forward-decl and one `_N = _N;` alias copy per phi per
+  // predecessor branch in the cross-compiled MSL.
+  uint32_t num_ad_stacks_{0};
```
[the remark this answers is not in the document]
duburcqa: fixed.

docs/source/user_guide/debug.md
```diff
@@ -90,7 +96,19 @@ def debug_kernel(a: qd.Template) -> None:
...
+Per-backend support:
+
+| Backend | Kernel `print()` |
+|---------|------------------|
+| CPU | yes |
+| CUDA | yes |
+| AMDGPU | no (compile error) |
+| Metal | no (silently dropped) |
+| Vulkan | yes (via debug-printf SPIR-V extension) |
+
```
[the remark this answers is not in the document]
duburcqa: fixed.

tests/python/test_adstack.py
```diff
@@ -1010,13 +1011,28 @@ def test_adstack_overflow_raises():
...
+    # a release-build app that still wants the deferred adstack-overflow error path) must get the same
+    # RuntimeError as a `debug=True` user. Regression coverage for the LLVM adstack-visitor codepaths gating on
+    # `check_out_of_bound` instead of `debug`.
+    # Excludes Metal / Vulkan: `Program::init` force-disables `check_out_of_bound` on arches without
+    # `Extension::assertion`, so the explicit-flag spelling alone cannot light up the bounds check there. The
+    # SPIR-V codegen gate ORs `debug` in to cover the `qd.init(debug=True)` path on those backends.
+    compute, _, _ = _overflowing_compute()
+    with pytest.raises((AssertionError, RuntimeError), match=r"[Aa]dstack overflow"):
+        compute.grad()
+        qd.sync()
```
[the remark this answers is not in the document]
duburcqa: fixed

PR #592, opened 2026-04-29
[Bug] Fix: promote Ndarray to AnyArray in build_Name for flattened struct fields

hughperkins:
populate_global_vars_from_dataclass stores raw unwrapped Ndarray instances in global_vars under flattened names (__qd_foo__qd_bar). These resolve via build_Name, which—unlike build_Attribute—was missing the _promote_ndarray_if_declared call. This caused qd.grouped() and struct-for loops to receive raw ScalarNdarray objects instead of AnyArray proxies, triggering a TypeError in begin_frontend_struct_for.

The bug is only observable after a field→ndarray backend switch (gs.destroy() + gs.init()) because in field-only mode the raw values are Field subclasses that pass the isinstance check directly.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
@claude review

PR #600, opened 2026-04-30
[Perf] Add cross-GPU subgroup.ballot(predicate) primitive

hughperkins:
Implement a portable ballot operation that returns a u32 bitmask where bit i is set if lane i's predicate is non-zero. Works across CUDA (__ballot_sync), AMDGPU (amdgcn_ballot.i32), and SPIR-V/Vulkan (OpGroupNonUniformBallot).

Follows the same cross-backend pattern as subgroup.shuffle: a single Python API (subgroup.ballot) dispatches to the appropriate backend intrinsic at codegen time. On AMDGPU CDNA with 64-wide wavefronts only the low 32 bits are returned, consistent with the u32 return type.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


docs/source/user_guide/subgroup.md
hughperkins, line 1: doc looks ok to me

PR #607, opened 2026-05-01
[Perf] Skip _recursive_set_args for all-Field frozen dataclass structs

hughperkins:
When a frozen dataclass struct has only qd.Field entries (no Ndarrays, no templates that consume launch-context slots), the entire struct contributes zero arguments to the kernel launch context.

Two changes eliminate the per-launch overhead for these structs:

1. _func_base.py: After computing the field plan, identify which entries are "non-field" (consume slots). Cache as _qd_non_field_plan on the struct instance. If empty, return (0, True) without recursion.

2. kernel.py: Before calling _recursive_set_args, check for a cached empty _qd_non_field_plan. If present, skip the call entirely, avoiding even function-call overhead.

On the anymal_uniform_kinematic benchmark (two robots, random control each step, field backend), this eliminates a -11% regression caused by struct-class annotations replacing qd.template() in kernel signatures.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/quadrants/lang/_func_base.py
+++ b/python/quadrants/lang/_func_base.py
@@ -143,6 +143,17 @@ def _get_frozen_dc_unwrapped(v: Any, fields_dict: dict) -> dict[str, Any]:
         object.__setattr__(v, "_qd_dc_unwrapped", unwrapped)
     except AttributeError:
         pass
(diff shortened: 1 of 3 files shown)


hughperkins:
@claude review

hughperkins:
@codex review

hughperkins:
@codex review

PR #618, opened 2026-05-04
[FEATURE] Support external Metal command queue via qd.init

hughperkins:
Allow callers (e.g. Genesis) to share PyTorch MPS's MTLCommandQueue with Quadrants so both frameworks dispatch GPU work on the same Metal queue. This eliminates the need for qd.sync() + torch.mps.synchronize() at every DLPack zero-copy interop point — Metal's sequential command buffer semantics guarantee ordering automatically.

Changes:
- CompileConfig: add external_metal_command_queue (uint64_t, default 0)
- MetalStream: add owns_queue_ flag; create_with_external_queue() factory
- MetalDevice: accept optional external queue; create_with_external_queue()
- MetalProgramImpl: route config→device creation
- export_lang.cpp: expose new field to Python
- field.py: gate qd.sync() and torch.mps.synchronize() on shared queue
- test_metal_shared_queue.py: 4 tests covering init, zero-copy, and sync gating

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/docs/source/user_guide/metal_shared_queue.md
+++ b/docs/source/user_guide/metal_shared_queue.md
@@ -98,7 +98,7 @@ qd.init(arch=qd.metal, external_metal_command_queue=queue_ptr)
 
 ## Lifetime and ownership
 
-The caller (your application) owns the command queue. Quadrants retains it for the duration of the runtime and does **not** release it on `qd.reset()`. You must keep PyTorch (and its MPS backend) alive for as long as the Quadrants runtime is active.
+The caller (your application) owns the command queue. Quadrants borrows the pointer without retaining it, so the caller must keep the queue alive for the lifetime of the Quadrants runtime. In practice this means keeping PyTorch (and its MPS backend) alive for as long as `qd.init()` is active.
 
 ## Fallback
 
--- a/docs/source/user_guide/metal_shared_queue.md
+++ b/docs/source/user_guide/metal_shared_queue.md
@@ -94,7 +94,7 @@ qd.init(arch=qd.metal, external_metal_command_queue=queue_ptr)
 | `f.to_torch(copy=False)` | `qd.sync()` called internally | no sync needed |
 | `f.to_torch(copy=True)` | `qd.sync()` + `torch.mps.synchronize()` | no sync needed |
 | Quadrants kernel after torch write | manual `torch.mps.synchronize()` required | automatic (same queue) |
-| `f.to_numpy()` | `qd.sync()` (always needed for CPU readback) | `qd.sync()` (still needed) |
+| `f.to_numpy()` | `qd.sync()` (always needed for CPU readback) | `qd.sync()` (still needed — the kernel-copy path is Quadrants-internal, not routed through MPS) |
 
 ## Lifetime and ownership
 
--- a/docs/source/user_guide/metal_shared_queue.md
+++ b/docs/source/user_guide/metal_shared_queue.md
@@ -7,9 +7,12 @@ The `external_metal_command_queue` option lets you pass PyTorch's command queue 
 ## Quick start
 
 ```python
+import torch
 import quadrants as qd
+from quadrants.interop import get_mps_command_queue
 
-queue_ptr = get_mps_command_queue()   # see below
+torch.zeros(1, device="mps")  # ensure MPS is initialised
+queue_ptr = get_mps_command_queue()
 qd.init(
     arch=qd.metal,
     external_metal_command_queue=queue_ptr,
@@ -32,56 +35,15 @@ You can still call `qd.sync()` when you need to read results back to the CPU (e.
 
 ## Extracting PyTorch's MTLCommandQueue
 
-PyTorch does not expose its MPS command queue through a public Python API. The following helper extracts it at runtime using `ctypes` and the Objective-C runtime, with no build-time PyTorch dependency:
+PyTorch does not expose its MPS command queue through a public Python API. Quadrants provides a built-in helper that extracts it at runtime using `ctypes` and the Objective-C runtime:
 
 ```python
-import ctypes
-import os
-import torch
+from quadrants.interop import get_mps_command_queue
 
-
-def get_mps_command_queue() -> int:
-    """Return PyTorch MPS's MTLCommandQueue* as a Python int."""
-    # Ensure MPS is initialised
-    torch.zeros(1, device="mps")
-
-    torch_lib = os.path.join(
-        os.path.dirname(torch.__file__), "lib", "libtorch_cpu.dylib"
-    )
-    handle = ctypes.CDLL(torch_lib)._handle
-
-    libdl = ctypes.CDLL(None)
-    dlsym = libdl.dlsym
-    dlsym.restype = ctypes.c_void_p
-    dlsym.argtypes = [ctypes.c_void_p, ctypes.c_char_p]
-
-    # at::mps::getDefaultMPSStream() -> MPSStream*
-    func_addr = dlsym(handle, b"_ZN2at3mps19getDefaultMPSStreamEv")
-    assert func_addr, "Cannot find getDefaultMPSStream — check PyTorch version"
-    stream_ptr = ctypes.CFUNCTYPE(ctypes.c_void_p)(func_addr)()
-
-    # MPSStream::commandBuffer() -> id<MTLCommandBuffer>
-    cb_addr = dlsym(handle, b"_ZN2at3mps9MPSStream13commandBufferEv")
-    assert cb_addr, "Cannot find MPSStream::commandBuffer"
-    cb_ptr = ctypes.CFUNCTYPE(ctypes.c_void_p, ctypes.c_void_p)(cb_addr)(
-        stream_ptr
-    )
-
-    # [commandBuffer commandQueue] via ObjC runtime
-    objc = ctypes.CDLL("/usr/lib/libobjc.A.dylib")
-    sel_reg = objc.sel_registerName
-    sel_reg.restype = ctypes.c_void_p
-    sel_reg.argtypes = [ctypes.c_char_p]
-    msg_send = objc.objc_msgSend
-    msg_send.restype = ctypes.c_void_p
-    msg_send.argtypes = [ctypes.c_void_p, ctypes.c_void_p]
-
-    queue_ptr = msg_send(cb_ptr, sel_reg(b"commandQueue"))
-    assert queue_ptr, "Failed to extract MTLCommandQueue"
-    return queue_ptr
+queue_ptr = get_mps_command_queue()  # returns int (raw pointer), or 0 on failure
 ```
 
-The C++ symbol `_ZN2at3mps19getDefaultMPSStreamEv` has been stable since PyTorch 1.13.
+The function returns the `MTLCommandQueue*` as a Python `int`. It returns `0` if extraction fails (e.g. non-macOS platform, PyTorch not installed, or unsupported PyTorch build). The underlying C++ symbol (`_ZN2at3mps19getDefaultMPSStreamEv`) has been stable since PyTorch 1.13.
 
 ## Init ordering
 
@@ -92,6 +54,8 @@ import torch
 torch.zeros(1, device="mps")        # trigger MPS init
 
 import quadrants as qd
+from quadrants.interop import get_mps_command_queue
+
 queue_ptr = get_mps_command_queue()
 qd.init(
     arch=qd.metal,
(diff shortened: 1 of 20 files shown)


docs/source/user_guide/metal_shared_queue.md
```diff
@@ -0,0 +1,114 @@
...
+
+The `external_metal_command_queue` option lets you pass PyTorch's command queue to Quadrants so that both frameworks share a single queue. Metal processes command buffers in commit order within a queue, so GPU-side ordering is automatic and the per-interop sync overhead is eliminated.
+
+## Quick start
+
+```python
+import quadrants as qd
+
+queue_ptr = get_mps_command_queue()   # see below
+qd.init(arch=qd.metal, external_metal_command_queue=queue_ptr)
```
duburcqa: Would it be possible to rather add an option 'share_torch_command_queue'?

This is a bit invasive (yet very shallow, in python scope), but then it brings guarantees for `to_torch`. Otherwise, someone specifying anything by torch mps queue, would have synchronise wrongly skipped.

docs/source/user_guide/metal_shared_queue.md
```diff
@@ -0,0 +1,114 @@
...
+- `to_torch(copy=True)` no longer calls `torch.mps.synchronize()` after the copy.
+- GPU work submitted by Quadrants and by PyTorch executes in the order it was committed — no manual sync needed between the two.
+
+You can still call `qd.sync()` when you need to read results back to the CPU (e.g. `to_numpy()`); what changes is that you no longer need *both* `qd.sync()` and `torch.mps.synchronize()` at every framework boundary.
+
+## Extracting PyTorch's MTLCommandQueue
+
+PyTorch does not expose its MPS command queue through a public Python API. The following helper extracts it at runtime using `ctypes` and the Objective-C runtime, with no build-time PyTorch dependency:
+
+```python
```
duburcqa, line 38: Not sure about the value of this in user-facing doc.

docs/source/user_guide/metal_shared_queue.md
```diff
@@ -0,0 +1,114 @@
...
+
+The C++ symbol `_ZN2at3mps19getDefaultMPSStreamEv` has been stable since PyTorch 1.13.
+
+## Init ordering
+
+PyTorch MPS must be initialised **before** `qd.init()` so that the command queue exists when Quadrants starts:
+
+```python
+import torch
+torch.zeros(1, device="mps")        # trigger MPS init
```
duburcqa: I think instantiating the device is sufficient. As you want.

duburcqa: Could you move device instantiation (ie torch init) in `get_mps_command_queue` ? I don't think this should be the responsibility of the user.

docs/source/user_guide/metal_shared_queue.md
hughperkins: well, you (Genesis), the user, have to provide this bit of code.

docs/source/user_guide/metal_shared_queue.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+## What changes with a shared queue
+
+| Scenario | Separate queues (default) | Shared queue |
+|----------|--------------------------|--------------|
+| `f.to_torch(copy=False)` | `qd.sync()` called internally | no sync needed |
+| `f.to_torch(copy=True)` | `qd.sync()` + `torch.mps.synchronize()` | no sync needed |
+| Quadrants kernel after torch write | manual `torch.mps.synchronize()` required | automatic (same queue) |
+| `f.to_numpy()` | `qd.sync()` (always needed for CPU readback) | `qd.sync()` (still needed) |
```
duburcqa: Really? I'm quite sure torch is doing this automatically when converting to numpy, and it should work just great when sharing the same command queue.

docs/source/user_guide/metal_shared_queue.md
hughperkins: I'd rather keep it general on the whole. and have the user pass in the command queue, keeps everything clean, on quadrants side. Open to an extra boolean like something like 'external_metal_command_queue_is_torch_queue=True`, or some such.

docs/source/user_guide/metal_shared_queue.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Lifetime and ownership
+
+The caller (your application) owns the command queue. Quadrants retains it for the duration of the runtime and does **not** release it on `qd.reset()`. You must keep PyTorch (and its MPS backend) alive for as long as the Quadrants runtime is active.
```
duburcqa: > Quadrants retains it for the duration of the runtime and does **not** release it on `qd.reset()`

If this is the case, it should probably be exposed as a standalone method and not as init parameter, but I can understand this is annoying.

docs/source/user_guide/metal_shared_queue.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Fallback
+
+If extracting the queue fails (e.g. on an older PyTorch version or a non-Apple system), fall back to the default separate-queue path:
+
+```python
+try:
+    queue_ptr = get_mps_command_queue()
+except (AssertionError, OSError):
+    queue_ptr = 0   # 0 means "create a new queue" (the default)
```
duburcqa: I would expect Quadrants to handle this, returning 0 with a warning. But this design is also ok-ish.

docs/source/user_guide/metal_shared_queue.md
hughperkins: Not sure I follow. How are you imagining this would be used, in this case?

python/quadrants/lang/field.py
```diff
@@ -218,7 +218,7 @@ def _try_zerocopy_torch(field: "Field", *, copy, device=None, is_scalar: bool =
         tc = torch.utils.dlpack.from_dlpack(field.to_dlpack())
     except RuntimeError as e:
         raise ValueError(f"Zero-copy not available: {e}") from None
-    if impl.current_cfg().arch == _ARCH_METAL:
+    if impl.current_cfg().arch == _ARCH_METAL and not impl.current_cfg().external_metal_command_queue:
```
duburcqa: This is not very efficient. It should probably be cached?

docs/source/user_guide/metal_shared_queue.md
duburcqa: > 'external_metal_command_queue_is_torch_queue=True`, or some such.

yes. I think this is necessary in this case.

docs/source/user_guide/metal_shared_queue.md
duburcqa: > well, you (Genesis), the user, have to provide this bit of code.

What? Wow that's painful. But ok I guess

docs/source/user_guide/metal_shared_queue.md
duburcqa: `qd.set_mps_command_queue(queue: int, is_torch_queue: bool = True)`

docs/source/user_guide/metal_shared_queue.md
hughperkins: `get_mps_command_queue()` is part of Genesis, not part of torch. (It doesnt itself involve Quadrants).

docs/source/user_guide/metal_shared_queue.md
hughperkins: I feel like it will be cleanest if the queue is set for the lifetime of qd.init(). I'm not even sure if it would work if we used 'set' like this. What do you see as an advanage of using 'set', and why do you feel this relates to 'retain'?

duburcqa: > I feel like it will be cleanest if the queue is set for the lifetime of qd.init()

For sure!

duburcqa: > What do you see as an advanage of using 'set', and why do you feel this relates to 'retain'?

The contract is that 'reset' cancels 'init'. If it does not, then for me 'init' is not the correct API.

hughperkins: reset does cancel init. In waht way do you feel that not calling 'retain' does not align with this contract?

hughperkins: oh I see, the doc says we call retain.

We dont.

I'll get the doc updaed.

hughperkins: updated the doc

docs/source/user_guide/metal_shared_queue.md
duburcqa: I guess you are against making this first case in Quadrants? This is acceptable, even though it would be nice to hide this complexity out of Genesis to avoid having to care about the internal.

There could be some generic cross-gpu-backend `force_shared_command_queue`. Since the same also applies to CUDA and AMDGPU, it just happens that it works by default.

docs/source/user_guide/metal_shared_queue.md
hughperkins: Per Opus:

If to_numpy() went through PyTorch (DLPack → MPS tensor → .cpu().numpy()), then PyTorch's own GPU→CPU transfer would synchronize
  automatically on the shared queue, and qd.sync() would indeed be redundant.

  But the actual Field.to_numpy() code path doesn't go through torch — it uses Quadrants' own tensor_to_ext_arr kernel to copy GPU→CPU, then calls qd.sync() to wait for that
  kernel. Since that's a Quadrants-internal command buffer, the CPU still needs to wait for it regardless of queue sharing.

hughperkins: Thoughts?

hughperkins: from Opus:

  Here's the step-by-step walkthrough of f.to_numpy() on Metal:

  Line 626 — arr = np.zeros(shape=self.shape, dtype=dtype)
  • Allocates a numpy array on CPU memory. This is the destination.

  Line 629 — tensor_to_ext_arr(self, arr)
  • This is a Quadrants kernel. It reads from self (a field backed by a Metal GPU buffer) and writes to arr (a numpy array in CPU memory).
  • Under the hood, Quadrants compiles this to a Metal compute shader. The kernel is dispatched as a command buffer on Quadrants' Metal command queue.
  • On Apple Silicon, Metal uses unified memory — the GPU and CPU share the same physical RAM. The numpy array's backing memory is accessible to the GPU. So the Metal shader can
     write directly into the numpy buffer.
  • tensor_to_ext_arr() returns immediately after submitting the command buffer. It does not wait for the GPU to finish.

  Line 630 — quadrants.lang.runtime_ops.sync()
  • This calls MetalStream::command_sync(), which does [cmdbuf waitUntilCompleted] for all pending command buffers.
  • This is where the CPU blocks until the GPU has actually finished writing to arr.
  • Without this sync, return arr on the next line would hand back a numpy array whose contents are still being written to by the GPU — you'd read garbage or partial data.

  Why the shared queue doesn't help here: the sync isn't about coordinating between two frameworks (Quadrants vs PyTorch). It's about waiting for the GPU to finish before the CPU
  reads the result. Even with a shared queue, the CPU still needs to wait for the Metal command buffer to complete. There's no PyTorch involvement in this path at all — it's
  purely Quadrants kernel → CPU readback.

  Contrast with to_torch(copy=False) where the shared queue does help: there the consumer is another GPU command buffer (PyTorch's), so Metal's in-queue ordering suffices without
  any CPU-side wait.

duburcqa: > But the actual Field.to_numpy() code path doesn't go through torch — it uses Quadrants' own tensor_to_ext_arr kernel to copy GPU→CPU

If this is true, it is a mistake that should be addressed.

hughperkins: why? that would add a torch dependency.

duburcqa: Because it is vastly more efficient. Without that, we cannot migrate Genesis to Quadrants' own 'to_numpy'. it is simply unusable.

duburcqa: I'm fine with keeping 'qd_to_numpy' in Genesis if for you it is not acceptable.

hughperkins: alright, well, one less thing to think about for me, if we just leave it in genesis for now :)

(otehrwise, we could have the quadrants vesrion switch depending on whether torch is avialble, add warnings if its not, etc, but taht would need a bunch of thought 😅  )

python/quadrants/lang/field.py
hughperkins: done

hughperkins:
@codex review

hughperkins:
re: deleted comment: Per Opus:

"we don't need it anymore — the comment was a FIXME describing the problem, and the shared queue feature is the fix. The function now has a proper docstring explaining when
   sync is needed (separate queues) and when it's not (shared queue). The FIXME is resolved."

quadrants/rhi/metal/metal_device.mm
```diff
@@ -1224,9 +1239,14 @@ DeviceCapabilityConfig collect_metal_device_caps(MTLDevice_id mtl_device) {
...
 MetalDevice *MetalDevice::create() {
   MTLDevice_id mtl_device = MTLCreateSystemDefaultDevice();
-
   return new MetalDevice(mtl_device);
 }
+MetalDevice *
+MetalDevice::create_with_external_queue(uint64_t external_queue_ptr) {
+  MTLDevice_id mtl_device = MTLCreateSystemDefaultDevice();
+  auto *queue = reinterpret_cast<MTLCommandQueue_id>(external_queue_ptr);
+  return new MetalDevice(mtl_device, queue);
```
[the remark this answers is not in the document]
hughperkins: Per Opus:

  My thoughts: Technically correct but practically irrelevant. Apple Silicon Macs have exactly one GPU — there is no multi-GPU scenario on macOS. MTLCreateSystemDefaultDevice()
  always returns the same (only) device. PyTorch MPS also uses MTLCreateSystemDefaultDevice(), so the queue will always belong to the same device. That said, it would be
  slightly more defensive to extract the device from the queue via [queue device]. Low priority — fine to leave as-is or address as a follow-up.

docs/source/user_guide/metal_shared_queue.md
hughperkins: These are the current parameters:

Screenshot 2026-05-04 at 15 33 18

Thoughts?

duburcqa: I know that these are the current parameters. I just find it a bit disappointing to have to manage command queue sharing for all backends manually on Genesis side. But the current approach does work in practice.

duburcqa:
ok to merge

PR #620, opened 2026-05-04
[Perf] Cache adstack-sizer metadata per task across SPIR-V + LLVM-GPU; per-snode / DeviceAllocation invalidation (Part 2/2)

duburcqa:
# Cache adstack-sizer metadata per task across SPIR-V + LLVM-GPU; per-snode / DeviceAllocation invalidation

> **Part 2 of 2** in the adstack-perf series. Builds on the generic launcher infrastructure landed in **Part 1** (`duburcqa/perf_persistent_context_jit_pointer`, #619). This PR adds the per-task metadata cache that turns `publish_adstack_metadata`'s sizer-kernel + DtoH-stride round-trip into a host-only republish on cache hit. The launcher-side persistent scratch from Part 1 is what makes the republish DMA cheap enough to be worth caching.

## TL;DR

Every reverse-mode launch with adstacks dispatches a sizer kernel per task to compute `adstack_offsets` / `adstack_max_sizes` / strides from the kernel's compile-time `SerializedSizeExpr` trees plus the current values of `FieldLoad` / `ExternalTensorRead` / `ExternalTensorShape` leaves. The sizer's output is a deterministic function of `(per-task SerializedSizeExpr*, snode write generation per leaf snode, ndarray data generation per leaf ndarray)`. A long sequence of reverse-mode launches over the same kernel pays the sizer dispatch + 3x DtoH-stride round-trips per launch even though the inputs almost never change between launches.

This PR caches the sizer's output keyed by `(AdStackSizingInfo *, dependency-generation snapshot)`. On cache hit the cached `(offsets, max_sizes, stride_combined, stride_float, stride_int)` is republished into the runtime struct via the same DMA path the host-eval branch already uses; the bytecode-encode + h2d + sizer-kernel + 3x DtoH stalls are skipped. Each launcher (CPU / CUDA / AMDGPU / SPIR-V) bumps `snode_write_gen_` / `ndarray_data_gen_` for everything its kernel can mutate before publish, so the snapshot catches any input change between launches and forces a fresh sizer dispatch.

```cpp
// AdStackCache: lives on Program, owned via unique_ptr.
class AdStackCache {
 public:
  // Per-task entries on each backend's cache key.
  bool try_llvm_per_task_ad_stack_cache_hit(const void *task_key, LaunchContextBuilder *ctx,
                                            LlvmPerTaskAdStackCacheEntry &entry);
  void record_llvm_per_task_ad_stack_cache(const void *task_key, LlvmPerTaskAdStackCacheEntry entry);
  // Same shape for SPIR-V (kept separate so the bytecode + sizer-kernel paths can diverge).
  // ...
  // Generation bumps - called by each launcher on every kernel that may write to the source.
  void bump_snode_write_gen(int snode_id);
  void bump_ndarray_data_gen(const DeviceAllocation &devalloc);
};
```

The metadata cache machinery (struct definitions, methods, generation maps) was extracted from `Program` into a standalone `AdStackCache` class living in `quadrants/program/adstack_size_expr_eval.{h,cpp}`, with `Program` holding a `std::unique_ptr<AdStackCache>` and a single getter. Net effect: the surface area changed in `Program` is small (forward decl + getter + ctor init); all the cache-table ownership lives in adstack-specific files.

## Why

Two costs the sizer dispatch was paying every launch:

1. **Bytecode encode + h2d + sizer-kernel + 3x DtoH stride read-back.** On AMDGPU / CUDA this is 4-5 driver hops per task per launch, plus the CPU stalls waiting on the DtoH. Genesis test_differentiable_rigid runs ~30+ adstack-bearing tasks per kernel; over a 100-substep horizon that's thousands of synchronous round-trips per training step.
2. **CPU host-eval also pays a per-launch fixed cost** - even though there's no kernel dispatch on CPU, the SizeExpr walk repeatedly reads the same `FieldLoad` / `ExternalTensorRead` values across launches whose inputs haven't changed.

The dependency snapshot (per-leaf snode write generation, per-leaf ndarray data generation) catches the actual mutation cases: a kernel that writes to `field_a` will bump `snode_write_gen_[field_a.snode_id]` before publish, so any cache entry depending on `field_a` is forced to re-evaluate. The cache amortises the dispatch + DMA cost to "once per generation bump" rather than "every launch".

## Surface API

Zero user-visible change. Same `qd.kernel.grad(...)` invocation; same gradient values; same per-launch semantics.

Internal API surface (for the launchers):

```cpp
// In each backend's KernelLauncher::launch_llvm_kernel (or gfx::Runtime::launch_kernel):
bump_writes_for_kernel_llvm(prog, &ctx, offloaded_tasks);   // CUDA / AMDGPU
bump_writes_for_kernel_llvm(prog, &ctx, snode_writes, arr_writes);  // CPU (precomputed at register time)
bump_writes_for_kernel_spirv(prog, &host_ctx, task_attribs, arr_access);  // gfx
```

Each helper iterates the kernel's compile-time write set (snode writes from codegen + ndarray writes from arg-access metadata) and bumps the matching generation counters on `prog->adstack_cache()`. The actual cache lookup happens inside `publish_adstack_metadata` (LLVM-GPU + CPU host-eval) and inside the gfx adstack-sizer-launch helper (SPIR-V).

### 1. `AdStackCache` class extraction

`quadrants/program/adstack_size_expr_eval.{h,cpp}` - new class. Holds:

- `SizeExpr` host-eval cache: `(SerializedSizeExpr *) -> {result, observed_reads}`. On hit, replays observed `FieldLoad` / `ExternalTensorRead` / `ExternalTensorShape` reads against the live state and confirms a match before reusing the result.
- `LlvmPerTaskAdStackCacheEntry`: `{offsets[], max_sizes[], stride_combined, stride_float, stride_int, snode_gen_snapshot, ndarray_gen_snapshot}` keyed by `AdStackSizingInfo *`.
- `SpirvBytecodeCacheEntry` + `PerTaskAdStackCacheEntry` for the gfx backend (kept separate from the LLVM entry so bytecode encode and sizer-launch paths can diverge per backend).
- Generation maps: `snode_write_gen_` (`int snode_id -> uint64`), `ndarray_data_gen_` (`DeviceAllocation -> uint64`).

`Program` holds `std::unique_ptr<AdStackCache> adstack_cache_;` with `adstack_cache()` getter. The forward declaration + getter is the only `Program` surface change; cache method bodies and table ownership live in adstack-specific files.

### 2. `bump_writes_for_kernel_llvm` / `bump_writes_for_kernel_spirv`

`quadrants/program/adstack_size_expr_eval.{h,cpp}`. Called by each launcher just before kernel dispatch. Walks the kernel's compile-time write set:

- snode writes: from codegen, attached to each `OffloadedTask` (LLVM-GPU) or `TaskAttributes` (gfx).
- ndarray writes: from `KernelContextAttributes::arr_access` (gfx) or precomputed `arg_writes_per_task` (CPU launcher).

For each, calls `prog->adstack_cache().bump_snode_write_gen(snode_id)` / `bump_ndarray_data_gen(devalloc)`. After this, any cache entry whose snapshot diverges from the new generation is invalidated on next lookup.

### 3. Cache lookup in `publish_adstack_metadata`

`quadrants/runtime/llvm/llvm_adstack_lazy_claim.cpp`. The GPU branch checks `prog->adstack_cache().try_llvm_per_task_ad_stack_cache_hit(...)` first; on hit, the cached payload republishes via `publish_metadata_pinned_async` (the same pinned-host async pipeline the host-eval branch uses) and skips the entire bytecode-encode + h2d + sizer-kernel + DtoH-stride pipeline. On miss, the original sizer dispatch runs and the resulting payload is recorded back into the cache via `record_llvm_per_task_ad_stack_cache(...)` along with the snapshot.

CPU host-eval has its own cache shape (no bytecode, just SizeExpr walk results); same pattern.

### 4. SPIR-V (gfx) adstack-sizer-launch helper

`quadrants/runtime/gfx/adstack_sizer_launch.cpp` - new file holding the SPIR-V sizer-launch path that previously lived inline in `gfx/runtime.cpp`. Lookup mirrors LLVM-GPU: bytecode cache hit short-circuits the encode + dispatch; per-task metadata cache hit short-circuits the entire sizer-kernel + DtoH pipeline.

### 5. Per-backend launcher integration

| Backend | Where bump fires | Where cache fires |
|---------|------------------|-------------------|
| CPU     | `quadrants/runtime/cpu/kernel_launcher.cpp` (host-eval branch of `publish_adstack_metadata`) | host-eval branch in `llvm_adstack_lazy_claim.cpp` |
| CUDA / AMDGPU | `quadrants/runtime/{cuda,amdgpu}/kernel_launcher.cpp` (just before `launch_offloaded_tasks`) | GPU branch in `llvm_adstack_lazy_claim.cpp` |
| SPIR-V (Metal / Vulkan) | `quadrants/runtime/gfx/runtime.cpp` (just before cmdlist record) | `quadrants/runtime/gfx/adstack_sizer_launch.cpp` |

| Backend | SizeExpr host-eval cache | Per-task metadata cache | Bytecode cache | Generation bump |
|---------|--------------------------|--------------------------|----------------|-----------------|
| CPU     | ✓                        | (host-eval only)         | (n/a, no bytecode) | ✓ (`bump_writes_for_kernel_llvm`) |
| CUDA    | (used by host-eval if reached) | ✓                  | ✓ (LLVM-GPU bytecode) | ✓ |
| AMDGPU  | (used by host-eval if reached) | ✓                  | ✓ (LLVM-GPU bytecode) | ✓ |
| Metal   | (used by host-eval if reached) | ✓ (gfx variant)    | ✓ (SPIR-V bytecode) | ✓ (`bump_writes_for_kernel_spirv`) |
| Vulkan  | (used by host-eval if reached) | ✓ (gfx variant)    | ✓ (SPIR-V bytecode) | ✓ |

## Side-effect audit

| Concern                                                          | Verdict                                                                              |
|------------------------------------------------------------------|--------------------------------------------------------------------------------------|
| Snode mutation between launches forces fresh sizer dispatch      | Each launcher bumps `snode_write_gen_[snode_id]` before launch for every snode the kernel writes; cache snapshot diverges, lookup misses. |
| Ndarray data mutation between launches forces fresh sizer dispatch | Each launcher bumps `ndarray_data_gen_[devalloc]` before launch; same mechanism.       |
| `SizeExprReadObservation` replay catches host-eval staleness     | On hit the cache replays the recorded `FieldLoad` / `ExternalTensorRead` reads and only short-circuits if every value matches; mismatch erases the entry and falls through to a full eval. |
| `AdStackCache` lifetime tied to `Program`                        | Owned by `std::unique_ptr<AdStackCache>` member; created in `Program` ctor, destroyed with the program. Forward decl in `program.h` keeps the include surface narrow. |
| Cross-backend cache-entry types do not collide                   | Distinct entry types per backend (`LlvmPerTaskAdStackCacheEntry` vs `PerTaskAdStackCacheEntry` vs `SpirvBytecodeCacheEntry`); `Program` does not see any of them. |
| Cache lookup happens after every snode/ndarray bump in the same launch | Bump calls fire **before** `launch_offloaded_tasks`; lookup fires inside `publish_adstack_metadata` (called from inside the offloaded-task loop). Order is bump-then-lookup so the snapshot read for hit-checking sees the post-bump generation. |
| Persistent device scratch from Part 1 still safe                 | Per-handle isolation (Part 1) means recursive `snode_reader_*` launches via host-eval `FieldLoad` cannot clobber the parent kernel's `arg_buffer` / `RuntimeContext` even when the cache miss path runs the sizer kernel that itself dereferences `ctx->arg_buffer`. |


duburcqa:
@claude review

docs/source/user_guide/autodiff.md
```diff
@@ -347,13 +349,13 @@ A large `ndrange` combined with several loop-carried variables multiplies quickl
...
+- **Loop bounds backed by a mutated ndarray.** A reverse-mode kernel with `for i in range(n[j])` requires `n[j]` to hold the same value at the forward call and at `.grad()`. Writes between those two points that go through Quadrants APIs (`Ndarray.write` / `Ndarray.fill` / kernel writes) are tracked: the cached adstack capacity is recomputed automatically, the kernel runs against the new value, gradients are correct. Writes that bypass Quadrants's tracking (DLPack zero-copy, raw pointer writes through non-torch consumers) are NOT tracked. If such a write grows `n[j]` between the forward call and `.grad()`, the kernel overflows the cached adstack and raises the *Adstack overflow* error described above. The error names this cause as the most common possibility. The safe rule: populate loop-bound ndarrays through Quadrants APIs, or before the forward call and leave them untouched until `.grad()` returns. Tape-based eager AD like [PyTorch's autograd](https://pytorch.org/docs/stable/notes/autograd.html) is not affected, since the trip count is recorded as the forward runs and reused at backward time.
```
duburcqa: This error is about a different problem. Please do not broaden the scope of this specific bullet point and just apply the necessary modification for it to remain valid (if any change must be made). You are already discussing adstack overflow separately. No need to repeat yourself.

docs/source/user_guide/autodiff.md
```diff
@@ -289,6 +289,8 @@ The evaluation happens in different places depending on the backend, but the res
 
 Either way, the per-thread stride and each adstack's offset / max-size land in a small buffer the main kernel reads on every push. The backing heap grows on demand to match the largest size any launch has needed so far, and is reused across subsequent launches - you do not need to reserve memory up front.
 
+The sized result is cached per task across launches and reused as long as nothing the formula reads has changed. Quadrants tracks invalidation automatically for writes that flow through a Quadrants API - kernel-side writes, `Ndarray.write` / `Ndarray.fill`, and field writes. Writes that bypass Quadrants (DLPack zero-copy mutations through a torch tensor sharing storage with a Quadrants ndarray, or raw pointer writes through a non-torch DLPack consumer) are NOT tracked: the next launch reuses the cached size, and if the bypass mutation grew the required capacity, the kernel overflows the adstack and raises a clear error at the next Quadrants Python entry. See [What can go wrong](#what-can-go-wrong) for the error contract and recovery flow.
```
duburcqa: Too much details.

docs/source/user_guide/autodiff.md
```diff
@@ -347,13 +349,13 @@ A large `ndrange` combined with several loop-carried variables multiplies quickl
...
+  2. *Sizer under-estimated the bound (Quadrants bug).* On unusually intricate nested loops - typically deeply nested `for i in range(arr[...])` with cumulative-index arithmetic - the sizer can compute a bound that is mathematically tighter than the actual push count. Please file a bug with `QD_DUMP_IR=1` set so the kernel IR ships with the report. Workarounds, in order of convenience: shorten the innermost dynamic loop; precompute its worst-case trip into a scalar field the kernel only reads; split the inner section into its own `@qd.kernel`; pass `ad_stack_size=N` to `qd.init()` with `N` large enough to cover the real push count (bypasses the sizer).
```
duburcqa: Too much details.

quadrants/program/adstack_size_expr_eval.cpp
```diff
@@ -950,8 +1322,96 @@ std::vector<uint8_t> encode_adstack_size_expr_device_bytecode_for_spirv(
...
+  for (const auto &task : offloaded_tasks) {
+    for (int snode_id : task.snode_writes) {
+      prog->adstack_cache().bump_snode_write_gen(snode_id);
+    }
+    for (int arg_id : task.arr_writes) {
+      ArgArrayPtrKey data_key{arg_id, TypeFactory::DATA_PTR_POS_IN_NDARRAY};
+      auto it = ctx->array_ptrs.find(data_key);
+      if (it != ctx->array_ptrs.end() && it->second != nullptr) {
+        prog->adstack_cache().bump_ndarray_data_gen(it->second);
+      }
```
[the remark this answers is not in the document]
duburcqa: Fixed in 93bac706e: `Ndarray::write` and `Ndarray::fill` now call `bump_ndarray_data_gen` directly so host-side mutations route through the same gen-counter path as kernel writes.

quadrants/program/adstack_size_expr_eval.h
```diff
@@ -2,17 +2,139 @@
...
+    ++ndarray_data_gen_[devalloc_ptr];
+  }
+
+ private:
+  std::unordered_map<const SerializedSizeExpr *, SizeExprCacheEntry> size_expr_cache_;
+  std::unordered_map<const void *, SpirvBytecodeCacheEntry> spirv_bytecode_cache_;
+  std::unordered_map<const void *, PerTaskAdStackCacheEntry> per_task_ad_stack_cache_;
+  std::unordered_map<const void *, LlvmPerTaskAdStackCacheEntry> llvm_per_task_ad_stack_cache_;
+  std::unordered_map<int, uint64_t> snode_write_gen_;
+  std::unordered_map<void *, uint64_t> ndarray_data_gen_;
```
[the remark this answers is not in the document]
duburcqa: Fixed in a36cd3f6e: `Ndarray::~Ndarray` calls `erase_ndarray_data_gen` so the holder address is dropped from the generation map before it can be reused by a future allocation, which prevents stale per-task cache entries from aliasing the new ndarray.

quadrants/program/adstack_size_expr_eval.cpp
```diff
@@ -950,8 +1322,96 @@ std::vector<uint8_t> encode_adstack_size_expr_device_bytecode_for_spirv(
...
+  }
+  for (const auto &task : offloaded_tasks) {
+    for (int snode_id : task.snode_writes) {
+      prog->adstack_cache().bump_snode_write_gen(snode_id);
+    }
+    for (int arg_id : task.arr_writes) {
+      ArgArrayPtrKey data_key{arg_id, TypeFactory::DATA_PTR_POS_IN_NDARRAY};
+      auto it = ctx->array_ptrs.find(data_key);
+      if (it != ctx->array_ptrs.end() && it->second != nullptr) {
+        prog->adstack_cache().bump_ndarray_data_gen(it->second);
```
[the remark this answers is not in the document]
duburcqa: Fixed in f19032ff7: `OffloadedTask` now carries an `arr_reads` set, and both `bump_writes_for_kernel_llvm` overloads bump `ndarray_data_gen` for read-only `DevAllocType::kNone` args, mirroring the SPIR-V `kone_h2d_blit` rule across CPU / CUDA / AMDGPU.

quadrants/runtime/llvm/runtime_module/runtime.cpp
```diff
@@ -1487,7 +1514,7 @@ void runtime_initialize(Ptr result_buffer,
   runtime->adstack_row_counters_capacity = 0;
   runtime->adstack_bound_row_capacities = nullptr;
   runtime->adstack_bound_row_capacities_capacity = 0;
-  runtime->adstack_overflow_flag = 0;
+  runtime->adstack_overflow_flag_dev_ptr = nullptr;
```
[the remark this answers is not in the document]
duburcqa: Fixed in f19032ff7: `runtime_initialize` now zeros `adstack_overflow_task_id_dev_ptr` alongside the flag pointer, so the cmpxchg path on overflow lands on a clean `nullptr` no-op when the setter never ran.

docs/source/user_guide/autodiff.md
```diff
@@ -289,6 +289,8 @@ The evaluation happens in different places depending on the backend, but the res
 
 Either way, the per-thread stride and each adstack's offset / max-size land in a small buffer the main kernel reads on every push. The backing heap grows on demand to match the largest size any launch has needed so far, and is reused across subsequent launches - you do not need to reserve memory up front.
 
+The sized result is cached per task and reused while the formula's inputs are unchanged. Although changing loop bounds at runtime is possible, it comes with some limitations for performance reasons. See [What can go wrong](#what-can-go-wrong) for details.
```
duburcqa: "formula's inputs" --> "loop bounds"

docs/source/user_guide/debug.md
```diff
@@ -38,7 +38,7 @@ a = x[-1]     # AssertionError in debug mode
 
 #### Adstack overflow
 
-`debug=True` also enables a deferred bounds check on the adstack used by reverse-mode autodiff. A push past the per-stack capacity (set via `qd.init(ad_stack_size=...)` or per-alloca by `determine_ad_stack_size`) raises `RuntimeError("[Aa]dstack overflow")` on the next `qd.sync()`. Without the check, an adstack overflow silently writes past the per-thread slab and produces a wrong gradient.
+The adstack overflow check on reverse-mode autodiff runs always, on every backend, regardless of `debug`. A push past the per-stack capacity (set via `qd.init(ad_stack_size=...)` or per-alloca by `determine_ad_stack_size`) raises `QuadrantsAssertionError("[Aa]dstack overflow")` at the next Quadrants Python entry - kernel launch, host-side field / ndarray read, or explicit `qd.sync()`. The error message describes the two possible causes (untracked tensor mutation between launches, or sizer under-estimate) and the recovery flow; see [Autodiff -> What can go wrong](autodiff.md) for the full description.
```
hughperkins: Is it guaranteed to be the next? I feel it is more likely to be the next launch or so on that happens after the error has occurred, (and not the next launch after the launch that led to a kernel erroring)?

hughperkins: eg, we launhc:
- A
- B
- C
- at this point, A finishes running, and it had an error
- D
     - receives the exception/error

?

hughperkins: (I suppose it depends on our definition of 'next')

docs/source/user_guide/autodiff.md
```diff
@@ -347,13 +349,11 @@ A large `ndrange` combined with several loop-carried variables multiplies quickl
...
-  - pass `ad_stack_size=N` to `qd.init()` with `N` large enough to cover the real push count (bypasses the sizer).
+- **Adstack overflow.** Surfaces as `QuadrantsAssertionError: Adstack overflow ...` at the next Quadrants Python entry. The message names the offending kernel + offload task and the most likely cause:
+  - *Untracked tensor mutation between launches.* A tensor backing a data-dependent loop bound was written to outside Quadrants's tracking - typically a DLPack zero-copy mutation through a torch tensor sharing storage with a Quadrants ndarray, or a raw pointer write through a non-torch consumer. The cached adstack capacity was sized against the value before the mutation; if the mutation grew the bound, the next launch overflows. Fix: route the write through a Quadrants API (`Ndarray.write` / `Ndarray.fill` / a kernel that writes the value). Alternatively, catch the exception and re-launch - Quadrants invalidates the cached bound on raise so the retry runs against the live state. Kernel state may be inconsistent after an overflow; do not retry the same step without addressing the cause and restarting from a clean state.
```
hughperkins: nice explanation I feel

docs/source/user_guide/autodiff.md
```diff
@@ -347,13 +349,11 @@ A large `ndrange` combined with several loop-carried variables multiplies quickl
...
+  - *Sizer under-estimated the bound (Quadrants bug).* On unusually intricate nested loops - typically deeply nested `for i in range(arr[...])` with cumulative-index arithmetic - the sizer can compute a bound that is mathematically tighter than the actual push count. File a bug with `QD_DUMP_IR=1` set so the kernel IR ships with the report. Workaround: pass a generous `ad_stack_size=N` to `qd.init()` with `N` large enough to cover the real push count (bypasses the sizer).
```
hughperkins: "File a bug with `QD_DUMP_IR=1` set". Whilst I can undersatnd what you mean, this seems to be a confusing way to state this? Maybe omsething like:

"File a bug report as follows, as an issue against the Quadrants repo:
- `rm /tmp/ir/*`
- run your job with the env var QD_DUMP_IR=1 defined
- zip the contents of /tmp/ir
- create an issue against the Quadrants repo
- attach the zip of /tmp/ir to the issue"

?

docs/source/user_guide/autodiff.md
```diff
@@ -347,13 +349,11 @@ A large `ndrange` combined with several loop-carried variables multiplies quickl
...
+- **Loop bounds backed by a mutated ndarray.** A reverse-mode kernel with `for i in range(n[j])` requires `n[j]` to hold the same value at the forward call and at `.grad()`. If anything writes to `n[j]` between those two points - the differentiable kernel itself, or any other kernel call - the computed gradient may come out silently wrong. The safe rule: populate loop-bound ndarrays before the forward call and leave them untouched until `.grad()` returns. The reason for that is Quadrants' adstack sizer design: it reads the loop bound separately at each dispatch, which includes forward and backward calls. Tape-based eager AD like [PyTorch's autograd](https://pytorch.org/docs/stable/notes/autograd.html) is not affected, since the trip count is recorded as the forward runs and reused at backward time.
```
hughperkins: 👀

hughperkins:
- left some comments on the doc
- waiting also for the agent-based checks to run
- this touches a ton of different files, so lets have a full Genesis unit test run, and benchmarks, pelase

docs/source/user_guide/autodiff.md
duburcqa: updated

docs/source/user_guide/autodiff.md
duburcqa: updated

docs/source/user_guide/debug.md
duburcqa: Only on CPU. Updated to be more accurate.

docs/source/user_guide/autodiff.md
```diff
@@ -347,13 +349,11 @@ A large `ndrange` combined with several loop-carried variables multiplies quickl
...
+  - *Sizer under-estimated the bound (Quadrants bug).* On unusually intricate nested loops - typically deeply nested `for i in range(arr[...])` with cumulative-index arithmetic - the sizer can compute a bound that is mathematically tighter than the actual push count. To file a bug: clear `/tmp/ir/`, rerun your script with `QD_DUMP_IR=1` set in the environment so Quadrants dumps the kernel IR there, then open an issue on the Quadrants repo with the contents of `/tmp/ir/` attached as a zip. Workaround: pass a generous `ad_stack_size=N` to `qd.init()` with `N` large enough to cover the real push count (bypasses the sizer).
```
hughperkins, line 354: Description of how to file the issue seems clear, complete and concise. Great! 🙌

hughperkins:
doc looks good to me 🙌  

I'm waiting now on:
- agent-based checks to run
- results of full Genesis unit test run (at least on field)
- results of Genesis benchmarks run (at least on field)

duburcqa:
|env                     |batch_size|backend|gjk_collision|constraint_solver|runtime_fps_main|runtime_fps_620|runtime_fps_delta_pct|
|anymal_random           |30000     |cuda   |-            |-                |9443191         |9390292        |-0.56                |
|anymal_uniform          |30000     |cuda   |-            |-                |12346008        |12387632       |+0.34                |
|anymal_uniform_kinematic|0         |cpu    |-            |-                |1982            |1948           |-1.72                |
|anymal_uniform_kinematic|30000     |cuda   |-            |-                |10631243        |10352282       |-2.62                |
|anymal_zero             |0         |cpu    |-            |-                |7267            |7139           |-1.76                |
|anymal_zero             |30000     |cuda   |-            |-                |19146598        |19301149       |+0.81                |
|box_pyramid_3           |4096      |cuda   |-            |-                |979160          |988222         |+0.93                |
|box_pyramid_4           |4096      |cuda   |-            |-                |384826          |387764         |+0.76                |
|box_pyramid_5           |4096      |cuda   |-            |-                |141026          |137764         |-2.31                |
|box_pyramid_6           |4096      |cuda   |False        |-                |58596           |58888          |+0.50                |
|box_pyramid_6           |4096      |cuda   |True         |-                |60147           |59231          |-1.52                |
|dex_hand                |4096      |cuda   |-            |-                |17240           |17173          |-0.39                |
|duck_in_box_easy        |30000     |cuda   |False        |-                |27101190        |26767881       |-1.23                |
|duck_in_box_easy        |30000     |cuda   |True         |-                |9526647         |9501811        |-0.26                |
|duck_in_box_hard        |0         |cpu    |-            |-                |5079            |5121           |+0.83                |
|duck_in_box_hard        |30000     |cuda   |False        |-                |10245189        |10305593       |+0.59                |
|duck_in_box_hard        |30000     |cuda   |True         |-                |3386673         |3416781        |+0.89                |
|franka                  |30000     |cuda   |-            |-                |22146368        |21472201       |-3.04                |
|franka_accessors        |0         |cpu    |-            |-                |1136            |1193           |+5.02                |
|franka_accessors        |30000     |cuda   |-            |-                |15354421        |15706391       |+2.29                |
|franka_free             |30000     |cuda   |-            |-                |32056108        |32810552       |+2.35                |
|franka_random           |0         |cpu    |-            |-                |6249            |6271           |+0.35                |
|franka_random           |30000     |cuda   |-            |CG               |16548461        |16839003       |+1.76                |
|franka_random           |30000     |cuda   |-            |Newton           |16720811        |16688250       |-0.19                |
|franka_random           |30000     |cuda   |False        |-                |16509323        |16401148       |-0.66                |
|franka_random           |30000     |cuda   |True         |-                |11570179        |11542843       |-0.24                |
|g1_fall                 |4096      |cuda   |-            |Newton           |907003          |922819         |+1.74                |
|go2                     |4096      |cuda   |False        |CG               |3646422         |3725625        |+2.17                |
|go2                     |4096      |cuda   |False        |Newton           |4624845         |4583004        |-0.90                |
|go2                     |4096      |cuda   |True         |-                |3330016         |3318628        |-0.34                |
|shadow_hand_cubes_sparse|0         |cpu    |-            |-                |65              |66             |+1.54                |

[speed_test_620.txt](https://github.com/user-attachments/files/27408609/speed_test_620.txt)
[speed_test_main.txt](https://github.com/user-attachments/files/27408610/speed_test_main.txt)

duburcqa:
All the unit tests are passing against https://github.com/Genesis-Embodied-AI/Genesis/pull/2742.

However, there is a bug in load_store elimination for static for-loops in `eliminate_recomputable_ad_stack_pushes` which breaks the unit test `test_differentiable_push` on genesis/main. It is unrelated to this PR. It was introduced by https://github.com/Genesis-Embodied-AI/quadrants/pull/621. I was not able to spot it because I ran the unit test against 2742 instead of main at that time... I suggest to fix this bug in a follow-up since it is unrelated?

duburcqa:
After applying https://github.com/Genesis-Embodied-AI/quadrants/pull/633, all the unit tests are passing on Genesis main 
`======= 642 passed, 3 skipped, 2 xfailed in 1086.07s (0:18:06) =======`

hughperkins:
checklist:
- docs look good to me
- agent based checks passed
- genesis unit tests passing
- genesis benchmarks ~neutral

=> ok to merge

PR #622, opened 2026-05-04
[CI] Encode Python-first testing policy in coverage-check prompt

hughperkins:
The coverage-check agent had no awareness of the team's preference for Python end-to-end tests over C++ unit tests, so it could flag missing C++ tests on changes that are already adequately covered from Python. Add a "Repo testing policy" section that:

- Defaults to Python tests under tests/python/.
- Lists narrow exceptions where a C++ test is still warranted: hardware capability gating / negative paths needing mocked device state, binary/IR-level assertions on generated artifacts, internal IR or codegen helpers with no Python entry point, and parity with an existing C++ test for an analogous function.
- Explicitly tells the agent not to flag missing C++ tests for code that is already covered by Python.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


duburcqa:
ok to merge

PR #623, opened 2026-05-04
[CI] Fix coverage report showing def lines as uncovered

hughperkins:
run_tests.py imported quadrants before pytest-cov started measuring, so all module-level code (def lines, class definitions, imports) ran before coverage began and was recorded as hits=0. Replace the early import with importlib.util.find_spec which locates the package path without executing any quadrants code.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

PR #629, opened 2026-05-04
[CI] Disable quadrants pytest plugin during quadrants internal coverage runs

hughperkins:
## Summary

- The quadrants package registers as a pytest plugin (`pytest11` entry point). When pytest loads this plugin, it imports the `quadrants` parent package, executing all module-level code **before** `pytest-cov` starts measuring coverage. This is the real root cause of `def` lines, imports, and module-level statements showing as uncovered in coverage reports.
- Disable the plugin with `-p no:quadrants` when collecting coverage in `run_tests.py`, and set `_QD_KCOV_ARC` directly (which the plugin would otherwise set).
- Follows up on #623 which addressed the `import quadrants` in `run_tests.py` itself but didn't fix the issue because the pytest plugin was the actual culprit.

## Diagnosis

The import ordering was verified locally:

| Method | Module-level lines covered | First HIT line |
|--------|---------------------------|----------------|
| `pytest --cov` (before fix) | 43 / 391 | line 321 |
| `pytest --cov -p no:quadrants` | 154 / 391 | line 4 |
| `coverage run -m pytest` | 154 / 391 | line 4 |

## Test plan

- [ ] CI coverage report should show `def` lines as covered (green) when their bodies are exercised
- [ ] Kernel coverage (`_QD_KCOV_ARC`) still works correctly

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -43,6 +43,11 @@ def _test_python(args, default_dir="python"):
             _spec = importlib.util.find_spec("quadrants")
             assert _spec is not None and _spec.origin is not None, "quadrants package not found"
             _cov_src = os.path.dirname(_spec.origin)
+            # Disable the quadrants pytest plugin (pytest11 entry point) during coverage runs.  Loading the plugin
+            # forces Python to import the quadrants parent package before pytest-cov starts measuring, so module-level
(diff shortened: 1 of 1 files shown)


hughperkins:
@codex review

PR #638, opened 2026-05-06
[Docs] Add user-guide page for qd.simt.block.* primitives

hughperkins:
## Summary

Adds a comprehensive user-guide page for `qd.simt.block.*` operations as `docs/source/user_guide/block.md`, plus a section on the closely related `qd.simt.grid.memfence()` so users picking between a block-scope and a device-scope fence can see both side by side.

Coverage:

- `block.sync()` / `block.mem_sync()` — barrier-vs-fence semantics, including the *Barrier vs fence: the distinction that matters* section.
- `block.sync_all_nonzero` / `_any_nonzero` / `_count_nonzero` — CUDA-only barrier-and-reduce predicate ops.
- `block.SharedArray(shape, dtype)` — shared-memory allocation.
- `block.global_thread_idx()` / `block.thread_idx()` — backend-specific thread indexing.
- `qd.simt.grid.memfence()` — device-scope memory fence.

Each op gets a row in the *What's available* support matrix, a *Semantics* section, and at least one worked *Examples* entry.

## Notes

- The CUDA `mem_sync()` caveat references the in-flight fix in #637, so the divergent-branch example is honest about current behavior on `main`. Once #637 lands, the caveat in the support-table footnote can be deleted.
- Also updates `docs/source/user_guide/index.md` to add `block.md` to the *SIMT primitives* toctree, and `docs/source/user_guide/parallelization.md` to fix the stale `qd.stmt.SharedArray` reference and cross-reference the new page.
- A separate PR adds the analogous expanded `subgroup.md`.

## Test plan

- [ ] `python python/tools/markdown_link_check.py docs` passes (verified locally).
- [ ] Sphinx HTML build (`make html` in `docs/`) succeeds in CI.

### DIFF
--- a/docs/source/user_guide/block.md
+++ b/docs/source/user_guide/block.md
@@ -24,12 +24,14 @@ Calling a backend marked "no" raises `ValueError` from the Python layer at trace
 
 \* On CUDA, `block.mem_sync()` currently lowers via `block_barrier` (i.e. `__syncthreads()`), which doubles as a memory fence but additionally requires thread convergence — meaning calling it from divergent control flow today deadlocks. A fix to lower `mem_sync()` to a pure `__threadfence_block()` is in flight as [quadrants#637](https://github.com/Genesis-Embodied-AI/quadrants/pull/637); once merged, the divergent-branch pattern shown in the `block.mem_sync()` semantics section below works as written. Until then, prefer calling `mem_sync()` from uniform control flow on CUDA.
 
+Naming note: `block.mem_sync()` is planned to be renamed to `block.mem_fence()` in a future release, to align with the project's "fence vs barrier" terminology. The new name is not yet available; this page uses the current name throughout.
+
 ## Barrier vs fence: the distinction that matters
 
 Two of these ops sound similar but have very different semantics, and mixing them up deadlocks the GPU. The summary:
 
 - `block.sync()` is a **thread-converging barrier**. Every thread in the block must reach the call site before any thread proceeds. It also implies a memory fence at block scope.
-- `block.mem_sync()` is a **memory fence only**, at block scope. It orders memory operations but does not require thread convergence — it is safe to call from divergent control flow (e.g. inside `if tid == 0`).
+- `block.mem_sync()` (to be renamed `block.mem_fence()` in a future release) is a **memory fence only**, at block scope. It orders memory operations but does not require thread convergence — it is safe to call from divergent control flow (e.g. inside `if tid == 0`).
 
 Concretely, on the SPIR-V backend `sync()` lowers to `workgroupBarrier` and `mem_sync()` lowers to `workgroupMemoryBarrier`. On CUDA, `sync()` lowers to `__syncthreads()`; `mem_sync()` is intended to lower to `__threadfence_block()` (a pure fence with no convergence requirement) — see the support-table caveat above. Calling `sync()` from a path that not all threads reach (a divergent `if`, an early `return`, etc.) is a classic GPU deadlock and applies to both backends.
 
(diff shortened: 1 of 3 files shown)


docs/source/user_guide/block.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Grid-scope fence: `qd.simt.grid.memfence()`
+
+`grid.memfence()` is the device-scope counterpart of `block.mem_sync()`. It orders memory operations across the entire grid, so writes made by one block become visible to other blocks after the fence. CUDA only today; lowers to `__threadfence()` (`nvvm_membar_gl`).
+
+Use it when you need cross-block coordination via global memory (decoupled look-back scan, inter-block flag publishing, single-pass reductions, etc.). For coordination within a single block, prefer `block.mem_sync()` — it is cheaper.
+
+There is no built-in grid-wide barrier (cooperative-groups-style); the only way to converge threads across blocks is to finish the kernel and launch a new one.
+
+## Examples
```
hughperkins: lets remove examples, so I dont have to review them...

docs/source/user_guide/block.md
```diff
@@ -0,0 +1,194 @@
...
+        tid = i % 32
+        if tid == 0:
+            buf[i // 32] = i // 32
+            qd.simt.grid.memfence()
+            qd.atomic_add(flag, 0, 1)
+```
+
+The `memfence()` here ensures that the data write to `buf` becomes visible to other blocks before the atomic increment of `flag` does, so any block reading `flag > N` is guaranteed to see at least `N` published `buf` entries. This is the building block of decoupled-look-back scan and other single-pass multi-block algorithms.
+
+## Performance and portability notes
```
hughperkins: lets remove this section. less stuff to review/maintain.

alanray-tech (commented):
## Review Summary

Well-written documentation page that accurately reflects the current source code, including the honest CUDA `mem_sync()` caveat referencing #637. The "Barrier vs fence" section is especially valuable. The `parallelization.md` fix for the stale `qd.stmt.SharedArray` reference is a welcome cleanup.

While reviewing the docs against the source, I found three pre-existing source-code issues (not introduced by this PR). I will file them as separate issues so they can be tracked independently:

1. **`global_thread_idx()` has an `or`-precedence bug** in `block.py:61` -- `if arch == _qd_core.cuda or _qd_core.amdgpu:` always evaluates to true because `_qd_core.amdgpu` is a truthy enum value, making the SPIR-V branch on line 63-64 unreachable. The doc describes this as "a control-flow quirk in the dispatch", which is accurate in effect but understates the root cause.

2. **`grid.memfence()` has no Python-level arch guard** -- unlike every other op in `block.py`, `grid.py:memfence()` calls the internal directly without checking the arch. The doc says "CUDA only" but a user calling it on a non-CUDA backend would get a less-than-helpful error from the C++ layer rather than a clean `ValueError`.

3. **`sync()` error message says "shared_array"** -- `block.py:19` raises `ValueError(f"qd.block.shared_array is not supported for arch {arch}")` inside the `sync()` function; the message should say `sync`.

Documentation-only nit: the anchor link in `parallelization.md` (`block.md#grid-scope-fence-qdsimtgridmemfence`) should be verified against the actual Sphinx-generated anchor to avoid a dead link.

Overall: LGTM as a docs PR. The content is accurate, well-structured, and covers every public op in `block.py` / `grid.py`.

alanray-tech (approved):

alanray-tech:
OK to merge

alanray-tech (commented):
**Simplified review** (supersedes my earlier longer comment):

LGTM as a docs PR. Content is accurate and well-structured.

Three pre-existing source bugs found during review — filed separately:
- #644 — `global_thread_idx()`: `or` precedence bug makes SPIR-V branch unreachable
- #645 — `grid.memfence()`: no Python-level arch guard (silently compiles as no-op on AMDGPU)
- #646 — `sync()`: error message incorrectly says `shared_array`

PR #639, opened 2026-05-06
[Docs] Expand qd.simt.subgroup user-guide page to cover every op

hughperkins:
## Summary

The existing `docs/source/user_guide/subgroup.md` only covered `shuffle` / `shuffle_down` / `reduce_add` / `reduce_all_add`. This PR expands it to document every name currently exposed in `python/quadrants/lang/simt/subgroup.py`, with a per-op backend support matrix so readers can tell which ops are usable on which target without grepping the codegens.

New *What's available* tables (grouped by category):

- **Data movement**: `shuffle`, `shuffle_down`, `shuffle_up`, `shuffle_xor` (TODO), `broadcast`, `broadcast_first` (TODO).
- **Identification and control**: `invocation_id`, `group_size`, `elect`, `barrier`, `memory_barrier`.
- **Voting and predicate ops**: `all_true`, `any_true`, `all_equal` (all TODO stubs today; CUDA-only fallbacks live under `qd.simt.warp.*`).
- **Reductions and scans**: `reduce_add`, `reduce_all_add` (portable), `inclusive_{add,mul,min,max,and,or,xor}` (SPV-only), `exclusive_*` (TODO).

Adds a *Semantics* subsection for every implemented op (`shuffle_up`, `broadcast`, `invocation_id`, `group_size`, `elect`, `barrier`, `memory_barrier`, `inclusive_*`) and a single rolled-up subsection covering the `# TODO` stubs so users know the names exist but cannot be relied on yet.

Adds one worked *Examples* entry for the SPV-only `inclusive_add` scan and a corresponding performance-notes bullet. Existing shuffle / reduce examples and notes are preserved verbatim.

## Notes

- Independent of #638 (block.md docs PR); does not link to that file. A follow-up can add a *Related* cross-link in either direction once both land.

## Test plan

- [ ] `python python/tools/markdown_link_check.py docs/source/user_guide/subgroup.md` passes (verified locally).
- [ ] Sphinx HTML build (`make html` in `docs/`) succeeds in CI.

### DIFF
--- a/docs/source/user_guide/subgroup.md
+++ b/docs/source/user_guide/subgroup.md
@@ -21,7 +21,7 @@ The full Python API is grouped here by category. The first column lists each op,
 
 \* AMDGPU `shuffle_down` (and therefore `reduce_add`, which is built on it) is currently emulated via `ds_bpermute` (~50 cycle latency).
 
-Use `shuffle` with an explicit lane index in the meantime — every shuffle pattern can be expressed that way.
+Use `shuffle` with an explicit lane index — every shuffle pattern can be expressed that way.
 
 ### Identification and control
 
--- a/docs/source/user_guide/subgroup.md
+++ b/docs/source/user_guide/subgroup.md
@@ -98,10 +98,11 @@ Lane `i` returns the `value` held by lane `i - offset`. Lanes near the bottom of
 
 ### `broadcast(value, index)`
 
-Every lane in the subgroup returns the `value` held by the lane whose subgroup-local id equals `index`. Equivalent to `shuffle(value, index)` but expresses intent ("read lane `index`") more directly and may map to a cheaper instruction where the backend has a dedicated broadcast.
+Every lane in the subgroup returns the `value` held by the lane whose subgroup-local id equals `index`. Expresses intent ("read lane `index`") more directly than `shuffle(value, index)` and on backends with a dedicated broadcast may map to a cheaper instruction.
 
 - Same dtype rules as `shuffle`.
 - Maps to `__shfl_sync` on CUDA, `ds_bpermute` on AMDGPU, and `OpGroupNonUniformBroadcast` on SPIR-V.
+- **Important: on SPIR-V, `index` must be dynamically uniform** — the same value on every lane in the subgroup. Passing a per-lane varying `index` is undefined behavior, because `OpGroupNonUniformBroadcast` requires its `Id` operand to be dynamically uniform across the subgroup. On CUDA / AMDGPU, `index` may vary per lane and the call is identical to `shuffle(value, index)`. If you need a varying source lane, use `shuffle` directly.
 
 ### Common to the data-movement ops
 
--- a/docs/source/user_guide/subgroup.md
+++ b/docs/source/user_guide/subgroup.md
@@ -33,6 +33,13 @@ The remaining shuffle flavours (`shuffle_up`, `shuffle_xor`) are exposed in the 
 | `subgroup.barrier()`                        | no   | no     | yes                     |
 | `subgroup.memory_barrier()`                 | no   | no     | yes                     |
 
+Naming note: two of the names above are planned to be renamed in a future release, to align with the project's naming conventions across scopes:
+
+- `subgroup.barrier()` will be renamed to `subgroup.sync()` (matching `block.sync()`).
+- `subgroup.memory_barrier()` will be renamed to `subgroup.mem_fence()` (matching the planned `block.mem_fence()` and `grid.mem_fence()`).
+
+The new names are not yet available; this page uses the current names throughout.
+
 ### Voting and predicate ops
 
 Although these functions exist, we will migrate them to have an additional `log2_size` parameter, similar to `reduce_add` — so that the vote is over the first `2**log2_size` lanes rather than the full subgroup.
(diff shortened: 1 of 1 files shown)


docs/source/user_guide/subgroup.md
```diff
@@ -1,23 +1,62 @@
...
+| `subgroup.shuffle(value, index)`            | yes  | yes    | yes                     | i32, u32, f32, f64, i64, u64 |
+| `subgroup.shuffle_down(value, offset)`      | yes  | yes\*  | yes                     | i32, u32, f32, f64, i64, u64 |
+| `subgroup.shuffle_up(value, offset)`        | no   | no     | yes                     | i32, u32, f32, f64, i64, u64 |
+| `subgroup.shuffle_xor(value, mask)`         | no   | no     | no                      | — (TODO stub on every backend) |
+| `subgroup.broadcast(value, index)`          | yes  | yes    | yes                     | i32, u32, f32, f64, i64, u64 |
+| `subgroup.broadcast_first(value)`           | no   | no     | no                      | — (TODO stub on every backend) |
 
 \* AMDGPU `shuffle_down` (and therefore `reduce_add`, which is built on it) is currently emulated via `ds_bpermute` (~50 cycle latency).
 
-The remaining shuffle flavours (`shuffle_up`, `shuffle_xor`) are exposed in the Python module but are not yet implemented across backends. Calling them will fail at codegen. Use `shuffle` with an explicit lane index in the meantime — every shuffle pattern can be expressed that way.
```
hughperkins, line 18: why'd we delete this?

alanray-tech (commented):
## Review Summary

Thorough expansion of `subgroup.md` that now documents every name in `subgroup.py`, with per-op backend support matrices, semantics sections, and clear signposting for TODO stubs. The categorization into Data movement / Identification / Voting / Reductions is well-chosen and makes the page scannable. Overall quality is high.

I verified the support-matrix claims against the C++ codegens (`codegen_cuda.cpp`, `codegen_amdgpu.cpp`, `spirv_codegen.cpp`). The claims that `broadcast` and `invocation_id` work on CUDA/AMDGPU are confirmed -- both are handled alongside `subgroupShuffle` in the CUDA and AMDGPU codegen visitors.

### Findings

**Pre-existing source bug (not introduced by this PR):**

`broadcast`, `group_size`, and `invocation_id` are all defined as public functions in `subgroup.py` but are **missing from `__all__`**. This means `from quadrants.lang.simt.subgroup import *` will not import them. They are still accessible via `qd.simt.subgroup.broadcast()`, so this is low severity, but it is inconsistent -- `broadcast_first` (a TODO stub) is in `__all__`, but the working `broadcast` is not.

**Documentation nit -- planned rename references non-existent names:**

The "Naming note" says:

> `subgroup.memory_barrier()` will be renamed to `subgroup.mem_fence()` (matching the planned `block.mem_fence()` and `grid.mem_fence()`)

But the current names are `block.mem_sync()` and `grid.memfence()` -- neither `block.mem_fence()` nor `grid.mem_fence()` exist today or in any open PR. This sentence could confuse readers into thinking those names are already available. Suggest either:
- Dropping the parenthetical, or
- Saying "matching a planned cross-scope `mem_fence` / `mem_sync` naming convention that is not yet finalized".

**Minor suggestions:**

1. The `inclusive_*` table rows appear to have slightly inconsistent leading whitespace compared to other rows in the same table (extra space before the first `|`). Purely cosmetic but worth fixing for clean rendering.

2. The "Reductions and scans" intro paragraph is a bit dense -- it tries to cover `log2_size` semantics, the plan to add `log2_size` to `inclusive_*`/`exclusive_*`, and the full-subgroup behavior all in one block. Consider splitting into a shorter intro + a note box for the planned API change.

3. Missing from the doc: same as `block.py` ops, none of the subgroup ops have Python-level arch guards. Functions like `barrier()`, `elect()`, `group_size()` are documented as "SPIR-V only" but will fail at codegen/link time rather than raising a clean `ValueError` at trace time. This is consistent with the pattern reported for `block.py` in #644/#645/#646, so no need for a separate issue, but worth noting for completeness.

### Checklist

- [x] Documentation accurately reflects the source code
- [x] Support matrix verified against C++ codegens
- [x] TODO stubs correctly documented as unusable
- [x] Existing shuffle/reduce examples preserved verbatim
- [x] New inclusive_add example is correct
- [ ] `broadcast`, `group_size`, `invocation_id` missing from `__all__` (pre-existing)
- [ ] Planned rename parenthetical references non-existent names

Overall: LGTM as a docs PR. One factual inaccuracy in the rename note should be corrected.

alanray-tech (commented):
Follow-up on the `broadcast` semantics section:

The doc says:

> Equivalent to `shuffle(value, index)` but expresses intent ("read lane `index`") more directly and may map to a cheaper instruction where the backend has a dedicated broadcast.

This omits a key correctness constraint: on SPIR-V, `broadcast` lowers to `OpGroupNonUniformBroadcast`, which requires `index` to be **dynamically uniform** (i.e. the same value across all invocations in the subgroup). Passing a non-uniform `index` is undefined behavior per the SPIR-V spec.

On CUDA/AMDGPU it doesn't matter -- both `shuffle` and `broadcast` go through the same `__shfl_sync` / `ds_bpermute` path. But on SPIR-V, the distinction is a correctness requirement, not just a hint.

Suggest adding something like:

> **Important:** on SPIR-V, `index` must be dynamically uniform (the same value on every lane in the subgroup). Passing a per-lane varying `index` is undefined behavior. On CUDA / AMDGPU, `index` may vary per lane (the call is identical to `shuffle`).

alanray-tech (commented):
Adding SPIR-V spec references for the `broadcast` dynamically-uniform requirement raised in my previous comment:

Per the [SPIR-V Specification (Khronos)](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html), the two instructions have different `index` constraints:

**`OpGroupNonUniformBroadcast`** (used by `subgroup.broadcast`):
> Before version 1.5, Invocation Id must come from a **constant instruction**. Starting with version 1.5, this restriction is lifted. However, **behavior is undefined when Invocation Id is not dynamically uniform**.

**`OpGroupNonUniformShuffle`** (used by `subgroup.shuffle`):
> Result is the Value of the invocation identified by the id Invocation Id.

No uniform or constant constraint on the Id -- it may vary per lane.

The spec changelog also records:
> Khronos Issue #402: Relax OpGroupNonUniformBroadcast Id from **constant to dynamically uniform**, starting with version 1.5.

This confirms that on SPIR-V, passing a non-uniform `index` to `broadcast` is UB, while `shuffle` has no such restriction. The doc should note this distinction, since it is a correctness requirement that users cannot infer from the current wording ("expresses intent ... more directly").

MuGdxy (approved):
OK to merge

PR #640, opened 2026-05-07
[Docs] Add user-guide page for atomics and bit operations

hughperkins:
## Summary

New user-guide page `docs/source/user_guide/atomics.md` covering the per-thread / scalar tier of Quadrants primitives — atomic read-modify-write ops (`qd.atomic_*`) and bit-counting helpers (`qd.math.popcnt`, `qd.math.clz`).

Companion to the in-flight subgroup (#639) and block (#638) docs; this is one of a series of tier-by-tier user-guide pages. Independent of the others — no cross-links to unmerged docs.

Includes:

- Backend / dtype support matrices for all 8 atomics and the two bit-count helpers, with caveats (`f64` fast path, `clz` u32 / u64 + SPV 32-bit hard-cap).
- Per-op semantics with the universal "returns the old value" / "per-location atomicity, not a fence on the rest of memory" framing.
- Worked examples: slot reservation (workhorse pattern for select / compact), histogram, popcnt sum, MSB-via-`clz`.
- Performance and portability notes — atomic contention, fence-pairing, `f64` and `atomic_mul` cost, `clz` portability.

Adds `atomics` to the SIMT-primitives toctree in `index.md`.

## Test plan

- [ ] `cd docs && make html` builds without warnings (no broken cross-links — all unmerged-doc references are plain prose, not Markdown links).
- [ ] Visual check rendered page on RTD preview.

Made with [Cursor](https://cursor.com)

PR #642, opened 2026-05-07
[Docs] Add user-guide page for qd.algorithms.* device-wide algorithms

hughperkins:
## Summary

New user-guide page `docs/source/user_guide/algorithms.md` documenting `qd.algorithms.parallel_sort` and `qd.algorithms.PrefixSumExecutor`.

Companion to the in-flight subgroup (#639), block (#638), atomics (#640), and grid (#641) docs; this is one of a series of tier-by-tier user-guide pages. Independent — no cross-links to unmerged docs.

Includes:

- Backend / dtype support matrix surfacing the most-hit limitation: `PrefixSumExecutor` is `i32`-only and CUDA + Vulkan only.
- Semantics for both ops with explicit constructor / API contracts (length cap, in-place behaviour, stability of `parallel_sort`).
- Worked examples: key-value sort by per-element key; scan-based select / compact via flag-prefix-sum.
- Performance notes — `O(N log² N)` vs radix, allocate-once / run-many for the executor, where the cost actually goes.

Adds a new "Algorithms" caption to the user-guide toctree in `index.md`.

## Test plan

- [ ] `cd docs && make html` builds without warnings.
- [ ] Visual check rendered page on RTD preview.

### DIFF
--- a/docs/source/user_guide/algorithms.md
+++ b/docs/source/user_guide/algorithms.md
@@ -32,7 +32,7 @@ qd.algorithms.parallel_sort(keys, vals)
 - **Key dtype.** Whatever the key field's dtype is, as long as `<` is meaningful for it (integer and float types).
 - **Stability.** Odd-even merge sort is *not* a stable sort — equal keys may be reordered relative to one another. If stability matters, encode tiebreakers into the keys (e.g. pack the original index into the low bits).
 - **Memory.** Strictly in-place — no auxiliary buffers from the caller's perspective.
-- **Performance characteristic.** Beats radix-style sorts for small N (roughly N ≲ 4K), losing to them at large N. For million-element key sets prefer a radix sort (qipc / CUB-style); for thousands or tens of thousands, this is a fine choice.
+- **Performance characteristic.** Beats radix-style sorts for small N (roughly N ≲ 4K), losing to them at large N. For million-element key sets prefer a radix sort; for thousands or tens of thousands, this is a fine choice.
 
 ### `qd.algorithms.PrefixSumExecutor`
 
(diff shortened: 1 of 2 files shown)


docs/source/user_guide/algorithms.md
```diff
@@ -0,0 +1,120 @@
...
+| `qd.algorithms.parallel_sort`   | Odd-even merge sort (in-place, key or key-value) | yes  | yes\*  | yes    | yes\* |
+| `qd.algorithms.PrefixSumExecutor` | Inclusive in-place prefix sum (i32 only)  | yes  | no     | yes    | no    |
+
+\* `parallel_sort` runs anywhere a Quadrants kernel runs; portability is inherited from the underlying kernel infrastructure. AMDGPU and Metal coverage is exercised less heavily than CUDA / Vulkan; report any failures.
+
+## Semantics
+
+### `qd.algorithms.parallel_sort(keys, values=None)`
+
+In-place sort. Reorders `keys` ascending; if `values` is provided, applies the same permutation to `values` (key-value sort). Both arguments must be 1-D `field` or `ndarray`.
```
alanray-tech: **Must fix — `parallel_sort` does not support `ndarray`, only `field`**

This line says:
> Both arguments must be 1-D `field` or `ndarray`.

But `parallel_sort` delegates to the `sort_stage` kernel, which directly accesses `keys.snode.ptr.offset` and `values.snode.ptr.offset` ([`_kernels.py:237–238`](https://github.com/Genesis-Embodied-AI/quadrants/blob/0fd07ef95e/python/quadrants/_kernels.py#L237-L238)). `ndarray` (`ScalarNdarray`) has no `snode` attribute — it uses an `arr` handle with a different storage mechanism. Passing an ndarray will fail during tracing with an `AttributeError`.

The test suite (`test_sort.py`) exclusively uses `qd.field()` and never passes `ndarray` to `parallel_sort`.

Suggested fix:
> Both arguments must be 1-D `field`.

docs/source/user_guide/algorithms.md
```diff
@@ -0,0 +1,120 @@
...
+# arr now holds the inclusive prefix sum: arr[i] = sum(arr_original[0..=i]).
+```
+
+Constructor:
+
+- `length: int` — the maximum number of elements the executor can scan. Internally allocates an auxiliary `qd.field(i32, shape=padded_length)` sized to the Kogge-Stone hierarchy (block size = 64).
+
+`run(input_arr)`:
+
+- `input_arr` must be a 1-D `qd.field(qd.i32, shape=(L,))` with `L <= length`.
```
alanray-tech: **Must fix — `L <= length` contract is incorrect; `L` must equal `length`**

This line says:
> `input_arr` must be a 1-D `qd.field(qd.i32, shape=(L,))` with `L <= length`.

But `run()` always uses the constructor's `length`, not the runtime array size. The relevant code path in [`_algorithms.py`](https://github.com/Genesis-Embodied-AI/quadrants/blob/0fd07ef95e/python/quadrants/algorithms/_algorithms.py):

```python
# line 75-76: length is fixed at construction time
def run(self, input_arr):
    length = self.sorting_length

# line 90: copies `length` elements FROM input_arr
blit_from_field_to_field(self.large_arr, input_arr, 0, length)

# line 114: copies `length` elements BACK to input_arr
blit_from_field_to_field(input_arr, self.large_arr, 0, length)
```

If `input_arr` has `L < length` elements, both `blit_from_field_to_field` calls read/write past the field's extent — out-of-bounds access with undefined behavior.

The constructor description at line 51 also says `length` is "the maximum number of elements", which reinforces the incorrect impression that shorter arrays are safe.

Suggested fix — either:
1. Change the doc to state that `input_arr` must have exactly `length` elements (and update line 51 accordingly), or
2. Fix the implementation to use `min(self.sorting_length, input_arr.shape[0])` so shorter arrays are actually supported.

alanray-tech (commented):
## Code Review

Two inline findings — both are factual inaccuracies that will mislead users into hitting runtime crashes. Backed by source-code evidence.

MuGdxy (dismissed):
LGTM — added a small fixup commit (import statement + rename `input` → `data`). Good to merge.

alanray-tech (approved):
LGTM — added a small fixup commit (import statement + rename `input` to `data`). Good to merge.

hughperkins:
Great!

PR #649, opened 2026-05-07
[Bug] Fix arg_buffer race when same kernel is called on different streams

hughperkins:
The per-handle persistent arg_buffer/runtime_context in the CUDA and AMDGPU kernel launchers is reused across launches.  When the same kernel is dispatched on two different explicit streams (qd_stream), the second call's memcpy can overwrite the buffer while the first kernel is still reading it, causing data corruption.

Fix: when active_stream != nullptr (explicit stream), allocate a per-call ephemeral buffer instead of reusing the persistent one. The stream-ordered mem_free_async ensures the memory stays live until the kernel finishes reading.  Default-stream launches keep the existing persistent-buffer fast path.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
At this point no more streams test failures, but 3 adstack test failures 🤔 

Screenshot 2026-05-07 at 16 05 05

hughperkins:
After update:

Screenshot 2026-05-07 at 16 23 05

PR #410, opened 2026-03-12
[Perf] Streams 1-4

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

PR #653, opened 2026-05-07
[Perf] Re-land Streams 1-4 with bug fixes

hughperkins:
## Summary

- Re-lands the streams merge (#410) that was reverted in #650
- Fixes arg_buffer race when the same kernel is called on different explicit streams (per-handle persistent buffers instead of launcher-global)
- Fixes adstack overflow diagnostic losing kernel name on fast GPUs (peek-then-sync instead of consuming both slots atomically)
- Fixes use-after-free in KernelLauncher: `contexts_` changed from `std::vector` to `std::deque` so that recursive `register_llvm_kernel` calls from `publish_adstack_metadata`'s host-eval path cannot invalidate references held by a parent `launch_llvm_kernel` frame

## Test plan

- [x] `test_adstack_triangular_ndrange_self_referential_push_idempotency` passes 10/10 with `QD_OFFLINE_CACHE=0`
- [x] Full adstack test suite (1108 tests) passes
- [ ] CI passes


hughperkins:
Benchmarks ok:

20260508-1705-streamsagain

hughperkins:
Genesis unit tests pass:

Screenshot 2026-05-08 at 18 06 37

PR #657, opened 2026-05-08
[Docs] Doc for existing atomics: switch support table to per-backend columns

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/docs/source/user_guide/atomics.md
+++ b/docs/source/user_guide/atomics.md
@@ -6,18 +6,22 @@ Atomic read-modify-write operations on a single memory location. They do not syn
 
 All atomic ops follow the same shape: `qd.atomic_op(x, y)` performs `x = op(x, y)` atomically and returns the **old** value of `x`. `x` must be a writable memory target (a field element, ndarray element, or matrix slot); scalars and constant expressions are not allowed.
 
-| Op             | Semantics                              | i32 | u32 | i64 | u64 | f32 | f64 |
-|----------------|----------------------------------------|-----|-----|-----|-----|-----|-----|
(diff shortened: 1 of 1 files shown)


alanray-tech:
The Codex bot's concern about CPU float `atomic_add` was partially valid — the table was indeed inconsistent, but the underlying codegen is correct.

**What the codegen actually does on CPU:**
- **f16 `atomic_add`**: Explicit CAS loop (via `atomic_op_using_cas` + `CreateFAdd`) — correctly a "CAS" entry.
- **f32 / f64 `atomic_add`**: Single `AtomicRMWInst::FAdd` — a single LLVM IR atomic instruction, same as integer min/max uses `AtomicRMWInst::Min/Max`.

**The inconsistency was in the docs table, not the codegen:**
Integer `min`/`max` (also single AtomicRMW instructions) were labelled "native" in the CPU column, while f32/f64 `atomic_add` (also a single AtomicRMW instruction) was lumped into "floats via CAS". Both are lowered identically at the IR level.

**Fix (b93301c):**
- CPU column for `atomic_add` now reads: "int / f32 / f64 native; f16 via CAS" — consistent with how integer min/max is treated.
- Added a cross-cutting note explaining that on CPU targets without hardware float atomics (e.g. x86), the compiler backend may lower these "native" ops to CAS in machine code, same as it does for integer min/max. The distinction from explicit "CAS" entries is that native ops can take advantage of hardware support where available.

alanray-tech (approved):
LGTM. The table restructure is a clear readability improvement, and the CPU column inconsistency for float atomic_add has been fixed in the follow-up commit.

hughperkins:
Great! Thank you. And thank you for the fix 🙌

PR #662, opened 2026-05-08
[Math] Make bitop operations portable cross-gpu

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_unary_ops.py
```diff
@@ -131,9 +128,6 @@ def test_u64(x: qd.uint64) -> qd.int32:
 
 @test_utils.test(arch=[qd.cpu, qd.metal, qd.cuda, qd.amdgpu, qd.vulkan])
```
hughperkins: why filter at all?

hughperkins: why filter at all?

docs/source/user_guide/math.md
```diff
@@ -48,12 +48,13 @@ def msb(x: qd.i32) -> qd.i32:
     return 31 - qd.math.clz(x)
 ```
 
-For `qd.u32` input on CUDA, cast first: `qd.math.clz(qd.bit_cast(x, qd.i32))`.
+For `qd.u32` input on CUDA or AMDGPU, cast first: `qd.math.clz(qd.bit_cast(x, qd.i32))`.
```
hughperkins: seems like this is not in fact general yet...

hughperkins:
@codex review

alanray-tech:
### Potential type mismatch when 64-bit popcnt/clz result feeds into a compound expression

The CUDA and AMDGPU codegen paths for 64-bit `popcnt`/`clz` truncate the LLVM value to i32 and mutate `stmt->ret_type` to `PrimitiveType::i32` **at codegen time**. However, the type-checking pass (`type_check.cpp`) has already set `ret_type` to the operand type (i64) and used that to make type-promotion decisions for downstream statements.

This means a compound expression like:

```python
@qd.kernel
def broken(x: qd.int64) -> qd.int64:
    return qd.math.popcnt(x) + qd.i64(1)
```

will fail at LLVM IR generation: the type checker saw `popcnt` as i64, decided no promotion was needed for the `+ i64(1)`, and emitted no `cast_value`. At codegen time, `CreateAdd(i32, i64)` is called, which triggers an LLVM assertion because `add` requires both operands to have the same type. The same issue applies to `clz`.

The existing tests only cover direct-return cases (`return qd.math.popcnt(x)`), which work by accident because the codegen-time `ret_type` mutation makes the downstream `cast_value` a no-op.

Note: this is not new to this PR — the CUDA `popcnt` i64 path has the same pattern pre-existing. This PR replicates it for AMDGPU and extends it to `clz`.

**Proposed fix:** set `ret_type = PrimitiveType::i32` for `popcnt`/`clz` in the **type-checking pass** (`type_check.cpp`), so all downstream type promotion and codegen operate on the correct type from the start. The per-file changes would be:

1. **`quadrants/transforms/type_check.cpp`** — In `visit(UnaryOpStmt*)`, add a special case after the existing `logic_not` handling:
   ```cpp
   } else if (stmt->op_type == UnaryOpType::popcnt || stmt->op_type == UnaryOpType::clz) {
       stmt->ret_type = PrimitiveType::i32;
   }
   ```
   This mirrors how `logic_not` already forces `ret_type = u1` regardless of operand width.

2. **`quadrants/codegen/llvm/codegen_llvm.cpp`** (CPU backend) — The `llvm.ctpop.i64` / `llvm.ctlz.i64` intrinsics return i64, but `ret_type` is now i32. Add a trunc for 64-bit inputs:
   ```cpp
   else if (op == UnaryOpType::popcnt) {
       llvm_val[stmt] = builder->CreateIntrinsic(llvm::Intrinsic::ctpop, {input_type}, {input});
       if (input_type->isIntegerTy(64)) {
           llvm_val[stmt] = builder->CreateTrunc(llvm_val[stmt], llvm::Type::getInt32Ty(*llvm_context));
       }
   }
   ```
   Same pattern for `clz`.

3. **`quadrants/codegen/cuda/codegen_cuda.cpp`** — Remove the `stmt->ret_type = PrimitiveType::i32` lines (type_check already handles this). `__nv_popcll` / `__nv_clzll` already return i32 from libdevice, so no other change needed.

4. **`quadrants/codegen/amdgpu/codegen_amdgpu.cpp`** — Remove the `stmt->ret_type = PrimitiveType::i32` lines. Keep the trunc (since `llvm.ctpop.i64` returns i64 but `ret_type` is now i32).

5. **`quadrants/codegen/spirv/spirv_codegen.cpp`** — For `popcnt`, `OpBitCount` on an i64 operand returns i64; add a cast to i32 after it (the `clz` path already computes in i32 and casts to `dst_type` at the end, which will now be i32 — no change needed there).

6. **`tests/python/test_unary_ops.py`** — Add compound-expression tests to cover the previously-broken case:
   ```python
   @qd.kernel
   def test_popcnt_compound_i64(x: qd.int64) -> qd.int64:
       return qd.math.popcnt(x) + qd.i64(1)

   @qd.kernel
   def test_clz_compound_i64(x: qd.int64) -> qd.int64:
       return qd.math.clz(x) + qd.i64(1)
   ```

This approach is consistent with how `logic_not` already forces `ret_type = u1` in `type_check.cpp`, and ensures all backends agree on the return type from the earliest stage of the pipeline.

hughperkins:
@codex review

alanray-tech:
LGTM, ok to merge

alanray-tech (approved):

PR #652, opened 2026-05-07
[GPU] Subgroup ops cross-gpu, at least for data movement ops

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

PR #665, opened 2026-05-09
[GPU] Subgroup ops cross-gpu

hughperkins:
Issue: #

### Brief Summary

Makes every operation in `qd.simt.subgroup` work across all GPU backends
(CUDA, AMDGPU, SPIR-V/Vulkan/Metal):

- **Renames + deprecation**: `barrier()` → `sync()`, `memory_barrier()` → `mem_fence()`. Old names kept as aliases that fire a one-shot DeprecationWarning. `sync()` and `mem_fence()` now have correct cross-backend implementations (not no-ops).
- **Data movement (universal)**: `shuffle_up`, `shuffle_xor`, `broadcast_first` now work on all three backends.
- **Identification + control**: `group_size()` and `elect()` (lane-0 always) implemented universally.
- **Reductions and scans (universal)**: 14 ops — `inclusive_{add,mul,min,max,and,or,xor}` and `exclusive_*` — implemented as portable Hillis-Steele scans over `shuffle_up` with an explicit `log2_size` parameter (no-arg form removed). SPIR-V `OpGroupNonUniformInclusiveScan` fast path dropped in favour of one portable `@qd.func` everywhere.
- **Voting and predicate ops (universal)**: `all_true`, `any_true`, `all_equal` as portable `shuffle_xor` butterflies with a CUDA-only `log2_size==5` shortcut to `__all_sync` / `__any_sync`. `all_equal` floats follow `OpGroupNonUniformAllEqual`: NaN ≠ NaN, +0.0 == -0.0.
- **Metal scan miscompilation fix** (from sibling agent, included): rework `_exclusive_scan` to do `shuffle_up → seed lane 0 → inclusive scan` so we avoid post-scan shuffles that MoltenVK miscompiles.

Tests cover every group across the full 64-lane launch (i.e. both 32-lane subgroups), all dtypes, all log2_sizes, plus float-equality and non-binary-truthy contract tests for the voting ops.

Validated on cluster + AMD cloud:

- CUDA (RTX-high): `test_simt.py` 379 passed / 1 skipped, `test_scan.py + test_sort.py` 65 passed
- Vulkan (RTX-high): `test_simt.py` 360 passed / 20 skipped, `test_scan.py + test_sort.py` 65 passed
- AMDGPU (MI300X, ROCm 7.2): `test_simt.py` 361 passed / 21 skipped (CUDA-only legacy `qd.simt.warp.*`), `test_scan.py + test_sort.py` 43 passed / 22 skipped
- Metal: handled in sibling commit (`fbdc91ae7`) — exclusive scans pass

Replaces #652 (auto-closed by the GitHub branch-rename API; previous branch name was `hp/subgroup-renames-and-universal-data-movement`).

### Walkthrough

copilot:walkthrough


hughperkins:
@codex review

quadrants/inc/internal_ops.inc.h
```diff
@@ -34,17 +34,9 @@ PER_INTERNAL_OP(subgroupShuffleDown)
...
-PER_INTERNAL_OP(subgroupInclusiveAdd)
-PER_INTERNAL_OP(subgroupInclusiveMul)
-PER_INTERNAL_OP(subgroupInclusiveMin)
-PER_INTERNAL_OP(subgroupInclusiveMax)
-PER_INTERNAL_OP(subgroupInclusiveAnd)
-PER_INTERNAL_OP(subgroupInclusiveOr)
-PER_INTERNAL_OP(subgroupInclusiveXor)
+// subgroupAdd / subgroupMul / subgroupMin / subgroupMax / subgroupAnd / subgroupOr / subgroupXor removed: use portable
+// Python `subgroup.reduce_add(value, log2_size)` (and equivalents) on top of `subgroupShuffleDown` / `subgroupShuffle`,
+// which work on all backends.  The inclusive-scan ops below remain SPIR-V-only pending portable replacements.
```
alanray-tech: Nit: the trailing sentence still says "The inclusive-scan ops **below** remain SPIR-V-only pending portable replacements", but all seven `subgroupInclusive*` entries have been removed in this PR. Should read something like: "The `subgroupInclusive` ops have also been removed — they are now portable `@qd.func` Hillis-Steele scans over `subgroupShuffleUp` in Python."

hughperkins: Opus said:

Fixed in 566a47b0a — rewrote the comment to cover both families uniformly: both `subgroupAdd*` / `subgroupMul*` / etc. and `subgroupInclusive*` / `subgroupExclusive*` are now removed in this PR, replaced by portable Python `@qd.func` Hillis-Steele scans on top of `subgroupShuffleDown` / `subgroupShuffleUp` / `subgroupShuffle`. Thanks for the catch.

python/quadrants/lang/simt/subgroup.py
```diff
@@ -89,76 +196,238 @@ def reduce_all_add(value, log2_size: template()):
...
+
+    The shuffle is in uniform CF (every lane participates); only the per-lane reduce step is conditional, matching the
+    contract for ``shuffle_up``.  Cross-group ``shuffle_up`` partners are masked out by ``lane_in_group >= offset``, so
+    groups smaller than the full subgroup compose correctly when ``log2_size < log2(group_size)``.
+
+    Note: ``qd.select`` cannot be used here instead of ``if`` because ``OpSelect`` on MoltenVK / Metal miscompiles when
+    one operand is an f32 produced by a shuffle intrinsic—the select silently returns the false-branch value regardless
+    of the condition.  The ``if`` form works correctly for the inclusive scan on its own; callers that issue further
+    subgroup ops after this scan (e.g. `_exclusive_scan`) must insert a ``sync()`` barrier to force reconvergence
+    before the next shuffle.
```
alanray-tech: Nit: the last sentence of this note — “callers that issue further subgroup ops after this scan (e.g. `_exclusive_scan`) must insert a `sync()` barrier to force reconvergence before the next shuffle” — describes a constraint from the **old** `_exclusive_scan` implementation (inclusive scan → shuffle_up on the result), which triggered MoltenVK’s reconvergence bug.

The current `_exclusive_scan` was restructured to avoid this entirely: it shuffles the *input* first, seeds lane 0 with the identity, then runs the inclusive scan as the final step — so there is no shuffle after the scan, and the reconvergence issue cannot arise. No `sync()` is inserted anywhere in the current code path, and all exclusive-scan tests pass on Metal.

Suggest trimming this to just the `qd.select` caveat and dropping the stale `sync()` advice, e.g.:

```
Note: ``qd.select`` cannot be used here instead of ``if`` because ``OpSelect`` on MoltenVK / Metal
miscompiles when one operand is an f32 produced by a shuffle intrinsic — the select silently returns
the false-branch value regardless of the condition.  The ``if`` form works correctly on its own.
```

hughperkins: Opus said:

Agreed — fixed in dd972537d. Verified `grep 'sync()' subgroup.py` finds zero call sites between any `_inclusive_scan` and a subsequent shuffle, and the current `_exclusive_scan` shape (shuffle input → identity-seed lane 0 → `_inclusive_scan`) has no shuffle-after-scan anywhere. Trimmed the docstring to just the `qd.select` / `OpSelect` caveat as you suggested.

Also fixed the same staleness in the prose block right above `_exclusive_scan` (line ~322 on the previous commit) which described the old `inclusive_scan` → `shuffle_up(result)` order — rewrote it to describe what the code actually does and why that ordering avoids the MoltenVK miscompile.

alanray-tech (approved):
LGTM, ok to merge

hughperkins:
Great! 🙌

PR #668, opened 2026-05-09
[AMDGPU] Hoist rand-state buffer to process lifetime

hughperkins:
Per `qd.init()` materialize_runtime allocates ~480 MiB of device memory for the per-thread RandState array (sizeof(RandState) * saturating_grid_dim * max_block_dim, ~19456 * 1024 * 24 B at default config), and frees it on `qd.reset()`. Under multi-process pytest-xdist contention this per-cycle hipMalloc/hipFree pair is a major contributor to AMDGPU `HSA_STATUS_ERROR_OUT_OF_RESOURCES` failures (see perso_hugh/doc/amdgpu_test_suite_oom_followup.md). RandState contents only depend on (num_states, starting_seed); both are stable across init/reset cycles for a given worker, so reusing the buffer changes no observable behavior.

This change introduces `PersistentRandStateBuffer`, a process-lifetime singleton that owns one device allocation (via `cuMemAlloc` / `hipMalloc` directly, bypassing `LlvmDevice` / `DeviceMemoryPool` so its lifetime is independent of any `LlvmRuntimeExecutor`). The buffer grows monotonically; it is intentionally leaked at process exit (matching the AMDGPU persistent-runtime-JIT cache pattern), since freeing driver memory at static-destructor time would race the GPU driver context's own teardown.

`runtime_initialize` and `runtime_get_memory_requirements` each gain one trailing parameter so the host launcher can pass the process-lifetime buffer pointer / signal that the rand-states bytes should not be carved out of `runtime_objects_chunk`. CPU path is byte-identical (passes nullptr / 0). A new
`runtime_get_rand_states_buffer_size` helper returns the page-aligned byte size to size the persistent allocation host-side.

Bisection on the C++ omnibus reproducer
(`experiments/launch_oor_repro/omnibus_repro.cpp`) showed that removing the per-cycle ~400 MiB realloc was sufficient to push the failure threshold past the workload that real `pytest -n 8` exercises.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/runtime/llvm/CMakeLists.txt
+++ b/quadrants/runtime/llvm/CMakeLists.txt
@@ -4,6 +4,7 @@ add_library(llvm_runtime)
 target_sources(llvm_runtime
   PRIVATE
     llvm_runtime_executor.cpp
+    persistent_rand_state_buffer.cpp
     adstack_lazy_claim/bound_eval.cpp
     adstack_lazy_claim/metadata_publish.cpp
     adstack_lazy_claim/heap_grow.cpp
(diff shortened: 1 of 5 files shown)


hughperkins:
@codex review

PR #669, opened 2026-05-09
[AMDGPU] Trim default mempool on qd.reset()

hughperkins:
`AMDGPUContext` (and `CUDAContext`, symmetrically) configures the device's default async-alloc mempool with a 128 MiB release threshold at construction. After `qd.reset()`, any device allocations that were serviced from this pool get returned to the pool, but pages remain cached in the pool until either an allocation request needs them or the threshold is exceeded. From the kernel driver's perspective this means up to 128 MiB of "freed" pages stay reserved per process, which to co-tenants on the same device looks indistinguishable from memory still in use.

This silently violates the user-facing contract of `qd.reset()` ("release everything I allocated") and on AMDGPU was a contributing factor to multi-process `HSA_STATUS_ERROR_OUT_OF_RESOURCES` failures under `pytest-xdist`: 8 workers each holding 128 MiB of stranded mempool cache adds up to ~1 GiB of aggregate VRAM that no process is actually using but no process can hand back without an explicit trim call (see perso_hugh/doc/amdgpu_test_suite_oom_followup.md).

Add a `trim_default_mem_pool()` method to both contexts that calls `hipMemPoolTrimTo` / `cuMemPoolTrimTo` with `min_bytes_to_keep = 0`, and invoke it from `LlvmRuntimeExecutor::finalize` after the existing `synchronize()`. No-op on devices without mempool support. The next `qd.init()` re-allocates from the driver instead of from the cached pool, which is irrelevant against the much heavier work `qd.init()` already does (LLVM JIT, HSACO load).

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/rhi/amdgpu/amdgpu_context.cpp
+++ b/quadrants/rhi/amdgpu/amdgpu_context.cpp
@@ -205,6 +205,16 @@ void AMDGPUContext::launch(void *func,
   }
 }
 
+void AMDGPUContext::trim_default_mem_pool() {
+  if (!supports_mem_pool_) {
+    return;
+  }
+  void *default_mem_pool = nullptr;
+  driver_.device_get_default_mem_pool(&default_mem_pool, 0 /*device ordinal*/);
+  // `min_bytes_to_keep = 0` releases every cached page back to the driver.
+  driver_.mem_pool_trim_to(default_mem_pool, 0u);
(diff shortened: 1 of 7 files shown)


hughperkins:
@codex review

PR #673, opened 2026-05-09
[GPU] Add atomic exchange

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


docs/source/user_guide/atomics.md
hughperkins, line 1: this doc change looks ok

hughperkins:
@codex review

quadrants/ir/stmt_op_types.h
```diff
@@ -62,7 +62,7 @@ enum class TernaryOpType : int { select, ifte, undefined };
 
 std::string ternary_type_name(TernaryOpType type);
 
-enum class AtomicOpType : int { add, sub, max, min, bit_and, bit_or, bit_xor, mul };
+enum class AtomicOpType : int { add, sub, max, min, bit_and, bit_or, bit_xor, mul, xchg };
```
alanray-tech, line 65: 🔴 **Critical**: Adding `xchg` to this enum without updating `expression_printer.h` causes an **out-of-bounds read** (undefined behavior).

In `expression_printer.h:159-167`, `names_table` has 8 entries (indices 0–7), but `xchg` has enum value 8. The guard `op_type > std::size(names_table)` (i.e. `> 8`) does **not** catch `xchg == 8`, so `names_table[8]` is an OOB access. Any IR dump / diagnostic that encounters an `atomic_exchange` node will read past the array.

Fix: add `"atomic_xchg"` to the end of `names_table`, and change the guard to `>=`:

```cpp
constexpr const char *names_table[] = {"atomic_add",     "atomic_sub",    "atomic_min",     "atomic_max",
                                       "atomic_bit_and", "atomic_bit_or", "atomic_bit_xor", "atomic_mul",
                                       "atomic_xchg"};
if (op_type >= std::size(names_table)) {
```

---

🟡 **Suggestion (pre-existing)**: While fixing that table, note that indices 2 and 3 are swapped relative to the enum. The enum order is `add=0, sub=1, max=2, min=3`, but the table has `"atomic_min"` at [2] and `"atomic_max"` at [3]. This has been wrong since before this PR but is worth fixing in the same change.

quadrants/ir/stmt_op_types.cpp
```diff
@@ -98,6 +98,7 @@ std::string atomic_op_type_name(AtomicOpType type) {
     REGISTER_TYPE(bit_and);
     REGISTER_TYPE(bit_or);
     REGISTER_TYPE(bit_xor);
+    REGISTER_TYPE(xchg);
```
alanray-tech, line 101: 🟡 **Suggestion**: `atomic_to_binary_op_type` (line 109–127 in this file) does not handle `AtomicOpType::xchg` — it falls through to the `default: QD_NOT_IMPLEMENTED` case. This is currently safe because `demote_atomics.cpp` special-cases xchg before calling `atomic_to_binary_op_type`, but it is a latent trap for anyone who later calls `atomic_to_binary_op_type(xchg)` without knowing about the special case.

Consider either:
- Adding a comment near `atomic_to_binary_op_type` explaining that `mul` and `xchg` intentionally have no binary-op mapping, or
- Adding an explicit `case AtomicOpType::xchg: QD_ERROR("xchg has no binary-op equivalent; handled separately in demote_atomics");`

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1665,7 +1665,21 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
...
     }
 
-    if (use_native_atomics) {
+    if (stmt->op_type == AtomicOpType::xchg && !dest_is_ptr && !dt->is_primitive(PrimitiveTypeID::f16)) {
+      // Float xchg: route through OpAtomicExchange on the uint-backed buffer pointer (already established by
+      // the addr_ptr block above when no native float-atomic-add cap is in play). OpAtomicExchange supports
+      // float operands per the SPIR-V spec, but going through uint avoids any spirv_has_atomic_float_* cap
+      // dependency and works on every backend (including MoltenVK / spirv-cross-msl on Apple Silicon).
+      // Shared (workgroup) float xchg and f16 xchg are not yet covered -- they would need uint-backing
+      // analogous to the add CAS path; out of scope for the initial xchg landing.
```
alanray-tech, line 1732: 🟡 **Suggestion**: The comment correctly documents that shared (workgroup) float xchg and f16 xchg are out of scope, but there is no runtime guard preventing those code paths from being reached. If someone calls `qd.atomic_exchange` on a `SharedArray` float or an f16 field on a SPIR-V backend today, control falls through to `shared_float_atomic()` or `float_atomic()` (~line 1690/1692), which almost certainly do not handle `AtomicOpType::xchg` and will produce incorrect SPIR-V or crash.

Consider adding a guard before the fallback paths, e.g.:

```cpp
    } else if (dest_is_ptr) {
      QD_ASSERT_INFO(stmt->op_type != AtomicOpType::xchg,
                     "atomic_exchange on shared float arrays is not yet supported in SPIR-V");
      val = shared_float_atomic(*ir_, stmt->op_type, addr_ptr, data, dt);
    } else {
      QD_ASSERT_INFO(stmt->op_type != AtomicOpType::xchg || !dt->is_primitive(PrimitiveTypeID::f16),
                     "Unreachable: global float xchg should be handled by the uint-bitcast path above");
      val = ir_->float_atomic(stmt->op_type, addr_ptr, data, dt);
    }
```

This gives a clear error message instead of a silent miscompile.

hughperkins:
Per Opus, above comments address:

Screenshot 2026-05-11 at 17 31 15

hughperkins:
Will make one single PR for new QIPC atomic ops
=> closing this one

PR #675, opened 2026-05-09
[SPIR-V] dispatch_max_reducers: register each task with the real kernel name

duburcqa:
# [SPIR-V] `dispatch_max_reducers`: register each task with the real kernel name

> One-line plumbing fix. SPIR-V's `dispatch_max_reducers` was passing `kernel_name=""` to `register_adstack_sizing_info`, which was harmless before #671 turned that registration call into a content-stable hash of `(kernel_name, task_id_in_kernel)`. Under the hash, every SPIR-V task with the same task index across distinct kernels collapses onto one registry slot, polluting the per-spec `max_reducer_cache_` and surfacing as `Adstack sizer shader returned an implausibly large per-thread stride` on Apple Metal in Genesis-style reverse-mode workloads.

```diff
 mutable_attribs.registry_id = cache->register_adstack_sizing_info(
-    static_cast<const void *>(&mutable_attribs), /*kernel_name=*/std::string{}, static_cast<int>(ti),
+    static_cast<const void *>(&mutable_attribs), kernel_name, static_cast<int>(ti),
     std::move(allocated_max_sizes), std::move(size_exprs));
```

`kernel_name` is plumbed from `runtime.cpp::launch_kernel` (already on hand for `publish_adstack_metadata_spirv` one line below). LLVM path was always passing the real name; only the SPIR-V launcher needed fixing.

## Tests

`tests/python/test_adstack.py::test_max_reducer_per_kernel_registry_id_isolation` - two `qd.template()` instantiations of the same `@qd.kernel` produce identical adstack trees and identical `(stack_id, mor_node_idx)` coords; pre-fix the second instantiation's first-launch dispatch counter stays at 0 (cache-hits the first's entries via shared `registry_id`), post-fix it advances. Genesis-shaped failure is reproduced by `repro_metal_rigid_step.py` in the issue thread.

### DIFF
--- a/quadrants/runtime/gfx/adstack_max_reducer_launch.cpp
+++ b/quadrants/runtime/gfx/adstack_max_reducer_launch.cpp
@@ -95,7 +95,8 @@ uint64_t pack_max_reducer_key(uint32_t registry_id, int32_t stack_id, int32_t mo
 MaxReducerResultMap GfxRuntime::dispatch_max_reducers(LaunchContextBuilder &host_ctx,
                                                       DeviceAllocationGuard *args_buffer,
                                                       const std::unordered_map<int, DeviceAllocation> &ndarray_allocs,
-                                                      const std::vector<spirv::TaskAttributes> &task_attribs) {
+                                                      const std::vector<spirv::TaskAttributes> &task_attribs,
+                                                      const std::string &kernel_name) {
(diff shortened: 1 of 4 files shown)


hughperkins:
@codex review

hughperkins:
ok to merge

PR #676, opened 2026-05-10
[BREAKING][GPU] New QIPC ops for subgroups

hughperkins:
## Summary

Note: this is BREAKING because we rename the tiled reduce operations to have a _tiled suffix, and put back the reductions without hte log2_size parameter, wihout a suffix.

Adds a single, consistent set of new SIMT-subgroup query / inter-processor-communication ops to ``qd.simt.subgroup``, all working portably across CUDA, AMDGPU, and SPIR-V (Vulkan / Metal):

- **Reductions (lane 0 / broadcast)** — ``reduce_min``, ``reduce_max`` (lane 0 of each ``2**log2_size`` group), ``reduce_all_min``, ``reduce_all_max`` (broadcast to every lane).
  - Built on the existing ``shuffle_down`` tree / ``shuffle_xor`` butterfly patterns; same shape as ``reduce_add`` / ``reduce_all_add``.
  - NaN handling for floats is implementation-defined and is documented.
- **Ballot** — ``subgroup.ballot(predicate)`` returns a ``u32`` bitmask (bit ``i`` set iff lane ``i``'s predicate is non-zero).
  - Single hardware instruction on every backend: ``__ballot_sync`` (CUDA), ``v_ballot_b32`` (AMDGPU), ``OpGroupNonUniformBallot`` (SPIR-V).
  - Cherry-picked from #600 (which is now superseded by this PR).
- **Segmented reduce** — ``subgroup.segmented_reduce_add`` / ``segmented_reduce_min`` / ``segmented_reduce_max`` (``value, head_flag, log2_size``) run a per-lane inclusive scan that resets at every non-zero ``head_flag``, scoped to ``2**log2_size`` consecutive lanes.
  - One ``ballot`` to materialise the head bitmask, one ``clz`` to find each lane's segment head, then a Hillis-Steele inclusive scan bounded by ``distance >= offset``. Cost: ``1 ballot + 1 clz + log2_size shuffles + log2_size ops``.
  - No identity argument is required (unlike ``exclusive_min`` / ``exclusive_max``): the per-lane ``distance >= offset`` guard ensures the scan never crosses a segment boundary, so a partner from another segment is never combined with the local value.
  - The group's first lane is treated as an implicit head, so a segment that starts mid-group is still reduced correctly.
- **Lane masks** — ``subgroup.lanemask_lt(lane_id)`` / ``_le`` / ``_eq`` / ``_gt`` / ``_ge``: closed-form ``u32`` masks parametrised by a lane id, mirroring CUDA's ``__lanemask_{lt,le,eq,gt,ge}`` but generalised to take any lane_id (pass ``invocation_id()`` for the CUDA built-in form).
  - Pure ``@qd.func`` arithmetic — no backend intrinsic, no shuffle, no ballot — so per-lane-varying ``lane_id`` works the same as a uniform one.
  - Caller contract: ``lane_id`` in ``[0, 31]``; on AMDGPU CDNA wave64 the mask covers only the low 32 lanes (build a 64-bit mask from two ``u32`` ballots if needed).

## Drive-by fixes (required by ``segmented_reduce_*``, but useful in their own right)

``qd.clz`` is the first user of ``clz`` in the codebase, and exposed bugs on every backend:

- **CUDA** — ``__nv_clz`` / ``__nv_clzll`` are declared on signed types but operate on the underlying bit pattern; route u32 / u64 through them so ``qd.clz(u32(...))`` no longer hits ``QD_NOT_IMPLEMENTED``.
- **AMDGPU** — ``emit_extra_unary`` had no ``clz`` case; map to LLVM's ``Intrinsic::ctlz`` with ``is_zero_undef=0``.
- **SPIR-V** — split GLSL.std.450 ``FindSMsb`` (#74, signed) and ``FindUMsb`` (#75, unsigned). The unsigned form is required for u32 / u64 inputs whose top bit may be set; ``FindSMsb`` is undefined for those (treats them as negative; "most-significant 0-bit" doesn't exist for ``0xFFFFFFFF``). Cast the result back to ``i32`` before the ``32 - msb - 1`` subtraction so SPIR-V's strict-type ``sub`` is happy.

## Stacking

This PR is stacked on top of #665 (``hp/cross-gpu-subgroup``).

It supersedes #600 (``hp/cross-gpu-ballot``), whose three commits are cherry-picked here unchanged. #600 can be closed once this lands (or once #665 lands, whichever is convenient).

## Test plan

- [x] Local pre-commit (black, clang-format, ruff, pylint) passes.
- [x] ``pyright`` (project config) clean for new code; pre-existing errors in untouched files unchanged.
- [x] CUDA (cluster, RTX PRO 6000 Blackwell): all 103 new lanemask + segmented_reduce_min/max tests pass; full ``test_simt.py`` (586 passed, 1 skipped) green.
- [x] Vulkan (same cluster node): all 103 new tests pass; full ``test_simt.py`` (567 passed, 20 skipped) green.
- [x] AMDGPU (amddesktop, Radeon RX 7900 XTX, gfx1100, ROCm): all 103 new tests pass; full ``test_subgroup_*`` (566 passed, 21 deselected) green.
- [x] Comment / docstring widths audited against the project's 120c target via ``find_underwrapped.py``.


hughperkins:
Opus wrote:

Thanks @chatgpt-codex-connector — both comments were correct.

**P2 wave64 in `_segment_head_distance`** — fixed in `ddb8e25f9`. Concrete failure on wave64 lane 32 with `log2_size == 5`:

- `group_mask = bits_in_group << 32` is u32 shift-out-of-range.
- `inclusive_mask = 0xFFFFFFFF >> (31 - 32)` is `>> -1` reinterpreted as `>> 0xFFFFFFFF`.
- `head_mask = ballot_first_n(p, 32)` doesn't contain lanes 32..63's predicates anyway.

Fix: switch to `ballot_full_subgroup` (u64) and run the rest of the algorithm in half-local 32-lane coordinates (`lane_in_half = lane - (lane & ~31)`). The `2**log2_size <= 32` cap guarantees segments never cross the half boundary, so half-local distance equals absolute distance and downstream `shuffle_up` partners stay in-half. On wave32 the new code folds back to the original arithmetic (high 32 bits of `full_mask` always zero, `half_base == 0`).

Added a new `test_subgroup_segmented_reduce_add_block64` parametric over `log2_size in [0, 5]` that uses `block_dim = 64` with identical patterns in both 32-lane halves. On wave32 hardware it passes regardless (two 32-lane subgroups); on wave64 it'll exercise the new half-local path. Wave32 results: 1187/1187 `test_simt` on CUDA + Vulkan, 584/584 (+ 21 skipped) on AMDGPU RDNA3. Wave64 (MI300X) testing pending hardware availability.

**P2 `math.md` clz support matrix** — fixed in `25a921707`. Updated the support table to reflect that `qd.math.clz` now works on i32/u32/i64/u64 on both CUDA (no more `bit_cast` workaround) and AMDGPU (was unsupported), with the SPIR-V 64-bit-truncation caveat preserved as the only remaining limitation. Dropped the AMDGPU `xfail` from `test_clz` and added new `test_clz_u32`/`_i64`/`_u64` cases covering values with the top bit set.

hughperkins:
Tests passing on metal:

hughperkins:
Opus said:

## MI300X (wave64) validation — found and fixed an LLVM AMDGPU isel bug

Now that `amdcloud` is back, ran the new ballot ops on real wave64 hardware (gfx942 / MI300X). Initial run had **10 failures**, mostly with `LLVM Fatal Error: Cannot select: AMDGPUISD::SETCC ... zero_extend i1 ... Register:i1` during kernel compile.

**Root cause:** LLVM AMDGPU's `llvm.amdgcn.ballot.i32` fails to select on gfx942 when the predicate is a non-constant `i1`. PR [llvm/llvm-project#71556](https://github.com/llvm/llvm-project/pull/71556) (LLVM 18) documented `ballot.iN` on a mismatched wavefront width as well-defined (SETCC at wavefront width, then trunc/zext to the requested type), and that's what the comment in `codegen_amdgpu.cpp` cited. In practice the bug persists in LLVM 20 *and* 22.1.0 — minimal repro:

```llvm
declare i32 @llvm.amdgcn.ballot.i32(i1)
define amdgpu_kernel void @bug(ptr %out, i32 %p) {
  %x = icmp ne i32 %p, 0
  %r = call i32 @llvm.amdgcn.ballot.i32(i1 %x)
  store i32 %r, ptr %out
  ret void
}
```

`llc -mcpu=gfx942` on this — `LLVM ERROR: Cannot select: AMDGPUISD::SETCC`. Same module with `ballot.i64` selects cleanly.

**Fix** (commit a5319f655):

- AMDGPU codegen now lowers `subgroupBallotU32` as `ballot.i64` followed by `trunc to i32`. Same single `v_cmp_*_e64` plus low-half store as the i32 form would have emitted (the trunc folds during isel), so zero runtime cost on either wave32 or wave64. Comment block in `codegen_amdgpu.cpp` updated with the bug context.
- Two `ballot_full_subgroup_*` tests had `block_dim=32` baked in, which silently violates the "all lanes active" contract on wave64 (only the low half of each wavefront is active). Bumped them to `block_dim=64` so each wave actually has every lane active on both wave32 (two full 32-lane waves) and wave64 (one full 64-lane wave).

**Validation** (post-fix):

| host | arch | suite | result |
| --- | --- | --- | --- |
| amdcloud (MI300X, wave64) | amdgpu | ballot | 16 passed, 1 skipped |
| amdcloud (MI300X, wave64) | amdgpu | segmented_reduce | 102 passed |
| amdcloud (MI300X, wave64) | amdgpu | full simt | 584 passed, 21 skipped |
| amddesktop (RX 7900 XTX, wave32) | amdgpu | ballot | 16 passed, 1 skipped |
| amddesktop (RX 7900 XTX, wave32) | amdgpu | segmented_reduce | 102 passed |

The wave64-specific `_segment_head_distance` rewrite (commit ddb8e25f9) also got real wave64 coverage via `test_subgroup_segmented_reduce_add_block64` — passes cleanly.

docs/source/user_guide/subgroup.md
```diff
@@ -19,16 +19,17 @@ The full Python API is grouped here by category. The first column lists each op,
 | `subgroup.broadcast(value, index)`          | yes  | yes    | yes                     | i32, u32, f32, f64, i64, u64 |
 | `subgroup.broadcast_first(value)`           | yes  | yes    | yes                     | i32, u32, f32, f64, i64, u64 |
 
-\* AMDGPU `shuffle_down` / `shuffle_up` (and therefore `reduce_add`, which is built on `shuffle_down`) are currently emulated via `ds_bpermute` (~50 cycle latency).
+\* AMDGPU `shuffle`, `shuffle_down`, and `shuffle_up` are lowered through `ds_bpermute` (~50 cycle latency, LDS-routed). On wave64 a cross-half read (target lane in the other SIMD32) additionally takes one `v_permlane64_b32` plus a per-lane select — see [AMDGPU wave64 cross-half lowering](#amdgpu-wave64-cross-half-lowering). Same-half reads have no overhead vs. CUDA's `__shfl_*_sync`.
```
hughperkins: Is this only on RDNA? Or on CDNA too?

docs/source/user_guide/subgroup.md
```diff
@@ -62,19 +63,28 @@ Why it composes exactly: the underlying `subgroup.shuffle` / `subgroup.shuffle_d
 
 #### Picking `log2_size` for a "full subgroup" reduction
```
hughperkins: do we need this section any more?

docs/source/user_guide/subgroup.md
```diff
@@ -62,19 +63,28 @@ Why it composes exactly: the underlying `subgroup.shuffle` / `subgroup.shuffle_d
...
+
+The two `ballot` variants are full-subgroup only (no `log2_size` parameter): `ballot_first_n(predicate, n)` returns a `u32` covering lanes `[0, n)` (with `n` a compile-time constant `<= 32`), and `ballot_full_subgroup(predicate)` returns a `u64` covering every lane in the subgroup (32 lanes on wave32, 64 on wave64).
 
 | Op                                          | CUDA          | AMDGPU | SPIR-V (Vulkan / Metal) |
 |---------------------------------------------|---------------|--------|-------------------------|
+| `subgroup.ballot_first_n(predicate, n)`     | yes           | yes    | yes                     |
+| `subgroup.ballot_full_subgroup(predicate)`  | yes (u32 zext'd to u64) | yes (wave32 / wave64) | yes (uvec4 hi:lo packed to u64) |
 | `subgroup.all_true(predicate, log2_size)`   | yes (fast at `log2_size==5`) | yes | yes |
```
hughperkins: where are the _full versions?

docs/source/user_guide/subgroup.md
```diff
@@ -86,6 +96,13 @@ CUDA shortcut: when `log2_size == 5` (full warp), `all_true` / `any_true` lower
 |---------------------------------------------|------|--------|-------------------------|------------------------------|
 | `subgroup.reduce_add(v, log2_size)`         | yes  | yes\*  | yes                     | any type supporting `+`      |
 | `subgroup.reduce_all_add(v, log2_size)`     | yes  | yes    | yes                     | any type supporting `+`      |
+| `subgroup.reduce_min(v, log2_size)`         | yes  | yes\*  | yes                     | integer + float              |
```
hughperkins: where are the _full versions?

docs/source/user_guide/subgroup.md
```diff
@@ -119,37 +136,37 @@ Each lane returns the `value` held by the lane whose subgroup-local id equals `i
 Lane `i` returns the `value` held by lane `i + offset`. Lanes near the top of the subgroup — where `i + offset >= subgroup_size` — receive an implementation-defined value (typically their own `value`), so reduction patterns must only trust lane 0's final result, or mask out the out-of-range lanes.
 
 - `value` and `offset` dtypes: same as `shuffle` above; `offset` is a `u32`.
-- Maps to `__shfl_down_sync` on CUDA and `OpGroupNonUniformShuffleDown` on SPIR-V. On AMDGPU it is currently emulated with `ds_bpermute` (see the support matrix above).
+- Maps to `__shfl_down_sync` on CUDA and `OpGroupNonUniformShuffleDown` on SPIR-V. On AMDGPU it is emulated with `ds_bpermute`; wave64 cross-half offsets (any `offset >= 32` for low-half lanes, or any non-zero `offset` for high-half lanes that lands across the SIMD32 boundary) go through the same `permlane64 + ds_bpermute + select` lowering as `shuffle` — see [AMDGPU wave64 cross-half lowering](#amdgpu-wave64-cross-half-lowering).
```
hughperkins: again, on amdgpu, is this for both RDNA and CDNA, or just for RDNA?

docs/source/user_guide/subgroup.md
```diff
@@ -119,37 +136,37 @@ Each lane returns the `value` held by the lane whose subgroup-local id equals `i
...
 
 ### `shuffle_up(value, offset)`
 
 Lane `i` returns the `value` held by lane `i - offset`. Lanes near the bottom of the subgroup — where `i - offset < 0` — receive an implementation-defined value (typically their own `value`), so the bottom `offset` lanes' results should be ignored or masked.
 
 - Same dtype rules as `shuffle` / `shuffle_down`; `offset` is a `u32`.
-- Maps to `__shfl_up_sync` on CUDA and `OpGroupNonUniformShuffleUp` on SPIR-V. On AMDGPU it is currently emulated with `ds_bpermute((lane - offset) * 4, value)` (same fast-path FIXME as `shuffle_down`).
+- Maps to `__shfl_up_sync` on CUDA and `OpGroupNonUniformShuffleUp` on SPIR-V. On AMDGPU it is emulated with `ds_bpermute((lane - offset) * 4, value)`; wave64 cross-half cases go through the [AMDGPU wave64 cross-half lowering](#amdgpu-wave64-cross-half-lowering) (same `permlane64 + ds_bpermute + select` sequence as `shuffle` / `shuffle_down`).
```
hughperkins: on amdgpu, is this for both RDNA and CDNA, or just for RDNA?

docs/source/user_guide/subgroup.md
```diff
@@ -119,37 +136,37 @@ Each lane returns the `value` held by the lane whose subgroup-local id equals `i
...
+- Implemented portably as a `@qd.func` over `shuffle`: every backend that lowers `shuffle` therefore lowers `shuffle_xor` with no additional codegen path. Inlines at compile time into a single `shuffle(value, u32(invocation_id()) ^ mask)`.
 - The XOR partner must be inside the active subgroup; behaviour outside that range is implementation-defined (same caveat as `shuffle`).
 
 ### `broadcast(value, index)`
 
 Every lane in the subgroup returns the `value` held by the lane whose subgroup-local id equals `index`. Expresses intent ("read lane `index`") more directly than `shuffle(value, index)` and on backends with a dedicated broadcast may map to a cheaper instruction.
 
 - Same dtype rules as `shuffle`.
-- Maps to `__shfl_sync` on CUDA, `ds_bpermute` on AMDGPU, and `OpGroupNonUniformBroadcast` on SPIR-V.
+- Maps to `__shfl_sync` on CUDA, `ds_bpermute` (plus a `permlane64`-driven cross-half select on wave64) on AMDGPU, and `OpGroupNonUniformBroadcast` on SPIR-V. See [AMDGPU wave64 cross-half lowering](#amdgpu-wave64-cross-half-lowering) for the wave64 mechanics.
```
hughperkins: on amdgpu, is this for both RDNA and CDNA, or just for RDNA?

alanray-tech (commented):
## Code Review Findings

Overall this is a high-quality, well-tested PR. The algorithms (tree reduce, XOR butterfly, Hillis-Steele scan, ballot-based segmented reduce) are correct and the three-backend coverage is thorough. A few items worth addressing or acknowledging:

### SPIR-V `visit(InternalFuncStmt)` missing default error path

`spirv_codegen.cpp` line ~1481-1611: if `stmt->func_name` matches none of the `if/else if` branches, `val` is never assigned but `ir_->register_value(stmt->raw_name(), val)` still executes at line 1611, registering an uninitialised `spirv::Value` into the IR. This is UB unless the front-end guarantees exhaustiveness. Suggest adding a `QD_ERROR` before the final `register_value` call (same pattern as the shuffle branch's `QD_ERROR("Unsupported operation: ...")` at line 1594).

### Documentation: `log2_size` valid range inconsistency

`docs/source/user_guide/subgroup.md` line ~200 says `log2_size` is in `[0, 5]` for reductions/scans/votes, and `[0, 6]` for `segmented_reduce_*_tiled` only. But the codebase and tests use `log2_size == 6` for non-segmented `_tiled` ops on AMDGPU wave64 (`test_subgroup_*_log2_size_6`). The `[0, 5]` claim is incorrect for wave64 non-segmented ops. Suggest changing to `[0, log2_group_size()]` (i.e. up to 6 on AMDGPU) to match reality.

### `subgroupBallotU64` on CUDA: high-zero semantics

`codegen_cuda.cpp` line ~769-774: `subgroupBallotU64` is implemented as `ZExt(ballot_i32)`, so high 32 bits are always zero on CUDA. This is correct and intentional, but the user-facing docs in `subgroup.md` could be more prominent about this -- a user calling `ballot(predicate)` on CUDA might expect lanes 32-63 to be populated if they're thinking in terms of the u64 return type. The code comment is good; consider a brief callout in the `ballot(predicate)` semantic section as well.

### `subgroup.py` / `reductions.py` circular import fragility

`subgroup.py` line ~398: the `from quadrants.lang.simt.reductions import *` at the bottom of `subgroup.py` creates a circular dependency that works today only because every symbol `reductions.py` imports from `subgroup` is defined before that line. If someone reorders `subgroup.py` in the future, the import will silently break. The existing comment explains the ordering requirement, which is good -- but consider adding a short note like `# WARNING: reordering definitions above this line may break the circular import with reductions.py` to make this more visible to future contributors.

### AArch64 host: VGPR asm fence absent for AMDGPU bitcode

`runtime.cpp` line ~1088-1090: the `+v` asm constraint is gated on `__x86_64__ || __i386__ || __amdgcn__`, so AArch64-hosted builds produce AMDGPU bitcode without the VGPR hint. The comment documents this well. If ARM-built wheels that run on gfx11+ RDNA3 are a shipping combination, it may be worth a CI regression test covering constant-index shuffle patterns on that build matrix to confirm LLVM's uniformity analysis still emits `ds_bpermute` rather than folding to `v_readlane`.

### Vulkan `PrefixSumExecutor` assumes subgroup width == 32

`_kernels.py` line ~288-295: `subgroup_inclusive_add_warp_i32` hardcodes `log2_size=5`, and `_algorithms.py` routes the Vulkan path through it. This matches the existing CUDA assumption (`WARP_SZ = 32`), but on a Vulkan driver that reports `subgroup_size != 32` (e.g. 16 on some Intel iGPUs, or 64 on AMDGPU via MoltenVK), the prefix-sum would silently compute wrong results. This is a pre-existing limitation (the kernel design assumes 32-wide warps), but now that subgroup intrinsics are wired in for Vulkan it becomes more relevant. Worth a brief comment or a runtime assert that `group_size() == 32` in the adapter.

alanray-tech (approved):
LGTM

PR #684, opened 2026-05-10
[GPU] New QIPC ops for block

hughperkins:
## Summary

Closes the block-level QIPC gaps listed in [perso_hugh/doc/qipc/qipc_gaps_block.md](https://github.com/Genesis-Embodied-AI/perso_hugh/blob/main/doc/qipc/qipc_gaps_block.md) by adding three new families of block-scope primitives under `qd.simt.block`, stacked on the new subgroup primitives branch (`hp/new-qipc-ops-subgroup`).

- **`block.reduce_{add,min,max}` / `block.reduce_all_{add,min,max}`** — CUB-style two-stage warp-reduction tree (per-warp `subgroup.reduce_*` into shared memory, second pass reduces the per-warp partials).  `_all` variants broadcast the result to every thread via shared memory.
- **`block.{inclusive,exclusive}_{add,min,max}`** — CUB-style block scan via warp scans (`subgroup.{inclusive,exclusive}_*` per warp, exclusive scan of warp aggregates in shared memory, fold prefix back into per-thread results).
- **`block.radix_rank_match_atomic_or`** — CUB's `BlockRadixRankMatchEarlyCounts` ATOMIC_OR path: per-warp histogram via `atomic_add` on shared memory, block-wide exclusive prefix over digit columns, intra-warp match via `atomic_or` on a per-digit lane mask, leader election via `clz`/`popcnt`, leader's `atomic_add` plus `subgroup.shuffle` broadcast for the warp offset.

## Files

- `python/quadrants/lang/simt/block.py` — implementations (~370 LoC)
- `tests/python/test_simt.py` — 78 new parametrized tests (36 reduce + 36 scan + 6 radix-rank), all `arch=qd.gpu`
- `docs/source/user_guide/block.md` — user-facing docs with API tables, semantics, costs, and examples for each new primitive

## Notes / gotchas resolved during development

- **CUDA cache staleness on first build.**  After pulling the subgroup branch, the qdcache contained pre-shuffle-up kernels and surfaced as `LLVMRuntime function subgroupShuffleUp not found`.  `rm -rf ~/.cache/quadrants/qdcache/*` fixes it; not a code bug.
- **SPIR-V `FindSMsb` vs `FindUMsb` (commit `d3e73bab4`).**  `leader = 31 - clz(cast(bin_mask, i32))` produced wrong leaders on Vulkan when `bin_mask` had bit 31 set (e.g., a single-lane match in lane 31): `FindSMsb` on a negative i32 returns the position of the highest 0-bit, not the 1-bit, so the leader was elected as lane 1 instead of lane 31 and non-leader lanes read warp_offset 0 from a non-leader lane.  Passing `u32` directly to `clz` dispatches `FindUMsb` and matches CUDA's `__clz`.
- **SPIR-V `subgroupBarrier` has no memory semantics.**  `OpControlBarrier(ScopeSubgroup, ScopeSubgroup, MemorySemantics=0)` does *not* publish shared-memory writes, unlike CUDA's `__syncwarp`.  `radix_rank` pairs every warp-sync inside the match phase with `subgroup.mem_fence()` (`_warp_sync_fence` helper) so the lane-mask publish/clear sequence has correct release/acquire ordering on Vulkan / Metal.
- **`lane_id = invocation_id()` was wrong on wave64 (commit `bac3f9f41`).**  On CDNA AMDGPU the hardware wave is 64 lanes wide but log2_warp=5 partitions a block into 32-lane logical warps, so `invocation_id() == 0` only fires once per hardware wave instead of once per logical warp — every other logical warp's aggregate was silently dropped.  Replaced with `tid & (WARP_SIZE-1)` (CUB's `threadIdx.x & 31` recipe).  Wave32 backends are unaffected.

## Tests

- **CUDA** (RTX, single-GPU rtx-high): **82/82** `-k block` passing
- **Vulkan** (RTX, single-GPU rtx-high): **81 passed + 1 skipped**, `-k block`
- **AMDGPU CDNA** (MI300X amdcloud): **partial** — see follow-up below

## AMDGPU CDNA wave64 follow-up (not blocking this PR)

`block.reduce` / `block.scan` work on AMDGPU only when the block is exactly one logical warp (`block_dim == 32` with `log2_warp=5`).  Multi-warp blocks fail because `shuffle_down` on AMDGPU's `ds_bpermute` lowering wraps OOB indices modulo the wave size — i.e. on a wave64 wave, `shuffle_down(value, 16)` from lane 48 reads lane 0 instead of returning the lane's own value (CUDA's `__shfl_down_sync` semantics with `width=32`).  `subgroup.reduce_*` / `subgroup.{inclusive,exclusive}_*` inherit the same constraint; the parent subgroup branch documents `log2_size` must equal the hardware wave size on CDNA.

`block.radix_rank_match_atomic_or` is wave32-only by construction: the bin_mask atomic_or + `clz` + `popcnt` leader pattern is keyed on a u32 lane mask.  Wave64 would need a u64 bin_mask, u64 `clz` / `popcnt`, and 64-lane `subgroup.shuffle` for the warp_offset broadcast.

Both follow-ups are tracked separately and are not in scope for closing the QIPC block gaps (QIPC currently consumes these on CUDA and Vulkan).

## Test plan

- [x] CUDA: `QD_WANTED_ARCHS=cuda pytest tests/python/test_simt.py -k block`
- [x] Vulkan: `QD_WANTED_ARCHS=vulkan pytest tests/python/test_simt.py -k block`
- [x] AMDGPU: documented as wave32-only / single-warp follow-up
- [ ] Metal CI


docs/source/user_guide/block.md
```diff
@@ -18,6 +18,11 @@ The closely-related grid-level fence (`qd.simt.grid.memfence()`) is documented a
 | `block.SharedArray(shape, dtype)`               | yes  | yes    | yes                     |
 | `block.global_thread_idx()`                     | yes  | yes    | —                       |
 | `block.thread_idx()`                            | no   | no     | yes                     |
+| `block.reduce_{add,min,max}(v, tid, ...)`       | yes  | yes    | yes                     |
```
hughperkins: - why do we need to pass in tid?

docs/source/user_guide/block.md
```diff
@@ -18,6 +18,11 @@ The closely-related grid-level fence (`qd.simt.grid.memfence()`) is documented a
...
+| `block.reduce_all_{add,min,max}(v, tid, ...)`   | yes  | yes    | yes                     |
+| `block.inclusive_{add,min,max}(v, tid, ...)`    | yes  | yes    | yes                     |
+| `block.exclusive_{add,min,max}(v, tid, ...)`    | yes  | yes    | yes                     |
+| `block.radix_rank_match_atomic_or(...)`         | yes  | yes    | yes                     |
 | `grid.memfence()` (device-scope, see below)     | yes  | no     | no                      |
```
hughperkins: why is this still in block.md?

docs/source/user_guide/block.md
```diff
@@ -102,6 +107,100 @@ On CUDA / AMDGPU this is the natural way to identify which work-item a thread sh
...
 
+### `block.reduce_{add,min,max}(value, tid, block_dim, log2_warp, dtype)`
+
+Block-scope reductions following CUB's `BLOCK_REDUCE_WARP_REDUCTIONS` strategy: each warp reduces its lanes via a `shuffle_down` tree, lane 0 of each warp publishes the warp aggregate to shared memory, then thread 0 sequentially folds the warp aggregates with the same operator. The result is valid in **thread 0 only**; other threads retain partial values. For the broadcast-to-every-thread variants see `block.reduce_all_{add,min,max}` below.
+
+Arguments:
+
+- `value`: per-thread input.
+- `tid`: calling thread's block-local index. Pass `i % block_dim` from a `qd.loop_config(block_dim=...)` kernel, or `qd.simt.block.thread_idx()` on backends that expose it.
+- `block_dim`: threads per block (compile-time `template()`; must be a multiple of `2**log2_warp`).
```
hughperkins: do we need to pass in block_dim

alanray-tech (commented):
## Review

LGTM overall - high-quality port of the CUB block primitives with thorough tests and good docs.

One minor doc request before merge:

**`block.inclusive_scan` / `block.exclusive_scan` generic-op contract should document that `op` must be commutative (or at least note the fold direction).**

The docstrings say "custom monoids" / "custom associative operators", which implies non-commutative ops like matrix-multiply would work. But the underlying `_inclusive_scan_tiled` / `_exclusive_scan_tiled` fold as `op(current, previous)` rather than `op(previous, current)`, so non-commutative operators will silently produce wrong prefixes. The typed `_add` / `_min` / `_max` wrappers are all commutative so they're fine - it's only the generic entry points that could mislead.

Suggest adding a one-liner to the `inclusive_scan` / `exclusive_scan` docstrings, e.g.:

> **Note:** `op` must be both associative and commutative. Non-commutative monoids (e.g. matrix multiply) will produce incorrect results due to the per-subgroup shuffle fold direction.

With that doc fix: LGTM, good to merge.

alanray-tech (approved):

PR #690, opened 2026-05-12
[Atomics] New QIPC ops for atomics

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


alanray-tech (changes requested):
## Review

The two issues flagged by Codex are both real and **must be fixed before merge**:

### 1. Tensor CAS scalarization will crash (vector/matrix `atomic_cas`)

`type_check` in `frontend_ir.cpp:982-984` accepts tensor-typed `expected` (peeling it to element type), but neither `scalarize.cpp:531` nor `lower_matrix_ptr.cpp:419` propagates `expected` when splitting atomic ops into per-component statements -- they both use the 3-arg `AtomicOpStmt` constructor which sets `expected = nullptr`. Any call like `qd.atomic_cas(vec_field[None], Vector([0, 0]), Vector([1, 1]))` will hit `QD_ASSERT(stmt->expected != nullptr)` in codegen/demotion and crash.

**Fix**: Either reject tensor destinations for CAS in `type_check` (smallest contact area), or update both scalarizers to split and thread `expected_i` into the 4-arg constructor. The former is safer for a first landing.

### 2. CAS `expected` operand is never type-cast to the pointee type

`type_check` validates that `expected` is integral but never inserts a cast to match the destination type. Meanwhile `val` gets implicit widening via `type_check_store` in `irpass::type_check` (`type_check.cpp:74`). A common call like `qd.atomic_cas(i64_field[None], 0, 1)` leaves `expected` as `i32` while memory holds `i64`, producing invalid IR for both LLVM `CreateAtomicCmpXchg` and SPIR-V `OpAtomicCompareExchange` (both require all operands to have the same type as the value in memory).

**Fix**: Add a `type_check_store`-equivalent for `stmt->expected` in the `AtomicOpStmt` visitor in `type_check.cpp`, or cast `expected` alongside `val` in `AtomicOpExpression::type_check`.

---

A few additional suggestions are left as inline comments.

quadrants/ir/frontend_ir.h
```diff
@@ -673,10 +673,17 @@ class AtomicOpExpression : public Expression {
...
   Expr dest, val;
+  // Only meaningful when `op_type == AtomicOpType::cas`. Empty Expr otherwise. CAS is the only atomic op
+  // with three operands -- (dest, expected, val) -- and is the only place this slot is populated.
+  Expr expected;
 
   AtomicOpExpression(AtomicOpType op_type, const Expr &dest, const Expr &val) : op_type(op_type), dest(dest), val(val) {
   }
 
+  AtomicOpExpression(AtomicOpType op_type, const Expr &dest, const Expr &expected, const Expr &val)
+      : op_type(op_type), dest(dest), val(val), expected(expected) {
```
alanray-tech: Nit: the member initializer list order `dest(dest), val(val), expected(expected)` doesn't match the parameter order `(dest, expected, val)`. C++ initializes in declaration order (`dest, val, expected`), so runtime behavior is correct, but the mismatch between parameter semantics and init order is a readability trap for maintainers. Consider either reordering the member declarations to `dest, expected, val` or swapping the parameter order to `(dest, val, expected)` to make the intent obvious.

quadrants/codegen/spirv/spirv_codegen.cpp
```diff
@@ -1723,19 +1723,58 @@ void TaskCodegen::visit(AtomicOpStmt *stmt) {
...
   } else if (is_integral(dt)) {
+    if (stmt->op_type == AtomicOpType::cas) {
+      // OpAtomicCompareExchange takes (scope, sem_eq, sem_neq, value, comparator) and returns the value
+      // originally at `addr_ptr`. We surface that prior value to the user; success is recovered with
+      // `(returned == expected)`. Matches CUDA atomicCAS semantics.
+      QD_ASSERT(stmt->expected != nullptr);
+      spirv::Value expected_val = ir_->query_value(stmt->expected->raw_name());
+      ir_->make_inst(spv::OpMemoryBarrier, ir_->const_i32_one_,
+                     ir_->uint_immediate_number(ir_->u32_type(), spv::MemorySemanticsAcquireReleaseMask |
+                                                                     spv::MemorySemanticsUniformMemoryMask));
```
alanray-tech: This `OpMemoryBarrier` is inserted before the CAS but none of the other atomic ops (including `xchg`, `atomic_add`, `atomic_min`, etc.) have it. `OpAtomicCompareExchange` already provides its own memory semantics via the `sem_eq` / `sem_neq` operands (which are `const_i32_zero_` here -- Relaxed). If the barrier is genuinely needed for CAS correctness, it should also be needed for the other atomic ops (they all use Relaxed semantics too). If it's not needed, it adds an unnecessary performance cost and an asymmetry that will confuse readers. Please either remove it or add a comment explaining why CAS specifically needs an extra barrier that no other atomic op does.

python/quadrants/lang/ops.py
```diff
@@ -1450,6 +1450,92 @@ def atomic_xor(x, y):
...
+        >>>     if old == 0:
+        >>>         # we won the race; the slot is ours
+        >>>         ...
+    """
+    if impl.is_python_backend():
+        old = x.item()
+        if old == expected:
+            x[()] = desired
+        return old
+    if not (is_quadrants_expr(x) and x.ptr.is_lvalue()):
```
alanray-tech, line 1556: Minor: `@writeback_binary` (used by all other atomic ops) has an extra guard `if isinstance(a, Field) or isinstance(b, Field): return NotImplemented` that lets Python's operator protocol fall through cleanly. `atomic_cas` inlines the lvalue check but skips this `Field` guard. Since `atomic_cas` is a regular function call (not an operator), the `NotImplemented` path isn't strictly needed, but if someone passes a raw `Field` instead of `field[None]`, they'll get a confusing `AttributeError` from `x.ptr` rather than the clear `QuadrantsSyntaxError`. Consider adding a similar early check or at least a guard like `if isinstance(x, Field): raise QuadrantsSyntaxError(...)`.

hughperkins:
Comments addressed, per Opus:

Screenshot 2026-05-12 at 12 32 57

alanray-tech (approved):
LGTM, ok to merge

PR #692, opened 2026-05-13
[AMDGPU] HIP graph runtime support for @qd.kernel(graph=True)

hughperkins:
## Summary

Ports the basic kernel-launch / instantiate / replay subset of the existing
CUDA graph manager to HIP. Each `@qd.kernel(graph=True)` kernel call now
builds a `hipGraphExec_t` on first invocation and replays it on subsequent
calls — collapsing per-launch overhead the same way the CUDA backend already
does.

Design doc (in [`perso_hugh`](https://github.com/Genesis-Embodied-AI/perso_hugh/blob/main/doc/graph_amd.md)) was iterated through the implementation and reflects the final
shape.

### What's in

- Driver bindings for `hipGraphCreate` / `hipGraphAddKernelNode` /
  `hipGraphInstantiate` / `hipGraphLaunch` / `hipGraphDestroy` /
  `hipGraphExecDestroy` (plus `hipFuncSetAttribute` for shared-memory
  opt-in).
- New `runtime/amdgpu/graph_manager.{h,cpp}`: per-`launch_id` `CachedGraph`
  cache, `resolve_ctx_ndarray_ptrs` for device-only ndarray gating, linear
  kernel-node chaining.
- `runtime/amdgpu/amdgpu_utils.{h,cpp}`: lifts `on_amdgpu_device` to a free
  function so both the launcher and graph manager can share it (mirrors
  `cuda_utils.{h,cpp}`).
- Launcher gates the new path on `ctx.use_graph &&
  ctx.graph_do_while_arg_id < 0` and forwards `get_graph_*` introspection
  accessors so existing tests can introspect cache size / hits / total
  builds the same way they already do on CUDA.

### Two AMDGPU-specific gotchas worth flagging at review

1. **`RuntimeContext` is device-staged, not host-pointer-relied-on.**
   The CUDA graph path passes `&cached.persistent_ctx` (a host pointer) to
   kernels and relies on UVA / HMM. AMDGPU on RDNA3 / Ubuntu doesn't have
   that shortcut by default and the non-graph AMDGPU launcher already
   `hipMalloc`s + `memcpy_host_to_device`s the `RuntimeContext` per launch.
   `CachedGraph` owns a persistent `device_runtime_ctx` buffer and stages
   into it exactly once at graph build (the fields are stable for the
   cached graph's lifetime).
2. **HIP's AMD backend takes kernel args via `extra`, not
   `kernelParams`.** The non-graph AMDGPU launcher already uses the
   `HIP_LAUNCH_PARAM_BUFFER_POINTER` / `BUFFER_SIZE` / `END` byte-buffer
   convention via `hipModuleLaunchKernel`. The graph kernel nodes must use
   the same convention. Passing `kernelParams` instead silently corrupts
   kernel arg loads on RDNA3 and the launched kernels fault asynchronously,
   surfacing as `hipErrorIllegalAddress` at the next host-visible sync. The
   second commit on this branch documents and fixes this — see
   `CachedKernelArgs` in `graph_manager.h` and the `params.extra = ...` /
   `params.kernelParams = nullptr` setup in `add_kernel_node`.

Stream-parallel groups (`stream_parallel_group_id != 0`) are silently
serialized inside the graph, matching the CUDA implementation. A parallel
DAG inside the graph is a possible follow-up.

`graph_do_while` (device-side iteration) is NOT covered by this PR: HIP has
no conditional / while graph nodes as of ROCm 7.2. AMDGPU `graph_do_while`
continues to use the existing host-side loop in
`launch_offloaded_tasks_with_do_while`.

## Test plan

- [x] AMDGPU: `tests/python/test_graph.py` + `tests/python/test_graph_do_while.py` — **24 / 24 passed** on amddesktop (Radeon RX 7900 XTX, `gfx1100` / RDNA3, ROCm 7.2.0).
- [x] AMDGPU regression smoke: `test_ndarray*` + `test_streams.py` — **110 passed, 75 skipped**.
- [x] CUDA regression: `tests/python/test_graph.py` + `tests/python/test_graph_do_while.py` — **24 / 24 passed** on the cluster (RTX PRO 6000 Blackwell).
- [x] User-facing docs updated (`docs/source/user_guide/graph.md`) — renamed "CUDA Graph" → "GPU Graphs", clarified AMDGPU support and `graph_do_while` fall-back behaviour.
- [ ] (deferred) MI300X / `gfx942` verification on `amdcloud` — only RDNA3 / `gfx1100` was hardware-tested here.
- [ ] (deferred) Performance measurement of the AMDGPU graph speed-up; the path is correctness-tested only.

### DIFF
--- a/docs/source/user_guide/graph.md
+++ b/docs/source/user_guide/graph.md
@@ -1,6 +1,8 @@
-# CUDA Graph
+# GPU Graphs (CUDA / HIP)
 
-CUDA graphs reduce kernel launch overhead by capturing a sequence of GPU operations into a graph, then replaying it in a single launch. On non-CUDA platforms, the cuda graph annotation is simply ignored, and code runs normally.
+GPU graphs reduce kernel launch overhead by capturing a sequence of GPU operations into a graph, then replaying it in a single launch. Quadrants supports basic `graph=True` on both CUDA (via CUDA graphs) and AMDGPU (via HIP graphs). On backends without a graph runtime (e.g. CPU, Vulkan), the `graph=True` annotation is silently ignored and code runs normally.
+
+Device-side iteration with `graph_do_while` is currently only fully native on CUDA SM 9.0+ (Hopper); on AMDGPU and older CUDA GPUs it falls back to a host-side loop (see [GPU-side iteration with `graph_do_while`](#gpu-side-iteration-with-graph_do_while) below).
 
 ## Basic usage
 
@@ -30,10 +32,14 @@ my_kernel(x, y)  # first call: builds and caches the graph
 my_kernel(x, y)  # subsequent calls: replays the cached graph
 ```
 
+This works the same way on CUDA and AMDGPU. The cache is keyed per (compiled-kernel-specialization, launch-id), so different template instantiations (different field bindings, etc.) get their own cached graph.
+
 ### Restrictions
 
-- **No struct return values.** Kernels that return values (e.g. `-> qd.i32`) cannot use CUDA graphs. An error is raised if `graph=True` is set on such a kernel.
+- **No struct return values.** Kernels that return values (e.g. `-> qd.i32`) cannot use GPU graphs. An error is raised if `graph=True` is set on such a kernel.
 - **Primal kernels only.** The `graph=True` flag is applied to the primal (forward) kernel only, not its adjoint. Autodiff kernels use the normal launch path.
+- **Device-resident ndarrays.** Graph mode bakes device pointers into the cached graph, so all ndarray arguments must be on the GPU. Passing a host-resident ndarray raises an error.
+- **`qd_stream` is incompatible** with `graph=True`. Choose one or the other.
 
 ### Passing different arguments
 
@@ -77,8 +83,8 @@ solve(x, counter)
 
 The argument to `qd.graph_do_while()` must be the name of a scalar `qd.i32` ndarray parameter. The loop body repeats while this value is non-zero.
 
-- On SM 9.0+ (Hopper), this uses CUDA conditional while nodes — the entire iteration runs on the GPU with no host involvement.
-- On older CUDA GPUs and non-CUDA backends, it falls back to a host-side do-while loop.
+- On CUDA SM 9.0+ (Hopper), this uses CUDA conditional while nodes — the entire iteration runs on the GPU with no host involvement.
+- On older CUDA GPUs, AMDGPU, and non-GPU backends, it falls back to a host-side do-while loop (see [Caveats](#caveats)). HIP does not currently expose conditional / while graph nodes, so AMDGPU cannot run the loop fully on the device today.
 
 ### Patterns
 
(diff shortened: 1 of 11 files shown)


docs/source/user_guide/graph.md
```diff
@@ -39,10 +41,14 @@ my_kernel(x, y)  # first call: builds and caches the graph
 my_kernel(x, y)  # subsequent calls: replays the cached graph
 ```
 
+This works the same way on CUDA and AMDGPU. The cache is keyed per (compiled-kernel-specialization, launch-id), so different template instantiations (different field bindings, etc.) get their own cached graph.
+
 ### Restrictions
 
 - **No struct return values.** Kernels that return values (e.g. `-> qd.i32`) cannot use graphs. An error is raised if `graph=True` is set on such a kernel.
 - **Primal kernels only.** The `graph=True` flag is applied to the primal (forward) kernel only, not its adjoint. Autodiff kernels use the normal launch path.
+- **Device-resident ndarrays.** Graph mode bakes device pointers into the cached graph, so all ndarray arguments must be on the GPU. Passing a host-resident ndarray raises an error.
```
duburcqa, line 50: Doing this is not even possible no? Like if you use AMDGPU backend, you cannot allocate ndarray on the CPU?

docs/source/user_guide/graph.md
```diff
@@ -137,12 +143,14 @@ However, other parameters can be any supported Quadrants kernel parameter type.
 
 ### Caveats
 
-On currently unsupported GPU platforms, such as AMDGPU at the time of writing, the value of the `graph_do_while` parameter will be copied from the GPU to the host each iteration, in order to check whether we should continue iterating. This causes a GPU pipeline stall. At the end of each loop iteration:
+On platforms without native device-side conditional graph nodes — currently CUDA pre-SM 9.0 and **AMDGPU** (HIP has no conditional / while node API as of ROCm 7.2) — the value of the `graph_do_while` parameter will be copied from the GPU to the host each iteration, in order to check whether we should continue iterating. This causes a GPU pipeline stall. At the end of each loop iteration:
```
duburcqa, line 146: The wording "GPU pipeline stall" is a bit misleading. Like, it will exhaust the queue, copy from device to host, and resume work. Not exactly what I would call stall. But maybe I'm getting it work.

hughperkins: The stall is diagrammed here:

https://github.com/hughperkins/cltorch#point-tensors-reduce-pipeline-stalls

Screenshot 2026-05-13 at 18 05 21

hughperkins: (written before torch had scalar tensors :) )

docs/source/user_guide/graph.md
hughperkins: For example, a numpy tensor is on the host.

duburcqa: Hum.. Ok I see. Not very limiting.

docs/source/user_guide/graph.md
duburcqa: Yeah sure. Not as bad as I would expect when I read "stale".

Could be nice to clarify where is the "stale" time in your linear bullet point enumeration.

duburcqa:
ok to merge.

docs/source/user_guide/graph.md
hughperkins: It's pretty bad otherwise:
1. we wouldnt have spent several hours one day discussing how to handle dynamic iterations of the solver without causing exactly this sync/stall
2. I wouoldnt have implemented conditional graphs in the first place

What are you imaginging I write, concretely?

duburcqa: I see. Ok so if it is this bad. The current framing is just right. Nothing to change. Maybe just stating explicitly what you just said, that "On platforms without native device-side conditional graph nodes", using graph in such a case is probably pointless and not really helping?

hughperkins: Using graph is still useful, probably, on platforms without conditional nodes. But it wont be as fast as on a platform with conditional nodes.

There are three levels basically:
- no graph, no conditional
    - kernels launched by c++ on host, as if we hadnt said it was using a graph at all
         - if the gpu runtime is sufficient, this will hide any launch latency
    - if there is a graph_while loop:
        - loop evaluated on the host, causes the pipeline stall as above
        - i.e. loss of hiding of launch latency
- graph, no conditional (CUDA, AMD)
     - if there is no graph_while loop:
         - all kernels launched with a single host-side command, run on gpu with no further host-side command
     - if there is a graph_while loop:
         - loop evaluated on the host, causes the pipeline stall as above
         - all kernels within each iteration launched by a single host-side command
- graph, with conditional (CUDA SM90+)
     - entire quadrants kernel contents run entirely on gpu, including loop, and conditional evaluation
         - including any graph_while loop

PR #693, opened 2026-05-13
[GPU] New device-level ops for QIPC

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


tests/python/test_algorithms.py
```diff
@@ -0,0 +1,1532 @@
...
+
+def _is_64bit_unsupported_on_current_arch(dtype):
+    """Return True iff ``dtype`` is an 8-byte type the current backend can't represent.
+
+    Metal has no f64 / i64 / u64 in buffer-backed I/O at all; Vulkan-on-Darwin (MoltenVK) has neither f64 nor 64-bit
+    integers. CUDA / Vulkan-on-Linux / AMDGPU support all three. The device-tier algorithms inherit the underlying
+    lang's dtype support.
+    """
+    arch = qd.lang.impl.current_cfg().arch
+    if dtype in (qd.i64, qd.u64) and arch == qd.metal:
```
hughperkins: this is super hard to read. lets group by system, ie arch and platform, then underneath each arch/platform pair, state the supported dtypes

tests/python/test_algorithms.py
```diff
@@ -0,0 +1,1532 @@
...
+    if dtype in (qd.i64, qd.u64) and arch == qd.vulkan and platform.system() == "Darwin":
+        return True
+    if dtype == qd.f64 and arch == qd.metal:
+        return True
+    if dtype == qd.f64 and arch == qd.vulkan and platform.system() == "Darwin":
+        return True
+    return False
+
+
+def _skip_if_64bit_unsupported(dtype):
```
hughperkins: why do we have both this and _is_64bit_unsupported_on_current_arch ?

tests/python/test_algorithms.py
```diff
@@ -0,0 +1,1532 @@
...
+        return
+    if dtype in (qd.i64, qd.u64):
+        pytest.skip(f"64-bit integer type {dtype} not supported on the current backend")
+    pytest.skip(f"f64 not supported on the current backend")
+
+
+def _is_apple_gpu_backend():
+    """Metal or MoltenVK (Vulkan-on-Darwin)."""
+    arch = qd.lang.impl.current_cfg().arch
+    if arch == qd.metal:
```
hughperkins: one line

return arch == qd.metal or (arch == qd.vulkan and platform.system() == "Darwin")

tests/python/test_algorithms.py
```diff
@@ -0,0 +1,1532 @@
...
+        return True
+    if arch == qd.vulkan and platform.system() == "Darwin":
+        return True
+    return False
+
+
+def _skip_if_radix_sort_large_n_on_apple_gpu(N):
```
hughperkins, line 133: why are we skipping large n on apple gpu?

tests/python/test_algorithms.py
```diff
@@ -0,0 +1,1532 @@
...
+
+# ---------------------------------------------------------------------------
+# Device select / compact
+# ---------------------------------------------------------------------------
+
+# Sizes that exercise: single-block path, two-pass path with even split,
+# off-by-one tile (B0 ends mid-block), three-pass recursion. Default 5 MB
+# scratch holds N + ceil(N/256) + ... <= ~1.3M u32 slots, so the largest
+# size below (200_000) fits comfortably with ~780-slot partials.
+_SELECT_SIZES = [1, 7, 255, 256, 257, 1023, 1024, 1025, 65536, 65537, 200_000]
```
hughperkins: lets move all these various globals to the top of the module

tests/python/test_algorithms.py
```diff
@@ -0,0 +1,1532 @@
...
+        # without papering over actual algorithmic regressions. The two asserts below pin the contract in
+        # plain-language form so a reader doesn't have to crunch sqrt(N) in their head: at <= 100 elements we still
+        # demand < 0.1% rel; at <= 100K elements < 1% rel; the qipc hot-path N=1M lands at ~2% rel. ``atol=1e-3``
+        # only kicks in for cumsum values near zero (the first handful of elements), where rtol * |desired| would
+        # underflow.
+        if dtype == qd.f32:
+            rtol = 2e-5 * math.sqrt(N)
+            assert rtol <= 1e-3 or N > 100, f"f32 scan rtol={rtol:g} too loose for small N={N}"
+            assert rtol <= 1e-2 or N > 100_000, f"f32 scan rtol={rtol:g} too loose for medium N={N}"
+            atol = 1e-3
```
hughperkins: lets move all definitions or calculations of rtol or atol to the top of the module
- global constants as global constnats
- formula as reusable formula for atol and rtol

python/quadrants/algorithms/_select.py
```diff
@@ -0,0 +1,194 @@
...
+    """Stream-compact ``arr`` by ``flags``: copy ``arr[i]`` to a dense prefix of ``out`` for every ``i`` where
+    ``flags[i] != 0``, in stable input order. Write the count of selected elements to ``num_out[0]``.
+
+    Args:
+        arr: 1-D tensor of any element dtype that Quadrants supports field-element assignment for: scalars
+            (``i32`` / ``u32`` / ``f32`` / ``i64`` / ``u64`` / ``f64``) and structs (``qd.Struct.field({...})`` or
+            ``qd.types.struct(...)`` - e.g. the libuipc ``Vector{2,3,4}i`` shapes). The scatter is
+            ``dst[idx] = src[i]``, which lowers per-field for struct dtypes, so no scratch reinterpretation is
+            needed for wider / composite element types.
+        flags: 1-D ``i32`` tensor, same shape as ``arr``. Each entry is treated as a boolean (``!= 0`` selects).
```
alanray-tech: The docstring says `!= 0` selects, but the implementation prefix-sums the raw `i32` flag values as counts â€” so `flags = [2, 0, 1]` would produce scanned indices `[0, 2, 2]`, giving wrong output and `num_out = 3` instead of 2.

Suggestion: tighten the contract to require strictly 0/1 flags and document it as such (both here and in `algorithms.md`). No need to add a normalization pass â€” just make it the caller's responsibility.

Proposed wording for the `flags` arg:

> **Every entry must be exactly `0` or `1`** (`1` selects). Non-0/1 values produce incorrect results. Caller-built â€” populate with a kernel applying your predicate, writing `1` for selected and `0` otherwise.

Same update needed in `algorithms.md` (overview table + constraints bullet).

python/quadrants/algorithms/_radix_sort.py
```diff
@@ -0,0 +1,503 @@
...
+    # needs roughly `hist_len / BLOCK_DIM + ...` extra u32 slots - well under hist_len / 100.
+    if hist_len + (hist_len + BLOCK_DIM - 1) // BLOCK_DIM > cap:
+        needed = hist_len + (hist_len + BLOCK_DIM - 1) // BLOCK_DIM
+        raise RuntimeError(
+            f"device_radix_sort on N={N} needs >= {needed} u32 scratch slots "
+            f"({needed * 4} bytes), but only {cap} are configured "
+            f"({cap * 4} bytes). Call quadrants._scratch.set_scratch_bytes(...) "
+            f"before any algorithm runs to raise the cap. For N=1M expect to "
+            f"need ~5 MB; for N=10M ~50 MB."
+        )
```
alanray-tech, line 459: This capacity check only accounts for one level of scan partials (`hist_len + ceil(hist_len / 256)`), but `_exclusive_scan_inplace_u32(..., n=hist_len, partials_cursor=hist_len)` recursively needs deeper levels too (e.g. `+ ceil(ceil(hist_len/256)/256)` and so on).

If a caller sets scratch to exactly the size reported here for a large signed/f32 sort, the later recursive scan check raises *after* `_twiddle_pass` has already transformed `keys` in place, leaving the user's data corrupted with no way to recover.

Suggestion: compute the full recursive scan footprint before accepting the call. Something like:

```python
def _scan_total_scratch(n: int, partials_cursor: int) -> int:
    total = partials_cursor
    while n > BLOCK_DIM:
        B = (n + BLOCK_DIM - 1) // BLOCK_DIM
        total = partials_cursor + B
        partials_cursor = total
        n = B
    return total
```

Then use `needed = _scan_total_scratch(hist_len, hist_len)` here, and crucially this check already runs before `_twiddle_pass` which is correct â€” just the footprint calculation needs the recursive accounting.

hughperkins: addressed in [ee41c47](https://github.com/Genesis-Embodied-AI/quadrants/pull/693/commits/ee41c473ead1bcaba21db44b04264d770d802777)

python/quadrants/algorithms/_select.py
hughperkins: addressed in https://github.com/Genesis-Embodied-AI/quadrants/pull/693/commits/de22f8cdd06ea17c6800954e06c094ac41b1293d

alanray-tech (approved):
LGTM

hughperkins:
Great! 🙌

PR #697, opened 2026-05-14
[Lang] @qd.func with dataclass annotation accepts dataclass values with extra fields (structural subset)

duburcqa:
# Accept structurally-wider dataclasses at `@qd.func` call sites

> Lets a `@qd.func` annotated with a `FooSmall` parameter be called with a `FooBig` value as long as `FooBig` contains every field `FooSmall` declares. Extra fields on the caller side are silently dropped.

## TL;DR

When a kernel holds a value of dataclass type `FooBig` and calls a `@qd.func` declaring `foo: FooSmall`, the call site previously flattened the argument by the caller-side type (emitting every field of `FooBig`). The callee's `arg_metas_expanded` is built from `FooSmall` and therefore expects fewer positional / keyword args, producing `QuadrantsSyntaxError: Too many arguments` at compile time.

The fix narrows the call-site flattening to the callee's declared annotation when one is available, so only the fields the callee actually expects are emitted.

## Why

Practical templating: callers often share a fat dataclass (full state) but downstream `@qd.func`s only need a slice. Re-typing the dataclass at every boundary forced users to either restructure code or manually unpack fields. The flattening was already field-name-based; the change makes positional / keyword flattening respect the declared parameter shape instead of the actual value shape.

## Mechanism

`call_transformer.build_Call` now resolves the callee's declared `arg_metas` (skipping the implicit `self` for `BoundQuadrantsCallable`) and threads them into:

- `_expand_Call_dataclass_args(callee_annotations=...)` - per-positional override
- `_expand_Call_dataclass_kwargs(callee_annotations_by_name=...)` - per-keyword override

When the override at a given position / name is itself a dataclass type, field iteration uses that type rather than the caller-side `val`. Nested dataclasses recurse with the callee's nested `field.type`, so deeper subtyping works too. When no override is present (legacy callers / non-`QuadrantsCallable` targets), behaviour is unchanged.

## Tests

`tests/python/test_py_dataclass.py::test_func_accepts_dataclass_with_extra_fields` covers the positional, keyword, and same-shape (`FooSmall -> FooSmall`) paths. Verified the test fails on the base branch with the original `Too many arguments` error and passes after the fix; the full `test_py_dataclass.py` + AST-transformer suites (174 tests) stay green.

### DIFF
--- a/python/quadrants/lang/ast/ast_transformers/call_transformer.py
+++ b/python/quadrants/lang/ast/ast_transformers/call_transformer.py
@@ -166,20 +166,31 @@ class CallTransformer:
 
     @staticmethod
     def _expand_Call_dataclass_args(
-        ctx: ASTTransformerFuncContext, args: tuple[ast.stmt, ...]
+        ctx: ASTTransformerFuncContext,
+        args: tuple[ast.stmt, ...],
+        callee_annotations: tuple[Any, ...] | None = None,
     ) -> tuple[tuple[ast.stmt, ...], tuple[ast.stmt, ...]]:
         """
         We require that each node has a .ptr attribute added to it, that contains
-        the associated Python object
+        the associated Python object.
+
+        callee_annotations, when provided, supplies the declared parameter type for each positional
+        arg. If the callee declares a dataclass that is a structural subset of the caller-side
+        value, iteration is driven by the callee's fields so extra caller-side fields are dropped.
         """
         args_new = []
         added_args = []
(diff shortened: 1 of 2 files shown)


hughperkins:
Ideally, I think it would be good to strive for structures in Quadrants that look as similar to Python as possible, and are theoretically pyright compliant. We are aiming to make a compiler for python, that transforms python into code that can run on GPUs. We are not striving to invent a new language I feel.

PR #700, opened 2026-05-15
[block] sync(): fix unsupported-arch error message

### LINKED ISSUE #646: [Bug] sync() error message incorrectly says shared_array
## Summary

`block.sync()` raises a `ValueError` with an incorrect function name in the error message.

## Details

In `python/quadrants/lang/simt/block.py`, line 19:

```python
def sync():
    arch = impl.get_runtime().prog.config().arch
    if arch == _qd_core.cuda or arch == _qd_core.amdgpu:
        return impl.call_internal("block_barrier", with_runtime_context=False)
    if arch_uses_spv(arch):
        return impl.call_internal("workgroupBarrier", with_runtime_context=False)
    raise ValueError(f"qd.block.shared_array is not supported for arch {arch}")
```

The error message says `qd.block.shared_array` but it should say `qd.simt.block.sync` (or at minimum `qd.block.sync`).

## Suggested fix

```python
raise ValueError(f"qd.simt.block.sync is not supported for arch {arch}")
```

## Found during

Review of #638 (docs for `qd.simt.block.*`).

hughperkins:
## Summary

`block.sync()` was raising `qd.block.shared_array is not supported for arch ...` when the arch wasn't CUDA / AMDGPU / Vulkan / Metal -- a copy-paste leftover from `SharedArray`. Switch to `qd.block.sync` so the message matches the function and follows the convention used by every other primitive in `block.py` (`sync_all_nonzero`, `mem_fence`, `thread_idx`, `global_thread_idx`, ...).

Closes #646.

## Test plan

- Code-only diff in an unreachable `raise ValueError` path; no behaviour change on supported archs.

### DIFF
--- a/python/quadrants/lang/simt/block.py
+++ b/python/quadrants/lang/simt/block.py
@@ -38,7 +38,7 @@ def sync():
         return impl.call_internal("block_barrier", with_runtime_context=False)
     if arch_uses_spv(arch):
         return impl.call_internal("workgroupBarrier", with_runtime_context=False)
-    raise ValueError(f"qd.block.shared_array is not supported for arch {arch}")
+    raise ValueError(f"qd.block.sync is not supported for arch {arch}")
 
 
 def sync_all_nonzero(predicate):


alanray-tech (approved):

hughperkins:
🙌

PR #702, opened 2026-05-15
[volatile_load] add qd.volatile_load primitive (closes #648)

### LINKED ISSUE #648: [Feature] Add volatile_load primitive for spin-wait patterns
## Summary

Quadrants has no volatile load or atomic load primitive. This makes it impossible to correctly and efficiently implement spin-wait patterns (e.g. decoupled-look-back scans) that require repeatedly reading a memory location until it changes.

## Problem

A spin-wait loop like:

```python
while flags[prev] == STATE_INVALID:
    pass
```

relies on the compiler re-reading `flags[prev]` from global memory on each iteration. Without a volatile load, the compiler may hoist the load out of the loop or cache the value in a register, turning the spin into an infinite loop.

Current workarounds are all suboptimal:

| Workaround | Correct? | Performance |
|-----------|----------|-------------|
| `grid.memfence()` inside the loop | Yes (acts as compiler barrier) | Bad — full device-scope cache drain per iteration |
| `atomic_add(flags[prev], 0)` | Yes (forces memory round-trip) | Bad — read-modify-write overhead, contention |
| Do nothing and hope the compiler doesn't optimize | Fragile | N/A |

## Proposed solution

Add a `qd.volatile_load(target, *indices)` primitive (or equivalent) that guarantees the load goes to memory on every call. The implementation maps cleanly to existing backend primitives:

- **LLVM IR** (CUDA / AMDGPU): emit `load volatile` instead of `load` — LLVM guarantees it cannot be eliminated, merged, or hoisted. On CUDA, LLVM lowers this to `ld.volatile.global` in PTX.
- **SPIR-V** (Vulkan / Metal): emit `OpLoad` with the `Volatile` Memory Access bit (`0x1`) — prevents expression forwarding and value caching.

No new hardware capability is needed; every backend already supports this at the instruction level.

## Use cases

- Decoupled-look-back scans (Onesweep-style) — spin on a flag array
- Producer-consumer patterns between blocks via global memory
- Any kernel that polls a shared memory location written by another thread/block

## Found during

Review of #641 (docs for `qd.simt.grid.*`), where the lookback_scan example relies on re-reading `flags[prev]` in a spin loop but has no way to guarantee it.

hughperkins:
## Summary

Adds `qd.volatile_load(target)`, a top-level frontend primitive (alongside `qd.atomic_*`) that maps to a `GlobalLoadStmt` carrying a new `is_volatile` flag.  Required for spin-wait correctness: without it, a loop like

```python
while flags[prev] == STATE_INVALID:
    pass
```

is undefined at the LLVM-IR / SPIR-V level (the optimiser may hoist the load and turn the spin into an infinite loop).  The portable workarounds (`grid.mem_fence()` inside the loop, `atomic_add(flags, 0)`) are correct but pay an order-of-magnitude perf tax over a real volatile read.

Closes #648.

## Codegen, all four backends

* **CUDA / AMDGPU** -- LLVM `LoadInst::setVolatile(true)`, lowering to `ld.volatile.global` (PTX) / unhoistable `global_load_*` (AMDGPU).  CUDA additionally suppresses the `__ldg` / `!invariant.load` read-only-cache fast path for volatile loads (the two are contradictory).
* **Vulkan / Metal** -- SPIR-V `OpLoad` with the `Volatile` `MemoryAccess` mask via a new `IRBuilder::load_variable_volatile` helper.  Per-load opt-in on top of the existing Metal-global `use_volatile_buffer_access_` flag.

## Transforms guarded

* `cache_loop_invariant_global_vars` -- skip volatile loads (would hoist).
* `simplify` -- skip the redundant-load search for volatile loads (the cell may have been written by another thread / block since the prior load).

## Design choices

* **Top-level `qd.volatile_load`** rather than `qd.simt.grid.volatile_load` -- matches `qd.atomic_*` placement and is consistent with how callers will pair the two ops in spin-wait code.
* **Flag on `GlobalLoadStmt`** rather than a dedicated `VolatileGlobalLoadStmt` -- minimal blast radius.  Only two transforms (`cache_loop_invariant`, `simplify`) and three codegen backends needed to learn the flag; everything else handles `is_volatile=false` (the default) unchanged.
* **`volatile_store` deferred** -- the issue is read-only and can be paired in a follow-up PR if Alan / maintainers want symmetry.  No commitment either way -- happy to add it here on request.

## Test plan

`tests/python/test_volatile_load.py` (9 tests, all pass on the cluster on x64 / cuda / vulkan):

* `test_volatile_load_basic_value` -- functional smoke on every backend (catches gross plumbing bugs).
* `test_volatile_load_emits_volatile_marker` -- IR-level: asserts `global load volatile` appears in dumped IR on every backend (the primary signal that `is_volatile` survived the transform pipeline and reached codegen).
* `test_volatile_load_survives_loop_invariant_caching` -- regression: kernel where a non-volatile load would be a hoist target, asserts the value is read correctly.
* `test_volatile_load_rejects_non_lvalue` / `test_volatile_load_rejects_bare_field` -- frontend guards mirror `qd.atomic_cas`.

Sanity-checked the broader `test_atomic / test_simt / test_offline_cache` suite on the cluster: 1183 pass, 18 fail.  All 18 failures reproduce on `main` without these changes (pre-existing f16-precision flake on `test_atomic_float_ops[cuda-f16-sub]` -- 3-of-5 attempts fail on main; Vulkan `shaderSubgroupExtendedTypes` validation issues on 64-bit subgroup ops -- correlates with the cluster Vulkan driver's missing extension, unrelated to this change).

Cluster GPU coverage: x64 + cuda + vulkan (cluster has no AMDGPU / Metal hardware).  AMDGPU lowering is identical to CUDA via the shared `TaskCodeGenLLVM::create_global_load`; Metal goes through the SPIR-V path tested on Vulkan plus the same `MemoryAccessVolatileMask` lowering MoltenVK already consumes.


alanray-tech (approved):
LGTM — clean design, well-documented, good test coverage. No concerns.

PR #704, opened 2026-05-16
[DataOriented] Fix ndarrays on data oriented

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


docs/source/user_guide/compound_types.md
```diff
@@ -5,21 +5,19 @@
 It can be useful to combine multiple ndarrays or fields together into a single struct-like object that can be passed into kernels, and into @qd.func's.
 
 The following compound types are available:
-- `dataclasses.dataclass` — **recommended**
-- `@qd.data_oriented` — for classes that define `@qd.kernel` methods, cannot contain ndarrays
-- `@qd.struct` / `@qd.dataclass` — legacy, field-only
+- `@dataclasses.dataclass` — lightweight container of tensors and primitives; can contain ndarrays
+- `@qd.data_oriented` — for creating objects with `self` that define `@qd.kernel` methods
```
duburcqa: That is a weird description I feel. 

In my view, it unlocks object-oriented programming with Quadrants by allowing decorating instance method with `@qd.kernel`, which avoids passing all tensors either as input arguments or global fields.

docs/source/user_guide/compound_types.md
```diff
@@ -5,21 +5,19 @@
...
+- `@qd.dataclass` — for structures that are embedded into the kernel, and don't contain ndarrays
```
duburcqa, line 10: > for structures that are embedded into the kernel, and don't contain ndarrays

Could be nice to explain how to choose one or the other, beyond strict description.

docs/source/user_guide/compound_types.md
hughperkins: Yes I agree. Will rephrase following your proposal.

docs/source/user_guide/compound_types.md
```diff
@@ -5,21 +5,19 @@
...
-|------------------------------------|:---------------------------:|:-------------------------:|:--------------------:|:------------------:|:--------------:|:-------------------------:|
-| `dataclasses.dataclass`            | yes                         | yes                       | yes                  | yes                | yes            | no [*1]                   |
-| `@qd.data_oriented`               | yes                         | yes                       | no                   | yes                | yes            | yes                       |
-| `@qd.struct`, `@qd.dataclass`     | yes                         | yes                       | no                   | yes                | yes            | yes                       |
+| property                            | `@dataclasses.dataclass`     | `@qd.data_oriented`              | `@qd.dataclass`                     |
+|-------------------------------------|:----------------------------:|:--------------------------------:|:-----------------------------------:|
+| Kernel-side representation          | none (flattened away)        | none (flattened away)            | real type with fixed memory layout  |
```
duburcqa: What "Kernel-side representation" means exactly?

docs/source/user_guide/compound_types.md
```diff
@@ -5,21 +5,19 @@
...
+| Can be tensor element type          | no                           | no                               | yes                                 |
```
duburcqa: "Can be used as tensor element type" sounds more natural to me, but maybe it is just me.

docs/source/user_guide/compound_types.md
```diff
@@ -123,10 +121,14 @@ Sub-struct passing supports:
...
 - positional and keyword call sites (`f(s.inner)` and `f(inner=s.inner)`)
 - call sites both directly inside `@qd.kernel` bodies and inside other `@qd.func` bodies
-- pruning of the sub-struct's leaf fields that the callee never reads
+- pruning of the sub-struct's leaf members that the callee never reads
 
 Note: assigning a sub-struct to a local variable and then passing it (`t = s.inner; touch_inner(t)`) is **not** supported. Pass the attribute access directly at the call site.
 
+### Under the hood
+
+A `dataclasses.dataclass` is a Python-only container. The compiler reads it at compile time and flattens its members into individual kernel parameters — the container itself has no memory layout and doesn't exist on the kernel side. That's why members are read-only: the values are captured at compile time and re-assigning them afterwards has no effect on running kernels.
```
duburcqa: "That's why members are read-only: the values are captured at compile time and re-assigning them afterwards has no effect on running kernels."

What this means is not very clear. Like, the tensor members are read-write, but you cannot overwrite the member itself?

docs/source/user_guide/compound_types.md
```diff
@@ -148,14 +150,146 @@ sim.step()
...
+
+### Under the hood
+
+Like `dataclasses.dataclass`, a `@qd.data_oriented` object is Python-only — the compiler flattens it into individual kernel parameters and the object itself has no kernel-side representation. Unlike `dataclasses.dataclass` it needs no member annotations: the compiler reads the live instance's attributes directly. Primitive members are baked into the kernel as constants, so each distinct primitive value compiles a new specialised kernel.
+
+## qd.dataclass / qd.types.struct
+
+Unlike `@qd.data_oriented` and `@dataclasses.dataclass`, `@qd.dataclass` creates a struct that is available within the kernels themselves. The former types are only used for structure on the python side, before compilation. `@qd.dataclass` can be used as the element type of fields. One key downside of `@qd.dataclass` is that they can only be used with fields and primitives, not with ndarray. This is because tensors are embedded in the struct by value, not as a reference pointer.
```
duburcqa: "This is because tensors are embedded in the struct by value, not as a reference pointer."

Can you elaborate?

docs/source/user_guide/compound_types.md
```diff
@@ -148,14 +150,146 @@ sim.step()
...
+| Annotation | Kernel-arg walker | Notes |
+|---|---|---|
+| `qd.types.NDArray[...]`           | ndarray slot                                       | leaf-level only |
+| `MyDataclass` (dataclass type)    | per-member flatten using annotations               | needs every member to have a quadrants-typed annotation |
+| `qd.template()`                   | value-driven walk of `vars(self)` / dataclass members | supports the full nesting matrix above |
+
+Two practical consequences:
+
+- **Containers with `@qd.data_oriented` anywhere in the tree** must be passed via `qd.template()` (or be the `self` of a `@qd.kernel` method on a `@qd.data_oriented` class). Using a typed-dataclass annotation on the outermost arg errors.
+- **A non-frozen `dataclasses.dataclass`** can be passed via the typed-dataclass annotation, but cannot be the outer `qd.template()` arg — `qd.template()` uses the instance as a dict key inside the template-mapper and a non-frozen dataclass has `__hash__ = None`. Add `frozen=True` if you need to pass it as `qd.template()` (for example, when it holds `@qd.data_oriented` children).
```
duburcqa: "—" on purpose?

duburcqa: It is a bit late to introduce "frozen".  It should be moved in "## dataclasses.dataclass".

docs/source/user_guide/compound_types.md
```diff
@@ -148,14 +150,146 @@ sim.step()
...
+
+### Reassigning ndarray members
+
+For both `dataclasses.dataclass` and `@qd.data_oriented` containers passed via `qd.template()`, reassigning an ndarray member between kernel launches is supported, including changes to `dtype`, `ndim`, or layout. A new specialised kernel is compiled and cached for the new shape; subsequent launches with the original shape continue to use the original cached kernel.
```
duburcqa: Are you 100% sure about this? I think we have a shortcut for frozen dataclasses that does not allow  reassigning an ndarray member between kernel launches, which anyway is forbidden by frozen.

docs/source/user_guide/compound_types.md
```diff
@@ -148,14 +150,146 @@ sim.step()
...
+
+### Restrictions
+
+- **`@qd.dataclass` (the Quadrants `StructType` decorator) cannot contain ndarrays.** This is a legacy field-only type. Use `dataclasses.dataclass` or `@qd.data_oriented` instead. (The function-form factory `qd.types.struct(...)` produces the same `StructType` and has the same restrictions.)
```
duburcqa: > This is a legacy field-only type

Really? it is considered legacy? I think it has use-cases, just not what we want in Genesis because it is field-only.

docs/source/user_guide/compound_types.md
```diff
@@ -148,14 +150,146 @@ sim.step()
...
+- **A typed-dataclass kernel-arg annotation cannot have a `@qd.data_oriented` member type** (see [\*1] above) — errors clearly at compile time.
+- **An outer `qd.template()` arg of dataclass type must be `frozen=True`** — non-frozen dataclasses are unhashable and the template-mapper cannot use them as cache keys.
```
duburcqa: You are repeating yourself.

docs/source/user_guide/compound_types.md
```diff
@@ -148,14 +150,146 @@ sim.step()
...
+- **Declare all ndarray members on a `@qd.data_oriented` class in `__init__`.** The template-mapper caches the set of ndarray-attribute paths reachable from the first instance walked, per class. Adding *new* ndarray attributes on later instances of the same class is safe — the per-instance weakref in the spec key disambiguates them, and the compile-time walker registers all reachable ndarrays. But:
```
duburcqa, line 323: That is more or less unsupported in standard Python. Any type checker would throw an error. But it is not enforced by the langage itself.

docs/source/user_guide/compound_types.md
hughperkins: Yes, fair. Apparently we are using it now for `qd.unpacked_array` ...

docs/source/user_guide/fastcache.md
```diff
@@ -172,3 +142,31 @@ print(obs.cache_stored)         # True if the compiled kernel was stored to cach
...
+
+**`@qd.data_oriented`:** the walker descends into `vars(obj)`. For each child:
+
+- `qd.ndarray` member — `(dtype, ndim, layout)` is included in the cache key. Element values are not.
+- Primitive (`int` / `float` / `bool` / `enum.Enum`) member — value is baked into the kernel (same semantics as a `qd.Template` primitive). Two instances of the same class with different primitive member values get different cache entries.
+- Nested `@qd.data_oriented` member — recurses.
+- Nested `dataclasses.dataclass` member — recurses (with the dataclass rules below).
+- `qd.field` member — fastcache is disabled for the entire kernel call. The kernel still runs via normal compilation; a warn-level log line is emitted.
+
+**`dataclasses.dataclass`:** the walker descends into the declared members. For each member, only the *type* is included in the cache key by default — **not** the value. To include a member's value, annotate it:
```
duburcqa, line 160: This section is great, but should probably be moved in Appendix, because it is not supposed to be relevant for users.

docs/source/user_guide/compound_types.md
hughperkins: It's actually not clear to me. I think the objective description is clearest for now.

I suppose, in my head, currently, the way I see the world, personally:
- data_oriented: use this for classes that contain kernels, similar to standard classes with methods in python-land
- dataclasses.dataclass: use this for dataclasses, that are used by the data_oriented classes, similar to how dataclasses are used to hold data in python-land
- qd.dataclass: I almost never use these to be honest. The fields-only limitation is quite serious for us. The only time I've used these lately is for the Tiles, which only exist inside the kernel. No access to them from outside the kernel

I could write this out, but it's *very* subjective. I think objective is the way to go.

docs/source/user_guide/compound_types.md
hughperkins: Like, inside the GPU kernel, rather than in the uncompiled python code.

For example, in datatclasses.dataclass, in the uncompiled python code, it looks like a normal python dataclass, that one can pass around.

In reality, it doesnt even exsit in the compiled kernels. It's broken up into individual kernel parameters, which are passed in individually, into the kernel, and are used as individual ndarrays etc, inside the kernel. The kernel doesn't "know" about the dataclasses.dataclass at all. Doesnt exist inside the kernel.

docs/source/user_guide/compound_types.md
hughperkins: is this a comment about non-ascii?

docs/source/user_guide/compound_types.md
hughperkins: what do you mean by 'this' here? assigning things in `__init__` or assigning things outside of `__init__`?

docs/source/user_guide/compound_types.md
```diff
@@ -123,10 +121,32 @@ Sub-struct passing supports:
...
-- pruning of the sub-struct's leaf fields that the callee never reads
+- pruning of the sub-struct's leaf members that the callee never reads
 
 Note: assigning a sub-struct to a local variable and then passing it (`t = s.inner; touch_inner(t)`) is **not** supported. Pass the attribute access directly at the call site.
 
+### Frozen vs non-frozen
+
+A `dataclasses.dataclass` may be either non-frozen (the default) or frozen (`@dataclass(frozen=True)`). For most cases you don't need `frozen=True`: a regular dataclass can be passed to a kernel through a typed-dataclass annotation (`def k(s: MyStruct)`) without any restriction.
+
+`frozen=True` becomes necessary when you want to pass the dataclass through `qd.template()`. `qd.template()` uses the instance as a dict key inside the template-mapper, and a non-frozen dataclass has `__hash__ = None`, so the template-mapper can't use it as a cache key. The typical case where you need this is when the dataclass holds a `@qd.data_oriented` child — those must reach the kernel via `qd.template()` (see [Nesting compatibility](#nesting-compatibility)).
```
hughperkins: does frozen=True also improve hot path time?

docs/source/user_guide/compound_types.md
hughperkins: modified the row heading (and added a bunch of new rows)

docs/source/user_guide/compound_types.md
hughperkins: I'm not sure I see the contradiction?
- frozen=True is faster (which I've documented above now, in the frozen vs non-frozen section header)
- frozen=False will also work (just slower)

docs/source/user_guide/compound_types.md
duburcqa: What about some advice / recommendation section? I understand that these suggestions are subjective, but as a new comer that knows nothing about Quadrants, I would like to have some advice about what to use. Otherwise you may feel a bit lost. if it is already hard to tell for you, imagine someone that knows nothing about Quadrants x) But definitely it does not have to be part of this section, maybe some FAQ I don't know?

docs/source/user_guide/compound_types.md
```diff
@@ -5,21 +5,26 @@
...
 
-**Use `dataclasses.dataclass` for new code.** It supports both fields and ndarrays, can be nested, and uses standard Python — no Quadrants-specific decorator needed.
+> ⚠️ **Deprecation: `@dataclasses.dataclass` instance passed via `qd.Template`.**
+> Passing a `@dataclasses.dataclass` instance into a `qd.Template`-annotated kernel parameter was never intended to be supported, and only works inadvertently. As of this release the combination emits a `DeprecationWarning` at compile time; in a future release it will become an error. The recommended annotation for a `@dataclasses.dataclass` is the dataclass type itself (`def k(s: MyStruct)`), which has a fast flatten-to-args path. For `@qd.data_oriented` containers continue to use `qd.Template` as before.
```
duburcqa: > Passing a `@dataclasses.dataclass` instance into a `qd.Template`-annotated kernel parameter was never intended to be supported, and only works inadvertently.

I recommend not documenting the past. It does not age well.

docs/source/user_guide/compound_types.md
```diff
@@ -123,10 +128,18 @@ Sub-struct passing supports:
...
 - positional and keyword call sites (`f(s.inner)` and `f(inner=s.inner)`)
 - call sites both directly inside `@qd.kernel` bodies and inside other `@qd.func` bodies
-- pruning of the sub-struct's leaf fields that the callee never reads
+- pruning of the sub-struct's leaf members that the callee never reads
 
 Note: assigning a sub-struct to a local variable and then passing it (`t = s.inner; touch_inner(t)`) is **not** supported. Pass the attribute access directly at the call site.
 
+### Frozen vs non-frozen
+
+A `dataclasses.dataclass` may be either non-frozen (the default) or frozen (`@dataclass(frozen=True)`). Both work as kernel arguments, but **kernel launch is faster with `frozen=True`** (because it enables some optimizations that would otherwise not be possible). Recommend `frozen=True` unless you specifically need to rebind members after construction.
```
duburcqa: "Recommend `frozen=True` unless you specifically need to rebind members after construction"

---> which is almost always a strong signal of a bad design.

docs/source/user_guide/compound_types.md
```diff
@@ -123,10 +128,18 @@ Sub-struct passing supports:
...
+
+### Under the hood
+
+A `dataclasses.dataclass` is a Python-only container. The compiler reads it at compile time and flattens its members into individual kernel parameters — the container itself has no memory layout and doesn't exist on the kernel side. Inside a kernel, tensor members are read-write through indexing (`s.x[i] = ...`) since their contents live on the device, but the member *binding* itself (`s.x = other_tensor`) cannot be reassigned from inside a kernel.
```
duburcqa: "since their contents live on the device" 

--> I don't get your point, it would just remove this information

docs/source/user_guide/compound_types.md
```diff
@@ -148,14 +161,157 @@ sim.step()
...
+
+```python
+vec3 = qd.types.vector(3, qd.f32)
+Particle = qd.types.struct(pos=vec3, vel=vec3, mass=qd.f32)
+particles = Particle.field(shape=(N,))
+```
+
+### Under the hood
+
+Unlike the other two compound types, `@qd.dataclass` is a real kernel-side type with a fixed memory layout. Each instance is laid out contiguously in memory, members are stored by value, and a tensor of the struct can be allocated (`Particle.field(...)`). Storing by value is also why ndarrays can't be members — ndarrays are heap-allocated buffers with dynamic shape and don't fit into a fixed-size cell.
```
duburcqa, line 289: I still don't understand what "members are stored by value" means. Like, the value of every single element is baked constant in the kernel and cannot be changed? Or just the address and shape?

docs/source/user_guide/compound_types.md
```diff
@@ -148,14 +161,157 @@ sim.step()
...
+
+This table summarises which member types are allowed inside which container type. "yes" means the member is walked correctly when the container is passed to a kernel; "no" means the member is ignored or the combination raises an error.
+
+| Container ↓ &nbsp;&nbsp;&nbsp; / &nbsp;&nbsp;&nbsp; Member → | `qd.ndarray` | `qd.field` | primitive | `dataclasses.dataclass` | `@qd.data_oriented` | `@qd.dataclass` |
+|---|:---:|:---:|:---:|:---:|:---:|:---:|
+| `dataclasses.dataclass`         | yes | yes | yes | yes | deprecated [\*1] | yes |
+| `@qd.data_oriented`             | yes | yes | yes | yes | yes      | yes |
+| `@qd.dataclass`                 | no  | yes | yes | no  | no       | yes |
+
+[\*1] A `dataclasses.dataclass` whose member type is `@qd.data_oriented` is currently deprecated. It only ever worked through the `qd.Template` outer-arg path (typed-dataclass annotations error out because `@qd.data_oriented` carries no per-member annotations to flatten), and that path is itself deprecated — see the [deprecation notice in Overview](#overview). To hold a `@qd.data_oriented` instance, make the outer container `@qd.data_oriented` too, rather than a `@dataclasses.dataclass`.
```
duburcqa: Still not a fan of documenting the past. What is deprecated is just not supported? Unless you are planning to maintain this deprecation, which sounds like a lot of effort for very limited value given our user base no?

docs/source/user_guide/compound_types.md
duburcqa: assigning things outside of `__init__` is softly unsupported in standard Python

docs/source/user_guide/compound_types.md
hughperkins: added a section

docs/source/user_guide/compound_types.md
hughperkins: It's supported in python, but not supported by pyright static type checker, is my understanding?

similarly, data_oriented allows assignment outside of `__init__` AFAIK, though we both agree it's poor practice, e.g. fails pyright.

docs/source/user_guide/compound_types.md
hughperkins: sure. You feel we should state that here?

docs/source/user_guide/compound_types.md
hughperkins: Here is how a qd.struct containing a vec3 and vec4 is laid out in memory:

```
- vec3[0]
- vec3[1]
- vec3[2]
- vec4[0]
- vec4[1]
- vec4[2]
- vec4[3]
```

The struct doesn't contain references or pointers to the tensors. It contains directly the values of the tensors themselves. And you cant change the size of any of these 'tensors', because you'd have to shift all of the 'tensors' after them.

It's actually a lot like how C structs work. For example the llvm runtime module is laid out in the same way https://github.com/Genesis-Embodied-AI/quadrants/blob/8439388d62721c5885bc4d4aab2c11a30c4d32ef/quadrants/runtime/llvm/runtime_module/llvm_runtime.h#L106-L125 .

docs/source/user_guide/compound_types.md
duburcqa: I think it would be nice to hit that having to specify 'frozen=False' is fishy, instead of maintaining a neutral tone.

docs/source/user_guide/compound_types.md
duburcqa: If this only contains hardcoded values that cannot be changed, what is the usecase? It sounds extremely niche. I guess I'm still missing something.

hughperkins: The values aren't hard-coded.

Here is an example C program that illustrates structs:

```
#include <stdio.h>

struct MyStruct {
	int a;
	int b;
	int c[4];
};

int main() {
	printf("hello\n");

	struct MyStruct s;
	s.a = 3;
	s.b = 5;
	s.c[1] = 4;

	printf("s.c[1] %i\n", s.c[1]);

	printf("&s.a %li\n", (long)&s.a);
	printf("&s.b %li\n", (long)&s.b);
	printf("&s.c[0] %li\n", (long)&s.c[0]);
	printf("&s.c[1] %li\n", (long)&s.c[1]);
	printf("&s.c[2] %li\n", (long)&s.c[2]);
	printf("&s.c[3] %li\n", (long)&s.c[3]);

	return 0;
}


```

Ouptut:
```
(quadrants) ~/git/quadrants (hp/data-oriented-ndarray-fix|✔) $ gcc /tmp/test.c -o /tmp/test && /tmp/test
hello
s.c[1] 4
&s.a 6096301716
&s.b 6096301720
&s.c[0] 6096301724
&s.c[1] 6096301728
&s.c[2] 6096301732
&s.c[3] 6096301736
```

hughperkins: Compared with using a vector in c++, which is more like normal python lists etc, which are by reference:

```
#include <iostream>

class Foo {
public:
	int a;
	int b;
	std::vector<int> myvec;
};

int main() {
	Foo foo;
	foo.a = 3;
	foo.myvec.push_back(123);
	foo.myvec.push_back(124);
	foo.myvec.push_back(130);
	std::cout << "foo.a " << foo.a << std::endl;
	std::cout << "&foo.a " << (long)&foo.a << std::endl;
	std::cout << "&foo.myvec[0] " << (long)&foo.myvec[0] << std::endl;
	std::cout << "&foo.myvec[1] " << (long)&foo.myvec[1] << std::endl;
	std::cout << "&foo.myvec[2] " << (long)&foo.myvec[2] << std::endl;
	return 0;
}

```

output:

```
(quadrants) ~/git/quadrants (hp/data-oriented-ndarray-fix|✔) $ g++ /tmp/test.cpp -o /tmp/test && /tmp/test
foo.a 3
&foo.a 6125956744
&foo.myvec[0] 4346993136
&foo.myvec[1] 4346993140
&foo.myvec[2] 4346993144
```

In the case of a vector, you can see the values are not even close to the other members of the class, in memory. They are allocated on the heap, and the class just references/points to them. And you can freely resize the vector as you like. Which is not possible in a C struct.

hughperkins: (really, we should rename qd.dataclass to qd.struct, and deprecate qd.dataclass. I feel it will be clearer. Unfortunately qd.Struct is already taken:

```
import quadrants as qd

qd.init(qd.gpu)

@qd.Struct
class Foo:
	...

@qd.kernel
def k1() -> None:
	...

k1()
```
output:
```
(quadrants) ~/git/quadrants (hp/data-oriented-ndarray-fix|✔) $ python /tmp/test.py
[Quadrants] version 0.0.0, llvm 22.1.0, commit 8a7ead44, osx, python 3.10.20
[Quadrants] Starting on arch=metal
Traceback (most recent call last):
  File "/tmp/test.py", line 6, in <module>
    class Foo:
  File "/Users/hugh/git/quadrants/python/quadrants/lang/struct.py", line 68, in __init__
    raise QuadrantsSyntaxError(
quadrants.lang.exception.QuadrantsSyntaxError: Custom structs need to be initialized using either dictionary or keyword arguments
```

But qd.struct doesn't exist, so maybe we could use that. (qd.Struct creates a qd.dataclass, so qd.struct and qd.Struct would be highly related).

).

hughperkins: Tweaked the doc to:
- mention that a qd.dataclass is analogous to a C struct
- discuss that vectors and matrices are allocated within the body of the struct, and change its size

docs/source/user_guide/compound_types.md
hughperkins: added some doc about best practices, and linters.

docs/source/user_guide/compound_types.md
hughperkins: I think I addressed this elsewhere in the doc now?

duburcqa:
ok to merge

PR #705, opened 2026-05-18
[DataOriented] Fastcache, perf, pruning

hughperkins:
## Summary

Extends `@qd.data_oriented` with kernel-pruning-driven fastcache argument hashing, an opt-in `stable_members=True` launch-time perf hint, walker robustness fixes (cycle-safe, MRO-safe), and `@qd.func`-from-kernel support for dataclass-typed args.

See the user-guide doc changes for the user-facing surface:
- [`compound_types.md`](docs/source/user_guide/compound_types.md) — `@qd.data_oriented` ndarray members, fastcache section, `stable_members=True`.
- [`fastcache.md`](docs/source/user_guide/fastcache.md) — pruning-driven argument hashing rules and the compound-type cache-keying table.


docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,34 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+- **L2** (L1 + narrow argument hash): stores the compiled artifact under a key that only hashes the arg paths in the L1 pruning set.
+
+#### Two rules
+
+The args hasher enforces two strict invariants:
+
+1. **The cache key may only include contributions from kernel-pruned paths.** A path is "pruned" (in the pruning-info sense) if Quadrants's compiler recorded the kernel reading it. Pruning info covers:
+   - Dataclass-flattened param accesses (`__qd_some_dc__qd_field` Names produced by `FlattenAttributeNameTransformer`, marked via `build_Name`).
+   - Ndarray accesses on data_oriented / template args (`struct_ndarray_launch_info`, folded into the pruning set by `Kernel._fold_struct_nd_paths_into_pruning`).
+   - Any other attribute-chain access on a kernel arg (`self.dofs.x`, `cfg.n`, …) recorded by `ASTTransformer.build_Attribute`'s `_qd_arg_chain` tracking and folded by `Kernel._fold_kernel_arg_chain_paths_into_pruning`. This covers primitive members baked into the kernel, nested struct paths, and accesses through `@qd.func` callees (propagated by `Pruning.record_after_call`).
```
duburcqa: I don't understand why you are mentioning `FlattenAttributeNameTransformer`, `build_Name`, `ASTTransformer.build_Attribute`, `Kernel._fold_kernel_arg_chain_paths_into_pruning`...etc and all these internals. For me it has nothing to do in a user-facing doc;

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,34 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+
+#### Practical implications
+
+- **Kernel-unused members do not affect the cache key.** If a `@qd.data_oriented` container has an opaque metadata member (UUID, Pydantic config, parent-solver back-pointer), and the kernel never reads it, the member is *not* hashed. Changes to `self._uid` or `self.cfg` don't disturb the cache key of a kernel that reads `self.dofs_state.x`.
+- **Kernel-unused members of unrecognised types are also fine.** Pruning narrowing skips them before the type-recognition check runs.
+- **Kernel-read members of unrecognised types fail fastcache loudly.** Either add explicit handling in `quadrants/lang/_fast_caching/args_hasher.py::stringify_obj_type` (for new tensor-like types whose dtype/shape matter), or move the access out of the kernel-read path (for opaque metadata that shouldn't be there in the first place).
```
duburcqa: > Either add explicit handling in `quadrants/lang/_fast_caching/args_hasher.py::stringify_obj_type` (for new tensor-like types whose dtype/shape matter)

Why I you inviting users to do such a thing?

docs/source/user_guide/fastcache.md
hughperkins: fair

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,34 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+Fastcache uses a **two-level cache**:
+
+- **L1** (source + config only): stores the set of *flat names* the kernel actually reads — e.g. `__qd_state__qd_x` for a kernel that reads `state.x`. This is the kernel's pruning info, computed at compile time by the AST builder.
+- **L2** (L1 + narrow argument hash): stores the compiled artifact under a key that only hashes the arg paths in the L1 pruning set.
+
+#### Two rules
+
+The args hasher enforces two strict invariants:
+
+1. **The cache key may only include contributions from kernel-pruned paths.** A path is "pruned" (in the pruning-info sense) if Quadrants's compiler recorded the kernel reading it. Pruning info covers:
```
hughperkins: To AI: lets not use this 'pruning-info' sense of pruning, that you are defining to mean 'not pruned' 😂  Let's just describe the world as the user sees it.

In fact, let's not even mention the 'pruning' concept.

The only thing the user needs to know, as far as strict invariant 1, is:
- if the kernel does not read or write a variable, it is entirely ignored by fastcache. It will not cause fastcache to fail, nor emit a warning, nor emit an error.

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,34 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+   - Dataclass-flattened param accesses (`__qd_some_dc__qd_field` Names produced by `FlattenAttributeNameTransformer`, marked via `build_Name`).
+   - Ndarray accesses on data_oriented / template args (`struct_ndarray_launch_info`, folded into the pruning set by `Kernel._fold_struct_nd_paths_into_pruning`).
+   - Any other attribute-chain access on a kernel arg (`self.dofs.x`, `cfg.n`, …) recorded by `ASTTransformer.build_Attribute`'s `_qd_arg_chain` tracking and folded by `Kernel._fold_kernel_arg_chain_paths_into_pruning`. This covers primitive members baked into the kernel, nested struct paths, and accesses through `@qd.func` callees (propagated by `Pruning.record_after_call`).
+
+   Paths *not* in the pruning set are skipped by the args hasher — they are guaranteed not to affect kernel codegen because the kernel cannot read them.
```
hughperkins: again, dont mention pruning set. Lets skip this paragraph.

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,34 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+
+2. **Unrecognised types at kernel-read paths must not be silently dropped or hashed by type-name.** If pruning says the kernel reads a path and the value at that path is a type the args hasher doesn't explicitly handle (Pydantic models, UUIDs, third-party tensor wrappers, …), fastcache is disabled for the call with a one-shot `[FASTCACHE][UNKNOWN_TYPE]` warning identifying the offending type plus an `[INVALID_FUNC]` log line confirming the cache is off. Capturing type identity without type parameters (dtype/shape on a hypothetical tensor type) would silently mask a value-affecting change.
```
hughperkins: to AI: remove 'Capturing type identity without type parameters (dtype/shape on a hypothetical tensor type) would silently mask a value-affecting change.'

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,34 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+
+#### Practical implications
+
+- **Kernel-unused members do not affect the cache key.** If a `@qd.data_oriented` container has an opaque metadata member (UUID, Pydantic config, parent-solver back-pointer), and the kernel never reads it, the member is *not* hashed. Changes to `self._uid` or `self.cfg` don't disturb the cache key of a kernel that reads `self.dofs_state.x`.
```
hughperkins: To AI: this is a duplciate of what you've already said. Remove this bulllet point (ie "kernel-unusd members...'

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,34 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+- **Kernel-unused members of unrecognised types are also fine.** Pruning narrowing skips them before the type-recognition check runs.
```
hughperkins: to AI: duplciate, skip this bullet point ("kernel-unused members of unrecognised type...")

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,34 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+   - Any other attribute-chain access on a kernel arg (`self.dofs.x`, `cfg.n`, …) recorded by `ASTTransformer.build_Attribute`'s `_qd_arg_chain` tracking and folded by `Kernel._fold_kernel_arg_chain_paths_into_pruning`. This covers primitive members baked into the kernel, nested struct paths, and accesses through `@qd.func` callees (propagated by `Pruning.record_after_call`).
+
+   Paths *not* in the pruning set are skipped by the args hasher — they are guaranteed not to affect kernel codegen because the kernel cannot read them.
+
+2. **Unrecognised types at kernel-read paths must not be silently dropped or hashed by type-name.** If pruning says the kernel reads a path and the value at that path is a type the args hasher doesn't explicitly handle (Pydantic models, UUIDs, third-party tensor wrappers, …), fastcache is disabled for the call with a one-shot `[FASTCACHE][UNKNOWN_TYPE]` warning identifying the offending type plus an `[INVALID_FUNC]` log line confirming the cache is off. Capturing type identity without type parameters (dtype/shape on a hypothetical tensor type) would silently mask a value-affecting change.
+
+#### Practical implications
```
hughperkins: To AI, just remove the entire 'practical implications' seciton. You're just restating the above I feel?

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,34 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+
+- **Kernel-unused members do not affect the cache key.** If a `@qd.data_oriented` container has an opaque metadata member (UUID, Pydantic config, parent-solver back-pointer), and the kernel never reads it, the member is *not* hashed. Changes to `self._uid` or `self.cfg` don't disturb the cache key of a kernel that reads `self.dofs_state.x`.
+- **Kernel-unused members of unrecognised types are also fine.** Pruning narrowing skips them before the type-recognition check runs.
+- **Kernel-read members of unrecognised types fail fastcache loudly.** Either add explicit handling in `quadrants/lang/_fast_caching/args_hasher.py::stringify_obj_type` (for new tensor-like types whose dtype/shape matter), or move the access out of the kernel-read path (for opaque metadata that shouldn't be there in the first place).
+
+`qd.field` / `ScalarField` / `MatrixField` are *recognised-but-unsupported*: encountering one at a kernel-read path disables fastcache for the call (with a warn-level diagnostic).
```
hughperkins: To AI: we can simply remove this I think? We already have a table of supported/nsupported types earlier right?

docs/source/user_guide/fastcache.md
```diff
@@ -143,13 +174,11 @@ print(obs.cache_stored)         # True if the compiled kernel was stored to cach
...
 On the first run you'll see `cache_stored=True` but `cache_loaded=False`. On the second run (after `qd.init`), `cache_loaded=True`.
 
-## Appendix
-
 ### Compound-type cache keying
 
 The args hasher walks compound-type kernel parameters recursively. For each leaf member it decides what (if anything) contributes to the cache key. The headline rules:
 
-**`@qd.data_oriented`:** the walker descends into `vars(obj)`. For each child:
+**`@qd.data_oriented`:** the walker descends into `vars(obj)`, narrowed by pruning info. For each walked child:
```
hughperkins: To AI: "narrowed by pruning info" => "narrowed to only include variables accessed by the kernel"

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,21 @@ Each compiled artifact is stored under a key derived from all of the following:
...
+### How the cache key is computed
+
+Fastcache uses a **two-level cache**:
+
+- **L1** (source + config only): identifies which kernel-arg variables the kernel actually reads or writes.
+- **L2** (L1 + narrow argument hash): stores the compiled artifact under a key that only hashes those variables.
+
+#### Two rules
+
+The args hasher enforces two strict invariants:
```
hughperkins: to AI: Is this true systematically? Or only for data oriented? (I dont remember).

docs/source/user_guide/compound_types.md
```diff
@@ -136,7 +136,7 @@ Sub-struct passing supports:
 - arbitrary nesting depth (`f(s.a.b.c)` where each level is a dataclass)
 - positional and keyword call sites (`f(s.inner)` and `f(inner=s.inner)`)
 - call sites both directly inside `@qd.kernel` bodies and inside other `@qd.func` bodies
-- pruning of the sub-struct's leaf members that the callee never reads
+- skipping of the sub-struct's leaf members that the callee never reads
```
hughperkins: To AI: in this case, I prefer pruning. Lets revert this change.

docs/source/user_guide/fastcache.md
```diff
@@ -120,6 +123,12 @@ Each compiled artifact is stored under a key derived from all of the following:
 
 When any of these change, the resulting key is different, so a new compilation occurs and a new entry is stored. Previous entries remain on disk — multiple cached versions coexist. You do not need to manually clear the cache when making code changes — the hash mismatch causes a transparent recompilation.
 
+### Two strict invariants
+
+1. **If the kernel does not read or write a variable, it is entirely ignored by fastcache.** It will not cause fastcache to fail, nor emit a warning, nor emit an error.
+
+2. **Unrecognised types at variables the kernel reads or writes must not be silently dropped or hashed by type-name.** If the value at such a variable has a type fastcache doesn't explicitly handle (Pydantic models, UUIDs, third-party tensor wrappers, …), fastcache is disabled for the call with a one-shot `[FASTCACHE][UNKNOWN_TYPE]` warning identifying the offending type plus an `[INVALID_FUNC]` log line confirming the cache is off.
```
hughperkins: To AI: shouldnt this be 'value of such a variable', not 'value at such a variable'?

docs/source/user_guide/fastcache.md
```diff
@@ -143,32 +152,25 @@ print(obs.cache_stored)         # True if the compiled kernel was stored to cach
 
 On the first run you'll see `cache_stored=True` but `cache_loaded=False`. On the second run (after `qd.init`), `cache_loaded=True`.
 
-## Appendix
-
 ### Compound-type cache keying
 
-The args hasher walks compound-type kernel parameters recursively. For each leaf member it decides what (if anything) contributes to the cache key. The headline rules:
+For `@qd.data_oriented` and `dataclasses.dataclass` kernel parameters, fastcache walks members recursively, only descending into members the kernel actually reads or writes (per the [strict invariants](#two-strict-invariants) above). Member-by-member behavior:
```
hughperkins: "only descending into members the kernel actually reads or write" => this seems hard to read/understand. Maybe remove this clause, start a new sentence and say "Any members that are not themselves read or written by the kernel, nor contain members read or written by the kernel, are skipped during the walk."

docs/source/user_guide/fastcache.md
```diff
@@ -143,32 +152,25 @@ print(obs.cache_stored)         # True if the compiled kernel was stored to cach
...
+    @dataclasses.dataclass
+    class SimConfig:
+        num_layers: int = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+        dt: float = dataclasses.field(metadata={FIELD_METADATA_CACHE_VALUE: True})
+    ```
 
-Note the asymmetry: `@qd.data_oriented` primitive members are baked into the kernel automatically (same semantics as `qd.Template`); `dataclasses.dataclass` members contribute only their *type* to the cache key unless you opt in per-member.
+    Annotate any member whose *value* (not just type) affects the compiled kernel — for example, an `int` used as a loop bound the compiler bakes into generated code. Without the annotation, two `SimConfig` instances with different `num_layers` values share a fastcache key, and the second silently loads a kernel compiled for the wrong value.
```
hughperkins: waht is SimConfig? Lets replace the entire "for example clause" with a sentence "Primarily this means any variable used inside qd.static."

docs/source/user_guide/compound_types.md
```diff
@@ -169,6 +169,33 @@ sim.step()
...
+```python
+@qd.data_oriented(stable_members=True)
+class Simulation:
+    def __init__(self, n):
+        self.x = qd.ndarray(qd.f32, shape=(n,))
+        self.v = qd.ndarray(qd.f32, shape=(n,))
+        # ... more ndarray / field / primitive members
+```
+
+(Equivalent: set `_qd_stable_members = True` as a class attribute.)
```
duburcqa: I don't think this should be advertised.

hughperkins:
Note: this is ready for next round of review.

docs/source/user_guide/compound_types.md
```diff
@@ -169,6 +169,31 @@ sim.step()
...
+
+Microbenchmark on an RTX PRO 6000 Blackwell with a container holding 30 `qd.ndarray` members across two nesting levels, calling a trivial kernel that takes the container as a `qd.template()` arg:
+
+| | Per-launch Python overhead |
+|---|---|
+| `stable_members=False` (default) | 18.5 µs/call |
+| `stable_members=True` | 13.5 µs/call |
+| | **−5 µs/call (−28%)** |
+
+**Trade-off:** with `stable_members=True`, reassigning an ndarray member on an instance is undefined behavior — the previously compiled kernel will be reused even if the new ndarray has a different `dtype`, `ndim`, or layout, silently bit-reinterpreting the new array's storage. Set it only on classes whose ndarray members are allocated once (typically in `__init__`) and never rebound. See [Reassigning ndarray members](#reassigning-ndarray-members) below for the supported alternative.
```
duburcqa: Maybe this behaviour should be enforced via post-init freeze. But I think it is tricky to get right without annoying boilerplate on user-side and/or performance penalty. 

Even Python dataclass does not offer a clean way to customise __init__ when frozen.

https://docs.python.org/3/library/dataclasses.html#dataclasses
```
frozen: If true (the default is False), assigning to fields will generate an exception. This emulates read-only frozen instances. See the [discussion](https://docs.python.org/3/library/dataclasses.html#dataclasses-frozen) below.

If [__setattr__()](https://docs.python.org/3/reference/datamodel.html#object.__setattr__) or [__delattr__()](https://docs.python.org/3/reference/datamodel.html#object.__delattr__) is defined in the class and frozen is true, then [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) is raised.
```

docs/source/user_guide/fastcache.md
```diff
@@ -143,32 +152,25 @@ print(obs.cache_stored)         # True if the compiled kernel was stored to cach
 
 On the first run you'll see `cache_stored=True` but `cache_loaded=False`. On the second run (after `qd.init`), `cache_loaded=True`.
 
-## Appendix
-
 ### Compound-type cache keying
 
-The args hasher walks compound-type kernel parameters recursively. For each leaf member it decides what (if anything) contributes to the cache key. The headline rules:
+For `@qd.data_oriented` and `dataclasses.dataclass` kernel parameters, fastcache walks members recursively. Any members that are not themselves read or written by the kernel, nor contain members read or written by the kernel, are skipped during the walk (per the [strict invariants](#two-strict-invariants) above). Member-by-member behavior:
```
duburcqa: This is redundant with:

> 2. **Unrecognised types at variables the kernel reads or writes must not be silently dropped or hashed by type-name.** If the value of such a variable has a type fastcache doesn't explicitly handle (Pydantic models, UUIDs, third-party tensor wrappers, …), fastcache is disabled for the call with a one-shot `[FASTCACHE][UNKNOWN_TYPE]` warning identifying the offending type plus an `[INVALID_FUNC]` log line confirming the cache is off.

hughperkins: Is it? I think they're highly related, but my reading is:

> Unrecognised types at variables the kernel reads or writes must not be silently dropped or hashed by type-nam

Talks about what happens if the args hasher hits an unrecognized type. And says that an unrecognized type should trigger some kind of warning, not just be silently ignored.

> Any members that are not themselves read or written by the kernel, nor contain members read or written by the kernel, are skipped during the walk

This talks about what happens to members that the kernel never reads nor writes: tehy are pruned from all consideration.

Now, these two things are related but different.

They are related because typically the various things that might be unreocgnized, suhc as id etc, are pruned, because the kernel does not read/write them, and therefore don't lead to a warning.

But they are not the same, and in fact pull in differennt directions. The first thing posts warnings for unreognized things. The second thing removes those unregonzied things altogether, avoidign the warnings.

Not sure if the doc should be restrutured, or rephrased, somehow to make these things clearer somehow?

docs/source/user_guide/compound_types.md
hughperkins: Ok, so, just to confirm, this comment is more like an observation, and no action is expected on my part?

duburcqa: I would say it is a design flaw / issue that is worth tracking somewhere even if it is not going to be fixed anytime soon.

docs/source/user_guide/fastcache.md
duburcqa: Personally I think it would make sense to pack these 2 informations in one paragraph or section, because these 2 informations are the two faces of the same coin: everything becomes obvious once you read both.

duburcqa:
ok to merge

hughperkins:
@codex review

hughperkins:
@codex review

python/quadrants/lang/_fast_caching/src_hasher.py
```diff
@@ -93,53 +128,195 @@ def _resolve_intenum_member(qualname: str | None, fallback: int | None) -> int |
...
-) -> str | None:
-    """
-    cache key takes into account:
-    - arg types
-    - cache value arg values
-    - kernel function (but not sub functions)
-    - compilation config (which includes arch, and debug)
-    - device capabilities (so a cached entry is not reused on a device whose caps change codegen)
+def make_source_config_key(kernel_source_info: FunctionSourceInfo) -> str:
+    """Build the L1 cache key: source + config + version, with no dependence on args.
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Confirmed and fixed in a370de261 - thanks, this was a real staleness bug, not a theoretical one.

Repro (single process, three `qd.init` cycles sharing one cache dir; `@qd.data_oriented` container `s` with a `flag` primitive and two ndarray members, kernel body `if qd.static(s.flag): s.x[0] = 1 else: s.y[0] = 2`):

1. `flag=True` compiles. L1 records `{__qd_s__qd_flag, __qd_s__qd_x}`.
2. `flag=False`, `y: f32` compiles. Its L2 key is narrowed by that set, so nothing about `s.y` is in the key.
3. `flag=False`, `y: i32`. Same L2 key -> `cache_loaded=True`, the f32 artifact is launched against the i32 ndarray, and `y[0]` reads back `1073741824` instead of `2`.

Fix in `persist_l1_and_set_l2_key`: after a cold compile, if this specialization read paths the shared L1 entry does not list, the entry is grown to the union and the L2 key phase 2 derived from the too-narrow set is discarded and recomputed from the union. That keeps the invariant that every stored L2 entry's key was derived from a path set covering the reads of the specialization it names, so a dtype / ndim / baked-value change at any path that specialization reads always changes its key. It converges rather than recompiling forever: each sibling specialization recompiles once after the union grows, then hits (verified over 8 runs of the probe - runs 4-8 are all `cache_loaded=True` with correct values). The cost is over-invalidation only: a path read by one specialization also enters its siblings' keys.

Same commit:

- `_CACHE_VALUE_SCHEMA_VERSION` -> `cachevalue-v5-pruning-union`, since pre-fix on-disk entries can name an artifact whose key omits paths it depends on.
- The fastcache restore path no longer falls back to L1's `graph_do_while_levels`. That table has the same per-specialization property (a `qd.graph_do_while` nested in a `qd.static` branch belongs only to the specializations that compile that branch), and with the schema version in both keys an L2 entry always carries its own table.
- Regression test `test_src_ll_cache_pruning_union_across_static_branches` (steps 1-3 above, plus the convergence assertion), and invariant 1 in `docs/source/user_guide/fastcache.md` now says an unread member is ignored only if *no* specialization reads it.

I did not key the L1 entry by specialization, which was your other suggestion: that needs the template / arg-feature fingerprint before the args walk, i.e. the wide walk the two-level cache exists to avoid.

python/quadrants/lang/kernel_impl.py
```diff
@@ -357,46 +369,54 @@ def data_oriented(cls=None, *, template_primitives: bool = True):
...
+    def make_kernel_indirect(fun, is_property, attr_name):
+        # Capture the primal at decoration time so the per-call path skips the ``_BoundedDifferentiableMethod``
+        # allocation. The class itself is validated when ``_BoundedDifferentiableMethod`` is invoked via the
+        # ``.grad()`` path; for the common primal call here we replicate the check inline.
+        primal = fun._primal
+
+        @wraps(fun)
+        def _kernel_indirect(self, *args, **kwargs):
+            try:
+                return primal(self, *args, **kwargs)
```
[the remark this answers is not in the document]
hughperkins: Agent says:

The check does still run on that path today, but you found something real next to it.

`QuadrantsCallable.__init__` ends with `update_wrapper(self, fn)`, which copies `fun.__dict__` over the attributes just assigned - including `fun`'s own `wrapper`. So `make_kernel_indirect`'s `_kernel_indirect` was never installed: every member-kernel call, bound or unbound, still routed through `wrapped_classkernel` -> `wrapped_func`, and `wrapped_classkernel` is what raises. Measured on the branch head before the change:

- `Decorated.__dict__["step"].wrapper.__code__.co_name == "wrapped_classkernel"`
- `Decorated.step(plain_object)` -> `QuadrantsSyntaxError: Please decorate class Plain with @qd.data_oriented`

So the comment was wrong in a different way than reported (the fast path did not exist, rather than existing without a check), and the saving it describes was not being taken either.

Fixed both halves in a370de261: `ret.wrapper` is now set to `_kernel_indirect` explicitly, with a comment about the clobbering, and the owner check is replicated inline there - so the diagnostic is unchanged while the call path loses two Python frames. Added `test_member_kernel_rejects_foreign_owner`, which pins both entry points: `Decorated.write(Undecorated(), a)` and an undecorated class called the ordinary way.

python/quadrants/lang/kernel_impl.py
```diff
@@ -275,7 +275,7 @@ def grad(self, *args, **kwargs) -> "Kernel":
         return self._adjoint(self._kernel_owner, *args, **kwargs)
 
 
-def data_oriented(cls):
+def data_oriented(cls=None, *, stable_members: bool = False):
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Docs are in sync now, no code change needed for this one.

`docs/source/user_guide/compound_types.md` has a `stable_members` section covering when to use it, the decorator form, the microbenchmark it buys (18.5 -> 13.5 us/call on a 30-ndarray container), and an explicit statement that reassigning an ndarray member on such a class is undefined behaviour - the previously compiled kernel is reused even if the new ndarray has a different dtype, ndim, or layout.

The point you raise about the existing text is handled too: the "Reassigning ndarray members" section now opens by scoping that support to `@qd.data_oriented` classes *without* `stable_members=True`, and links back to the trade-off. The `_qd_stable_members = True` class attribute mentioned in the runtime docstring is documented in the same section, including the fact that it is read from the class itself rather than from its bases, so a subclass has to set it (or be decorated) too.

hughperkins:
@codex review

hughperkins:
Ok, so we need:
- [X] codex green
- [x] Genesis benchmarks
- [x] Genesis unit tests
- [x] CI green
- [x] check docs
- [x] check comments / code ratio

hughperkins:
(removing ok to merge, whilst I fix up comments/code ratio etc)

hughperkins:
docs look ~ok

hughperkins:
running genesis benchmarks on bench1 (oh... we probably need to migrate genesis to use dataclasse instead of data orietned, for static first, but .... maybe with the new bnechmarks, we dont? 🤷 )

hughperkins:
running genesis unit tests on tests0

hughperkins:
cut a bunch of comments

hughperkins:
genesis unit tests passing:

Screenshot 2026-08-31 at 10 40 50

hughperkins:
benchmark look ok-ish:

20260831_do_0946 (1)

hughperkins:
lets run before/after for dex_hand, g1_fall_accessors, g1_fall

hughperkins:
(running before/after on bench1)

hughperkins:
before/after reasultss:

```
=== before/after interleaved (3 runs each, single node) ===
before qd=f6df02c2e  after qd=5b75c0421  genesis=19f56d65

dex_hand                 4096  cuda                 2        32669.5   1.65        32631.5   0.06    -0.12      1.65
g1_fall                  4096  cuda                 2      1892491.5   1.54      1892153.5   1.27    -0.02      1.99
g1_fall_accessors        4096  cuda                 2      1266250.5  25.68      1282734.0  17.48     1.30     31.19
[10:08:55] Wrote /home/hugh/git/tmp/20260831_do_1240/meta.yaml
[10:08:55] Done. NO significant before/after difference
```

PR #707, opened 2026-05-18
[Vulkan] Declare GroupNonUniform SPIR-V caps and enable shaderSubgroupExtendedTypes

hughperkins:
The new QIPC subgroup/block work (#676, #684) broadened how often the SPIR-V codegen emits `OpGroupNonUniform*` ops and the `SubgroupLocalInvocationId` BuiltIn, but the corresponding Vulkan feature / SPIR-V capability boilerplate was never wired up:

- `VkPhysicalDeviceShaderSubgroupExtendedTypesFeatures::shaderSubgroupExtendedTypes` was not requested at device creation, so the validation layer rejects the shader with VUID-RuntimeSpirv-None-06275 whenever an `OpGroupNonUniform*` is invoked with an 8/16/64-bit integer or 16-bit float.

- `OpCapability GroupNonUniform{,Shuffle,ShuffleRelative,Vote,Arithmetic,Ballot}` were never declared in the SPIR-V module header, so strict validators (and any driver that enforces the letter of the spec) reject the module at `vkCreateShaderModule` with VUID-VkShaderModuleCreateInfo-pCode-08737. Permissive drivers tolerated the omission for 32-bit ops but the bug was waiting to bite as soon as wider types were used.

This commit gates the new feature/cap emissions on the existing `spirv_has_subgroup_*` / `VkPhysicalDevice...ExtendedTypes` device queries, so no behavioural change occurs on platforms that don't expose them.

Tests on Vulkan x i64 subgroup/block exclusive_min/max still fail because of a separate NVIDIA-driver-level miscompile of `OpStore`/`OpSelect` on function-scope `%long` variables when the stored value is an i64 constant with high bits set (e.g. `INT64_MAX`). That failure is not made worse by this commit and now produces the correct numeric value on every backend other than the NVIDIA Vulkan driver currently shipped here.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/codegen/spirv/spirv_ir_builder.cpp
+++ b/quadrants/codegen/spirv/spirv_ir_builder.cpp
@@ -107,6 +107,19 @@ void IRBuilder::init_header() {
     // (and the SPIR-V spec marks both as required for any `OpGroupNonUniformShuffle{,Up,Down}` /
     // `OpGroupNonUniformBroadcast` emission), so we tie them to `spirv_has_subgroup_basic` rather
     // than introducing a separate device cap.
+    //
+    // FIXME: Vulkan's `VkPhysicalDeviceSubgroupProperties::supportedOperations` exposes
+    // `VK_SUBGROUP_FEATURE_SHUFFLE_BIT` / `VK_SUBGROUP_FEATURE_SHUFFLE_RELATIVE_BIT` as bits separate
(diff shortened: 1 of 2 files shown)


hughperkins:
@codex review

PR #708, opened 2026-05-18
[AutoDiff] Reject recycled identity_key in AdStackCache::register_adstack_sizing_info

hughperkins:
Stop the `same identity_key -> same registry_id` idempotent fast path from collapsing two unrelated kernels into a single registry slot when the allocator recycles an `OffloadedTask::ad_stack` address across `qd.reset()` / re-codegen cycles. The previous behavior overwrote the old entry's kernel_name in place and returned the existing id, so the per-spec `max_reducer_cache_` keyed by `(registry_id, stack_id, mor_node_idx)` then served stale results to the new kernel.

Symptom: `test_max_reducer_per_kernel_registry_id_isolation` flaked (~40% pass rate even in isolation) whenever the allocator happened to reuse the address across the two `qd.template()` instantiations of `compute`. `test_adstack_overflow_diagnostic_and_auto_recovery` flaked the same way — its enriched diagnostic resolved the cmpxchg-recorded id to whatever kernel had most recently overwritten the slot, e.g. `snode_reader_4_kernel` instead of the actual `compute_..._reverse_grad`.

Fix: guard the idempotent path on `(kernel_name, task_id_in_kernel)` matching the stored entry. On mismatch, drop the stale reverse-lookup entry and fall through to the content-stable hash path, which mints (or finds) the correct id for the new kernel. The previous registry entry stays alive so any still-live owner (e.g. a kernel whose IR baked the old id) keeps resolving via `lookup_adstack_sizing_info`.

The C++ unit test for this codepath updated to pin both invariants: true idempotent re-registration (same pointer + same name/task → same id, metadata updated in place; matches the production START + FINALIZE codegen flow) and pointer-recycled re-registration (same pointer + different name/task → different ids, previous entry preserved).

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/program/adstack/cache.cpp
+++ b/quadrants/program/adstack/cache.cpp
@@ -481,13 +481,28 @@ uint32_t AdStackCache::register_adstack_sizing_info(const void *identity_key,
   // Idempotent re-registration: same `identity_key` yields the same id across re-compiles and updates the entry's
   // metadata + size_exprs in place. The key is just an opaque dedup token - the registry never dereferences it; all
   // data needed by the diagnose path is copied into the entry below.
+  //
+  // BUT `identity_key` is the address of an `AdStackSizingAttribs` / `OffloadedTask::ad_stack` held by a launcher's
+  // tasks vector or by a transient `current_task` during codegen. When the previous owner is freed and the allocator
+  // recycles the address for a brand-new task that belongs to a DIFFERENT logical kernel, the raw-pointer lookup
+  // succeeds against the stale entry. Returning the previous id would then cause the new task to inherit the old
+  // task's `registry_id`, and the per-spec `max_reducer_cache_` (keyed by `(registry_id, stack_id, mor_node_idx)`)
+  // would serve a stale result to the new kernel — observed as `assert dispatch_count > after_first` failing in
+  // `test_max_reducer_per_kernel_registry_id_isolation` whenever the allocator happens to recycle the same address
+  // across the two `qd.template()` instantiations. Guard the idempotent path on `(kernel_name, task_id_in_kernel)`
+  // matching so a recycled address falls through to the content-stable hash path below, which produces (or finds)
+  // the correct id for the new kernel.
   if (auto it = adstack_sizing_info_id_by_ptr_.find(identity_key); it != adstack_sizing_info_id_by_ptr_.end()) {
     auto &entry = adstack_sizing_info_registry_[it->second];
-    entry.kernel_name = kernel_name;
-    entry.task_id_in_kernel = task_id_in_kernel;
-    entry.allocated_max_sizes = std::move(allocated_max_sizes);
-    entry.size_exprs = std::move(size_exprs);
-    return it->second;
+    if (entry.kernel_name == kernel_name && entry.task_id_in_kernel == task_id_in_kernel) {
+      entry.allocated_max_sizes = std::move(allocated_max_sizes);
+      entry.size_exprs = std::move(size_exprs);
+      return it->second;
+    }
+    // Recycled pointer for a different kernel. Drop the stale reverse-lookup entry so the fall-through below treats
+    // this as a fresh registration (the registry entry itself stays alive because the previously-registered owner
+    // may still resolve its id through `lookup_adstack_sizing_info` on the overflow diagnose path).
+    adstack_sizing_info_id_by_ptr_.erase(it);
   }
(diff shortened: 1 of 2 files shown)


quadrants/program/adstack/cache.cpp
```diff
@@ -481,13 +481,28 @@ uint32_t AdStackCache::register_adstack_sizing_info(const void *identity_key,
...
+  //
+  // BUT `identity_key` is the address of an `AdStackSizingAttribs` / `OffloadedTask::ad_stack` held by a launcher's
+  // tasks vector or by a transient `current_task` during codegen. When the previous owner is freed and the allocator
+  // recycles the address for a brand-new task that belongs to a DIFFERENT logical kernel, the raw-pointer lookup
+  // succeeds against the stale entry. Returning the previous id would then cause the new task to inherit the old
+  // task's `registry_id`, and the per-spec `max_reducer_cache_` (keyed by `(registry_id, stack_id, mor_node_idx)`)
+  // would serve a stale result to the new kernel — observed as `assert dispatch_count > after_first` failing in
+  // `test_max_reducer_per_kernel_registry_id_isolation` whenever the allocator happens to recycle the same address
+  // across the two `qd.template()` instantiations. Guard the idempotent path on `(kernel_name, task_id_in_kernel)`
+  // matching so a recycled address falls through to the content-stable hash path below, which produces (or finds)
```
duburcqa, line 493: When is this issue supposed to happen? I guess after Quadrants reset and re-init? If so, I would rather recommend clearer entire adstack cache at program termination instead of checking whether the cache entry is still valid in adstack itself.

hughperkins: Opus writes:

  My take: the premise is wrong, so the suggested alternative doesn't fix it.

  The recycling happens within a single Program lifetime, not across qd.reset() / re-init. During normal codegen the LLVM backends do, per task:

  current_task = std::make_unique<OffloadedTask>(...);
  register_adstack_sizing_info(&current_task->ad_stack, ...);   // START registration
  // ... codegen ...
  offloaded_tasks.push_back(*current_task);   // copy by value into the vector
  current_task = nullptr;                      // free the original

  The next make_unique<OffloadedTask> then often lands at the same address the previous one just vacated, with a different kernel_name. My debug trace from
  test_max_reducer_per_kernel_registry_id_isolation (single qd.init, no reset) shows it clearly — same address recycling across ext_arr_to_ndarray_c18_0_kernel →
  compute_c50_0_kernel → compute_c51_0_reverse_grad_kernel → compute_c50_1_kernel → compute_c51_1_reverse_grad_kernel inside one test run.

  So "clear the entire adstack cache at program termination" wouldn't fix this — the cache is already empty at the start of the buggy sequence; the corruption builds up _during_
   codegen of a single program.

  The other option that would have worked is the inverse: invalidate the adstack_sizing_info_id_by_ptr_ reverse-lookup entry every time an OffloadedTask::ad_stack is freed (i.e.
   hook the destructor or every current_task = nullptr site). But that spreads the contract across many call sites in CPU/CUDA/AMDGPU codegens, plus the launcher's vector copy
  semantics, vs. one localized check at the single ingress point. The targeted check in register_adstack_sizing_info keeps the contact area minimal (per AGENTS.md).

  The codex bot's comment on the same PR is actually the real follow-up worth doing: the offline-cache ensure_runtime_registry_ids_for_max_reducer() path short-circuits on
  is_adstack_sizing_info_registered(&ad_stack) _before_ reaching my mismatch check, so a recycled address on the cache-reload path still skips registration. Worth fixing in a
  follow-up PR.

hughperkins: Thoughts?

duburcqa: I think the issue reported by codex should be fixed as part of this PR.

hughperkins: The issue with the codex flagged issue is that it's not testable currently, which is why I asked AI to add a FIXME, rather than trying to fix it.

If we try to fix it, two things might happen:
1. we might not actually fix it; we only THINK we fixed it,
2. might create new bugs

So I would prefer to wait until we hit some actual related bug, and then fix it, if that situation arises

duburcqa: Ok let's merge

duburcqa:
ok to merge

PR #710, opened 2026-05-20
[Lang] Add axes= to ndrange

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/docs/source/user_guide/parallelization.md
+++ b/docs/source/user_guide/parallelization.md
@@ -48,29 +48,29 @@ def fill(a: qd.Template) -> None:
 
 `I` is a `qd.Vector` with one element per dimension.
 
-### Controlling iteration order with `layout=`
+### Controlling iteration order with `axes=`
 
-By default, `qd.ndrange(d0, d1, ..., dN-1)` makes the **last argument the innermost (fastest-varying) axis** in the flat parallel loop: adjacent flat threads differ in the last index. The `layout=` keyword lets you choose a different iteration-nesting order. It's a tuple of `int` listing the **canonical axis index at each successive iteration-nesting level, outermost first**, and must be a permutation of `range(N)` where `N` is the number of arguments to `qd.ndrange`:
+By default, `qd.ndrange(d0, d1, ..., dN-1)` makes the **last argument the innermost (fastest-varying) axis** in the flat parallel loop: adjacent flat threads differ in the last index. The `axes=` keyword lets you choose a different iteration-nesting order. It's a tuple of `int` listing the **canonical axis index at each successive iteration-nesting level, outermost first**, and must be a permutation of `range(N)` where `N` is the number of arguments to `qd.ndrange`:
 
 ```python
 @qd.kernel
 def k():
     # axis 1 is outermost (slowest-varying), axis 0 is innermost (fastest-varying)
-    for i, j in qd.ndrange(M, N, layout=(1, 0)):
+    for i, j in qd.ndrange(M, N, axes=(1, 0)):
         ...
 ```
 
-The yielded loop variables (`i`, `j`, ...) are still bound to canonical axes 0, 1, ... — only the visit order changes. `layout=None` (the default) and the identity permutation `(0, 1, ..., N-1)` are equivalent and reproduce the default last-argument-innermost order. Mismatched length and non-permutation values are rejected up front with `qd.QuadrantsSyntaxError`.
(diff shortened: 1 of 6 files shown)


docs/source/user_guide/parallelization.md
```diff
@@ -48,6 +48,32 @@ def fill(a: qd.Template) -> None:
 
 `I` is a `qd.Vector` with one element per dimension.
 
+### Controlling iteration order with `layout=`
```
duburcqa: You prefer 'layout' over 'axis' ?

hughperkins: I think it should be consistent with qd.Tensor.

hughperkins: (honestly though, in the absence of numpy using axis, I do actually prefer layout).

hughperkins: axis in numpy is a single integer, not a tuple. I think the meaning is slightly different:

Screenshot 2026-05-20 at 09 42 40

duburcqa: Layout is weird. There is no layout involved here.

duburcqa: Like, it has nothing to do with qd.Tensor layout, why the name should be the same?

duburcqa: Numpy uses axes: https://numpy.org/doc/2.2/reference/generated/numpy.permute_dims.html

hughperkins: its the exact same concept
- if you check the commit history AI was trying to exactly tie the motivation of using this functionality on ndarray to using the similar functionality on qd.Tensor
- in practice, they will tend to be used together
     - without using the similar functionality on qd.Tensor, there is no obvious reason why we would want to be able to switch the dimensions on ndarray at runtime I feel?

I feel that our options are:
- use layout for both Tensor and ndarray, or
- use axes for both Tensor and ndarray

(I'm tempted to compromise with using layout for Tensor and axes for ndarray, but it's inconsistent, and inconsistency within Quadrants is worse than inconsistency between Quadrants and some other library I feel).

hughperkins: *ndrange

hughperkins: I suppose you could argue that one is permuting the physical layout, and one is permuting iteration order, and those are different things. But it's still about permutating dimensions I feel.

duburcqa: > I suppose you could argue that one is permuting the physical layout, and one is permuting iteration order, and those are different things. 

Yes, it is exactly what I thinking.

duburcqa: That being said, I don't think we should argue too much about this. I prefer axes, you prefer layout. Your call. I can live with layout, it is just the name of an argument, nothing dramatic.

duburcqa:
ok to merge

docs/source/user_guide/parallelization.md
```diff
@@ -48,6 +48,32 @@ def fill(a: qd.Template) -> None:
...
+
+By default, `qd.ndrange(d0, d1, ..., dN-1)` makes the **last argument the innermost (fastest-varying) axis** in the flat parallel loop: adjacent flat threads differ in the last index. The `layout=` keyword lets you choose a different iteration-nesting order. It's a tuple of `int` listing the **canonical axis index at each successive iteration-nesting level, outermost first**, and must be a permutation of `range(N)` where `N` is the number of arguments to `qd.ndrange`:
+
+```python
+@qd.kernel
+def k():
+    # axis 1 is outermost (slowest-varying), axis 0 is innermost (fastest-varying)
+    for i, j in qd.ndrange(M, N, layout=(1, 0)):
+        ...
+```
```
duburcqa, line 61: Could be nice to add in comments above '...' the first 4 iterations, which implies replacing M and N by concrete values, eg (2,  10).

docs/source/user_guide/tensor.md
```diff
@@ -113,6 +113,8 @@ b[i, j] = ...                   # canonical indexing in kernels still works
 
 Any permutation is supported, up to Quadrants' `quadrants_max_num_indices` (currently 12). `layout=None` and the identity permutation (`(0, 1, ..., N-1)`) are equivalent and forward no permutation to the underlying allocator.
 
+For best performance, pair `qd.tensor(..., layout=...)` with a matching iteration order via `qd.ndrange(..., layout=...)` (see [`parallelization`](parallelization.md#controlling-iteration-order-with-layout)): the permutation has the same meaning on both APIs, and using the same value on both lines adjacent flat threads up with adjacent physical memory slots.
```
duburcqa: > see [`parallelization`](parallelization.md#controlling-iteration-order-with-layout)

This does exist ?! I must have missed that.

hughperkins: what do you mean by 'this', in this context?

PR #712, opened 2026-05-23
[Perf] Add `qd.types.UnpackedVector[dtype, count]` for indexed register access

hughperkins:
Adds a new ``qd.types.UnpackedVector[dtype, count]`` annotation for ``@qd.dataclass`` that exposes a logical N-element group as ``obj.r[i]`` while storing N individually-named synthetic scalar fields (``_r0..._r{N-1}``) under the hood. For python-int indices (including ``qd.static(range(N))``-unrolled loop variables), the AST transformer rewrites ``obj.r[i]`` directly to ``obj._r{i}``, so the generated LLVM IR / PTX is byte-identical to a hand-rolled named-field struct.

Motivation: today's idiomatic ``r: qd.types.vector(N, dtype)`` group field is the *packed* layout -- one ``alloca`` covers all N slots. Packed storage is fine until register pressure rises -- once LLVM SROA fails to decompose the packed ``alloca`` (e.g. two concurrent tiles in a Cholesky + TRSM kernel), the whole group spills to local memory as a unit, causing runtime regressions via ``ld.local`` / ``st.local`` traffic. The named-field cascade pattern avoids the spill but balloons source size (32-way ``if k == N: self.rN = val`` write cascades duplicated at every callsite). ``UnpackedVector`` is the named-field layout with the ergonomic indexed syntax restored: the storage is unpacked into N independent ``alloca``s so SROA + ``mem2reg`` can promote each slot independently, and source-side ``obj.r[i]`` collapses to one AST node per callsite.

### Usage

```python
import quadrants as qd

@qd.dataclass
class Tile:
    r: qd.types.UnpackedVector[qd.f32, 32]

@qd.kernel
def k(out: qd.types.NDArray[qd.f32, 1]) -> None:
    t = Tile()
    for i in qd.static(range(32)):
        t.r[i] = qd.f32(i)
    out[0] = t.r[3]
```

### Naming history

- First proposed as ``qd.field_array``, renamed to ``qd.register_array`` during review (``qd.field`` already refers to global tensor fields, so ``field_array`` was ambiguous).
- Renamed to ``qd.unpacked_array`` -- "register" reads as imperative ("register the array") and overcommits the type to an implementation effect (register residency) that is the optimiser's call, not a guarantee; "unpacked" describes the actual storage layout and contrasts cleanly with the implicit packed-vector default.
- Renamed to ``qd.types.UnpackedVector`` and switched from function-call form to subscript form ``[dtype, count]``: matches ``qd.types.NDArray[dtype, ndim]`` as a parameterised type, lives in the same namespace as its packed sibling ``qd.types.vector``, and the subscript spelling makes the marker visually read as a type annotation (not a runtime value), which heads off the common ``@dataclasses.dataclass`` vs ``@qd.dataclass`` confusion.

### Changes

- ``lang/unpacked_vector.py``: new module with ``UnpackedVector`` class (subscriptable via ``__class_getitem__``), ``_expand_unpacked_vector_naming`` helper, ``_UnpackedVectorRef`` transient proxy, and misuse guards (``__getitem__`` / ``__call__`` on the marker raise a clear ``QuadrantsSyntaxError`` pointing at the ``@qd.dataclass`` requirement). No dependency on ``struct.py``.
- ``lang/struct.py``: imports from ``unpacked_vector``; ``StructType.__init__`` expands an ``UnpackedVector`` annotation into ``count`` synthetic scalar fields plus ``_unpacked_groups`` metadata; ``StructType.__call__`` propagates the metadata onto the Struct instance as ``_qd_unpacked_groups``.
- ``lang/impl.py``: ``expr_init`` preserves the ``_qd_unpacked_groups`` metadata across the ``Struct`` rewrap, via ``setattr`` to satisfy pyright.
- ``lang/ast/ast_transformer.py``: ``build_Attribute`` returns a ``_UnpackedVectorRef`` for group access; ``build_Subscript`` resolves it to a direct field reference for python-int / ``qd.static``-resolved indices.
- ``types/__init__.py``: re-exports ``UnpackedVector`` at the bottom of the module (load-order is load-bearing -- documented inline).
- ``tests/python/test_unpacked_vector.py``: 8 tests (construction, static python-int subscript, qd.static loop-var subscript, runtime-index rejection, static-index OOB rejection, plus 3 misuse-guard tests for the marker).
- ``tests/python/test_api.py``: golden-list update.
- ``docs/source/user_guide/unpacked_vector.md``: new user-guide page (problem statement, how-to, common pitfalls, constraints, comparison table vs ``qd.f32`` / ``qd.types.vector``). Linked from ``index.md`` Core Concepts and cross-linked from ``compound_types.md``.

### Constraints

- Static indices only: ``t.r[k]`` must resolve at compile time. Runtime ``k`` is rejected with a friendly error pointing at ``qd.static``; existing cascade helpers continue to handle the runtime case by spelling out the ``_rN`` fields directly. Adding runtime-int support is a small follow-up.
- Storage only: no vector arithmetic (no ``t.r + other``, no ``t.r.dot(...)``). Use ``qd.types.vector(N, dtype)`` if you need those.
- ``count`` is baked into the struct layout when ``@qd.dataclass`` processes the class.

Verified on a port of genesis ``_tile32.py``: PTX byte-identical to the named-field baseline on both ``chol_kernel`` and ``chol_trsm_kernel``; zero local-memory spills; ~25% compile-time reduction on the single-tile harness; source dropped from 1068 to 515 lines (-52%).

All pre-existing tests in test_py_dataclass.py + test_complex_struct.py + test_struct.py continue to pass; the 8 new tests pass in ~2.5s total. Pyright clean. Pre-commit (black, clang-format, ruff, pylint, trailing-whitespace, eof) clean.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
Need some user-facing doc.

hughperkins:
(Also, might want to revisit the name)

docs/source/user_guide/compound_types.md
```diff
@@ -159,3 +159,5 @@ class Particle:
     vel: qd.types.vector(3, qd.f32)
     mass: qd.f32
 ```
+
+For larger statically-indexed groups (e.g. a row or tile of scalars that you want SROA to register-promote independently), see {doc}`unpacked_array`.
```
duburcqa: What is 'SROA' ?

"to register-promote independently" ---> what does this means? It is hardcore.

hughperkins: instead of storing in global memory, it stores the variable in a register. orders of magnitudes faster.

hughperkins: (Note: technically it would be `local` memory, not `global` memory. But `local` vs `global` is a logical distinction. Physically they are both stored off-chip, in DRAM, far far away from the cores).

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+
+The intuitive way to group `N` scalars on a per-thread struct is to declare a vector member:
+
+```python
+@qd.dataclass
+class Tile:
+    r: qd.types.vector(32, qd.f32)
+```
+
+`qd.types.vector(N, dtype)` lays the group out as a single packed `alloca`. LLVM's scalar-replacement-of-aggregates pass (SROA + `mem2reg`) tries to decompose that `alloca` into per-slot SSA values so each one can live in a register — but once a kernel's register pressure crosses a threshold (e.g. two concurrent `32x32` tiles in a Cholesky + triangular solve), SROA bails out on the packed `alloca` and the whole group spills to local memory as a unit. Each access then turns into a `ld.local` / `st.local` and the kernel slows down dramatically.
```
duburcqa: > `qd.types.vector(N, dtype)` lays the group out as a single packed `alloca`.

Not clear what this means.

> LLVM's scalar-replacement-of-aggregates pass (SROA + `mem2reg`) tries to decompose that `alloca` into per-slot SSA values so each one can live in a register — but once a kernel's register pressure crosses a threshold (e.g. two concurrent `32x32` tiles in a Cholesky + triangular solve), SROA bails out on the packed `alloca` and the whole group spills to local memory as a unit.

That is ultra hardcore.

> Each access then turns into a `ld.local` / `st.local` 

What does this means?

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+```python
+@qd.dataclass
+class Tile:
+    r0: qd.f32
+    r1: qd.f32
+    # ... 30 more lines ...
+    r31: qd.f32
+```
+
+Now each slot has its own `alloca`, and SROA + `mem2reg` can promote each one independently. The optimiser is also free to spill only the slots it has to, instead of the whole group as a unit. The cost is that every index becomes a cascade in source:
```
duburcqa: > Now each slot has its own `alloca`, and SROA + `mem2reg` can promote each one independently. The optimiser is also free to spill only the slots it has to, instead of the whole group as a unit. 

That's also hardcore.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+```python
+def get_r(t, k):
+    if k == 0:
+        return t.r0
+    elif k == 1:
+        return t.r1
+    # ... 30 more branches ...
+```
+
+…which is duplicated at every call site that wants to read or write the group.
```
duburcqa: '…'

duburcqa: > which is duplicated at every call site that wants to read or write the group.

So what?

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+
+`qd.unpacked_array` is the named-field layout with the ergonomic indexed syntax restored:
+
+```python
+@qd.dataclass
```
duburcqa, line 35: > @qd.dataclass

It should be `qd.unpacked_array` no?

docs/source/user_guide/unpacked_vector.md
hughperkins: if you have a vector, rather than a qd.unpacked_array, it either stores the entire vector in registers, or the entire vector in local memory (meaning, unintuitively: in off-chip DRAM, sloowwww 🐌).

The unpacked_array can be partially in registers, partially in 'local' memory, not either/or.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+class Tile:
+    r: qd.unpacked_array(32, qd.f32)
+```
+
+The annotation expands at struct-definition time into the `N` synthetic scalar fields. The AST transformer rewrites `obj.r[i]` (for any python-int / `qd.static`-resolved `i`) into a direct reference to the synthetic field `obj._r{i}`. The IR / PTX matches the hand-rolled named-field version exactly.
```
duburcqa: > The IR / PTX matches the hand-rolled named-field version exactly.

That is a bit low-level. Just saying you get the best of both worlds: clean syntax and performance (runtime or compile time?).

docs/source/user_guide/unpacked_vector.md
hughperkins: Same as above:

"if you have a vector, rather than a qd.unpacked_array, it either stores the entire vector in registers, or the entire vector in local memory (meaning, unintuitively: in off-chip DRAM, sloowwww 🐌).

The unpacked_array can be partially in registers, partially in 'local' memory, not either/or."

docs/source/user_guide/unpacked_array.md
```diff
@@ -0,0 +1,135 @@
...
+
+```python
+@qd.dataclass
+class Tile:
+    r: qd.unpacked_array(32, qd.f32)
+```
+
+The annotation expands at struct-definition time into the `N` synthetic scalar fields. The AST transformer rewrites `obj.r[i]` (for any python-int / `qd.static`-resolved `i`) into a direct reference to the synthetic field `obj._r{i}`. The IR / PTX matches the hand-rolled named-field version exactly.
+
+The name "unpacked" is a contrast with the packed-vector default: a packed group is one `alloca`, an unpacked group is `N` `alloca`s, one per slot. Whether the slots end up in registers is the optimiser's call; `unpacked_array` removes the layout obstacle that was preventing it.
```
duburcqa: I would avoid the notion of 'alloca' in user facing doc

docs/source/user_guide/unpacked_vector.md
hughperkins: Yes, this is what we have currently. "fun".

hughperkins: well, it's slow for one.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+    scale: qd.f32
+```
+
+The generated struct has 65 scalar members (`_a0..._a31`, `_b0..._b31`, `scale`).
+
+## When to reach for it
+
+Use `unpacked_array` when:
+
+- the group is *small and statically-sized*, and
```
duburcqa: What 'small' means?

docs/source/user_guide/unpacked_vector.md
hughperkins: slow to compile.

(and also slower at runtime).

(both, or one, or the other, depending on exact configuraiton of code)

docs/source/user_guide/unpacked_vector.md
hughperkins: no

docs/source/user_guide/unpacked_vector.md
duburcqa: What does 'statically-sized' means? Like, dynamic allocation is not supported by Quadrants from what I know.

docs/source/user_guide/unpacked_vector.md
hughperkins: the dataclass contains the unpacked array

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+- the kernel body accesses it with python-int / `qd.static`-resolved indices (typically unrolled inner loops), and
+- you have measured (or strongly suspect) that an equivalent `qd.types.vector(N, dtype)` is leaving slots in local memory under register pressure.
```
duburcqa: > - you have measured (or strongly suspect) that an equivalent `qd.types.vector(N, dtype)` is leaving slots in local memory under register pressure.

This is quite vague.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+
+A good signal is `ptxas` reporting non-zero "bytes spill stores / loads" for the kernel, or `ld.local` / `st.local` instructions in the generated PTX that don't correspond to a deliberate shared-memory access.
```
duburcqa: > A good signal is `ptxas` reporting non-zero "bytes spill stores / loads" for the kernel, or `ld.local` / `st.local` instructions in the generated PTX that don't correspond to a deliberate shared-memory access.

How one is supposed to check this?

docs/source/user_guide/unpacked_vector.md
hughperkins: I think it's the clearest way of stating that the generated code is identical across both versions? (I suppose we could say 'generated code is identical across both versions', but it's ambiguous about *which* generated code. Like, there is intermediate generated code within quadrants that will be different at certain points)

docs/source/user_guide/unpacked_array.md
hughperkins: Yeah, I agree. I remember saying the same on your own docs :). Will ponder.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+
+Prefer `qd.types.vector(N, dtype)` for small groups where register pressure is low and runtime indexing is needed — vectors keep all the usual arithmetic conveniences (element-wise ops, dot products, etc.) that `unpacked_array` does not.
+
+## Constraints and limitations
+
+- **Static indices only.** `t.r[k]` must resolve at compile time, i.e. `k` is a python-int literal or a `for k in qd.static(range(N)):` loop variable. A runtime-int index raises a `QuadrantsSyntaxError` at compile time with a message pointing at the `qd.static` requirement. If you need a runtime index over the group, spell out the cascade explicitly (`if k == 0: ...`).
```
duburcqa: You already mentioned this. It is weird to repeat yourself.

docs/source/user_guide/unpacked_array.md
```diff
@@ -0,0 +1,135 @@
...
+- you have measured (or strongly suspect) that an equivalent `qd.types.vector(N, dtype)` is leaving slots in local memory under register pressure.
+
+A good signal is `ptxas` reporting non-zero "bytes spill stores / loads" for the kernel, or `ld.local` / `st.local` instructions in the generated PTX that don't correspond to a deliberate shared-memory access.
+
+Prefer `qd.types.vector(N, dtype)` for small groups where register pressure is low and runtime indexing is needed — vectors keep all the usual arithmetic conveniences (element-wise ops, dot products, etc.) that `unpacked_array` does not.
+
+## Constraints and limitations
+
+- **Static indices only.** `t.r[k]` must resolve at compile time, i.e. `k` is a python-int literal or a `for k in qd.static(range(N)):` loop variable. A runtime-int index raises a `QuadrantsSyntaxError` at compile time with a message pointing at the `qd.static` requirement. If you need a runtime index over the group, spell out the cascade explicitly (`if k == 0: ...`).
+- **Static out-of-bounds is rejected at compile time.** `t.r[7]` on an `unpacked_array(4, ...)` group raises `QuadrantsSyntaxError: unpacked_array index out of bounds: r[7] (count=4)`.
```
duburcqa: That is a feature, not a "constraints and limitations"

docs/source/user_guide/unpacked_vector.md
hughperkins: small relative to the available register memory for each thread.

docs/source/user_guide/unpacked_array.md
```diff
@@ -0,0 +1,135 @@
...
+- **Storage only.** An `unpacked_array` group has no vector arithmetic. There is no `t.r + other`, no `t.r.dot(...)`, no broadcast operations. If you want those, use `qd.types.vector(N, dtype)` instead.
+- **`count` is fixed at struct-definition time.** It must be a positive python-int literal.
```
duburcqa: How is that supposed to be specific to 'unpacked_array'? It is the case for all Quadrants tensors no?

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,135 @@
...
+- **Static out-of-bounds is rejected at compile time.** `t.r[7]` on an `unpacked_array(4, ...)` group raises `QuadrantsSyntaxError: unpacked_array index out of bounds: r[7] (count=4)`.
+- **Storage only.** An `unpacked_array` group has no vector arithmetic. There is no `t.r + other`, no `t.r.dot(...)`, no broadcast operations. If you want those, use `qd.types.vector(N, dtype)` instead.
+- **`count` is fixed at struct-definition time.** It must be a positive python-int literal.
+- **Naming.** The synthetic fields use the convention `_{group_name}{i}` (e.g. `_r0`, `_r1`, ..., `_r31`). Avoid declaring your own field with a name that collides with one of those, or `StructType` will report a duplicate member.
```
duburcqa: How is that supposed to be a "constraints and limitations"

docs/source/user_guide/unpacked_vector.md
hughperkins: ndarrays and fields can both have parametrized allocation.
```
my_tensor = ndarray(some_count, i32)
my_tensor = field(some_count, i32)
```
(where some_count is a runtime variable)

docs/source/user_guide/unpacked_vector.md
hughperkins: It seems specific to me. It's the exact use-case. You think that the vector is ending up in local memory, not in registers, and thus making your program super sloowww 🐌 .

docs/source/user_guide/unpacked_array.md
hughperkins: ndarrays can be arbitrarly sizes, fields too, but cannot be used with `@qd.dataclass` anyway.

Within a `@qd.dataclass`, you can only use primtivies, vectors, matrix and unpacked_array. (maybe it should be unpacked_vector actually?). matrix, vector, unpacked_array and primitives all have fixed sizes, which are fixed at the time of defining the `@qd.dataclass`.

hughperkins: I think I'm going to:
- move unpacked_array into types namespace
- change it to use annotation syntax `qd.types.UnpackedArray[ti.i32, 32]`
- rename it to `unpacked_vector`, to avoid confusion

e.g. AI gets confused with unpacked array:

Screenshot 2026-05-29 at 10 21 45
Screenshot 2026-05-29 at 10 21 53

hughperkins: (the reason for moving into 'type's namespace was because, just now, when trying to craft examples, I did thigns like
```
t  = qd.unpacked_array()
```
... which is meaningless. I mean, it's just creating an annotation. But it *sound* like it's constructing an actual unpacked array.)

I'll make the changs first, then we can see how taht looks.

docs/source/user_guide/unpacked_vector.md
hughperkins: because you can't declare your own fields with the same name as the unpacked_array plus a conflicting numericl suffix. It's arguably not a 'limitation', but it seems clearly a 'constriant'?

docs/source/user_guide/unpacked_array.md
hughperkins: fair. we should probalby ust remove this bullet point altogeterh.

docs/source/user_guide/unpacked_vector.md
hughperkins: (so no more than around 64 or so, on the whole, (but i'ts not an exact number. It's a trade-off with occupancy etc)

docs/source/user_guide/unpacked_vector.md
hughperkins: note to self: explore this point.

docs/source/user_guide/unpacked_vector.md
duburcqa: It could be nice to add some introducing to this concept in the doc for "newbies", or guys like me willing to make best use of Quadrants.

hughperkins: Note: some more backgroud:

Screenshot 2026-05-29 at 11 00 47

docs/source/user_guide/unpacked_vector.md
hughperkins: Ok, so, more background:
- we are using the unpacked vector inside qd.dataclass, because we are targeting the use-case of Tiles, which are qd.dataclass
- there is currently no reuirement for something like unpacked vector direclty inside a kernel, and that would be a separate, and additional piece of work

docs/source/user_guide/unpacked_vector.md
hughperkins: and more:

Screenshot 2026-05-29 at 11 40 01

docs/source/user_guide/unpacked_vector.md
hughperkins: Good point. Adding a section for this.

hughperkins: (done)

docs/source/user_guide/unpacked_vector.md
hughperkins: (not sure if this was the best thread to choose to paste these screenshots. this thread became "the doc should explain this stuff better, which these comments arent addressing)

docs/source/user_guide/unpacked_vector.md
hughperkins: oh, you are commenting on the non-ascii? I see

docs/source/user_guide/compound_types.md
```diff
@@ -159,3 +159,5 @@ class Particle:
     vel: qd.types.vector(3, qd.f32)
     mass: qd.f32
 ```
+
+For larger statically-indexed groups (e.g. a row or tile of scalars that you want SROA to register-promote independently), see {doc}`unpacked_vector`.
```
duburcqa: You are still using the acronym SROA without defining it anywhere.

docs/source/user_guide/compound_types.md
duburcqa: I don't think this comment is address for now. Still not understandable by an average user.

docs/source/user_guide/unpacked_vector.md
duburcqa: Did you add these details in the doc?

duburcqa: Close this conversation if you think it is resolved.

docs/source/user_guide/unpacked_vector.md
duburcqa: Did you add these details in the doc? Close this conversation if you think it is resolved.

docs/source/user_guide/unpacked_vector.md
duburcqa: yes

duburcqa: Did you add these details in the doc? Close this conversation if you think it is resolved.

docs/source/user_guide/unpacked_vector.md
duburcqa: Did you add these details in the doc? Close this conversation if you think it is resolved.

docs/source/user_guide/unpacked_vector.md
duburcqa: What about starting with "generated code is identical across both versions" then add "i.e. the IR / PTX matches the hand-rolled named-field version exactly." for advanced users that understand what is truly going on? This way to have multiple level of understanding, to give valuable information of users of varying expertise.

docs/source/user_guide/unpacked_vector.md
duburcqa: I'm still not sure to understand.

docs/source/user_guide/unpacked_vector.md
duburcqa: Sure, but it is vague in a sense that I have no idea how to "measure or strongly suspect that an equivalent qd.types.vector(N, dtype) is leaving slots in local memory under register pressure" in practice

duburcqa: So this advice is not really helping me personally.

docs/source/user_guide/unpacked_vector.md
duburcqa: Close this conversation if you think it is resolved.

docs/source/user_guide/unpacked_array.md
duburcqa: Close this conversation if you think it is resolved.

docs/source/user_guide/unpacked_array.md
duburcqa: understood.

docs/source/user_guide/unpacked_vector.md
duburcqa: > because you can't declare your own fields with the same name as the unpacked_array plus a conflicting numericl suffix.

Ok now I get it. It is clearly a limitation indeed. Maybe the wording could be improved because I did not get it before your explanation.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
+# Unpacked vector
+
+`qd.types.UnpackedVector[dtype, N]` is a `@qd.dataclass` field annotation that declares a group of `N` scalars exposed under a single name via indexed syntax (`t.r[i]`), laid out as `N` independent slots so the compiler can keep each one in a register independently.
```
duburcqa, line 3: "the compiler can keep each one in a register independently" --> for efficiency?

docs/source/user_guide/unpacked_vector.md
hughperkins: Sounds good.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
...
+
+## What problem does it solve?
+
+The natural way to group `N` scalars on a per-thread struct is a vector member:
```
duburcqa, line 7: "The natural way" --> "The naive way" ?

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
...
+
+```python
+@qd.dataclass
+class Tile:
+    r: qd.types.vector(32, qd.f32)
+```
+
+A `qd.types.vector(N, dtype)` is stored as a single group. The compiler either keeps the whole group in registers, or, if registers are tight, spills the whole group to "local" memory (which despite the name lives in off-chip DRAM and is slow). It is all-or-nothing per group.
```
duburcqa, line 15: "if registers are tight" --> what does this mean? Maybe it is just my English that is bad.

docs/source/user_guide/unpacked_vector.md
hughperkins: 'natural' is the correct word.

In general, when we want to group things in a kernel, we use a qd.vector. For example, Genesis uses qd.vectors extensively, to represent positions and quaternions I think?

docs/source/user_guide/unpacked_vector.md
hughperkins: it all comes back to https://github.com/Genesis-Embodied-AI/quadrants/pull/712#discussion_r3324490769 Let me know your thoughts on how best to rephrase this.

hughperkins: (see also the screenshots later in that thread)

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
...
+
+`qd.types.UnpackedVector` is the named-field layout with the indexed syntax restored:
+
+```python
+@qd.dataclass
+class Tile:
+    r: qd.types.UnpackedVector[qd.f32, 32]
+```
+
+The annotation expands at class-definition time into the `N` named-field equivalent, and the AST transformer rewrites each `obj.r[i]` (for python-int / `qd.static`-resolved `i`) into a direct reference to the i-th slot. So you get the best of both worlds: a clean indexed write/read at the source level, and the per-slot register residency of the hand-rolled named-field version.
```
duburcqa, line 40: "the AST transformer rewrites each " --> I don't think users are aware of this "AST transformer" in the first place.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
...
+
+## When to reach for it
+
+Use `UnpackedVector` when:
+
+- the group is *small relative to the per-thread register budget* (roughly speaking, a few dozen slots on current NVIDIA GPUs; an SM gives each thread up to 255 32-bit registers, shared with all live state in the kernel, so what counts as "small" depends on what else the kernel is holding), and
+- the kernel body accesses it with python-int / `qd.static`-resolved indices (typically unrolled inner loops), and
+- an equivalent `qd.types.vector(N, dtype)` version of the same kernel is, or might be, spilling the group to local memory under register pressure. See [How to check for spills](#how-to-check-for-spills) below for the concrete workflow to confirm a spill is actually happening.
+
+Prefer `qd.types.vector(N, dtype)` for small groups where register pressure is low and runtime indexing is needed -- vectors keep all the usual arithmetic conveniences (element-wise ops, dot products, etc.) that `UnpackedVector` does not.
```
duburcqa: " is low and runtime indexing is needed" -->  OR, not and ?

duburcqa: "--" on purpose ?

docs/source/user_guide/unpacked_vector.md
hughperkins: "if available registers are few".

i.e. when a kernel is running on a gpu, each thread has a certain number of registers available. The number of registers isn't actually fixed to a specific amount per se, though it cannot be more than some number like 255, as far as I know. But hte more registers each thread uses, the poorer is the occupany of that kernel, which slows down performance, because it reduces latency hiding. So we'll tend to use way fewer than 255 registers in practice, like maybe 32-64 or so. That limits the number of things you can store in registers. Anything that won't fit in registers will need to be stored in 'local' memory instead, which, despite the name, is the same physical memory as global memory: i.e. off-chip, DRAM, super slow.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Constraints and pitfalls
+
+- **Use `@qd.dataclass`, not `@dataclasses.dataclass`.** The `UnpackedVector[dtype, count]` annotation is only expanded by `@qd.dataclass`; on a stdlib dataclass the annotation is inert metadata and the indexed-access syntax will not work. Subscripting or calling the marker outside that context raises a `QuadrantsSyntaxError`.
```
duburcqa, line 102: "stdlib" is not standard for python.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
...
+- **Subscript syntax, not a function call.** Write `r: qd.types.UnpackedVector[qd.f32, 32]`. There is no `qd.types.UnpackedVector(qd.f32, 32)` call form. The subscript spelling makes the marker visually read as a type annotation, not a runtime value.
```
duburcqa: Not sure if this is necessary to explain. Like, you never mentioned it was possible and Python by default does not support this, so users would simply expect it does not work anyway.

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
...
+- **Static indices only.** `t.r[k]` must resolve at compile time, i.e. `k` is a python-int literal or a `for k in qd.static(range(N)):` loop variable. A runtime-int index raises a `QuadrantsSyntaxError` at compile time. If you need a runtime index over the group, spell out the cascade explicitly. This is by design: the whole point of the layout is that each slot can be promoted to its own register, which requires every access site to name a specific slot at compile time.
```
duburcqa: This was already mentioned no?

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
...
+- **No vector arithmetic.** An `UnpackedVector` group is storage only. There is no `t.r + other`, no `t.r.dot(...)`, no broadcast operations. If you want those, use `qd.types.vector(N, dtype)` instead.
+- **Synthetic field-name collisions are rejected.** The expansion uses the convention `_{group_name}{i}` (e.g. `_r0`, `_r1`, ..., `_r31`). If you declare your own field with one of those names, `@qd.dataclass` raises a `QuadrantsSyntaxError` pointing at the collision.
```
duburcqa, line 104: When will this exception be risen?

docs/source/user_guide/unpacked_vector.md
```diff
@@ -0,0 +1,192 @@
...
+| `qd.types.vector(N, dtype)`         | one slot holding `N` packed scalars     | yes              | small groups with vector arithmetic   |
+| `qd.types.UnpackedVector[dtype, N]` | `N` independent slots                   | no               | groups that need to stay register-resident under pressure |
+
+Under low register pressure the three options generate similar code. Under high register pressure `UnpackedVector` is the one most likely to stay in registers because the optimiser can promote each slot independently.
+
+# Advanced
+
+## How the layout works at the LLVM level
+
+A `qd.types.vector(N, dtype)` field lowers to a single `alloca` of `N` packed scalars. LLVM's SROA + `mem2reg` passes attempt to decompose that `alloca` into `N` per-slot SSA values so each can live in a register, but the decomposition is conservative: under high register pressure (e.g. two concurrent `32x32` tiles in a Cholesky + triangular solve), SROA bails out on the packed `alloca` and the whole group spills to local memory as a unit. Each access then becomes a `ld.local` / `st.local`.
```
duburcqa, line 120: I understand that this section is advanced, but could you make this a little bit more accessible? With 2 levels of details maybe, as mentioned before.

docs/source/user_guide/unpacked_vector.md
hughperkins: good spot. should be 'is low or runtime indexing is not needed", I think, right.

hughperkins: @ai please remove double hyphens.

ai: Please, don't use my user name for LLM command prefix

docs/source/user_guide/unpacked_vector.md
hughperkins: 🤔

docs/source/user_guide/unpacked_vector.md
hughperkins: 😅

docs/source/user_guide/unpacked_vector.md
duburcqa: No problem!

docs/source/user_guide/unpacked_vector.md
duburcqa: This is interesting. I don't know if there is a place to mention this in the doc. Maybe some introduction doc about how GPU works where you explain how to get the most out of Quadrants?

docs/source/user_guide/compound_types.md
hughperkins: addressed now (by hand) maybe 🤞

docs/source/user_guide/compound_types.md
hughperkins: addressed

hughperkins:
closing this for now, to reduce confusion ,and keep pr list clean

PR #714, opened 2026-05-28
[Perf] Move register-tile Cholesky optimizations from Genesis back into quadrants

hughperkins:
Apply the inlined column-setter cascades + qd.static-unrolled cholesky_ body (split partial sums + per-lane running row-norm) to `_Tile16x16` and add a matching `_Tile32x32` in `_tile16.py` / `_tile32.py`. The slice proxies (`_TileSliceProxy`, `_VecSliceProxy`, `_TileRefProxy`, `_OuterProduct`) are factored to be tile-size-agnostic, and `tile_slicing.py` now dispatches on either tile cache, so the existing `tile[:] = arr[r:r2, c:c2]` syntax works identically for both 16x16 and 32x32.

Code originates from `genesis/utils/_tile16.py` / `_tile32.py`, which were local copies maintained in genesis pending this move.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


docs/source/user_guide/tile.md
```diff
@@ -1,10 +1,15 @@
-# Tile16x16: register-resident 16x16 tiles
+# Register-resident tiles: `Tile16x16` and `Tile32x32`
 
-`Tile16x16` provides a 16x16 matrix tile that lives entirely in registers, distributed across 16 threads in a subgroup (warp). Each thread holds one row as 16 scalar registers. Cross-thread communication uses subgroup shuffles — no shared memory needed.
+Quadrants provides two register-resident matrix tile types:
 
-This is useful for implementing blocked linear algebra kernels (Cholesky, triangular solve, etc.) where you want to keep working data in registers for maximum throughput.
+- `qd.simt.Tile16x16` — a 16x16 tile distributed across 16 threads in a subgroup (one row per thread, 16 scalar registers per thread).
+- `qd.simt.Tile32x32` — a 32x32 tile distributed across 32 threads in a subgroup (one row per thread, 32 scalar registers per thread).
```
duburcqa, line 6: Out of curiosity, are they both deriving from some `BaseTile` or `AbstractTile` ?

docs/source/user_guide/tile.md
```diff
@@ -167,8 +172,35 @@ Pass `dtype=qd.f64` for double precision:
 t = qd.simt.Tile16x16.zeros(dtype=qd.f64)
 ```
 
+## `Tile32x32`
+
+The 32x32 sibling is used the same way as `Tile16x16`, just with `block_dim=32` and `SIZE == 32`:
```
duburcqa, line 177: You already mentioned "everything carries over to `Tile32x32` by swapping the class name and using `SIZE == 32` / `block_dim=32`. The [`Tile32x32` section](#tile32x32) below has the dispatch rules and a worked example." ?

docs/source/user_guide/tile.md
```diff
@@ -167,8 +172,35 @@ Pass `dtype=qd.f64` for double precision:
...
+### When to pick 32x32 vs 16x16
+
+| Size  | Threads per block | Registers per thread | Best for |
+|-------|------------------:|---------------------:|----------|
+| 16x16 | 16                | 16                   | Small problems where occupancy from many narrow blocks matters; very small N (e.g. N≤16 or N≈48) where the 32-tile would waste lanes |
+| 32x32 | 32                | 32                   | Larger problems (N ≳ 32) where bigger tiles cut the number of blocked passes and the FMA chain inside `cholesky_` amortizes the larger register file |
+
 ## Method reference
 
+The table below shows the 16x16 API; the `Tile32x32` API is identical, swap the class name and use `SIZE == 32`.
```
duburcqa: Already mentioned no?

duburcqa:
OK to merge.

docs/source/user_guide/tile.md
hughperkins: They are not. Does this affect your choices when using the tiles?

docs/source/user_guide/tile.md
hughperkins: softened the language in the previous reference.

docs/source/user_guide/tile.md
duburcqa: No. Just I think it is a design issue if it is not the case. Because it means that there is 0 guarantee to have compatible API between both, and it must be implemented specifically in both cases, which is likely to cause code duplication. Finally, It makes typing more annoying, because one has to use Union of all variants instead of just base class.

hughperkins: > It makes typing more annoying, because one has to use Union of all variants instead of just base class

I feel like this contradicts your earlier 'No'.

I feel like code duplication etc is out of scope, as long as the implementations match the doc.

However, typing is a plausible reason.

But... these classes are only used inside kernels. Aren't we currently simply not using annotations when we call into qd.func's?

duburcqa: > I feel like code duplication etc is out of scope, as long as the implementations match the doc.

Out of scope of what exactly? I'm reviewing your PR.

duburcqa: > But... these classes are only used inside kernels. Aren't we currently simply not using annotations when we call into qd.func's?

We are, but it would be nice to slowly move towards full typing I think.

duburcqa: > I feel like this contradicts your earlier 'No'.

Not really. it is a minor inconvenience in terms of usability.

hughperkins: > Out of scope of what exactly? I'm reviewing your PR.

Reviewing code, rather than doc/utilization.

hughperkins: > We are, but it would be nice to slowly move towards full typing I think.

Yes, I agree. Strongly.

Will take a look. I'm not sure if it's actually possible to have base classes for these `qd.dataclass` types. I've barely used them, but I suspect that, since they are structs (by value), there is no inheritance, the same as for C structs.

Note that I don't think we support union types in Quadrants either. Though I'm up for handling that sometime somehow too.

hughperkins: Well, with a few modifications (not in this branch, nor in main), it appears inheritance works ok for qd.dataclass:


```
import quadrants as qd

qd.init(qd.cpu)

@qd.dataclass
class Parent:
	a: int

	@qd.func
	def foo(self):
		print('foo', self.a)

@qd.dataclass
class Child1(Parent):
	b: int

	@qd.func
	def bar(self):
		print('bar', self.b)

@qd.dataclass
class Child2(Parent):
	c: int

@qd.kernel
def k1(p: Parent, c: Child1, d: Child2) -> None:
	print('k1')
	print(p.a)
	p.foo()
	print(c.a, c.b)
	print(d.a, d.c)
	c.foo()
	c.bar()

p = Parent()
print(p.a)

c = Child1()
print(c.a, c.b)

d = Child2()
print(d.a, d.c)

k1(p, c, d)
```

output:
```
[Quadrants] version 0.0.0, llvm 22.1.0, commit 8a7ead44, osx, python 3.10.20
[Quadrants] Starting on arch=arm64
0
0 0
0 0
k1
0
foo 0
0 0
0 0
foo 0
bar 0
```

hughperkins: Mind you, that doesnt yet show that it's possibel to have an abstract base class 🤔

hughperkins: Unfortunately, we cna't re-use methods across classes, because they all contain things like:

Screenshot 2026-06-02 at 11 53 16

This might change after we merge https://github.com/Genesis-Embodied-AI/quadrants/pull/712

But I want to first move the existing Tiles code back into Quadrants, from Genesis, before messing around with it, ideally.

PR #718, opened 2026-06-02
[Type] Add unpacked form for qd.vector for indexed register access

hughperkins:
## Summary

Adds an `unpacked=True` per-thread storage-layout kwarg to `qd.types.vector` for use on `@qd.dataclass` field annotations:

```python
@qd.dataclass
class Tile:
    r: qd.types.vector(32, qd.f32, unpacked=True)
```

The value type is a vector. The kwarg declares that *storage* is unpacked: `N` independent scalar slots (one `alloca` each) instead of one packed slot. The optimiser can then promote each slot to a register independently and spill only the ones it has to, instead of having to commit to keeping or spilling the whole group as a unit. The motivation is high register pressure -- e.g. several concurrent 32x32 tiles in a Cholesky + triangular solve -- where LLVM's SROA + `mem2reg` bails out on a packed alloca and the whole vector ends up in local memory.

Today only indexed access (`obj.r[i]` for python-int / `qd.static`-resolved `i`) is exercised on an unpacked field; vector arithmetic / reductions / pass-by-value / runtime indexing / shared-memory copy as a unit on unpacked fields are planned follow-ups -- they only need an AST-side materialisation of the field as a vector value, not new types.

## What's added

- **API surface**: `unpacked: bool = False` kwarg on `qd.types.vector(...)` and `VectorType.__init__`. Plain `qd.types.vector(N, dtype)` is unchanged.
- **Field expansion**: when a `@qd.dataclass` field is annotated with a vector type whose `_is_unpacked` flag is set, `StructType.__init__` expands it into `N` synthetic scalar members `_r0.._r{N-1}`.
- **AST rewriting**: `ASTTransformer.build_Attribute` recognises `obj.r` as an unpacked-group access via a `_qd_unpacked_groups` tag on the struct, and `build_Subscript` resolves `obj.r[i]` to a direct reference to the synthetic `_r{i}` field.
- **Metadata propagation**: the `_qd_unpacked_groups` tag rides on every `Struct` / `StructField` instance the type produces -- through nested `cast()`, `filled_with_scalar()`, and per-index `impl.subscript` paths -- so `outer.inner.r[i]` and `tile_field[n].r[i]` (including nested) all resolve correctly.
- **Misuse guards**: a vector type declared with `unpacked=True` refuses to be instantiated as a value (`T(...)`, `T.field(...)`, `T.ndarray(...)`) outside a `@qd.dataclass` field annotation, with an error pointing at the dataclass-annotation usage. Synthetic-field-name collisions (a user-declared field shadowing a future `_r{i}`, or vice versa) raise `QuadrantsSyntaxError` at class-decoration time.
- **User-guide doc**: `docs/source/user_guide/unpacked.md` covers motivation, usage, constraints, the layout-vs-LLVM relationship, and a workflow for confirming spills (PTX dump + `ptxas -v`, `ld.local` grep, Nsight Compute, post-optimisation LLVM IR).

## Test plan

- [x] `pre-commit run -a` clean (black / clang-format / trim-whitespace / ruff / pylint).
- [x] `tests/python/test_unpacked.py`: 13/13 pass on CUDA (RTX PRO 6000 Blackwell) -- 8 kernel tests cover the full AST-rewrite + NVPTX codegen path; 5 cover the python-side misuse / collision guards.
- [x] `tests/python/test_api.py`: 18/18 pass on CUDA -- confirms public-surface golden is unchanged (the kwarg form adds no new top-level symbols).

## Follow-ups (not in this PR)

- AST-side materialisation of `obj.r` as a vector value, so vector arithmetic / `.norm()` / runtime indexing / pass-by-value across `@qd.func` boundaries / shared-memory copy as a unit all start working on unpacked-layout fields.
- Kernel-local unpacked allocations (i.e. usable outside of `@qd.dataclass`).


docs/source/user_guide/matrix_vector.md
```diff
@@ -98,3 +98,130 @@ large = qd.field(qd.f32, shape=(64, 64))
...
+
+| If you...                                                                                                | Use                              |
+|----------------------------------------------------------------------------------------------------------|----------------------------------|
+| want vector arithmetic (`a + b`, `a.dot(b)`, `a.norm()`, swizzle, broadcast) at the use site             | `unpacked=False` (the default)   |
+| need to index with a *runtime* integer (`v[k]` where `k` isn't a python-int)                             | `unpacked=False`                 |
+| need to pass the vector across a `@qd.func` boundary, copy it into shared memory, or instantiate it as a kernel-local value | `unpacked=False`                 |
+| are storing the field in a smallish struct, register pressure is low, and the kernel works as-is        | `unpacked=False`                 |
+| only ever access the field as `obj.r[i]` with python-int / `qd.static`-resolved `i` (unrolled inner loops), the group is small relative to the per-thread register budget, *and* the packed version is spilling under register pressure | `unpacked=True`                  |
```
duburcqa: I'm not sure it is the correct framing. basically, `unpacked=True` is strictly more limited on all axes compared to `unpacked=False`, so basically it is only relevant in one specific case, to avoid spilling under register pressure. If this happens, the way to go may be to adapt the code so that using `unpacked=True` because possible, at the cost of increasing the code complexity for extra perf boost. Which adaptation the code may require is exactly what you mention in this list.

hughperkins: Not sure I follow
- why do you feel using unpacked=True incresaes code complexity?
- why do you feel this framing is not correct? How aer you thinking of framing it?

duburcqa: What increases the code complexity is doing all the necessary modification to accommodate the requirements for being able to use `unpacked=True`.

To frame it differently: always use `unpacked=False`, but if for some reason to need to squeeze more performance and you discover that you have a spilling issue to due register pressure, then you may consider switching to `unpacked=True` and doing all the necessary modifications to comply with the requirements to mentioned in this table, which is likely to increase code complexity somehow.

hughperkins: empirically, in the one case we use them currently, i.e. tiles, we don't need any of the things in the table. No increased complexity. This is why I feel the proposed framing is not clearly accurate.

hughperkins: I'm fine with "To frame it differently: always use unpacked=False, but if for some reason to need to squeeze more performance and you discover that you have a spilling issue to due register pressure, then you may consider switching to unpacked=True". But I dont agree with "doing all the necessary modifications". They aren't really 'modifications', it's more like, if you need to do any of these things, then you cannot use unpacked I think?

hughperkins: Note:
- I think "need to pass the vector across a `@qd.func` boundary," is false, since it's inlined anyway.
- the arithmetic bit is a bit of a 'straw man', since we could implment arithmetic on unpacked; we just dont have any such requirement currently
- "are storing the field in a smallish struct, register pressure is low, and the kernel works as-is" isnt really a cosntraint, doesnt imply increasing code complexity

The only thing we are left with really is:
- "need to index with a *runtime* integer (`v[k]` where `k` isn't a python-int".

This isn't really a 'complexity' issue. If you need this, you cannot use unpacked.

hughperkins: I think the main trade-off might be compile time versus runtime. We have three optoins:
- use vector => fastest compile time, slowest runtime
- use hand coded scalars => slowest compile time, fastest runtime
- use unpacked => medium compile time, also fastest runtime

hughperkins: (also, hand coded scalars is a maintenance nightmare...)

duburcqa: I think with all these information, it draws a pretty faithful picture of the situation. Would you mind structure this and add it to the documentation?

duburcqa:
ok to merge

PR #722, opened 2026-06-05
[Tile] Use unpacked vector field for Tile16x16/Tile32x32 register storage

hughperkins:
## Summary

Replace the hand-rolled ``r0..rN-1: dtype`` field declarations and their matching cascades in `Tile16x16` / `Tile32x32` with a single

```python
r: qd.types.vector(_TILE, dtype, unpacked=True)
```

field, accessed as `self.r[k]`.  With python-int / `qd.static`-resolved indices the unpacked vector still maps to one independent register slot per use, so the generated PTX/LLVM IR is unchanged — but the source shrinks dramatically (net -870 lines).

Also drops the now-redundant private helpers (`_get_col`, `_set_col`, `_r`) and the `_REGS` field-name table.  These were all `_`-prefixed and only used internally to the two tile modules.

## Test plan

- `pre-commit run -a` (black, ruff, pylint): clean
- `pyright python/quadrants/lang/simt/_tile16.py python/quadrants/lang/simt/_tile32.py`: 0 errors
- `python tests/run_tests.py -v -t1 test_tile` on an RTX PRO 6000 cluster node: **732 passed, 182 skipped, 0 failed** (~10 min); covers cuda+vulkan, f32+f64, ndarray+field for both tile sizes, including `cholesky_`, `solve_triangular_`, `qd.outer(...)` rank-1 updates, slice load/store, and the blocked-Cholesky demo
- Existing 68 `PURE.VIOLATION` warnings on `TILE` / `SIZE` test globals are pre-existing and unrelated


hughperkins:
Benchmarks on genesis:

20260609_tiles_unpacked

dex_hand regression seems concerning 🤔

hughperkins:
tests pass at least:

Screenshot 2026-06-08 at 12 45 19

hughperkins:
Genesis benchmarks:

20260629_tiles

hughperkins:
Looking ok now (but we need to merge the latest quadrants relase into Genesis, so we have a fair comparison)

20260714_tiles

hughperkins:
Benchmarks looking ok:

20260715_tiles_1157

hughperkins:
Genesis unit tests passing:

Screenshot 2026-07-15 at 14 09 28

duburcqa:
ok to merge

PR #725, opened 2026-06-05
[Graph] Add qd.checkpoint

hughperkins:


docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
 Therefore on unsupported platforms, you might consider creating a second implementation, which works differently. e.g.:
 - fixed number of loop iterations, so no dependency on gpu data for kernel launch; combined perhaps with:
 - make each kernel 'short-circuit', exit quickly, if the task has already been completed; to avoid running the GPU more than necessary
+
+## Checkpoints with `qd.checkpoint` *(experimental)*
```
hughperkins, line 180: Note: after the backwards-incompatible disaster taht was the algorithms.md changes for qipc, that kept changing 😅  I think I'd like to mark things as 'experimental' for a few weeks/months, until we are confident the api is stable.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+> **Status:** Implemented end-to-end on every supported backend. The Python contract (`qd.checkpoint`, `GraphStatus`, `kernel.resume(from_checkpoint=…)`) is identical everywhere; only the device-side lowering differs:
+>
+> - **CUDA SM 9.0+**: each checkpoint becomes a CUDA-graph IF conditional node. The gate kernel (`_qd_checkpoint_if_gate`) reads `*resume_point` and writes the IF handle; `yield_on=` checkpoints append a tiny device-side yield-check kernel inside the IF body that atomic-CASes `yield_signal`. The WHILE-with-yield condition kernel reads `yield_signal` for early exit.
```
hughperkins: I think we can probalby cut all this junk (or move it to an 'advanced' section perhaps?)

hughperkins: (we already have the table at the top of this doc right?)

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+> - **AMDGPU**: GPU-side via the same codegen-prologue + flat HIP graph approach used on pre-Hopper CUDA. HIP 7.2 has no conditional-node API and no indirect-dispatch primitive, so every cp_id >= 0 body kernel carries an LLVM-IR early-return prologue that reads `RuntimeContext::checkpoint_resume_point_ptr` / `checkpoint_yield_signal_ptr`, and a pre-built yield-check kernel (bundled HSACO via `scripts/build_checkpoint_yield_check_hsaco.py`) is inlined after each yielding checkpoint's body to atomically update `yield_signal`. The streaming path (graph_do_while + checkpoint) uses the same prologue and yield-check kernel.
+
+`qd.checkpoint()` marks a section of a graph kernel as a *skippable, optionally yieldable stage*. The intended use is the qipc-style re-entrant Newton loop, where each iteration is divided into a handful of named stages (assemble, broadphase, line search, ...) and the host may need to grow a pre-allocated buffer when a stage detects overflow.
```
hughperkins: Dont reference qipc in doc.

hughperkins: You can mention the abstract general form of the usage, just not mention qipc.

Like:

"An example use-case is an algorithm implemented in cuda graph, where you might need to allocate addiitonal memory in the middle of the graph, and the graph operations are in-place, so simply retrying the entire graph is not an option."

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+                pass
+        with qd.checkpoint():                       # cp_id 2: line search
+            for i in range(arr.shape[0]):
+                # ...
+                pass
+```
+
+Each `with qd.checkpoint(...)` block gets a `cp_id` assigned by declaration order (0, 1, 2, ... flat across the whole kernel, independent of whether the checkpoint is inside or outside a `qd.graph_do_while`).
+
+### Yield mechanism (CUDA, Vulkan, Metal, CPU/x64)
```
hughperkins: if it runs on all plaatforms, then dont put a list of the supported platforms at all.

hughperkins: Wait, are we missing amd? 🤔

docs/source/user_guide/graph.md
hughperkins: Note to AI: this is not a request to you, it is an obseration for other human reviewers.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+If `yield_on=foo` is supplied, the body may write a non-zero value into the `foo` ndarray to signal "the host needs to handle something" (typically: pre-allocated buffer too small). At the end of the checkpoint body the framework injects a small yield-check step that:
```
hughperkins: replace 'typically' with 'for example'

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+1. Reads `*foo` and skips out early if it's `0`.
```
hughperkins: what is `*foo`? We are writing code in python here.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+2. Atomically claims `yield_signal` with this checkpoint's `cp_id` (first yielder in declaration order wins).
+3. Disables every later checkpoint in the same launch (subsequent gate steps see "I'm past the yield point" and short-circuit their IF bodies).
```
hughperkins: this all sounds really compliated, and is talking about how it's implemetned behind hte scenes. Can you describe what it does conceptually.

docs/source/user_guide/graph.md
hughperkins: I think you mean `foo[()]`, right?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+4. Resets `*foo` to `0` so the next launch starts clean without host intervention.
+
+Combined with `qd.graph_do_while`, the framework also handles WHILE early-exit: as soon as a yield has been observed, the loop terminates. Without that the loop body would re-enter, see "post-yield" state in every gate, skip every checkpoint, never decrement the counter, and spin forever.
```
hughperkins: again, this seems like something you can express more succinctly, "qd.checkpoint will exit any surrounding graph do while"

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+All GPU backends implement the yield-check as a tiny device kernel:
```
hughperkins: far too complicated for this high level section. probably can cut altogether. The user ony needs to know wehther the qd.checkpoint runs on the host or the gpu, and same for the yield.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+- **CUDA SM 9.0+**: `_qd_checkpoint_yield_check` kernel appended inside each checkpoint's IF conditional body; the conditional gate skips it when the checkpoint is skipped.
+- **CUDA SM < 9.0**: the same `_qd_checkpoint_yield_check` kernel inserted inline after the checkpoint's last body kernel in the flat graph; it self-gates on `*resume_point` / `*yield_signal` so a skipped checkpoint's yield-check is a no-op.
+- **Vulkan / Metal**: SPIR-V `checkpoint_yield_check_shader` indirect-dispatched with the same per-checkpoint `dim3` buffer the body uses, so a `(0, 0, 0)` dispatch (skipped checkpoint) no-ops.
+- **CPU**: a host statement at the end of each checkpoint body that reads the `yield_on=` flag and atomic-CASes `yield_signal`.
+
+Per-launch lifecycle is identical across backends: `yield_signal` and `resume_point` reset on the first yield-capable launch, first-yielder-wins, `*flag = 0` reset, WHILE early-exit, per-iteration `resume_point` reset.
```
hughperkins: `flag[()]`

hughperkins: if something is identical across backedns, you can remove this sentence altogether.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+- **CUDA SM < 9.0**: the same `_qd_checkpoint_yield_check` kernel inserted inline after the checkpoint's last body kernel in the flat graph; it self-gates on `*resume_point` / `*yield_signal` so a skipped checkpoint's yield-check is a no-op.
+- **Vulkan / Metal**: SPIR-V `checkpoint_yield_check_shader` indirect-dispatched with the same per-checkpoint `dim3` buffer the body uses, so a `(0, 0, 0)` dispatch (skipped checkpoint) no-ops.
+- **CPU**: a host statement at the end of each checkpoint body that reads the `yield_on=` flag and atomic-CASes `yield_signal`.
+
+Per-launch lifecycle is identical across backends: `yield_signal` and `resume_point` reset on the first yield-capable launch, first-yielder-wins, `*flag = 0` reset, WHILE early-exit, per-iteration `resume_point` reset.  
+The aggregated `yield_signal` is read back to the host once per launch (`cuMemcpyDtoH` on CUDA, `Device::readback_data` on Vulkan / Metal, a plain pointer read on CPU). Pinned-host yield flags are a future enhancement to remove the read-back stall.
```
hughperkins: remove all this advanced stuff. Just say how it behaves, from the user point of view. Remove the 'future enhancements' bit.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+### Host-side yield / resume loop
+
+Kernels with at least one `yield_on=` checkpoint return a `qd.GraphStatus` from every launch (and from `kernel.resume(...)`). This is true on every backend that implements the gate (CUDA SM 9.0+ via IF conditional nodes, CUDA SM < 9.0 via codegen prologue + flat graph, AMDGPU via codegen prologue + flat HIP graph, Vulkan / Metal via gate shader + indirect dispatch, CPU via host-branch gating). The status carries two fields:
```
hughperkins: remove all this detailed stuff about all platforms. Only the first sentence is needed from this paragraph. And the "The status carries two fields:' bit.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+                         from_checkpoint=status.checkpoint)
+```
+
+Kernels with `qd.checkpoint()` but no `yield_on=` keep their previous return contract (typically `None`) — the `GraphStatus` surface is opt-in via `yield_on=`.
+
+### Restrictions (enforced at kernel compile time)
+
+- Must be used inside `@qd.kernel(graph=True)`.
+- `yield_on=` (when supplied) must be the bare name of a kernel parameter that is a 0-d `qd.types.ndarray(qd.i32, ndim=0)`.
+- Checkpoints cannot be nested inside other checkpoints. A checkpoint inside a `qd.graph_do_while` body is fine and is the expected pattern.
```
hughperkins: remove the second sentence, i.e. "A checkpoint inside a `qd.graph_do_while` body is fine and is the expected pattern."

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+- Cannot be combined with `qd.stream_parallel()` in the same kernel.
```
hughperkins: this is alreayd implied by using it with graph, so remove 'Cannot be combined with `qd.stream_parallel()` in the same kernel.'

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+### Authoring tip: every statement inside a checkpoint must live in a top-level `for` loop
```
hughperkins: why is this a 'tip'? it sounds like a requirement, so remvoe 'authoring tip'

docs/source/user_guide/graph.md
```diff
@@ -154,3 +158,107 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+The checkpoint gate (a CUDA-graph IF node on CUDA SM 9.0+, a codegen-emitted LLVM-IR early-return on CUDA SM < 9.0, an indirect-dispatch gate shader on Vulkan / Metal, a host task-loop branch on CPU) routes work into a checkpoint's body based on whether the statement was lowered to a `range_for` task with the matching `checkpoint_id`. Bare scalar statements (e.g. `counter[()] -= 1`) fall into the offloader's pending-serial bucket, which currently loses the surrounding `checkpoint_id` and emits the work as a `serial` task with `checkpoint_id == -1`. Tasks with `cp_id == -1` run unconditionally, outside every checkpoint gate — meaning a yielding checkpoint won't actually skip them on subsequent checkpoints in the same launch.
```
hughperkins: this is all way too omplicated. just say what the user actually needs to know in order to use the feature.

hughperkins (commented):
adding comments

docs/source/user_guide/graph.md
```diff
@@ -4,14 +4,13 @@ Graphs reduce kernel launch overhead by capturing a sequence of GPU operations i
 
 ## Backend support
 
-Both features run on every backend. They are *hardware accelerated* on CUDA (via CUDA graphs) and AMDGPU (via HIP graphs); `graph_do_while` additionally requires CUDA SM 9.0+ / Hopper for its hardware-accelerated path. On other backends, `graph=True` is silently ignored and the kernel runs via the normal launch path, and `graph_do_while` falls back to a host-side do-while loop that copies the condition value GPU → host each iteration (causing a pipeline stall — see [Caveats](#caveats)).
+All three features run on every backend. `graph=True` is *hardware accelerated* on CUDA (via CUDA graphs) and AMDGPU (via HIP graphs), and silently ignored on other backends (the kernel runs via the normal launch path). `graph_do_while` is hardware accelerated only on CUDA SM 9.0+ / Hopper; everywhere else it falls back to a host-side do-while loop that copies the condition value GPU → host each iteration (causing a pipeline stall — see [Caveats](#caveats)). `qd.checkpoint` gating is fully GPU-side on every GPU backend.
```
hughperkins: why state this when it's already in the table? Be more concise.

docs/source/user_guide/graph.md
```diff
@@ -4,14 +4,13 @@ Graphs reduce kernel launch overhead by capturing a sequence of GPU operations i
...
 
 | Feature | `qd.cuda` SM 9.0+ | `qd.cuda` < SM 9.0 | `qd.amdgpu` | `qd.metal` | `qd.vulkan` | `qd.cpu` |
 | --- | --- | --- | --- | --- | --- | --- |
 | `graph=True` | hardware accelerated | hardware accelerated | hardware accelerated | runs (no acceleration) | runs (no acceleration) | runs (no acceleration) |
 | `graph_do_while` | hardware accelerated | host fallback | host fallback | host fallback | host fallback | host fallback |
-
-AMDGPU `graph_do_while` falls back to the host-side loop because HIP does not currently expose conditional / while graph nodes (as of ROCm 7.2).
+| `qd.checkpoint` (skip + `yield_on=`) | GPU-side gating | GPU-side gating | GPU-side gating | GPU-side gating | GPU-side gating | host-branch gating |
```
hughperkins: why add 'gating'? Can we remove it? Also, can we change 'host-branch' to 'host-side', for consistency?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +153,77 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
 Therefore on unsupported platforms, you might consider creating a second implementation, which works differently. e.g.:
 - fixed number of loop iterations, so no dependency on gpu data for kernel launch; combined perhaps with:
 - make each kernel 'short-circuit', exit quickly, if the task has already been completed; to avoid running the GPU more than necessary
+
+## Checkpoints with `qd.checkpoint` *(experimental)*
+
+`qd.checkpoint()` marks a section of a graph kernel as a *skippable, optionally yieldable stage*. An example use-case is an algorithm implemented as a graph where you might need to allocate additional memory part-way through, the graph operations are in-place, and simply retrying the whole graph from the start is not an option. `qd.checkpoint` lets the kernel pause at a known point, surface the reason to the host, let the host fix things up, and resume from that point on the next launch.
```
hughperkins: I wonder if 'pause' is the right word? Maybe 'break'? and is it really a 'known point'? Maybe simply 'at some point in the graph'?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +153,77 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+            for i in range(arr.shape[0]):
+                # ...
+                pass
+        with qd.checkpoint():                       # cp_id 2
+            for i in range(arr.shape[0]):
+                # ...
+                pass
+```
+
+Each `with qd.checkpoint(...)` block gets a `cp_id` assigned by declaration order (0, 1, 2, ... flat across the whole kernel, independent of whether the checkpoint is inside or outside a `qd.graph_do_while`).
```
hughperkins: explain what is the purpose of the cp_id. Why do we need it?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +153,77 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+### Yield mechanism
+
+If `yield_on=foo` is supplied, the body may write a non-zero value into `foo[()]` (for example, when a pre-allocated buffer is too small) to signal "the host needs to handle something before this checkpoint can complete". When that happens:
+
+1. The framework records the checkpoint that yielded (first yielder in declaration order wins).
+2. Every later checkpoint in the same launch is skipped.
+3. `qd.checkpoint` will exit any surrounding `qd.graph_do_while`.
+4. `foo[()]` is reset to `0` so the next launch starts clean without host intervention.
```
hughperkins: remove 'so the next launch starts clean without host intervention'

docs/source/user_guide/graph.md
```diff
@@ -154,3 +153,77 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+    status = step.resume(arr, overflow_flag, newton_cond,
+                         from_checkpoint=status.checkpoint)
+```
+
+Kernels with `qd.checkpoint()` but no `yield_on=` keep their previous return contract (typically `None`) — the `GraphStatus` surface is opt-in via `yield_on=`.
+
+### Restrictions
+
+- Must be used inside `@qd.kernel(graph=True)`.
+- `yield_on=` (when supplied) must be the bare name of a kernel parameter that is a 0-d `qd.types.ndarray(qd.i32, ndim=0)`.
```
hughperkins: "bare name" is confusing. Isnt it simply a kernel parameter?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +153,77 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+A bare statement like `counter[()] -= 1` placed directly inside a `qd.checkpoint` body is not currently associated with the checkpoint's `cp_id` and will run unconditionally, even when the checkpoint is meant to be skipped. Wrap such statements in a one-iteration `for` loop:
+
+```python
+with qd.checkpoint(yield_on=overflow_flag):
+    for _ in range(1):
+        counter[()] = counter[()] - 1
+```
+
+The runtime cost of an extra 1-iteration loop is negligible compared to the kernel-launch overhead the rest of the graph saves.
```
hughperkins: remove this paragraph.

hughperkins: I wonder if we should somehow detect violations and throw an exception 🤔

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,75 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+### Restrictions
+
+- Must be used inside `@qd.kernel(graph=True)`.
+- `yield_on=` (when supplied) must be a kernel parameter that is a 0-d `qd.types.ndarray(qd.i32, ndim=0)`.
+- Checkpoints cannot be nested inside other checkpoints.
+
+### Every statement inside a checkpoint must live in a top-level `for` loop
+
+A bare statement like `counter[()] -= 1` placed directly inside a `qd.checkpoint` body would not pick up the checkpoint's `cp_id` and would run unconditionally even when the checkpoint is meant to be skipped, so the AST transformer rejects it at compile time. Wrap such statements in a one-iteration `for` loop:
```
hughperkins: this is way too much deatil about the implementation. Just state that bare statements are not allowed inside qd.checkpoint, and will raise an Exception. And to wrap such statements in a one-iteration `for` loop.

But.. if we can detect these things, in order to throw the exception, maybe we can wrap them in the one iteration for loop automaticaly?

docs/source/user_guide/graph.md
hughperkins: Start of with the high level summary.

"Each `with qd.checkpoint(...)` block gets a `cp_id` assigned. You can use the `cp_id` to identify which checkpoint yielded and which checkpoint to resume from — see [Host-side yield / resume loop](#host-side-yield--resume-loop) below.".  How the cp_id is assigned is un-necessary detail here, as well as th edetails about graph do while.

docs/source/user_guide/graph.md
```diff
@@ -4,14 +4,11 @@ Graphs reduce kernel launch overhead by capturing a sequence of GPU operations i
 
 ## Backend support
 
-Both features run on every backend. They are *hardware accelerated* on CUDA (via CUDA graphs) and AMDGPU (via HIP graphs); `graph_do_while` additionally requires CUDA SM 9.0+ / Hopper for its hardware-accelerated path. On other backends, `graph=True` is silently ignored and the kernel runs via the normal launch path, and `graph_do_while` falls back to a host-side do-while loop that copies the condition value GPU → host each iteration (causing a pipeline stall — see [Caveats](#caveats)).
-
 | Feature | `qd.cuda` SM 9.0+ | `qd.cuda` < SM 9.0 | `qd.amdgpu` | `qd.metal` | `qd.vulkan` | `qd.cpu` |
```
hughperkins, line 9: why state this when it's already in the table? Be more concise.

hughperkins:
Note: this PR is ready for review

docs/source/user_guide/graph.md
```diff
@@ -4,14 +4,11 @@ Graphs reduce kernel launch overhead by capturing a sequence of GPU operations i
 
 ## Backend support
 
-Both features run on every backend. They are *hardware accelerated* on CUDA (via CUDA graphs) and AMDGPU (via HIP graphs); `graph_do_while` additionally requires CUDA SM 9.0+ / Hopper for its hardware-accelerated path. On other backends, `graph=True` is silently ignored and the kernel runs via the normal launch path, and `graph_do_while` falls back to a host-side do-while loop that copies the condition value GPU → host each iteration (causing a pipeline stall — see [Caveats](#caveats)).
```
duburcqa: Why is this removed? Because of the following table?

docs/source/user_guide/graph.md
```diff
@@ -4,14 +4,11 @@ Graphs reduce kernel launch overhead by capturing a sequence of GPU operations i
...
-
 | Feature | `qd.cuda` SM 9.0+ | `qd.cuda` < SM 9.0 | `qd.amdgpu` | `qd.metal` | `qd.vulkan` | `qd.cpu` |
 | --- | --- | --- | --- | --- | --- | --- |
 | `graph=True` | hardware accelerated | hardware accelerated | hardware accelerated | runs (no acceleration) | runs (no acceleration) | runs (no acceleration) |
 | `graph_do_while` | hardware accelerated | host fallback | host fallback | host fallback | host fallback | host fallback |
-
-AMDGPU `graph_do_while` falls back to the host-side loop because HIP does not currently expose conditional / while graph nodes (as of ROCm 7.2).
+| `qd.checkpoint` (skip + `yield_on=`) | GPU-side | GPU-side | GPU-side | GPU-side | GPU-side | host-side |
```
duburcqa: There is no 'qd.' for 'graph_do_while'. This is inconsistent.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
 Therefore on unsupported platforms, you might consider creating a second implementation, which works differently. e.g.:
 - fixed number of loop iterations, so no dependency on gpu data for kernel launch; combined perhaps with:
 - make each kernel 'short-circuit', exit quickly, if the task has already been completed; to avoid running the GPU more than necessary
+
+## Checkpoints with `qd.checkpoint` *(experimental)*
```
duburcqa, line 180: If checkpoint is a graph-specific API, it should be prefixed by 'graph_', as for all the other functions.

duburcqa: Know that I am thinking, we should probably just have some new 'qd.graph.' submodule. That would like everything both simpler and more less confusing.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+> **Experimental.** `qd.checkpoint`, `qd.GraphStatus`, and `kernel.resume(from_checkpoint=...)` are experimental APIs. The shape of the public surface (the context-manager signature, the `GraphStatus` fields, the host-side resume loop, the error messages, and the cross-backend lowering details) may change in any future release without a deprecation cycle.
+
+`qd.checkpoint()` marks a section of a graph kernel as a *skippable, optionally yieldable stage*. An example use-case is an algorithm implemented as a graph where you might need to allocate additional memory part-way through, the graph operations are in-place, and simply retrying the whole graph from the start is not an option. `qd.checkpoint` lets the kernel break at some point in the graph, surface the reason to the host, let the host fix things up, and resume from that point on the next launch.
```
duburcqa: Mentioning "*skippable, optionally yieldable stage*" without introducing these concepts is impossible to follow.

duburcqa: "the graph operations are in-place, and simply retrying the whole graph from the start is not an option"

---> Why? I don't see the relationship between these 2 statements.

duburcqa: ---> `qd.checkpoint` lets the kernel break at some point in the graph

Only whenever some condition is triggered? I think that is one of the big advantages no?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+                pass
+        with qd.checkpoint(yield_on=overflow_flag): # cp_id 1 (can yield)
+            for i in range(arr.shape[0]):
+                # ...
+                pass
+        with qd.checkpoint():                       # cp_id 2
+            for i in range(arr.shape[0]):
+                # ...
+                pass
+```
```
duburcqa, line 214: This example arrives too early I think? We still have no idea what `qd.checkpoint()` without argument does.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+Each `with qd.checkpoint(...)` block gets a `cp_id` assigned. You can use the `cp_id` to identify which checkpoint yielded and which checkpoint to resume from — see [Host-side yield / resume loop](#host-side-yield--resume-loop) below.
+
+### Yield mechanism
+
+If `yield_on=foo` is supplied, the body may write a non-zero value into `foo[()]` (for example, when a pre-allocated buffer is too small) to signal "the host needs to handle something before this checkpoint can complete". When that happens:
+
+1. The framework records the checkpoint that yielded (first yielder in declaration order wins).
+2. Every later checkpoint in the same launch is skipped.
+3. `qd.checkpoint` will exit any surrounding `qd.graph_do_while`.
+4. `foo[()]` is reset to `0`.
```
duburcqa: I'm not in favour of this behaviour. It is inconsistent with everything else.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+            for i in range(arr.shape[0]):
+                # ...
+                pass
+```
+
+Each `with qd.checkpoint(...)` block gets a `cp_id` assigned. You can use the `cp_id` to identify which checkpoint yielded and which checkpoint to resume from — see [Host-side yield / resume loop](#host-side-yield--resume-loop) below.
+
+### Yield mechanism
+
+If `yield_on=foo` is supplied, the body may write a non-zero value into `foo[()]` (for example, when a pre-allocated buffer is too small) to signal "the host needs to handle something before this checkpoint can complete". When that happens:
```
duburcqa: "before this checkpoint can complete"

I don't understand. Based on what is written next, it seems that the entire while block is aborted (like a break basically), so it is more invasive / aggressive than what is advertised.

Implementing the resuming logics is user responsibility apparently?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+Kernels with at least one `yield_on=` checkpoint return a `qd.GraphStatus` from every launch (and from `kernel.resume(...)`). The status carries two fields:
+
+- `status.yielded` — `True` iff some `yield_on=` flag was non-zero during this launch.
+- `status.checkpoint` — `cp_id` of the first (in declaration order) checkpoint that fired its flag, or `None` when `yielded` is `False`.
+
+Resume by calling `kernel.resume(..., from_checkpoint=status.checkpoint)`. Every `qd.checkpoint` with `cp_id < from_checkpoint` is skipped on the resume launch; the rest run normally. The canonical host loop:
+
+```python
+status = step(arr, overflow_flag, newton_cond)
```
duburcqa, line 241: How does this interplay with user-specify return values? It is not supported by graph?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+```python
+@qd.kernel(graph=True)
+def step(
+    arr: qd.types.ndarray(qd.f32, ndim=1),
+    overflow_flag: qd.types.ndarray(qd.i32, ndim=0),
+    newton_cond: qd.types.ndarray(qd.i32, ndim=0),
+):
+    while qd.graph_do_while(newton_cond):
+        with qd.checkpoint():                       # cp_id 0
```
duburcqa: It would be nice to be able to specify some enum-based ID, to allow cleaner implementation of checkpoint recovery.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+4. `foo[()]` is reset to `0`.
+
+### Host-side yield / resume loop
+
+Kernels with at least one `yield_on=` checkpoint return a `qd.GraphStatus` from every launch (and from `kernel.resume(...)`). The status carries two fields:
+
+- `status.yielded` — `True` iff some `yield_on=` flag was non-zero during this launch.
+- `status.checkpoint` — `cp_id` of the first (in declaration order) checkpoint that fired its flag, or `None` when `yielded` is `False`.
+
+Resume by calling `kernel.resume(..., from_checkpoint=status.checkpoint)`. Every `qd.checkpoint` with `cp_id < from_checkpoint` is skipped on the resume launch; the rest run normally. The canonical host loop:
```
duburcqa: > Every `qd.checkpoint` with `cp_id < from_checkpoint` is skipped on the resume launch; the rest run normally.

Does this means the checkpoint for 'cp_id == from_checkpoint' will be executed once again entirely?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+```python
+status = step(arr, overflow_flag, newton_cond)
+while status.yielded:
+    handle_overflow_for(status.checkpoint, ...)
+    status = step.resume(arr, overflow_flag, newton_cond,
+                         from_checkpoint=status.checkpoint)
+```
+
+Kernels with `qd.checkpoint()` but no `yield_on=` keep their previous return contract (typically `None`) — the `GraphStatus` surface is opt-in via `yield_on=`.
```
duburcqa: I don't think it should be supported. That is basically a misusage no?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+  ```python
+  with qd.checkpoint():
+      for _ in range(1):
+          c[()] = c[()] + 1
+      for i in range(arr.shape[0]):
+          arr[i] = arr[i] + 1
+  ```
+
+  The restriction is by design: each top-level statement inside a checkpoint becomes its own GPU task / graph node, so silently auto-wrapping bare statements would hide a sequence of N field writes ballooning into N kernel launches. Forcing the user to write the `for`-wrap themselves keeps the lowering visible and gives a single obvious place to fuse multiple writes into one task by sharing a single wrapper.
```
duburcqa: Why indentation here?

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,78 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+                # ...
+                pass
+        with qd.checkpoint(yield_on=overflow_flag): # cp_id 1 (can yield)
+            for i in range(arr.shape[0]):
+                # ...
+                pass
+        with qd.checkpoint():                       # cp_id 2
+            for i in range(arr.shape[0]):
+                # ...
+                pass
```
duburcqa: Is it possible to only specify `with qd.checkpoint(yield_on=overflow_flag):` and remove the others? If not, this should be explicit in the documentation. I don't think remember reading this.

docs/source/user_guide/graph.md
hughperkins: yes. seemed not useful to reproduce the information that is clearer in the table as unclear wordy prose.

docs/source/user_guide/graph.md
hughperkins: let's say you have:
```
def iteration():
    data[0] += 1
    data[0] *= 2
```
... but after doing `data[0] += 1`, you realize you dn't have enough memory to do `data[0] *= 2`, so you need to pause for a moment, and allocate more memory.

you cannot restart the function from the start becuase we already incremented `data[0]` by 1 already.

you have to start from `data[0] *= 2`.

docs/source/user_guide/graph.md
hughperkins: which bit? the resetting foo to 0?

docs/source/user_guide/graph.md
hughperkins: when it yields, the caller is given a checkpoint id. They can pass the checkpoint id into the `resume` function, to resume at that checkpoint id, skipping everything before it.

docs/source/user_guide/graph.md
hughperkins: return values are not supported. by graph, correct, https://github.com/Genesis-Embodied-AI/quadrants/blob/d9331f4ba4e5db07f8e3d56319138100cb06bc61/docs/source/user_guide/graph.md?plain=1#L48

docs/source/user_guide/graph.md
hughperkins: Do you mean, instead of incremntally auto-assigning an integer id, require the id to be passed in, as an integer, and we could pass in an intenum value, converted to int?

docs/source/user_guide/graph.md
hughperkins: Yes.

docs/source/user_guide/graph.md
hughperkins: soo.... remember we don't have the technology to jump directly to a node. Or I suppose we might do, but that's not what we are doing here. We have a linear sequence of offloaded tasks:

```
@qd.kernel(graph=True)
def k1():
     task1()
     task2()
     task3()
     task4()
     task5()
```
We want to be able to jump out of 3, and then later restart from 3, so we wrap with qd.checkpoint:
```
@qd.kernel(graph=True)
def k1():
     task1()
     task2()
     with qd.checkpoint(yield=foo):
         task3()
     task4()
     task5()
```
When we resume, it will still execute task1 and task2 anyway, because they arent guarded with anything.

We need to guard task1 and task2 with with checkpoint, so they can be skipped.
```
@qd.kernel(graph=True)
def k1():
     with qd.checkpoint():
	     task1()
     with qd.checkpoint():
         task2()
     with qd.checkpoint(yield=foo):
         task3()
     task4()
     task5()
```
But this will still run task4 and task5 when we break out of task3. So we need to guard them too:
```
@qd.kernel(graph=True)
def k1():
     with qd.checkpoint():
	     task1()
     with qd.checkpoint():
         task2()
     with qd.checkpoint(yield=foo):
         task3()
     with qd.checkpoint():
         task4()
     with qd.checkpoint():
         task5()
```

But your line of questioning raises an interseting question. Let me check with AI:

Screenshot 2026-06-16 at 14 43 50

hughperkins: reply and my reply:

Screenshot 2026-06-16 at 14 47 20
Screenshot 2026-06-16 at 14 47 30

hughperkins: reply from AI:
Screenshot 2026-06-16 at 14 49 59

hughperkins: Screenshot 2026-06-16 at 15 34 01

hughperkins: thoughts?
- we ditch qd.checkpoint() (without params)
- we add qd.checkpoint_always_runs()
- anything not in qd.checkpoint_always_runs(), and not in qd.checkpoint(yield=) get implicitly wrapped in what used to be qd.checkpoint()
- this behavior is triggered for any graph which contains at least one qd.checkpoint(yeidl=)

hughperkins: (and you have to pass in the checkpoint id yourself, which you can pass in as the value of an enum)

docs/source/user_guide/graph.md
hughperkins: see discussion above https://github.com/Genesis-Embodied-AI/quadrants/pull/725#discussion_r3423185882

which I summarized here: https://github.com/Genesis-Embodied-AI/quadrants/pull/725#discussion_r3423515725

docs/source/user_guide/graph.md
hughperkins: see discussion later https://github.com/Genesis-Embodied-AI/quadrants/pull/725#discussion_r3423522920

duburcqa: ok I see, so you just mean that you have no guarantee that kernels being executed is not modifying the input data in-place. It would be nice to clarify this.

hughperkins: I mean, in qipc, it specifically IS modifying data in-place.

hughperkins: 'graph operations' here means the kernels that you are running, eg qipc kernels

duburcqa: > I mean, in qipc, it specifically IS modifying data in-place.

So what? That is not the point here.

> 'graph operations' here means the kernels that you are running, eg qipc kernels

Yes, like I said, this is not clear.

docs/source/user_guide/graph.md
hughperkins: Seems reasonable. But perhaps not in this PR?

docs/source/user_guide/graph.md
hughperkins: added "and therefore not idempotent, and therefore for which "

hughperkins: added "and therefore not idempotent, and therefore for which "

docs/source/user_guide/graph.md
duburcqa: In that case, make sure it is tracking somewhere. Still, it is weird to address this in another PR, since it is literally the PR that introduces this function. But I can understand you want to move faster and qipc is already relying on this one.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,90 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
 - fixed number of loop iterations, so no dependency on gpu data for kernel launch; combined perhaps with:
 - make each kernel 'short-circuit', exit quickly, if the task has already been completed; to avoid running the GPU more than necessary
+
+## Checkpoints with `qd.checkpoint` *(experimental)*
+
+> **Experimental.** `qd.checkpoint`, `qd.GraphStatus`, and `kernel.resume(from_checkpoint=...)` are experimental APIs. The shape of the public surface (the context-manager signature, the `@qd.kernel(checkpoints=True)` flag, the `GraphStatus` fields, the host-side resume loop, the error messages, and the cross-backend lowering details) may change in any future release without a deprecation cycle.
+
+`qd.checkpoint` lets a graph kernel pause partway through, surface a reason to the host, let the host fix things up, and resume from where it paused on the next launch. An example use-case is an algorithm implemented as a graph that may need to allocate additional memory partway through, where the graph operations are in-place, and therefore not idempotent, and therefore for which simply retrying the whole graph from the start is not an option.
```
duburcqa: As I already mentioned, "graph operations" is not a standard term, you cannot use it without defining what it means, or just reformulate this sentence to avoid using this terminology.

I would suggest to define in parentheses what "idempotent" means in programming. It is not really common knowledge and understanding what it means in this context is critical.

docs/source/user_guide/graph.md
duburcqa: yes

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,90 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+- `status.checkpoint` — the `cp_id` label of the yielding checkpoint (or `None` when `yielded` is `False`).
+
+Resume by calling `kernel.resume(..., from_checkpoint=label)`. Everything before `label` in source order is skipped on the resume launch; everything from `label` onward runs normally. The canonical host loop:
+
+```python
+overflow_flag[()] = 0  # initialise before the first launch
+status = step(arr, overflow_flag, newton_cond)
+while status.yielded:
+    handle_overflow_for(status.checkpoint, ...)
+    overflow_flag[()] = 0  # clear before resume, otherwise the same checkpoint yields again
```
duburcqa, line 244: I like it.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,90 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+The `cp_id` argument is the label you'll use to identify the checkpoint from the host (in `GraphStatus.checkpoint` and `kernel.resume(from_checkpoint=...)`). It must be an int literal or an `IntEnum` value; the framework preserves the value as-is, so `qd.checkpoint(Stage.SIM, ...)` round-trips as `Stage.SIM` rather than the raw int. Labels must be unique within a kernel.
+
+### Yield mechanism
+
+When the body of a checkpoint writes a non-zero value into `yield_on[()]`:
+
+1. Everything after the yielding checkpoint in the same launch is skipped.
+2. `qd.checkpoint` will exit any surrounding `qd.graph_do_while`.
+
+The framework never writes into your `yield_on` buffer — you own it end-to-end. That means:
```
duburcqa, line 225: Nice.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,90 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+
+- Before the **first** launch, initialise it to `0` (a freshly allocated `qd.ndarray` is not guaranteed to be zeroed).
+- Before each **resume** launch, reset it to `0` (otherwise the body of the same checkpoint sees the stale non-zero value and yields again on the same condition, looping forever).
```
duburcqa: Could be worth using some :warning: tag.

docs/source/user_guide/graph.md
```diff
@@ -154,3 +151,90 @@ Note: the basic `graph=True` path (without `graph_do_while`) does **not** stall
...
+overflow_flag[()] = 0  # initialise before the first launch
+status = step(arr, overflow_flag, newton_cond)
+while status.yielded:
+    handle_overflow_for(status.checkpoint, ...)
+    overflow_flag[()] = 0  # clear before resume, otherwise the same checkpoint yields again
+    status = step.resume(arr, overflow_flag, newton_cond,
+                         from_checkpoint=status.checkpoint)
+```
+
+### Restrictions
```
duburcqa, line 255: Could you explain / be more explicit about what happens during resume?

Stating clearly that the entire checkpoint block is re-executed, and that it is user-responsibility to ensure idempotent behaviour when checkpoint is needed? Because if the state is altered during the checkpoint block, resuming is not going to save you I guess? Whatever the answer, it should be very clear in the doc.

Beyond that, how does checkpointing works under the hood? Does it snapshot all the input data by copy before yielding, or it just return like this? If no copy is made, this means that resuming must be done "right away", without further altering the data in between, otherwise it is some kind of undefined behaviour.

Another important point, what is I don't want to resume in such a case and I just want to move on to another kernel and continue like this? Is it supported or resume must happen?

I think it is essentially to clarify all these points in the documentation.

docs/source/user_guide/graph.md
duburcqa: ok cool. I forgot about that.

docs/source/user_guide/graph.md
duburcqa: Yes exactly.

docs/source/user_guide/graph.md
duburcqa: This should be clearly stated if not already the case.

docs/source/user_guide/graph.md
duburcqa: I think this issue resolved itself with the new API design, which is great.

docs/source/user_guide/graph.md
hughperkins: checkpoint does NOT require idempotent behavior. This is the entire purpose of checkpoint: to be able to interrupt and resume graphs that are NOT idempotent.

docs/source/user_guide/graph.md
hughperkins: I would argue that idempotent is a standard programming term. It is not compiler specific. It is not physics siulation specific. I've seem the term used by many engineers in my previous company, during standard PRs.

docs/source/user_guide/graph.md
hughperkins: ah, hte checkpoint block itself. right.

hughperkins: the checkpoint block itself actually does not so much require idempotence, as requiring that it is atomic: it either succeeds completely, or fails without changing anything.

hughperkins: as an example, in the case of allocation issues, the checkpoint block looks like:

- do we have enough memory availbel?
     - no: exit now
     - yes: ok, lets proceed with running the sort etc

docs/source/user_guide/graph.md
hughperkins: - replaced 'graph operaitons' with 'operations in the graph'
- replcae 'not idempotent' with 'cannot be rerun'

docs/source/user_guide/graph.md
hughperkins: wouldnt that be non-ascii? 🤔

docs/source/user_guide/graph.md
hughperkins: added 'resume where' section

docs/source/user_guide/graph.md
duburcqa: No, it uses `:warning:`

docs/source/user_guide/graph.md
duburcqa: > the checkpoint block itself actually does not so much require idempotence, as requiring that it is atomic: it either succeeds completely, or fails without changing anything.

Yeah, this is exactly what I meant by « ensure idempotent behaviour **when checkpoint is needed** »

docs/source/user_guide/graph.md
duburcqa: ‘cannot be rerun´ without altering / corrupting the output

docs/source/user_guide/graph.md
hughperkins: "fails without changing anything" I feel is not idempotent? Idempotent means that calling the function multiple times is identical in effect to calling it once. But if it fails the first time, it would only be idempotent if it always failed thereafter I feel?

docs/source/user_guide/graph.md
hughperkins: done

docs/source/user_guide/graph.md
hughperkins: done

docs/source/user_guide/graph.md
hughperkins: In any case, I feel the 'resume where' section addresses your original concerns about ensuring that what happens during resume is docuemnted?

duburcqa:
ok to merge

hughperkins:
(running Genesis unit tests and benchmarks)

hughperkins:
Note: for the linewrapping in the .mm file, will open a separte PR to change .mm wrapping to 120c

hughperkins:
Genesis unit tests passing:

hughperkins:
latest benchmark:

2026022_chkpoint_1451

hughperkins:
Finally ok benchamrks:

20260622_chkpt_2039

hughperkins:
ran again, still ok:

2026022_chkpoint_2038

hughperkins:
unit tests pass:

Screenshot 2026-06-22 at 21 41 22

PR #728, opened 2026-06-06
[Graph] Graph do while can be nested

hughperkins:


docs/source/user_guide/graph.md
```diff
@@ -9,10 +9,13 @@ Both features run on every backend. They are *hardware accelerated* on CUDA (via
 | Feature | `qd.cuda` SM 9.0+ | `qd.cuda` < SM 9.0 | `qd.amdgpu` | `qd.metal` | `qd.vulkan` | `qd.cpu` |
 | --- | --- | --- | --- | --- | --- | --- |
 | `graph=True` | hardware accelerated | hardware accelerated | hardware accelerated | runs (no acceleration) | runs (no acceleration) | runs (no acceleration) |
-| `graph_do_while` | hardware accelerated | host fallback | host fallback | host fallback | host fallback | host fallback |
+| `graph_do_while` (single) | hardware accelerated | host fallback | host fallback | host fallback | host fallback | host fallback |
+| `graph_do_while` (nested / sibling) | hardware accelerated | host fallback | host fallback | host fallback | host fallback | host fallback |
```
hughperkins: do we really need to duplicate these identical lines 😂

docs/source/user_guide/graph.md
```diff
@@ -129,21 +132,115 @@ def converge(x: qd.types.ndarray(qd.f32, ndim=1),
 
 `graph_do_while` has **do-while** semantics: the kernel body always executes at least once before the condition is checked. This matches the behavior of CUDA conditional while nodes. The flag value must be >= 1 at launch time. Passing 0 with a kernel that decrements the counter will cause an infinite loop.
 
+> **Reset the flag outside the loop body, never inside it.** The counter / `keep_going` flag must reach a terminating value *as a result of the loop body*. If you (re)set the flag to a non-zero value **inside** the `while qd.graph_do_while(...)` body — e.g. `for _ in range(1): counter[()] = N` placed within the loop — it is re-applied on every iteration and the loop never terminates. Do the reset before the loop (a run-once top-level `for`, see [Loop-carried state](#loop-carried-state)) or on the host between launches (`counter.fill(N)`).
```
hughperkins: This seems obvous? Do we need this? 🤔

docs/source/user_guide/graph.md
```diff
@@ -129,21 +132,115 @@ def converge(x: qd.types.ndarray(qd.f32, ndim=1),
...
+
 ### ndarray vs field
 
 The parameter used by `graph_do_while` MUST be an ndarray.
 
 However, other parameters can be any supported Quadrants kernel parameter type.
 
+### Nested loops and mixing with for-loops
```
hughperkins, line 132: Lets mark this functionality as 'experimental'.

docs/source/user_guide/graph.md
```diff
@@ -9,10 +9,13 @@ Both features run on every backend. They are *hardware accelerated* on CUDA (via
...
 | `graph=True` | hardware accelerated | hardware accelerated | hardware accelerated | runs (no acceleration) | runs (no acceleration) | runs (no acceleration) |
-| `graph_do_while` | hardware accelerated | host fallback | host fallback | host fallback | host fallback | host fallback |
+| `graph_do_while` (single) | hardware accelerated | host fallback | host fallback | host fallback | host fallback | host fallback |
+| `graph_do_while` (nested / sibling) | hardware accelerated | host fallback | host fallback | host fallback | host fallback | host fallback |
 
 AMDGPU `graph_do_while` falls back to the host-side loop because HIP does not currently expose conditional / while graph nodes (as of ROCm 7.2).
 
+Nested and sibling `graph_do_while` loops (see [Nested loops](#nested-loops-and-mixing-with-for-loops)) work on every backend: hardware-accelerated as a single GPU-side graph on CUDA SM 9.0+, and via a host-side driver everywhere else (CPU, CUDA pre-SM 9.0, AMDGPU, Vulkan, Metal). On the host-fallback backends each loop-body pass is replayed from the host and the condition value is copied GPU → host between iterations (see [Caveats](#caveats)).
```
hughperkins: Does this show anything that isn't in the table? Lets remove it?

But lets make a new paragraph stating the nested graph do while is experimetnal for now.

docs/source/user_guide/graph.md
```diff
@@ -129,21 +132,115 @@ def converge(x: qd.types.ndarray(qd.f32, ndim=1),
...
+`while`-loop. Bare statements (assignments, `if`, etc.) at these levels are rejected with a
+`QuadrantsSyntaxError`. Wrap any such statement in a trivial loop:
+
+```python
+# Instead of:   counter[()] = counter[()] - 1
+for _ in range(1):
+    counter[()] = counter[()] - 1
+```
+
+This restriction keeps each offloaded task tagged with the exact loop level it belongs to. `for`-loops
```
hughperkins: remove '`for`-loops
and `graph_do_while` loops may otherwise be freely ordered, mixed, and nested.'

docs/source/user_guide/graph.md
```diff
@@ -135,15 +137,58 @@ The parameter used by `graph_do_while` MUST be an ndarray.
 
 However, other parameters can be any supported Quadrants kernel parameter type.
 
+### Nested loops and mixing with for-loops
+
+> **Experimental.** Nested / sibling `graph_do_while` loops, and mixing `graph_do_while` with top-level `for`-loops, are experimental for now. Single-loop `graph_do_while` is the stable path.
+
+`graph_do_while` loops can be **nested** inside one another, placed **side by side** (siblings), and freely **mixed with plain top-level for-loops**. Each loop has its own scalar `qd.i32` counter ndarray. On CUDA SM 9.0+ the whole nest runs entirely GPU-side as one graph of conditional while nodes; on the host-fallback backends the loop nest is driven from the host.
```
duburcqa: 'Each loop has its own scalar `qd.i32` counter ndarray' ---> what does this means exactly? one must specify different counters for both loops? What happens if one fails to do so?

docs/source/user_guide/graph.md
```diff
@@ -135,15 +137,58 @@ The parameter used by `graph_do_while` MUST be an ndarray.
...
+                x[i] = x[i] + 1
+            for _ in range(1):
+                inner[()] = inner[()] - 1
+        for _ in range(1):
+            outer[()] = outer[()] - 1
+```
+
+Important points for nested loops:
+
+- **Re-arm inner counters.** An inner loop's counter is not reset automatically between outer iterations. Reset it at the top of the enclosing loop body (as `inner[()] = 5` above), or the inner loop will only run during the first outer iteration.
```
duburcqa: It is weird to mention this. Like outer is not reset either. There is nothing new about this behaviour. Just classical programming. It is a tip / reminder at best.

docs/source/user_guide/graph.md
```diff
@@ -135,15 +137,58 @@ The parameter used by `graph_do_while` MUST be an ndarray.
...
+
+### Kernel structure restriction
+
+When a kernel uses `qd.graph_do_while()` anywhere, **every top-level statement** — both in the kernel body and inside each `graph_do_while` body — must be either a `for`-loop or a `qd.graph_do_while()` `while`-loop. Bare statements (assignments, `if`, etc.) at these levels are rejected with a `QuadrantsSyntaxError`. Wrap any such statement in a trivial loop:
+
+```python
+# Instead of:   counter[()] = counter[()] - 1
+for _ in range(1):
+    counter[()] = counter[()] - 1
+```
```
duburcqa: It is a good opportunity to explain what it does. I know it but it is not immediately obvious if you discover this pattern for the first time

docs/source/user_guide/graph.md
```diff
@@ -135,15 +137,58 @@ The parameter used by `graph_do_while` MUST be an ndarray.
...
+
+This restriction keeps each offloaded task tagged with the exact loop level it belongs to. A `graph_do_while` `while`-loop may only appear at the kernel top level or directly inside another `graph_do_while` body — it cannot be placed inside a `for`-loop.
```
duburcqa: "This restriction keeps each offloaded task tagged with the exact loop level it belongs to."

---> I don't understand what this means.

docs/source/user_guide/graph.md
```diff
@@ -135,15 +137,58 @@ The parameter used by `graph_do_while` MUST be an ndarray.
...
+
 ### Restrictions
 
-- The same physical ndarray must be used for the counter parameter on every
-  call. Passing a different ndarray raises an error, because the counter's
-  device pointer is baked into the graph at creation time.
+- The counter ndarray may be swapped between calls: the cached graph reads each counter through an indirection slot that is refreshed on every launch, so passing a different ndarray (or alternating between several) replays the cached graph without rebuilding it.
```
duburcqa: "the cached graph reads each counter through an indirection slot that is refreshed on every launch, so passing a different ndarray (or alternating between several) replays the cached graph without rebuilding it."

No idea what this means.

docs/source/user_guide/graph.md
```diff
@@ -135,15 +137,58 @@ The parameter used by `graph_do_while` MUST be an ndarray.
...
 
 ### Caveats
 
-On platforms without native device-side conditional graph nodes — currently CUDA pre-SM 9.0 and **AMDGPU** (HIP has no conditional / while node API as of ROCm 7.2) — the value of the `graph_do_while` parameter will be copied from the GPU to the host each iteration, in order to check whether we should continue iterating. This causes a GPU pipeline stall. At the end of each loop iteration:
+On platforms without native device-side conditional graph nodes — currently CUDA pre-SM 9.0, **AMDGPU** (HIP has no conditional / while node API as of ROCm 7.2), Vulkan, Metal, and CPU — the value of the `graph_do_while` parameter will be copied from the GPU to the host each iteration, in order to check whether we should continue iterating. This causes a GPU pipeline stall. For nested loops this host round-trip happens once per iteration of each loop level, and each loop-body task is replayed individually, so deeply nested loops on these backends pay correspondingly more host overhead (they remain correct, just slower than the CUDA SM 9.0+ native path). At the end of each loop iteration:
```
duburcqa: It looks like not usable beyond prototyping on anything else than CUDA? The cost of nested sync + device-to-host seems prohibitive otherwise? Or it is not that expensive? If it is prohibitive, I would state than very clearly to discourage using it in production for anything that is not CUDA. I'm even wondering if it should not rather raise an exception if it is considered a misusage.

docs/source/user_guide/graph.md
hughperkins: infinite loop, same as a normal while loop:

```
i = 4
while i != 0:
     while i != 0:
          print('i', i)
          i -= 1
     i -= 1
```
output, after ctrl-c as fast as I can:
```
i -263814
i -263815
i -263816
i -263817
i -263818
i -263819
^Ci -263822
i -263823
i -263824
```

with graph do while:
```
import quadrants as qd

qd.kernel(graph=True)
def k1(a: qd.types.NDArray):
    for _ in range(1):
         a[()] = 4
    while qd.graph_do_while(a):
         print('a', a)
         while qd.graph_do_while(a):
              print('a', a[()])
              a[()] = a[()] - 1
         a[()] = a[()] - 1

qd.init(qd.cpu)
a = qd.ndarray(qd.i32, shape=())
k1(a)

qd.init(qd.gpu)
a = qd.ndarray(qd.i32, shape=())
k1(a)
```

output:
```
a -33305
a -33306
a -33307
a -33308
a -33309
a -33310
a -33311
a -33312
^CTraceback (most recent call last):
```

I'm not sure why the AI bothered to write this bit really. I feel it's obvious/superfluous, and we should remove it?

hughperkins: (tested on cpu and cuda)

docs/source/user_guide/graph.md
hughperkins: yes, I agree. I'll remove.

docs/source/user_guide/graph.md
hughperkins: I think it's superfluous. I'll remove.

By the way, the context for a lot of these random, state-the-obvious, doc things is that at the start a lot of the obvious things were completely broken 😬  (notably on main, there's quite a lot of weirdness as far as when to reset stuff). And since I ask AI to update user facing docs as required after each change, it gradually accumulates these superflous stuff :)

docs/source/user_guide/graph.md
hughperkins: it means if you have multiple parts of your code that call into the exact same kernel function, that is a graph, that contains a graph do while, but you pass in a different ndarray counter tnesor from each location (which is a very normal and logical thing to do, since they're completely unconnected pieces of code) it won't crash/break/give incorrect results.

really, this is more something thatusers will just assume works, than new information.

hisotirclly though, this didnt used to be the case, and there used to be doc saying that this would NOT work. Because it did not.

I can probbalby simply remove it, as for all the other state-the-obvious things above :)

docs/source/user_guide/graph.md
hughperkins: It runs on anything.

It's no more prohibitive than our current solver, in Genesis, which already has nested while loops, just not explicilty coded using graph do while.

It is definitely NOT a misuage. Otherwise, we should just remove graph altoegeh, because it is not portable.

hughperkins: Remember that hte tight inner loop is what is going to take most of the time.

on cuda 90+, all loops get to run on the gpu. on other hardware, just as for genesis solver on all platforms right now, the outer loops run in python, and the inner loop runs on gpu.

hughperkins: Created a new performance section, that walks through the 3 scenarios I have encountered to date.

docs/source/user_guide/graph.md
hughperkins: Removed.

duburcqa:
ok to merge

quadrants/runtime/gfx/kernel_launcher.cpp
```diff
@@ -1,35 +1,76 @@
...
-void KernelLauncher::launch_offloaded_tasks_with_do_while(Handle handle, LaunchContextBuilder &ctx) {
-  const ArgArrayPtrKey key{ctx.graph_do_while_arg_id, TypeFactory::DATA_PTR_POS_IN_NDARRAY};
+int32_t KernelLauncher::readback_graph_do_while_flag(LaunchContextBuilder &ctx, int cond_arg_id) {
+  const ArgArrayPtrKey key{cond_arg_id, TypeFactory::DATA_PTR_POS_IN_NDARRAY};
   auto it = ctx.array_ptrs.find(key);
   QD_ASSERT(it != ctx.array_ptrs.end());
 
   auto *device = config_.gfx_runtime_->get_ti_device();
   DeviceAllocation alloc = *(static_cast<DeviceAllocation *>(it->second));
   DevicePtr dev_ptr = alloc.get_ptr(0);
```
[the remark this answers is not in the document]
hughperkins: AI says this is a false positive:

Screenshot 2026-06-16 at 12 38 02
Screenshot 2026-06-16 at 12 38 07

hughperkins:
Benchmark resutls:

20260616_1410_nestgraph

Genesis unit tests:

Screenshot 2026-06-16 at 19 56 48

hughperkins:
check with @duburcqa offline => he's ok to merge this (once tests pass etc)

hughperkins:
Finally 🙌 Genesis unit tests passing 🙏 

Screenshot 2026-06-17 at 20 30 07

hughperkins:
Benchmarks 🤔

20260616_1410_nestgraph (1)

PR #729, opened 2026-06-08
[Perf] Reduce compilation time on kernels with multiple offloaded tasks

hughperkins:
Once a kernel is offloaded, cfg_optimization now builds a separate control-flow
graph per offloaded task (per sub-block, with the correct parallel-execution
flag) instead of one whole-kernel CFG spanning all tasks. store-to-load
forwarding and dead-store elimination then run scoped to each task.

This is semantics-preserving: each offloaded task is a separate device launch,
so cross-task register forwarding is impossible anyway, and global memory
(fields, external tensors, and the global-temporary buffer that carries values
between tasks) is treated conservatively as live-in/live-out of every task by
the existing CFG boundary seeding (reaching-def start-node seed + live-var
final-node seed). The win is compile time: the reaching-definition / live-
variable dataflow becomes ~linear in total IR rather than super-linear in the
combined whole-kernel IR, which blows up for kernels that pack many stages into
one @qd.kernel.

Gated by CompileConfig::cfg_optimization_per_task (default true; env
QD_CFG_OPTIMIZATION_PER_TASK / qd.init kwarg). Pre-offload IR, function bodies,
and the real-matrix path fall back to the whole-kernel CFG unchanged.Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/docs/source/user_guide/optimization_passes.md
+++ b/docs/source/user_guide/optimization_passes.md
@@ -1,6 +1,8 @@
 # Optimization passes
 
-When you call a `@qd.kernel` function, Quadrants first translates the kernel into an internal form then rewrites it, step by step, into something equivalent but cheaper to run. This page explains, at a high level, what those rewrite steps (the *optimization passes*) do, which ones cost compile time, and how to control and inspect them. You do not need to understand any of this to use Quadrants but it helps when you are trading compile time against runtime, or debugging a surprising result.
+When you call a `@qd.kernel` function, Quadrants first translates the kernel into an internal form then rewrites it, step by step, into something equivalent but cheaper to run. "Equivalent" here means that they yield the exact same observable behavior and produce the exact same outputs [*1] for all valid inputs, regardless of how they are restructured. This page explains, at a high level, what those rewrite steps (the *optimization passes*) do, which ones cost compile time, and how to control and inspect them. You do not need to understand any of this to use Quadrants but it helps when you are trading compile time against runtime, or debugging a surprising result.
+
+[*1] To within floating-point precision, not bit exact. In particular, fast maths optimizations are fairly approximate.
 
 ## Key terms
 
--- a/docs/source/user_guide/optimization_passes.md
+++ b/docs/source/user_guide/optimization_passes.md
@@ -4,7 +4,7 @@ When you call a `@qd.kernel` function, Quadrants first translates the kernel int
 
 ## Key terms
 
-A handful of technical terms show up throughout this page. They are defined once here and used freely below.
+Let's start by defining terms that will be used throughout the page and are necessary in order to be able to understand it.
 
 - **Kernel** - a function you decorate with `@qd.kernel`. It is the unit that Quadrants compiles and launches.
 - **IR (intermediate representation)** - the compiler's internal version of your kernel: a flat list of small, explicitly-typed instructions, sitting between your Python source and the final machine code. Every pass reads and rewrites the IR; none of it is something you write by hand.
--- a/docs/source/user_guide/optimization_passes.md
+++ b/docs/source/user_guide/optimization_passes.md
@@ -63,7 +63,7 @@ A **control-flow graph** is a map of your kernel's basic blocks together with th
 
 Building and analysing the CFG is the most expensive optimization in the pipeline, which is why it runs at most once per simplify stage rather than every round.
 
-**Per-task scoping.** The CFG optimization runs **once per offloaded task**, on that task alone, rather than across the whole kernel. This is both faster to analyse and safe: because each task is a separate device launch, a value held in a register in one task cannot survive into the next one, so there is never anything to forward across a task boundary anyway. Anything written to global memory is treated as potentially read by a later task, so no store another task might need is dropped.
+**One CFG per offloaded task.** The CFG optimization is built and run separately for each offloaded task, over that task's IR alone — never over the whole `qd.kernel` at once. This is both faster to analyse and safe: because each task is a separate device launch, a value held in a register in one task cannot survive into the next one, so there is never anything to forward across a task boundary anyway. Anything written to global memory is treated as potentially read by a later task, so no store another task might need is dropped.
 
 ## Controlling the passes
 
--- a/docs/source/user_guide/optimization_passes.md
+++ b/docs/source/user_guide/optimization_passes.md
@@ -1,16 +1,16 @@
 # Optimization passes
 
-When you call a `@qd.kernel` function, Quadrants does not ship your Python straight to the GPU. It first translates the kernel into an internal form and rewrites it, step by step, into something equivalent but cheaper to run. This page explains, at a high level, what those rewrite steps (the *optimization passes*) do, which ones cost compile time, and how to control and inspect them. You do not need to understand any of this to use Quadrants — the defaults are good — but it helps when you are trading compile time against runtime, or debugging a surprising result.
+When you call a `@qd.kernel` function, Quadrants first translates the kernel into an internal form then rewrites it, step by step, into something equivalent but cheaper to run. This page explains, at a high level, what those rewrite steps (the *optimization passes*) do, which ones cost compile time, and how to control and inspect them. You do not need to understand any of this to use Quadrants but it helps when you are trading compile time against runtime, or debugging a surprising result.
 
 ## Key terms
 
-A handful of words show up throughout this page. They are defined once here and used freely below.
+A handful of technical terms show up throughout this page. They are defined once here and used freely below.
 
-- **Kernel** — a function you decorate with `@qd.kernel`. It is the unit that Quadrants compiles and launches.
-- **IR (intermediate representation)** — the compiler's internal version of your kernel: a flat list of small, explicitly-typed instructions, sitting between your Python source and the final machine code. Every pass reads and rewrites the IR; none of it is something you write by hand.
-- **Pass** — one transformation step over the IR. An *optimization* pass rewrites the IR into a form that produces the **same results** but runs faster or uses less memory. (Some passes are not optimizations but *lowering* steps — they translate high-level constructs into lower-level ones; this page focuses on the optimizations.)
-- **Basic block** — a straight-line run of instructions with no branches into or out of the middle. Control flow (`if`, loops) connects blocks together.
-- **Offloaded task** — after the *offload* step, your kernel is split into one or more tasks, and each task becomes a single device launch: one GPU grid launch on a GPU backend, or one parallel loop on CPU. A simple kernel is usually one task; a kernel with, say, a short serial preamble followed by a big parallel loop becomes several tasks that run back to back.
+- **Kernel** - a function you decorate with `@qd.kernel`. It is the unit that Quadrants compiles and launches.
+- **IR (intermediate representation)** - the compiler's internal version of your kernel: a flat list of small, explicitly-typed instructions, sitting between your Python source and the final machine code. Every pass reads and rewrites the IR; none of it is something you write by hand.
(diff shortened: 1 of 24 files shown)


docs/source/user_guide/optimization_passes.md
```diff
@@ -0,0 +1,90 @@
+# Optimization passes
+
+When you call a `@qd.kernel` function, Quadrants does not ship your Python straight to the GPU. It first translates the kernel into an internal form and rewrites it, step by step, into something equivalent but cheaper to run. This page explains, at a high level, what those rewrite steps (the *optimization passes*) do, which ones cost compile time, and how to control and inspect them. You do not need to understand any of this to use Quadrants — the defaults are good — but it helps when you are trading compile time against runtime, or debugging a surprising result.
```
duburcqa: "Quadrants does not ship your Python straight to the GPU."

This is a strange wording. I don't even understand what could be "ship your Python straight to the GPU" in practice.

I would recommend going straight to the point and explain directly that "When you call a `@qd.kernel` function, Quadrants first translates the kernel into an internal form and rewrites it, step by step, into something equivalent but cheaper to run."

I would also clarify what "something equivalent" means. Basically, without torch decorator, it could run out of the box in native PyTorch (well, to some extent, but you see what I mean, it is mostly valid PyTorch-based python code, with a few specific semantics to further optimise on GPU). The opposite is also mostly true. if you add @qd.kernel on top of the torch-based python native function, it is likely to run on Quadrants.

duburcqa: "This page explains, at a high level, what those rewrite steps (the *optimization passes*) do,"

---> It is a bit confusing. it makes it sounds like to entire translation work is about optimisation passes, which is not the case.

duburcqa: > You do not need to understand any of this to use Quadrants — the defaults are good —

Well, not really. We forcibly disable cfg_optimization in Genesis typically.

Moreover, you are using em-dash. I don't know if it is on purpose.

docs/source/user_guide/optimization_passes.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Key terms
+
+A handful of words show up throughout this page. They are defined once here and used freely below.
```
duburcqa: Weird wording I think but ok.

docs/source/user_guide/optimization_passes.md
```diff
@@ -0,0 +1,90 @@
...
+All of these are fields of `CompileConfig`, so you set them at `qd.init(...)` (or via the matching `QD_<UPPERCASE_NAME>` environment variable). See [qd.init options](./init_options.md) for the full list and the environment-variable convention.
+
+| Option | Default | Effect |
+|--------|---------|--------|
+| `cfg_optimization` | `True` | Turn the CFG optimization on/off. Turning it **off** makes compilation up to ~6× faster while costing ~1–5% of runtime speed — worth it when compile time dominates and the runtime delta is acceptable. |
+| `opt_level` | `1` | `0` disables the two heavier passes (CSE and CFG optimization). |
+| `advanced_optimization` | `True` | The fixed-point simplify loop above. Set to `False` to run just a single basic cleanup pass instead — much faster to compile, much less optimized. |
+| `constant_folding` | `True` | Enables the constant-folding pass. |
+| `fast_math` | `True` | Allows IEEE-relaxed floating-point rewrites (e.g. fusing a multiply and add). Covered in [qd.init options](./init_options.md#fast_math). |
+
```
duburcqa, line 81: I would add a section to explain that disabling this pass is not recommended, because it may cause unexpected failure. The tests are mostly validating the default optimisation options, which may hide bugs that would only surfaces when these optimisations are disabled.

I would recommend you to have a look at the code comments in Quadrants' init from genesis-world: https://github.com/Genesis-Embodied-AI/genesis-world/blob/92d9ae741c07279d44380b109beee391c1517bc8/genesis/__init__.py#L261

docs/source/user_guide/optimization_passes.md
hughperkins: I think it is fairly clear to my mind?:
- It first translates the kernel into an internal form
- then it optimizes it (the 'rewrite' steps)

Maybe the sequencing should be clearer?

"It first translates the kernel into an internal form, then it rewrites it, ... " ?

hughperkins: em-dash is not on purpose. will address

so remove "the defaults are good"?

docs/source/user_guide/optimization_passes.md
hughperkins: "words" => "specialist terms"?

docs/source/user_guide/optimization_passes.md
hughperkins: when you say 'pass', you mean 'fast_math' option? Or you mean passes/options in general?

docs/source/user_guide/optimization_passes.md
hughperkins: I'm not sure I follow your analogy about 'something equivalent'. This is describing the optimization steps, not the translation step. It means that eg CSE will remove the common subexpressions, and reuse one single calculated value. The code is equivaleent, but shorter and faster.

docs/source/user_guide/optimization_passes.md
duburcqa: it is the entire introduction that feels weird. Like I would expect just something explaining that you will start by rigorously defining a handful of specialist terms that will be used throughout this section and necessary to understand it. Something like this. 

Saying 'A handful of specialist terms show up throughout this page.' is already better than the current version, but not stressing the point that it is important to understand what it is referring to in this context to understand this section correctly.

docs/source/user_guide/optimization_passes.md
duburcqa: As you want. I think mentioning that Quadrants made some effort to set good default is a relevant information.

duburcqa: yes, I think it is worth mentioning that there is a first one-to-one translation pass, running before any optimization passes.

duburcqa: It must be clear that these optimizations are not necessary for the code to run. At the moment it is a bit confusing, it sounds like optimizing is the sole job of Quadrants.

duburcqa: I think the notion of equivalent should be made more explicit. It is not clear in the doc and it is definitely not trivial at all even if it sounds simple at first.

docs/source/user_guide/optimization_passes.md
duburcqa: I mean passes/options in general

docs/source/user_guide/optimization_passes.md
hughperkins: then I'm not clear what your request is. Above, you seem to sugest that the quadrnts defaults are not good? (Or at least, not perfect)

hughperkins: I could write: ""Equivalent" here means that they yield the exact same observable behavior and produce the exact same outputs for all valid inputs, regardless of how they are restructured. And "observable" only considers the ordering and values of I/O operations and volatile variables. ". But I dont think this improves the page, considering this is the introductino paragraph.

hughperkins: I feel the curent sequencing in the descriptoin makes it clear that quadrants does both translation and optimization.

duburcqa: You only guarantee floating point accuracy? Not bit exact I guess

duburcqa: I think the first part is good enough (with the floating point accuracy, no need to mention IO and stuffs)

duburcqa: Yes

duburcqa: lol you are right. We did some effort to set good default but they are biased toward maximum performance at the cost of longer compile time, which may be problematic for very complex kernels, so it is important to check these tunables if compile time is getting out of control.

hughperkins: Honestly I have no idea, but I don't think any of the changes are modifying associativity etc. Maybe the fast floating point counts as an optimization pass? And thta clearly changes floating point.

Actually lets ask AI:

"  Structural / memory passes are value-preserving (bit-exact regardless of fast_math). DIE, CSE, LICM, unreachable-code elimination, and the CFG pass's store-to-load forwarding
  + dead-store elimination don't change which arithmetic executes — CSE/LICM reuse the identical computation, store-to-load forwarding returns the exact bits that were stored,
  DSE only drops writes nobody reads. So these never perturb numerics.

"  Arithmetic-rewriting passes + backend codegen are gated on fast_math, and that's where bit-exactness is lost."

And:

"  One caveat I didn't fully trace: constant_fold evaluates ops at compile time. Basic + − × ÷ are correctly-rounded IEEE so they fold exactly, but compile-time evaluation of
  transcendentals could in principle differ from the device's math library.

"  So: other than fast_math, the optimizations are bit-exact except alg_simp's pow strength reduction (FP-precision, not bit-exact) and the signed-zero additive identity
  (sign-of-zero only) — both of which run unconditionally. Want me to trace exactly which float ops constant_fold handles, and/or note these caveats in the doc?"

How does this affect what you feel we should write about floating point?

hughperkins: so I can resolve this thread?

duburcqa: I would say there is no guarantee to be bit-exact. May be, may be not. It is option dependent. Only guaranteed if all optimization passes are disabled.

duburcqa: This is even worst on GPU where fast_math CANNOT be disabled reliably without using 'precise' keyword. So it is not even true to say that it is guaranteed with optimization passes disabled.

duburcqa: I think it would be great to mention the rational behind the default options, for the user to have a mental representation of what they are aiming for and when it is relevant to consider modifying them.

duburcqa:
ok to merge

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -17,10 +88,53 @@ bool cfg_optimization(const CompileConfig &config,
...
+    // redundant because the post-offload per-task cfg ("simplify_III" onward) redoes the intra-task
+    // store-to-load forwarding + dead-store elimination once tasks exist. So for exactly those phases we ditch
+    // cfg, keeping only the cheap dead-alloca cleanup. For ANY other caller of full_simplify on non-offloaded
+    // IR (unit tests, standalone blocks / function bodies that are never offloaded), we must still run the
+    // whole-kernel cfg below, or its forwarding/DSE would be silently lost -- so we fall through.
+    const bool pre_offload_compile_phase =
+        phase == "simplify_I" || phase == "simplify_II" || phase == "pre_autodiff" || phase == "post_autodiff";
+    if (pre_offload_compile_phase) {
+      die(root);
+      return false;
```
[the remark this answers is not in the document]
hughperkins: checked empirically, documented at https://github.com/Genesis-Embodied-AI/perso_hugh/blob/main/doc/cfg_per_offloaded_task_const_bounds_2026jun16.md

Summarized by AI here:

Screenshot 2026-06-16 at 12 20 17

hughperkins:
Genesis benchmark results:

20260616_cfgper

(i.e. neutral, no regression)

hughperkins:
Genesis unit tests passed (on field):

Screenshot 2026-06-16 at 12 54 31

PR #731, opened 2026-06-10
[Misc] Remove 2x2 and 3x3 sym_eig specializations

hughperkins:
The 3x3 closed-form path (dsytrd3 / dsyevq3 / _sym_eig3x3) was a port of Joachim Kopp's "Numerical diagonalization of 3x3 matrices", which is GNU LGPL -- shipped unattributed inside this Apache-2.0 library. (The 2x2 path is original/textbook and is kept.)

Route N=3 through the existing clean cyclic-Jacobi sym_eig_general (Golub & Van Loan, Apache-2.0) and delete the Kopp routines. The 3x3 tests already call qd.sym_eig, so they now exercise Jacobi (incl. the identity case that motivated Kopp's QL fallback); the obsolete cluster-Vulkan skip for the closed-form path is removed.

We also remove the 2x2 closed form, so that we consistently use the cyclic Jacobi form throughout.

User-facing docs (linalg_per_thread.md) updated: 2x2..12x12 cyclic Jacobi; also fixes the inaccurate "Eigen3 computeDirect" labels.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


docs/source/user_guide/linalg_per_thread.md
```diff
@@ -180,10 +178,10 @@ The rotation factor `R` from `A = R @ S` is the rigid alignment that minimises `
 ## Shapes, performance, portability
 
 - **Compile time.**
-  - **Closed-form ops** (`qd.svd`, `qd.sym_eig` 2×2/3×3, `qd.polar_decompose`, `qd.eig`, `qd.solve`) — each call is unrolled per thread into a moderately large block of straight-line code; compile time is generally fine at these shapes.
-  - **Cyclic Jacobi** (`qd.sym_eig` ≥4×4, `qd.make_spd`) — the per-pair Givens step is unrolled but the outer sweep loop is a runtime `range`, so compile time is roughly proportional to `N²` (number of `(p, q)` pairs per sweep) rather than `N² · MAX_SWEEPS`. Concrete numbers on CUDA + LLVM 22.1: ~3 s at N=4, ~30 s at N=6, ~3 min at N=9, ~2 min at N=12 (yes, faster than N=9 — the per-pair body is dominated by `if static(p < q):` filtering).
+  - **Closed-form ops** (`qd.svd`, `qd.polar_decompose`, `qd.eig`, `qd.solve`) — each call is unrolled per thread into a moderately large block of straight-line code; compile time is generally fine at these shapes.
```
duburcqa, line 181: I would suggest to focus on what matters of the end user.

docs/source/user_guide/linalg_per_thread.md
```diff
@@ -180,10 +178,10 @@ The rotation factor `R` from `A = R @ S` is the rigid alignment that minimises `
...
+  - **Cyclic Jacobi** (`qd.sym_eig` all sizes, `qd.make_spd`) — the per-pair Givens step is unrolled but the outer sweep loop is a runtime `range`, so compile time is roughly proportional to `N²` (number of `(p, q)` pairs per sweep) rather than `N² · MAX_SWEEPS`. Concrete numbers on CUDA + LLVM 22.1: ~3 s at N=4, ~30 s at N=6, ~3 min at N=9, ~2 min at N=12 (yes, faster than N=9 — the per-pair body is dominated by `if static(p < q):` filtering).
```
duburcqa: I would suggest to focus on what matters of the end user.

By the way, I don’t think taking >=2min to compile a kernel is really acceptable. Not sure if we should allow more than size 6.

docs/source/user_guide/linalg_per_thread.md
```diff
@@ -180,10 +178,10 @@ The rotation factor `R` from `A = R @ S` is the rigid alignment that minimises `
...
 - **Runtime cost.** Cyclic Jacobi at N=12 with `MAX_SWEEPS=12` does roughly `12 · 66 · 12 ≈ 9500` per-thread arithmetic ops — fast on any modern GPU, but if you're calling it inside a hot kernel for a million elements that's still ~10 GFLOP-equivalent. For larger matrices use a different algorithm (or call quadrants `linalg.*` for a sparse-aware path).
-- **Backend portability.** All ops compile cleanly on CUDA, AMDGPU, Vulkan, and Metal — they are pure register arithmetic with no SIMT primitives, so there is no codegen split. The `qd.sym_eig` ≥4×4 / `qd.make_spd` paths have been verified at N ∈ {4,5,6,9,12} × {f32, f64} × five symmetric-matrix factories on CUDA + Vulkan + AMDGPU; Metal coverage is via the same parametrized tests.
+- **Backend portability.** All ops compile cleanly on CUDA, AMDGPU, Vulkan, and Metal — they are pure register arithmetic with no SIMT primitives, so there is no codegen split. The `qd.sym_eig` / `qd.make_spd` paths have been verified at N ∈ {2,3,4,5,6,9,12} × {f32, f64} × five symmetric-matrix factories on CUDA + Vulkan + AMDGPU; Metal coverage is via the same parametrized tests.
```
duburcqa: Same. I would suggest to focus on high level info for users.

docs/source/user_guide/linalg_per_thread.md
hughperkins: I think you're contradicting yourself.

Compile time clearly *is* important to the end user.

These functions were made for qipc, and @alanray-tech prefers to have the faster runtime, than use a runtime loop, and increase runtime latency by ~20% or so.

docs/source/user_guide/linalg_per_thread.md
hughperkins: removed

docs/source/user_guide/linalg_per_thread.md
hughperkins: (simplified this pargarph)

hughperkins: (in qipc we are using size 12x12)

hughperkins: actually... qipc odesnt use the quadrants sym_eig. It uses Householder-something instead. I'll change this to use a rntime loop when size > 6

hughperkins: Turns out it was alreayd doing this actually 🤔  Anyway, I've change to to > 4, because even 30 seconds sounds high, for a default thing, to be honest.

Screenshot 2026-06-10 at 11 47 58

hughperkins: So, it turns out that even with the outer loop unrolled, N=12 takes 75 seconds to compile.

Apparently, per AI, the way to solve this is to use scratch memory, rather than using fixed sized Matrixes.

This would be an API change.

We should probably just limit sym_eig currently to N <= 4 6 (we only added it recently anyway). And if we want N larger than this, go the scratch route.

Screenshot 2026-06-10 at 12 17 03
Screenshot 2026-06-10 at 12 17 45

hughperkins: Updated to no longer support N > 6 at all.

Updated make_spd to also no longer support N > 6.

None of this is used by qipc in the end actually. We're using Householder. We probably use scratch (see above). I'm not migrating that into quadrants currenlty, because it's under MPL, which is not very copyleft, but it's more copyleft than Apache 2.

duburcqa: > I think you're contradicting yourself. Compile time clearly is important to the end user.

I was referring to "the per-pair Givens step is unrolled but the outer sweep loop is a runtime `range`" and "the per-pair body is dominated by `if static(p < q):` filtering"

docs/source/user_guide/linalg_per_thread.md
```diff
@@ -180,10 +178,9 @@ The rotation factor `R` from `A = R @ S` is the rigid alignment that minimises `
...
-- **Backend portability.** All ops compile cleanly on CUDA, AMDGPU, Vulkan, and Metal — they are pure register arithmetic with no SIMT primitives, so there is no codegen split. The `qd.sym_eig` ≥4×4 / `qd.make_spd` paths have been verified at N ∈ {4,5,6,9,12} × {f32, f64} × five symmetric-matrix factories on CUDA + Vulkan + AMDGPU; Metal coverage is via the same parametrized tests.
+  - **Closed-form ops** (`qd.svd`, `qd.polar_decompose`, `qd.eig`, `qd.solve`) — each call is unrolled per thread into a moderately large block of straight-line code; compile time is generally fine at these shapes.
+  - **Cyclic Jacobi** (`qd.sym_eig`, `qd.make_spd`) — the outer sweep loop is unrolled for `N ≤ 4` and a runtime loop for `N > 4` (unrolling it grows compile time quickly with N). Compile times for a single `qd.sym_eig` call on CUDA + LLVM 22.1 (`QD_OFFLINE_CACHE=0`): ~11 s at N=4 (unrolled); ~4 s at N=6 (runtime). Sizes above 6×6 are intentionally not supported — the inner `(p, q)` Givens steps are still unrolled, so compile time grows steeply with N (~26 s at N=9, ~125 s at N=12).
```
duburcqa: Same, I would remove details about unrolling to justify why it takes time to compile.

Notably, I would replace "Sizes above 6×6 are intentionally not supported — the inner `(p, q)` Givens steps are still unrolled, so compile time grows steeply with N (~26 s at N=9, ~125 s at N=12)." by "Sizes above 6×6 are intentionally not supported to keep compile time reasonable." "[as it grows steeply with N (~26 s at N=9, ~125 s at N=12).]." (if you want to explain a bit)

duburcqa:
ok to merge!

PR #732, opened 2026-06-10
[Algorithms] Revamp algorithms API and provide algos as qd.funcs

hughperkins:
Revamp algorithms API

Note: this is a BACKWARDS INCOMPATIBLE change

Changes:
- remove device_ prefix from device functions, to be consistent with previously existing algorithms, and because device_ is not very intuitive
- create @qd.func versions of each device function, with _func suffix
- user has to supply scratch buffer now
- provide _scratch_size functions to determine what scratch buffer size we need


docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,75 @@
...
...x_rank_match_atomic_or`, and `subgroup` reductions internally, and rely on the kernel-launch boundary (plus `atomic_add` in a few places) for cross-block synchronization rather than any in-kernel grid-scope barrier. Most are called from host (Python) code, not from inside a kernel. Some also ship a composable `@qd.func` form (the `_func` suffix) you can call at the **top level** of your own kernel so the op fuses with your other phases into one compiled kernel / captured graph - currently the device-wide reduce (`reduce_{add,min,max}_func`), the exclusive scan (`exclusive_scan_{add,min,max}_func`), stream compaction (`select_func`), reduce-by-key (`reduce_by_key_add_func`), and the LSB radix sort (`radix_sort_func`). These `_func` forms take the live element count as a **device-resident** `Expr` and a compile-time recursion `DEPTH` (the radix sort calls it `LOG256_MAX_N`), so one captured graph replays for every count up to the depth's capacity. Every algorithm also has a friendly host (Python) entry that validates inputs, derives the depth, and launches the work standalone; the host entry is the right choice unless you specifically need to fuse the op into your own captured graph.
```
hughperkins: not necessarily kernel launches

hughperkins: and in fact, it's only a single kernel launch (which wraps a single func).

hughperkins: this whole paragraph needs to be rewrittne from scratch really

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
 # Algorithms
 
-Device-wide algorithms are primitives that consume and produce whole arrays, executed as one or more kernel launches under the hood. They sit one tier above block and subgroup primitives: they *use* `block.reduce`, `block.exclusive_scan`, `block.radix_rank_match_atomic_or`, and `subgroup` reductions internally, and rely on the kernel-launch boundary (plus `atomic_add` in a few places) for cross-block synchronization rather than any in-kernel grid-scope barrier. The user calls them from host (Python) code, not from inside a kernel.
+Device-wide algorithms operate on whole arrays — reductions, scans, radix sort, stream compaction, and reduce-by-key — built on the block and subgroup primitives one tier below. Each op has a friendly host (Python) entry that validates its inputs, sizes the work, and launches it standalone. Most also ship a composable `@qd.func` variant (the `_func` suffix) you can call at the **top level** of your own kernel, so the op is captured into the same kernel / graph as your surrounding phases. Every op requires caller-owned **scratch**, covered in the next section.
```
duburcqa: What is a "Device-wide algorithm" ?

duburcqa: What is a "friendly host (Python) entry" ?

duburcqa: "Most also ship a composable"

ship is a bit weird in this context no? An algorithm (or op?) "ships" something?

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+| `qd.algorithms.reduce_by_key_add(keys_in, values_in, keys_out, values_out, num_runs, scratch)` | Collapse each consecutive run of equal keys into `(key, sum_of_values)`. Friendly host entry. | yes  | yes\*  | yes    | yes\* |
+| `qd.algorithms.reduce_by_key_add_func(keys_in, values_in, keys_out, values_out, num_runs, scratch, n, VALUE_DTYPE, LOG256_MAX_N)` | Same reduce-by-key as a `@qd.func`, to compose at the top level of your own kernel (device-resident count `n`, compile-time `LOG256_MAX_N`; `VALUE_DTYPE` only for the `values_out` zero-init). | yes  | yes\*  | yes    | yes\* |
+| `qd.algorithms.parallel_sort`                               | Odd-even merge sort (in-place, key or key-value). **Deprecated**: prefer `radix_sort`. | yes  | yes\*  | yes    | yes\* |
+| `qd.algorithms.PrefixSumExecutor`                           | Inclusive in-place prefix sum (i32 only). **Deprecated**: prefer `exclusive_scan_add`. | yes  | no     | yes    | no    |
```
duburcqa: Why having both instead of just, say 'reduce_by_key_add', and automatically determine if it is the top level loop or not, and keep the correct implementation accordingly?

duburcqa: Hum... I think I understand. 'reduce_by_key_add' is a kernel and 'reduce_by_key_add_func' is a func.

Rather than adding a suffix, I think it would be more natural to have some extra hierarchical level, eg `qd.algorithms.func.`. But not a strong opinion.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
 
-\* `device_reduce_*`, `device_exclusive_scan_*`, `device_select`, `device_radix_sort`, `device_reduce_by_key_add`, and `parallel_sort` run anywhere a Quadrants kernel runs; portability is inherited from the underlying block / subgroup primitives.
+\* `reduce_*`, `exclusive_scan_*`, `select`, `radix_sort` / `radix_sort_func`, `reduce_by_key_add`, and `parallel_sort` run anywhere a Quadrants kernel runs; portability is inherited from the underlying block / subgroup primitives.
+
+## Host entries and composable `_func` forms
```
duburcqa: I'm not sure "Host entries" is very clear for the average user. Even if it is accurate, it took me some time to understand what you mean.

duburcqa: Python entries sounds less confusing but maybe it is just me.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+
+Each algorithm comes in two forms:
+
+- **Host entry** (e.g. `reduce_add`, `radix_sort`) — called from Python. It validates its inputs, derives the recursion depth from the input length, and launches the work standalone. Use this unless you specifically need to fuse the op into your own captured graph.
+- **Composable `@qd.func`** (the `_func` suffix, e.g. `reduce_add_func`, `radix_sort_func`) — called at the **top level** of your own `@qd.kernel`, so the op is captured into the same kernel / graph as your surrounding phases.
+
+The two forms differ in how the input size is supplied:
+
+- For a `_func` form the **live element count** is a **device-resident** scalar (named `n`), so it can be data-dependent and never has to be read back to the host — which is what lets the op be captured into a graph that replays without re-tracing. The host entries read the count on the host instead.
```
duburcqa: What is the "live element count" you are referring to? Was it introduced before?

duburcqa: "so it can be data-dependent and never has to be read back to the host"

This is understandable but a bit abstract/mysterious at this point without a concrete example.

duburcqa: "which is what lets the op be captured into a graph that replays without re-tracing"

People mostly care about knowing these routines are graph-friendly. Not really why.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+- The **maximum capacity** of a `_func` form is a **compile-time** constant, `LOG256_MAX_N`: one captured graph replays for any live count up to that capacity. The host entries derive it for you; on the `_func` path you supply it — see [Capacity (`LOG256_MAX_N`)](#capacity-log256_max_n).
```
duburcqa: What is the capacity of an algorithm? This notion was introduced before?

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+
+Because a `_func` form runs entirely as device code it does no host-side validation (a size check would force a device→host read of the count that defeats graph capture), so you must size its `scratch` correctly up front — see [Scratch space](#scratch-space).
```
duburcqa: "device→host" ---> "device-to-host"

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+
+## Capacity (`LOG256_MAX_N`)
+
+The `_func` forms bake their **maximum capacity** in as a compile-time constant, `LOG256_MAX_N`. An op compiled for a given `LOG256_MAX_N` correctly handles any live count `n` with `0 <= n <= 256 ** LOG256_MAX_N`: every op fans in `BLOCK_DIM = 256` elements per staircase level, so the capacity is base-256 in the number of levels (for the reduce / scan family `LOG256_MAX_N` is the number of phases; for the radix sort it is the histogram-scan depth — same base-256 capacity either way). The `_func` template spells it `LOG256_MAX_N` (uppercase, because it is a compile-time `qd.template()`); the `*_scratch_slots` host helpers take the same value as a lowercase `log256_max_n` argument.
```
duburcqa: Why is this manually configured?

---> every op fans in `BLOCK_DIM = 256` elements per staircase level, so the capacity is base-256 in the number of levels (for the reduce / scan family `LOG256_MAX_N` is the number of phases

As an average user, I have no idea what you are talking about. What is a "staircase level"?

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+The capacity grows fast — each level multiplies it by 256:
+
+| `LOG256_MAX_N` | Max count (`256 ** LOG256_MAX_N`) |
+|----------------|-----------------------------------|
+| `1`            | `256` |
+| `2`            | `65,536` |
+| `3`            | `16,777,216` |
+| `4`            | `4,294,967,296` (≈ 4.3 billion) |
+
+`LOG256_MAX_N = 4` already exceeds what a 32-bit index can address (`256 ** 4 == 2 ** 32`), so in practice you rarely need more than 3–4.
```
duburcqa: 4 can be used even if it exceeds max 32-bit index?

docs/source/user_guide/algorithms.md
hughperkins: well, it contrasts with 'block' and 'subgroup' operations. I guess we could simply drop 'device-wide'?

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+
+It must be compile-time because it fixes the number and order of the internal launches: the staircase is statically unrolled to `LOG256_MAX_N` levels, and that fixed launch topology is exactly what lets **one captured graph replay for any count up to the capacity** without re-tracing. The live count `n` flows as a device value while `LOG256_MAX_N` is frozen at trace time.
```
duburcqa: I don't think this level of details is really relevant for the end users.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+
+```python
+def log256_max_n(capacity: int) -> int:
+    d = 1
+    while 256 ** d < capacity:
+        d += 1
+    return d
+```
+
+- **Over-specifying is safe.** A capacity larger than the live count just adds staircase levels that operate on length-1 buffers — harmless identity no-ops. The only cost is a few extra empty launches and marginally larger scratch, so when in doubt, round up.
```
duburcqa: It sounds so harmless that I don't understand why it is not forced to '4' systematically.

docs/source/user_guide/algorithms.md
hughperkins: Yeah, the phrasing is a bit odd. There are two versions of each funciton:
- a host-side python function, that is launched outside fo a kernel, and provides valdiation
- a kernel-side qd.func function, that is launched inside a kernel, is copataible with cuda graph, but does not provide valdiation, and generally is a bit 'closer to the metal'

I guess I can rephrase more like what I just wrote above.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+- **Under-specifying is a bug.** If `256 ** LOG256_MAX_N < n` the op cannot address the tail of the input, and the `_func` path has no host-side guard to catch it. Always cover your worst case.
```
duburcqa: What would happen to me if I don't do this?

docs/source/user_guide/algorithms.md
duburcqa: This may be a stupid question, but at this point I don't understand how I am supposed to specify LOG256_MAX_N.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
 
 ## Scratch space
 
-Every device-wide algorithm in this module decomposes into "per-block partial → cross-block combine → finalize" passes (tree reduction, three-pass Blelloch scan, four-pass radix sort, scan-then-scatter compaction). The per-block partials need somewhere to live between kernel launches - that buffer is called **scratch**. Rather than ask each algorithm to allocate its own (forcing a `qd.field(...)` per call and undermining the no-implicit-allocation contract of the rest of the API), `qd.algorithms` shares a single set of module-level scratch fields across every call.
+Every device-wide algorithm in this module decomposes into "per-block partial → cross-block combine → finalize" passes (tree reduction, three-pass Blelloch scan, four-pass radix sort, scan-then-scatter compaction). The per-block partials need somewhere to live between kernel launches - that buffer is called **scratch**, and **the caller owns it**. Every algorithm takes a mandatory `scratch` argument.
```
duburcqa: "→" is intended?

docs/source/user_guide/algorithms.md
hughperkins: To AI:

- replace 'Device-wide algorithms' with 'Algorithms'
- replace 'Each op has a friendly host (Python) entry that validates its inputs, sizes the work, and launches it standalone. Most also ship a composable `@qd.func` variant (the `_func` suffix) you can call at the **top level** of your own kernel, so the op is captured into the same kernel / graph as your surrounding phases.' with:

"There are two versions of each function:
- a host-side python function, that is launched outside of a kernel, and provides validation
- a kernel-side qd.func function, that is launched inside a kernel, is compatible with cuda graph, but does not provide valdiation, and generally is a bit 'closer to the metal'"

In a new paragraph put:

"Every op requires caller-owned **scratch**, covered in the next section."

hughperkins: replaced ,see my instructions to the AI, above.

docs/source/user_guide/algorithms.md
duburcqa: > Every device-wide algorithm in this module decomposes into "per-block partial → cross-block combine → finalize" passes (tree reduction, three-pass Blelloch scan, four-pass radix sort, scan-then-scatter compaction). The per-block partials need somewhere to live between kernel launches - that buffer is called **scratch**, and **the caller owns it**. Every algorithm takes a mandatory `scratch` argument.

I don't know if it is the best way to frame it. "per-block partial → cross-block combine → finalize" is not very clear, even if I get it because I'm already used to this pattern. "The per-block partials need somewhere to live between kernel launches" is also a surprising way to introduce the need to storing intermediary results.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+
+**Ask first, then allocate.** Each algorithm ships a companion `*_scratch_slots(N)` function - branch-free integer arithmetic, no device round-trip - that returns the minimum number of slots needed for a length-`N` input. Allocate at least that many; allocating more is fine. These functions are **host- and kernel-callable**: pass a Python `int` to size an allocation up front, or call the same function inside a `@qd.kernel` on a device-read `N` to recompute the requirement on-device and validate it against `scratch.shape[0]` (e.g. raise an overflow flag for a yield-and-realloc loop) without ever reading `N` back to the host. The fixed-depth ops follow the same convention: `reduce_scratch_slots(N, log256_max_n)`, `exclusive_scan_scratch_slots(N, log256_max_n)`, and `radix_sort_scratch_slots(N, log256_max_n)` are host- and kernel-callable when you pass the compile-time depth the op is compiled for; called as `reduce_scratch_slots(N)` / `exclusive_scan_scratch_slots(N)` / `radix_sort_scratch_slots(N)` (depth omitted) they auto-pick the minimal depth from `N` and are host-only.
```
duburcqa: That would be great to have an example at some point. Too many information without anything concrete at that point.

duburcqa: I think '_slots' suffix is sufficient, instead of '_scratch_slots'. But not a strong opinion.

duburcqa: Similarly, I'm wondering if some nested module would be better or not, eg `qd.algorithms.slots.select`.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+```python
+import quadrants as qd
 
-**Allocation is lazy.** A scratch field is only allocated on its first `get_scratch_*()` call from inside an algorithm. Programs that never touch `qd.algorithms.*` pay nothing; programs that only touch 4-byte algorithms never allocate the u64 buffer. (The default budget is therefore a per-field worst case, not a fixed cost: a 4-byte-only caller pays 5 MB, not 10 MB.)
+N = 1_000_000
 
-**`qd.reset()` invalidates every scratch field** via an `impl.on_reset` hook, and resets the byte budget back to `DEFAULT_SCRATCH_BYTES`. The next algorithm call after a `qd.init()` reallocates against the fresh runtime at the default capacity. This keeps `qd.init` / `qd.reset` a "clean slate" - all runtime-scoped state (resource handles *and* config) goes away on reset, by design. Apps that need a persistent bump should call `set_scratch_bytes` immediately after each `qd.init`.
+# 4-byte reduce: u32 scratch.
+scratch = qd.Tensor(qd.ndarray(qd.u32, shape=max(qd.algorithms.reduce_scratch_slots(N), 1)))
+qd.algorithms.reduce_add(inp_i32, out=out_i32, scratch=scratch)
```
duburcqa: Where are defined 'inp_i32' and 'out_i32'? It is nice if snippet can be executed.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
 
-**Tuning the budget.** Call `quadrants._scratch.set_scratch_bytes(N)` before any algorithm runs (or before any algorithm runs after a `qd.reset()`). Pass a larger value to cover bigger `N`, or a smaller value to reduce the resident footprint on memory-constrained devices:
+# 8-byte reduce: same slot count, u64 buffer.
+scratch64 = qd.Tensor(qd.ndarray(qd.u64, shape=max(qd.algorithms.reduce_scratch_slots(N), 1)))
+qd.algorithms.reduce_add(inp_f64, out=out_f64, scratch=scratch64)
+```
+
+Several sizing functions return `0` for the trivial / single-tile case (e.g. `reduce_scratch_slots(N)` for `N <= 1`, `exclusive_scan_scratch_slots(N)` for `N <= 256`); wrap the count in `max(..., 1)` so the `qd.Tensor` allocation stays legal, since the algorithm won't touch the buffer in those cases anyway.
```
duburcqa: Why it is the responsibility of the user of doing 'max(..., 1)' instead of being handled internally?

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
+
+**Too-small buffer.** The host-launched algorithms (`reduce_*`, `exclusive_scan_*`, `select`, `reduce_by_key_add`) validate the supplied `scratch` against `*_scratch_slots(N)` and raise `qd.algorithms.InsufficientScratchError` (a `RuntimeError` subclass) **before** launching anything, so a failed call never corrupts the caller's inputs. The exception carries the sizes programmatically:
```
duburcqa: I think this information is surfacing too late in the doc.

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
 
 ```python
-from quadrants import _scratch
-_scratch.set_scratch_bytes(20 << 20)   # 20 MB; covers N up to ~5M for device_select / radix sort
+try:
+    qd.algorithms.reduce_add(arr, out=out, scratch=scratch)
+except qd.algorithms.InsufficientScratchError as err:
+    scratch = qd.Tensor(qd.ndarray(qd.u32, shape=err.required_slots))   # err.required_bytes / err.provided_slots also available
+    qd.algorithms.reduce_add(arr, out=out, scratch=scratch)
```
duburcqa: When is this workflow supposed to be relevant and why are you advertising it?

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
 
-There are **two scratch fields**, one per element width that algorithm partials need to live in:
+| Algorithm | Sizing function | Scratch dtype |
+|-----------|-----------------|---------------|
+| `reduce_{add,min,max}` / `reduce_*_func` | `reduce_scratch_slots(N[, log256_max_n])` | `u32` (4-byte `arr`) / `u64` (8-byte `arr`) |
+| `exclusive_scan_{add,min,max}` / `exclusive_scan_*_func` | `exclusive_scan_scratch_slots(N[, log256_max_n])` | `u32` (4-byte `arr`) / `u64` (8-byte `arr`) |
+| `select` | `select_scratch_slots(N)` | `u32` (always) |
+| `reduce_by_key_add` | `reduce_by_key_scratch_slots(N)` | `u32` (always) |
+| `radix_sort` / `radix_sort_func` | `radix_sort_scratch_slots(N[, log256_max_n])` | `u32` (always, regardless of key width) |
```
duburcqa: How did you choose the name of these methods? Is it matching numpy and/or torch?

docs/source/user_guide/algorithms.md
```diff
@@ -1,50 +1,123 @@
...
 
-The per-algorithm sections below mention scratch only to call out per-algo footprint (so you can size the budget for a known `N`); the mechanics live here.
+The per-algorithm sections below restate the sizing function and footprint for each op.
 
 ## Semantics
 
-### `qd.algorithms.device_reduce_{add,min,max}(arr, out)`
+### `qd.algorithms.reduce_{add,min,max}(arr, out, scratch)`
 
 Device-wide tree reduction over a 1-D tensor: `out[0]` holds `sum(arr)` / `min(arr)` / `max(arr)`. The monoid identity is derived from `arr.dtype` automatically (`0` for `add`; largest representable value for `min` - `+inf` for floats, `INT{32,64}_MAX` for signed ints, `UINT{32,64}_MAX` for unsigned; smallest representable value for `max` - `-inf` for floats, `INT{32,64}_MIN` for signed ints, `0` for unsigned), mirroring the `block.reduce_min` / `subgroup.reduce_min` typed wrappers which don't take an identity for the same reason.
```
duburcqa: I guess it is out of scope, but this doc is hard to understand for an average user.

"Device-wide tree reduction"
"The monoid identity"

docs/source/user_guide/algorithms.md
```diff
@@ -53,32 +126,48 @@ import quadrants as qd
...
-- The last pass writes the final value to `out[0]` directly. The kernel launches are pipelined back-to-back; correctness relies on the kernel-boundary serialization that Quadrants provides between host-launched kernels.
+- Fixed-depth tree reduction emitted as a single kernel launch. The host entry derives the minimal depth `D` (smallest `D` with `256**D >= N`) from `N` and launches one `@qd.kernel` that emits `D` phases internally; each phase uses `BLOCK_DIM = 256` threads per block and reduces 256 elements per block via `block.reduce_{add,min,max}`. For `N <= 256` one phase suffices; up to `256^2 = 65536`, two; and so on. Out-of-range lanes contribute the monoid identity, derived in-kernel from the element dtype (no runtime identity argument).
+- Per-phase partials are written to the caller's `scratch` buffer (u32 for 4-byte dtypes, u64 for 8-byte dtypes; see [Scratch space](#scratch-space)); the final phase writes `out[0]` directly. The phases are separate offloaded launches inside the one kernel, so correctness relies on the same launch-boundary serialization as before.
```
duburcqa: What value do you see in this section for the end user?

docs/source/user_guide/algorithms.md
```diff
@@ -53,32 +126,48 @@ import quadrants as qd
...
+
+Composing the func inside your own kernel (qipc-style): call `reduce_{add,min,max}_func(arr, out, scratch, n, DTYPE, LOG256_MAX_N)` at the **top level** of your `@qd.kernel`. `n` is the live count read **on-device** (e.g. `count[0]`); `DTYPE` is the element dtype (an `ndarray` kernel argument exposes no `.dtype` inside the kernel, so pass it explicitly); `LOG256_MAX_N` is the compile-time phase count - the emitted reduce handles any count `<= 256**LOG256_MAX_N`, so a graph captured for a given `LOG256_MAX_N` is reusable across all such counts. Size `scratch` with `reduce_scratch_slots(capacity_n, LOG256_MAX_N)`. Never nest the call in ordinary runtime `for` / `if` / `while` control flow (that demotes the phase loops and drops the per-phase grid-wide barriers).
```
duburcqa: "(qipc-style)" ?

duburcqa: What is `DTYPE` here?

duburcqa: "ordinary runtime `for` / `if` / `while` control flow"

As opposed to an extra-ordinary runtime `for` / `if` / `while` control flow? I guess you mean non-static loosely speaking, but the wording is still funny I feel.

docs/source/user_guide/algorithms.md
```diff
@@ -238,19 +397,37 @@ Constraints:
...
+    num_runs: qd.types.ndarray(dtype=qd.i32, ndim=1),
+    scratch: qd.types.ndarray(dtype=qd.u32, ndim=1),
+    count: qd.types.ndarray(dtype=qd.i32, ndim=1),
+):
+    # ... (typically sort keys_in / values_in by key first) ...
+    qd.algorithms.reduce_by_key_add_func(keys_in, values_in, keys_out, values_out, num_runs, scratch, count[0], qd.f32, LOG256_MAX_N)
+    # ... more top-level phases ...
+```
+
+Scratch footprint: `reduce_by_key_scratch_slots(N)` ≈ `1.004 * N` u32 slots. See [Scratch space](#scratch-space).
```
duburcqa: I'm not sure this is the best way to present here algorithm. There is a lot of redundant information.

docs/source/user_guide/algorithms.md
```diff
@@ -324,9 +501,13 @@ def init() -> None:
         indices[i] = i
 
 init()
-qd.algorithms.device_radix_sort(
-    keys, tmp_keys=tmp_keys, values=indices, tmp_values=tmp_idx,
-)
+D = 1
+while 256 ** D < N:
+    D += 1
```
duburcqa: Why this must be done manually? Couldn't this be the default if 'D' is not specified in 'qd.algorithms.radix_sort_scratch_slots' and 'qd.algorithms.radix_sort'?

duburcqa: Unless 256 is not standard, I don't have the full context, just the diff.

docs/source/user_guide/algorithms.md
hughperkins: Could do. To AI: lets move qd.func's in algorithms to be underneath qd.algorithms.func. namespace, and remove the `_func` suffix.

docs/source/user_guide/algorithms.md
hughperkins: To AI: replace "## Host entries and composable `_func` forms" with "## Host-side python functions versus kernel-side qd.func forms"

docs/source/user_guide/algorithms.md
hughperkins: To AI:
- remove "that replays without re-tracing".
- I dont think there is any requirement taht the scalar is named n, only the parameter to the function. remove "(named `n`)"
- remove "so it can be data-dependent and never has to be read back to the host". It's not really a question of being data-dependent, the only thing we need is that it's device resident, which you already stated.
- remove " The host entries read the count on the host instead."
- add a new bullet point "Host-side functions receive the count as a primitive integer, from the host-side function-call parameters. If your code needs to first read the count from gpu-side, this will likely cause a gpu pipeline stall."

hughperkins: To AI: Lets put these two bullet points under a 'live element count' sub-heading, and then make a new sub-heading for 'maximum capacity', with its own bullet points.

docs/source/user_guide/algorithms.md
hughperkins: Yes. 4 doesnt itself cause any overflow. You'll only get overflow if you use `n` higher than your data type can support, i.e. you are using 32-bit integers, rather than 64-bit integers (these algorithms support both integer lengths)

docs/source/user_guide/algorithms.md
hughperkins: because it's slower. each additional 1 increases the number of launches. 4 launches is 4 times slower than 1 launch, and 33% slower than 3 launches.

docs/source/user_guide/algorithms.md
hughperkins: silent UB (no validation)

docs/source/user_guide/algorithms.md
hughperkins: for example, in qipc, we're basically hard-coding it to 3.

docs/source/user_guide/algorithms.md
hughperkins: > I'm wondering if some nested module would be better or not, eg qd.algorithms.slots.select.

Interesting

docs/source/user_guide/algorithms.md
hughperkins: I think the AI is verbalizing the 're-entrant' qd.checkpoint 'allocate on overflow' approach that QIPC is using. But I agree it's totally superfluous to this doc, and I'll look at removing this.

docs/source/user_guide/algorithms.md
hughperkins: It's what alan asked for https://github.com/Genesis-Embodied-AI/qipc/blob/main/docs/design/parallel-primitives.md#correctness I didn't actually stop and even think about the names, just used them verbatim. Open to input on names you feel could be good.

One thing though: I lean towards making the algorithm name explicit somehow, rather than just providing some sort, and then having to choose one algorithm that works for all users. There are several algorithms possible, and they each have their own trade-offs. As an end-user I'd like to be able to choose the tradeoff that works for me.

There are two ways we can make the algorithm choices explict:
- method 1. add to the name, e.g. `sort_radix`, `sort_bitonic`
- method 2, use some kind of enum thing, e.g .`sort(algorithm: qd.Template, ...)`, where algorithm is an enum like:
```
class SortAlgorithm(intEnum):
     radix = 1
     bitonic = 2
```

The second method is perhaps more appealing as an api, but way more fiddly to write I feel, and I think under the hood I would want to factorize into sepraate functions anywya, to keep things clean:

```
@qd.func
def sort(algorithm: qd.Template, ...):
    if qd.static(algorithm == SortAlgorithm.radix):
        _sort_radix( ... )
    if qd.static(algorithm == SortAlgorithm.bitonic):
        _sort_bitonic( ... )
```
... so we will still have at least as many functions, just an arguably slightly clean API perhaps?

docs/source/user_guide/algorithms.md
hughperkins: It's totally in-scope. You never reviewed the original algorithms.md doc. I'm totally open to your reviewing the rest of it here/now.

docs/source/user_guide/algorithms.md
hughperkins: fair

docs/source/user_guide/algorithms.md
hughperkins: It's a qaudrants type, such as `qd.i32`. I actually ha the same question whilst reading, so I'll look into making this more explicit.

hughperkins: well, also contrasts with graph do while.

docs/source/user_guide/algorithms.md
hughperkins: I'm not sure I follow "Unless 256 is not standard" 🤔 

In any case, this example is no longer valid, since we are going to remove the host-side functions.

docs/source/user_guide/algorithms.md
hughperkins: (and like, we are entirely revamping the entire API 😅 , so whilst we are at it, no obvious constraints on 'oh we already fixed this, cannot change this')

docs/source/user_guide/algorithms.md
hughperkins: yes, agreed

docs/source/user_guide/algorithms.md
hughperkins: probably can just remove. seems superfluous to the end user anyway.

docs/source/user_guide/algorithms.md
duburcqa: Numpy is passing `kind`: https://numpy.org/doc/stable/reference/generated/numpy.sort.html

duburcqa: For reference, torch opted for doing magic (https://discuss.pytorch.org/t/radix-sort-for-tensor/157550/2):
"You could directly call x.sort() and allow PyTorch to select the right sorting algorithm. Right now, it would call bitonicSort or radixSort depending on the size of the inputs as seen [here](https://github.com/pytorch/pytorch/blob/a85d1f0bcdd02cf18d3b0517337458cb51a18cdb/aten/src/ATen/native/cuda/Sort.cu#L269-L281)."

duburcqa: I don't know what is the best choice honestly. I think users do not care about which sorting method is being used, so if there is a reliable way to determine which one to pick, I would be slightly in favour of doing this, but if it is a risky move or we do not have the man power for this, adding an explicit argument with some default value is fine.

hughperkins: I suppose even in a qd.func, even with a runtime n, we could switch between functions depending on n. And it would be ok to switch to a different function each iteration, if n is right on the boundary. that wouldnt cause a perf issue.

Is n the only thing that determines algorithm though? I think it is NOT the only consideration.

One other consideraiton I think is: how sorted are the inputs already?

For example, in broadphase we are using insertion sort because it is fast when the tensor is alreayd nearly sorted (since the order is fairly stable between iterations).

docs/source/user_guide/algorithms.md
```diff
@@ -1,189 +1,259 @@
...
+| `qd.algorithms.PrefixSumExecutor`                           | Inclusive in-place prefix sum (i32 only). **Deprecated**: prefer `exclusive_scan_add`. | yes  | no     | yes    | no    |
 
-\* `device_reduce_*`, `device_exclusive_scan_*`, `device_select`, `device_radix_sort`, `device_reduce_by_key_add`, and `parallel_sort` run anywhere a Quadrants kernel runs; portability is inherited from the underlying block / subgroup primitives.
+\* `reduce_{add,min,max}`, `exclusive_scan_{add,min,max}`, `select`, `sort`, `reduce_by_key_add`, and `parallel_sort` run anywhere a Quadrants kernel runs; portability is inherited from the underlying block / subgroup primitives.
 
-## Scratch space
+## Composable `@qd.func` ops
+
+Every algorithm is a composable `@qd.func` (e.g. `reduce_add`, `sort`): call it at the **top level** of your own `@qd.kernel`, so the op is captured into the same kernel / graph as your surrounding phases. A `while qd.graph_do_while(...):` body counts as top level; never nest the call in ordinary runtime `for` / `if` / `while` control flow, which demotes the phase loops, collapses the per-phase grid-wide barriers, and corrupts the result.
```
hughperkins: should we throw an exception rather than being corrupted?

docs/source/user_guide/algorithms.md
hughperkins: personally, I feel that two functions, with the same name, in different namespaces, are likely two functions that do the same, in a different context. I don't think I would expect that they actually pair with each other.

docs/source/user_guide/algorithms.md
hughperkins: there are examples later on now

docs/source/user_guide/algorithms.md
```diff
@@ -1,256 +1,374 @@
...
+| `qd.algorithms.{reduce,exclusive_scan,select,reduce_by_key,sort}_scratch_slots(...)` | Host- and kernel-callable helpers returning the scratch slot count each op needs. | yes | yes |
+| `qd.algorithms.parallel_sort` | Odd-even merge sort (in-place, key or key-value). **Deprecated**: prefer `sort`. | no | yes |
+| `qd.algorithms.PrefixSumExecutor` | Inclusive in-place prefix sum (i32 only). **Deprecated**: prefer `exclusive_scan_add`. | no | yes |
 
-\* `device_reduce_*`, `device_exclusive_scan_*`, `device_select`, `device_radix_sort`, `device_reduce_by_key_add`, and `parallel_sort` run anywhere a Quadrants kernel runs; portability is inherited from the underlying block / subgroup primitives.
+These algorithms run on all of Quadrants' GPU backends (CUDA, AMDGPU, Vulkan, Metal); they are GPU-only and are not available on the CPU backend. (On Metal / MoltenVK only 32-bit dtypes are supported - no `i64` / `u64` / `f64`; `PrefixSumExecutor` is CUDA / Vulkan only.) The `*_scratch_slots` helpers are plain integer arithmetic, so they also run on the host / CPU.
```
duburcqa: No reason to mention 'MoltenVK' here.

docs/source/user_guide/algorithms.md
```diff
@@ -1,256 +1,374 @@
...
 
-## Scratch space
+## Composable `@qd.func` ops
+
+Every algorithm is a composable `@qd.func` (e.g. `reduce_add`, `sort`): call it at the **top level** of your own `@qd.kernel`, so the op is captured into the same kernel / graph as your surrounding phases. The function is annotated with `requires_top_level=True`, so attempting to use it outside of top level will throw an exception [FIXME: implement this]. Note that within a `qd.checkpoint`, `qd.loop_do_while` both count as being at top-level.
```
duburcqa: "[FIXME: implement this]" ?

---> `qd.checkpoint` is not merged.

docs/source/user_guide/algorithms.md
```diff
@@ -1,256 +1,374 @@
...
+`LOG256_MAX_N = 4` already exceeds what a 32-bit index can address (`256 ** 4 == 2 ** 32`), so in practice you rarely need more than 3-4.
 
-**Tuning the budget.** Call `quadrants._scratch.set_scratch_bytes(N)` before any algorithm runs (or before any algorithm runs after a `qd.reset()`). Pass a larger value to cover bigger `N`, or a smaller value to reduce the resident footprint on memory-constrained devices:
+It must be compile-time because it fixes the number and order of the internal launches. Each op runs as a *staircase*: a fixed ladder of `LOG256_MAX_N` levels where every level is a separate launch that shrinks the working set by a factor of `BLOCK_DIM = 256` (a level handling `k` elements feeds `ceil(k / 256)` to the next, until one element remains). Statically unrolling that staircase to `LOG256_MAX_N` levels gives a fixed launch topology, which is exactly what lets **one captured graph replay for any count up to the capacity** without re-tracing. The live count `n` flows as a device value while `LOG256_MAX_N` is frozen at trace time.
```
duburcqa: I'm not sure it is necessary to provide so many details to the end user.

docs/source/user_guide/algorithms.md
```diff
@@ -1,256 +1,374 @@
...
+    while 256 ** d < capacity:
+        d += 1
+    return d
 ```
 
-`set_scratch_bytes` raises `RuntimeError` if any scratch field has already been allocated in the current runtime cycle (re-`qd.init`-ing wipes that constraint). `scratch_bytes` must be a positive multiple of 8.
+- **Over-specifying is safe.** A capacity larger than the live count just adds staircase levels that operate on length-1 buffers - harmless identity no-ops. The only cost is a few extra empty launches and marginally larger scratch, so when in doubt, round up.
+- **Under-specifying is a bug.** If `256 ** LOG256_MAX_N < n` the op cannot address the tail of the input, and there is no host-side guard to catch it. Always cover your worst case.
+
+Size `scratch` with the **same** `LOG256_MAX_N` you compile the op for - `reduce_scratch_slots(capacity, log256_max_n)`, `exclusive_scan_scratch_slots(capacity, log256_max_n)`, `sort_scratch_slots(capacity, log256_max_n)` - see [Scratch space](#scratch-space). (The `*_scratch_slots` helpers can also derive the minimal depth from `N` for you when you omit the argument - but that path is host-only, so when composing the op into a kernel always pass the explicit depth.)
```
duburcqa: It is not clear from the doc why you are not forcing 3 by default, and adding some troubleshooting section to explain the exception for people exceeding the limit, explain why there is such limit and how to fix it. Here it seems that using 3 would just cover 99% of the use-cases without users having to bother about some capacity consideration, simplifying the doc for the majority.

docs/source/user_guide/algorithms.md
hughperkins: Yes. You want me to add a FIXME for this?

docs/source/user_guide/algorithms.md
hughperkins: This would mean adding a LOG256_MAX_N parameter to the scratch dimensioning functions, which might seem strange, since we are already providing n as a parmeter, so why cant the functions calculate LOG256_MAX_N themselves?

duburcqa: Hum... let me think. Intuitively if the same default is used everywhere, it would work just fine everywhere. Then in case for failure it does not sounds broken to have to update the capacity everywhere needed consistently for the algorithm and its associated scratch memory allocation.

By the way, is this scratch memory local to every algorithm call or is it persistent across calls?

docs/source/user_guide/algorithms.md
duburcqa: Why mentioning in the doc something that does not exist? This is weird.

hughperkins: I mean, it's going to exist, within a few days, once we merge qd.checkpoint.

docs/source/user_guide/algorithms.md
hughperkins: > By the way, is this scratch memory local to every algorithm call or is it persistent across calls?

No.

It's passed in by the client, who owns it. It will be optimum if they persist it between calls, but it's not required.

hughperkins: > in case for failure it does not sounds broken to have to update the capacity everywhere needed

Having to update it does not sound broken per se. But having to pass ing both log256_max_n and n sounds strange perhaps?

docs/source/user_guide/algorithms.md
duburcqa: I think the terminology is great, but It would be nice to clarify what it means in the doc the first time it is introduced

duburcqa: "Each op has a friendly host (Python) entry that validates its inputs, sizes the work, and launches it standalone."

--> this is no longer true no?

duburcqa: it looks like this entire paragraph requires rewriting now that host-side variants have been removed.

duburcqa: This thread is stale now that host-side variants have been removed.

docs/source/user_guide/algorithms.md
duburcqa: This thread is stale now that host-side variants have been removed.

docs/source/user_guide/algorithms.md
duburcqa: This thread is stale now that host-side variants have been removed.

docs/source/user_guide/algorithms.md
duburcqa: "of a `_func` form " 

--> only the "_func" form exists now.

duburcqa: "The host entries derive it for you; on the `_func` path you supply it" 

--> this is stale no?

docs/source/user_guide/algorithms.md
duburcqa: Same. It requires rewording if there is only the so-called "`_func` form" left now.

docs/source/user_guide/algorithms.md
duburcqa: This formulation is still weird. You introduce what is the capacity after explaining that it is a compile time constant.

docs/source/user_guide/algorithms.md
duburcqa: "The `_func` forms"

duburcqa: I see. Then I guess it should be the default?

docs/source/user_guide/algorithms.md
duburcqa: Is specifying any value larger than 4 raising an exception? Because I think it should be the case. Anything beyond 4 does not really make sense from my understanding.

docs/source/user_guide/algorithms.md
duburcqa: Then it should be mentioned explicitly.

docs/source/user_guide/algorithms.md
duburcqa: This should be mentioned explicitly.

docs/source/user_guide/algorithms.md
duburcqa: As you wish.

docs/source/user_guide/algorithms.md
duburcqa: I don't understand what this variable is. it is not defined in any example I could find. Why is it written in upper-case? I must be missing something.

duburcqa: clarify the doc please.

docs/source/user_guide/algorithms.md
```diff
@@ -1,256 +1,374 @@
...
+| `1`            | `256` |
+| `2`            | `65,536` |
+| `3`            | `16,777,216` |
+| `4`            | `4,294,967,296` (~ 4.3 billion) |
 
-**`qd.reset()` invalidates every scratch field** via an `impl.on_reset` hook, and resets the byte budget back to `DEFAULT_SCRATCH_BYTES`. The next algorithm call after a `qd.init()` reallocates against the fresh runtime at the default capacity. This keeps `qd.init` / `qd.reset` a "clean slate" - all runtime-scoped state (resource handles *and* config) goes away on reset, by design. Apps that need a persistent bump should call `set_scratch_bytes` immediately after each `qd.init`.
+`LOG256_MAX_N = 4` already exceeds what a 32-bit index can address (`256 ** 4 == 2 ** 32`), so in practice you rarely need more than 3-4.
 
-**Tuning the budget.** Call `quadrants._scratch.set_scratch_bytes(N)` before any algorithm runs (or before any algorithm runs after a `qd.reset()`). Pass a larger value to cover bigger `N`, or a smaller value to reduce the resident footprint on memory-constrained devices:
+It must be compile-time because it fixes the number and order of the internal launches.
```
duburcqa: "because it fixes the number and order of the internal launches." 

I don't understand what this means. It would be nice to clarify in the doc or remove details.

docs/source/user_guide/algorithms.md
duburcqa: I cannot find any mention of this anymore in the doc. What happened? This paragraph was important if still applicable, and if not, I would like to understand why it no longer is.

docs/source/user_guide/algorithms.md
duburcqa: Could you recap the final decision?

docs/source/user_guide/algorithms.md
duburcqa: Cool! Then 'Device-wide tree reduction' is still worth clarifying. What is a "tree reduction" was never introduced in the doc and still used once.

docs/source/user_guide/algorithms.md
```diff
@@ -306,56 +424,51 @@ The implementation is a Kogge-Stone hierarchical scan: per-block inclusive scan
...
-        indices[i] = i
-
-init()
-qd.algorithms.device_radix_sort(
-    keys, tmp_keys=tmp_keys, values=indices, tmp_values=tmp_idx,
-)
-# keys is now ascending; indices[k] is the original index of the k-th smallest key. (Stable: ties between equal keys preserve their input-order indices.)
-```
+- Fixed-depth tree reduction. Each phase uses `BLOCK_DIM = 256` threads per block and reduces 256 elements per block via `block.reduce_{add,min,max}`. `LOG256_MAX_N = 1` covers `N <= 256`; `2` covers up to `256^2 = 65536`; and so on. Out-of-range lanes contribute the [identity value](#common-conventions).
+- Per-phase partials are written to the caller's `scratch`; the final phase writes `out[0]` directly.
```
duburcqa: I don't understand what this means.

docs/source/user_guide/algorithms.md
duburcqa: I agree. But the doc should reflect the current state, not the future state.

docs/source/user_guide/algorithms.md
duburcqa: > But having to pass ing both log256_max_n and n sounds strange perhaps?

Yes definitely. If there is a way to resolve this, that would be great.

> It is not clear from the doc why you are not forcing 3 by default, and adding some troubleshooting section to explain the exception for people exceeding the limit, explain why there is such limit and how to fix it. Here it seems that using 3 would just cover 99% of the use-cases without users having to bother about some capacity consideration, simplifying the doc for the majority.

What is the current status regarding my original suggestion? Are we going to take this path? Or it is causing additional issue we have not resolved yet? Having a hard time to follow this thread, probably because I'm missing part of the context.

docs/source/user_guide/algorithms.md
hughperkins: I think you are looking at old code.

This iw what I see:

Screenshot 2026-06-18 at 10 09 16

docs/source/user_guide/algorithms.md
hughperkins: This code is marked 'outdated'. I think you are somehow commenting on out dated code.

hughperkins: This code is marked 'outdated'. I think you are somehow commenting on out dated code.

docs/source/user_guide/algorithms.md
hughperkins: This code is marked 'outdated'. I think you are somehow commenting on out dated code.

docs/source/user_guide/algorithms.md
hughperkins: This code is marked 'outdated'. I think you are somehow commenting on out dated code.

docs/source/user_guide/algorithms.md
hughperkins: can you clarify what you think it means and /or how you feel it might be useful to rphease this please?

maybe we should replace 'internal launches' with 'device kernel launches' or 'internal device kernel alunches' perhaps?

docs/source/user_guide/algorithms.md
hughperkins: out[0] is the first element of the output tensor

whilst working on calculating this, scratch space is used.

we can probably just delete 'Per-phase partials are written to the caller's `scratch`; the final phase', and replace with 'The output is written to out[0]' perhaps?

docs/source/user_guide/algorithms.md
hughperkins: I'm asking for claifiaction on the direction you feel we should go in.

The options I see are:

1. status quo:
- user provides log256_max_n to algorithm functions, along with n
- does not provide to scratch function (n is sufficient)

2. default to log256_max_n = 3
    - provide this to the scratch functions, along with n
    - provide this also to the algorithm functions, along with n

3. some other option we haven't come up with yet

It sounds like we should go with 2, but I need your confirmaiton you prfer 2 please. (or perhaps you can come up with option 3, that we didnt think of yet, which is partly the goal of pointing out the current concerns, so you have this opporutnity).

docs/source/user_guide/algorithms.md
hughperkins: migrated it to be a fixme

docs/source/user_guide/algorithms.md
hughperkins: done

docs/source/user_guide/algorithms.md
hughperkins: Again, I dont think this code still exists? This is what I see

Screenshot 2026-06-18 at 10 57 58

docs/source/user_guide/algorithms.md
duburcqa: > can you clarify what you think it means

I have absolutely no idea what it means.

docs/source/user_guide/algorithms.md
```diff
@@ -1,256 +1,374 @@
...
+- **maximum capacity**, `LOG256_MAX_N`: a **compile-time** constant. One compiled kernel can by used for any live count up to that capacity - see [Capacity (`LOG256_MAX_N`)](#capacity-log256_max_n).
 
-Every device-wide algorithm in this module decomposes into "per-block partial → cross-block combine → finalize" passes (tree reduction, three-pass Blelloch scan, four-pass radix sort, scan-then-scatter compaction). The per-block partials need somewhere to live between kernel launches - that buffer is called **scratch**. Rather than ask each algorithm to allocate its own (forcing a `qd.field(...)` per call and undermining the no-implicit-allocation contract of the rest of the API), `qd.algorithms` shares a single set of module-level scratch fields across every call.
+Because each op runs entirely as device code it does no host-side validation (a size check would force a device->host read of the count that defeats graph capture), so you must size its `scratch` correctly up front - see [Scratch space](#scratch-space).
```
duburcqa: device-to-host

docs/source/user_guide/algorithms.md
duburcqa: ok I think I finally understand. You are using 'UPPERCASE' conventions for input arguments that are supposed to be constant. I think this should be mentioned explicitly somewhere, because somehow I thought it was some kind of global constants, with the algorithms looking for it in local scope...

docs/source/user_guide/algorithms.md
duburcqa: Ok now I understand. You are just using UPPERCASE convention for static input argument. This confused me A LOT. Everything starts to finally make sense...

docs/source/user_guide/algorithms.md
hughperkins: I've renamed the function to `qd.sort`. For now there is just the one implementation, so no need for `kind` paraeterm. We can add that if we ever add other implemtnations.

docs/source/user_guide/algorithms.md
hughperkins: removed 'device-wide' and 'tree'

docs/source/user_guide/algorithms.md
hughperkins: I mean, I thought this convention odd too to be honest, but kind of went with the flow. Happy to change them to lower case if we prefer?

docs/source/user_guide/algorithms.md
hughperkins: this is marked 'out dated', and searching for 'Never nest' turns up nothing for me

docs/source/user_guide/algorithms.md
hughperkins: I wonder how big quadrants tensors can be 🤔 

ok, so apparently indexes are not only 32-bit, but *signed* 32-bit. I'm not quite sure why they are signed, but they are.

So yeah, even 4 goes up to twice the maximum capacity of a quadrants tensor. right

hughperkins: apparently it's signed ,because it just uses the default quadrants integer type, which is i32.

docs/source/user_guide/algorithms.md
duburcqa: Ok I think I finally understand what is happening here well enough to answer your question...

What about only passing N to scratch functions, and determine 'log256_max_n' from it (to be as tight as possible). Then, the user only has to specify manually 'log256_max_n' in the algorithm function to pick a different value than the default.

But if N yields to 2 in the scratch buffer while the algorithm is compiled for 3 by default what would happen? Is it compatible or it is necessary for log256_max_n to match exactly between the scratch functions and the algorithm functions? By the way, it would be great to add this information in the documentation!

docs/source/user_guide/algorithms.md
hughperkins: apparnetly changing the default integer type wont fix this though.

docs/source/user_guide/algorithms.md
duburcqa: Yes it is better :)

docs/source/user_guide/algorithms.md
hughperkins: added excpetion

docs/source/user_guide/algorithms.md
hughperkins: addressed

docs/source/user_guide/algorithms.md
hughperkins: addressed

docs/source/user_guide/algorithms.md
hughperkins: bcaues we dont have host-launched algos any more. No valdiation. No ecxeption.

docs/source/user_guide/algorithms.md
duburcqa: I would prefer please. The idea is not bad, but does not pay off in practice. Too disturbing.

docs/source/user_guide/algorithms.md
duburcqa: "ordinary runtime" is still mentioned.

docs/source/user_guide/algorithms.md
hughperkins: so AI points out that having default set to 3 doesnt stop people having to reason about the maximum: they still need to checck whether their N is no greater than 256 ^ 3. If it is, then not increasing log256_max_n is UB. The only thing adding a default does is to remove people having to pass it in, when N is <= 256 ^3. But by hiding the parameter, you make ti so peple dont notice it, and can easily enter UB regime, without knowing it.

hughperkins: I suppose we could do the reverse: make ti default to 4, and then allow people to pass in 3, for increased speed.

hughperkins: But then people get an algorithm which is 25% slower, without knowing why.

Perhaps explicit is good?

hughperkins: (or maybe we make a function to calculate log256_max_n from n, like basically:

```
def calc_log_256_max_n(max_n):
    return int(ceil(log(max_n) / log(256))
```
)

hughperkins: wait, perhaps you question is: 'can we get scratch() fucntion to return log_256_max_n, and the user passes that into the algorithm?'

Yes, that could work.

duburcqa: >  But by hiding the parameter, you make ti so peple dont notice it, and can easily enter UB regime, without knowing it.

if we cannot raise an exception (I guess because we cannot afford it), then hiding the parameter with a default value is not a viable option.

 > wait, perhaps you question is: 'can we get scratch() fucntion to return log_256_max_n, and the user passes that into the algorithm?'

this is tricky because this parameter must be static.

let’s stick to what we have.

docs/source/user_guide/algorithms.md
hughperkins: the func launches gpu kernels, inside of itself. the numbe of gpu kernels it launches is a function of log256_max_n (presuably a * log256_max_n + b, where a is positive integer, and b is non-negative integer)

docs/source/user_guide/algorithms.md
hughperkins: addressed

docs/source/user_guide/algorithms.md
duburcqa: Oh ok I see. Because it is a top level func. Would be great to clarify this on the documentation (typically i much prefer the terminology « offloaded tasks » over « launches ».)

duburcqa:
Ok to merge

docs/source/user_guide/algorithms.md
hughperkins: addressed a bit

hughperkins:
(running Genesis unit tests and benchmarks)

hughperkins:
Genesis benchmarks ok:

2026022_algo

hughperkins:
Genesis unit tests passing:

Screenshot 2026-06-22 at 08 59 12

PR #740, opened 2026-06-12
[lang] Add @qd.func(requires_top_level=True) to reject nested calls at compile time

hughperkins:


docs/source/user_guide/sub_functions.md
```diff
@@ -36,3 +36,44 @@ def compute(a: qd.Template) -> None:
     for i in range(10):
         increment(a, i)
 ```
+
+## Restricting a func to the top level (`requires_top_level=True`)
```
duburcqa, line 40: Not sure that `requires_` prefix is really needed. As you want if you like it.

docs/source/user_guide/sub_functions.md
```diff
@@ -36,3 +36,44 @@ def compute(a: qd.Template) -> None:
...
+
+**Experimental.** `requires_top_level` is an experimental feature and its behaviour or API may change in a future release.
+
+Some functions contain top-level for-loops, and thus must not be called from within top-level for-loops.
```
duburcqa: "Some functions contain top-level for-loops"

That does not really mean anything. A for-loop is a for-loop. It becomes a top-level for-loop it is used at the top-level. It is not intrinsically top-level. Yet, it may have been implemented in a way that assumes being used in a context where it is top-level.

docs/source/user_guide/sub_functions.md
```diff
@@ -36,3 +36,44 @@ def compute(a: qd.Template) -> None:
...
+
+To both enforce this, and to signal to potential users/AIs that a func must not be called from within a top-level for-loop, you can annotate a `qd.func` with `qd.func(requires_top_level=True)`. This will throws `QuadrantsSyntaxError` at compile time if used within a for-loop.
```
duburcqa: 'to potential users/AIs' --> I would remove this. it is unnecessary.

docs/source/user_guide/sub_functions.md
hughperkins: I prefer requires, because this makes it clear that this is something that the caller needs to consider. This parameter doesn't actually itself affect anything inside the func itself, and the requires makes this clear, I feel.

duburcqa:
ok to merge

hughperkins:
Genesis unit test results:

Screenshot 2026-06-16 at 17 31 53

hughperkins:
Genesis benchmark results:

20260616_1410_requiretop

PR #743, opened 2026-06-16
[Build] Use cached Windows clang download

hughperkins:
setup_clang() on Windows was the only download_dep() caller passing force=True, which unconditionally deleted and re-extracted the ~1GB LLVM/clang archive on every build.py invocation (and could trigger a full re-download when the cached zip's size check failed). Drop force=True and fold build_version into the cache dir name, matching setup_llvm() in llvm.py: download_dep's own cache check now skips re-extraction, while a rebuilt or version-bumped archive still lands in a fresh directory.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/scripts/ti_build/alter.py
+++ b/.github/workflows/scripts/ti_build/alter.py
@@ -12,7 +12,7 @@ import psutil
 # -- own --
 from . import misc
 from .cmake import cmake_args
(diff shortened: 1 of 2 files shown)


hughperkins:
@codex review

PR #746, opened 2026-06-17
[AMDGPU] Don't emit llvm.amdgcn.permlane64 on CDNA

hughperkins:
v_permlane64_b32 is an RDNA-only instruction: it exists on gfx11 (RDNA3) and
gfx12 (RDNA4) but on no CDNA part. quadrants previously enabled the
llvm.amdgcn.permlane64 intrinsic on gfx940/gfx941/gfx942 (CDNA3) as well. On
gfx942 the AMDGPU backend does not cleanly "Cannot select" the intrinsic -- it
selects the V_PERMLANE64_B32 pseudo, which has no valid MC opcode for CDNA, and
then crashes with a bare SIGSEGV in SIInstrInfo::getInstSizeInBytes during the
branch-relaxation pass. This made scene.build() segfault for any kernel using a
cross-half subgroup shuffle (genesis-world #2962).

Gate has_permlane64 to gfx11/gfx12 only, so every CDNA target (and gfx10.x
RDNA1/2) takes the existing LDS-roundtrip software emulation, which produces
correct cross-half results on wave64 hardware. Also drop the
QD_AMDGPU_FORCE_PERMLANE64_FALLBACK env-var escape hatch and correct the
related comments in llvm_context.cpp, runtime.cpp and test_simt.py.Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


ZJLi2013:
Validated this on real CDNA3 hardware — it fixes the gfx942 SIGSEGV. Please consider merging.

**Setup.** AMD **MI300X (gfx942 / CDNA3)**, host ROCm 6.3.2, `rocm/pytorch:rocm7.2.1_ubuntu24.04_py3.12`. Built this branch (commit `406a932`) with `CMAKE_ARGS=-DQD_WITH_AMDGPU=ON` → `quadrants 1.0.3.dev21+g406a932`.

**Result** (minimal Franka `scene.build()` from genesis-world#2962, which uses the cross-half shuffle path that emits `permlane64`):

| Genesis | Quadrants | `scene.build()` |
|---|---|---|
| 1.0.0 | 0.8.0 / 1.0.2 (stock) | 💥 SIGSEGV 139 (`SIInstrInfo::getInstSizeInBytes`) |
| 1.0.0 | **this PR (source)** | ✅ BUILD_OK / STEP_OK |

The only variable is whether quadrants carries this PR, so the fix is confirmed at the codegen level.

**Why it matters beyond that one case.** genesis-world 1.1.1 worked around the crash by rewriting *one* kernel (Franka self-collision) to use `subgroup.bitonic_sort_kv_tiled`, but that's per-kernel: on the released `genesis 1.1.1 + quadrants 1.0.2` stack, other rigid scenes still crash on gfx942 (e.g. the `go2` benchmark SIGSEGVs at build). With this PR the whole rigid benchmark suite (`dex_hand / franka / go2 / anymal / box_pyramid / g1_fall`) builds and runs on gfx942 — i.e. this PR is the general fix, not just for the reported scene.

**Minor build notes** (clean `rocm/pytorch` py3.12 image, `--no-build-isolation`): needs `apt-get install liblz4-dev libssl-dev libzstd-dev`, and `pip install "cmake>=3.17" ninja scikit-build-core nanobind pybind11 pybind11-stubgen ruamel.yaml`, with `/opt/rocm/lib/llvm/bin` on PATH. On py3.12 there's a `DeprecationWarning: ast.Str is deprecated` from `lang/ast/ast_transformers/function_def_transformer.py`.

releated to https://github.com/Genesis-Embodied-AI/genesis-world/issues/2962

hughperkins:
@codex review

hughperkins:
I skim-read through the user-facing doc changes, and look reasonable to me:
- no change for RDNA doc
- changes for CDNA doc corresponding to the earlier description

docs/source/user_guide/subgroup.md
```diff
@@ -110,14 +110,14 @@ Each lane returns the `value` held by the lane whose subgroup-local id equals `i
...
 
 - `value` and `offset` dtypes: same as `shuffle` above; `offset` is a `u32`.
-- Maps to `__shfl_down_sync` on CUDA and `OpGroupNonUniformShuffleDown` on SPIR-V. On AMDGPU it is emulated with `ds_bpermute`; wave64 cross-half offsets (any `offset >= 32` for low-half lanes, or any non-zero `offset` for high-half lanes that lands across the SIMD32 boundary) go through the same `permlane64 + ds_bpermute + select` lowering as `shuffle` - see [AMDGPU wave64 cross-half lowering](#amdgpu-wave64-cross-half-lowering). These operations are added on both RDNA and CDNA.
+- Maps to `__shfl_down_sync` on CUDA and `OpGroupNonUniformShuffleDown` on SPIR-V. On AMDGPU it is emulated with `ds_bpermute`; wave64 cross-half offsets (any `offset >= 32` for low-half lanes, or any non-zero `offset` for high-half lanes that lands across the SIMD32 boundary) go through the same wave64 cross-half lowering as `shuffle` - a single wave-wide `ds_bpermute` on CDNA, or a `permlane64 + ds_bpermute + select` sequence on RDNA - see [AMDGPU wave64 cross-half lowering](#amdgpu-wave64-cross-half-lowering). These operations are added on both RDNA and CDNA.
```
duburcqa: All this doc is very advanced. I guess it is not meant for the average user. Like, I don't think an average end-user like me care about the fact that it maps to `__shfl_down_sync` or whatever. I just want to know if it is highly optimised on all devices, and whether I should expect lower performance on older gen hardware, stuffs like this. Personally I cannot do anything with the information that is implemented as a `permlane64 + ds_bpermute + select` sequence on RDNA.

It seems that you have been following this pattern extensively in this documentation. And know that I'm taking a step back, I'm wondering if all of this is on the right track. Like, there should be a general high-level document explaining what users can expect from this submodule, but then, the documentation for every specific method should sit on the docstring of each specific method directly, not centralised on some markdown general doc.

Anyway, it is out of scope of this PR, but you should have this in mind and track it somewhere.

hughperkins: Yeah, I agree with your first paragraph.

And yes, probably should be handled in a separate PR from this one. Created ticket in internal tracker, as CMP-328

hughperkins: Seems this fails the new doc quality check anyway... so I'll probably have to create a PR to fix subgroup.md first, before merging this PR.

duburcqa: ok let's do this!

hughperkins:
@codex review

duburcqa:
ok to merge

PR #747, opened 2026-06-17
[Build] Skbuild core migration

hughperkins:
Switch `license = {text = "Apache-2.0"}` to the PEP 639 SPDX form
`license = "Apache-2.0"`, clearing the SetuptoolsDeprecationWarning about
project.license as a TOML table (supported since the required setuptools>=77).Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

PR #753, opened 2026-06-23
[Build] Use CMAKE_ARGS directly, drop legacy QUADRANTS_CMAKE_ARGS

hughperkins:
scikit-build-core's canonical CMake-args passthrough is CMAKE_ARGS, so the build.py tooling no longer needs a separate QUADRANTS_CMAKE_ARGS input that it mirrors into CMAKE_ARGS on writeback. Point CMakeArgsManager at CMAKE_ARGS for both input and output and drop the mirror.

Because the manager now parses the same variable it writes, a user who exports CMAKE_ARGS before `build.py wheel`/`--shell`/`-w` has those defines absorbed and re-rendered rather than clobbered (fixes the review concern on #747), while the toolchain options the manager injects are still applied.

Also switch the CI 2_build scripts to export CMAKE_ARGS. The Windows script was exporting the stale GSTAICHI_CMAKE_ARGS, which the manager never read, so Windows CI had been building with default options instead of the intended -DQD_WITH_VULKAN/CUDA/AMDGPU; this fixes that too. Docs updated accordingly.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


duburcqa:
ok to merge

hughperkins:
@codex review

duburcqa:
ok to merge

PR #754, opened 2026-06-23
[Doc] Add AMD GPU backend build instructions

hughperkins:
Document enabling the Linux-only QD_WITH_AMDGPU backend via CMAKE_ARGS, note that the ROCm libdevice is vendored (no ROCm needed to build), and that ROCm is only required at runtime.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/docs/source/user_guide/contributing.md
+++ b/docs/source/user_guide/contributing.md
@@ -110,17 +110,9 @@ export QUADRANTS_CMAKE_ARGS="$QUADRANTS_CMAKE_ARGS -DLLVM_ROOT=/path/to/llvm"
 The AMD GPU backend is Linux-only (it is force-disabled on macOS and Windows) and is off by default, so enable it explicitly through `CMAKE_ARGS`:
 
 ```
-export CMAKE_ARGS="-DQD_WITH_AMDGPU=ON -DQD_WITH_CUDA=OFF"
-./build.py wheel
+./build.py --shell
+CMAKE_ARGS="-DQD_WITH_AMDGPU=ON -DQD_WITH_CUDA=OFF" pip install --no-build-isolation -e . -v
 ```
-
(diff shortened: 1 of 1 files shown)


docs/source/user_guide/contributing.md
```diff
@@ -105,6 +105,15 @@ export LLVM_DIR="/path/to/llvm/"
...
 ```
 
+### Building with the AMD GPU backend (Linux)
+
+The AMD GPU backend is Linux-only (it is force-disabled on macOS and Windows) and is off by default, so enable it explicitly through `CMAKE_ARGS`:
+
+```
+./build.py --shell
+CMAKE_ARGS="-DQD_WITH_AMDGPU=ON -DQD_WITH_CUDA=OFF" pip install --no-build-isolation -e . -v
+```
```
duburcqa, line 115: I would rather generalise the build section to introduce the flags for all platforms and ask users to toggle ON/OFF what they need.

hughperkins: What other platforms do we have? The only other one is Vulkan, which we don't seem to be actively using internally. We get Metal and CUDA by default.

duburcqa: Metal.

hughperkins: Metal builds by default. No args required.

duburcqa: Sure, but you may want to avoid that?

hughperkins: I've never ever felt the need to not build metal. That would be a very esoteric edge case I feel.

For esoteric edge cases, I think the better approach is to use ccmake. Though apparently thats very hidden away in some unobvious seciont https://github.com/Genesis-Embodied-AI/quadrants/blob/main/docs/source/user_guide/contributing.md#building-the-package-for-release-purposes

Thoughts on my breaking out using ccmake to view and set all available options into a separate section, and leaving the amdgpu instructions as-is?

duburcqa: Let's stick to this then.

duburcqa:
ok to merge

PR #762, opened 2026-06-24
[Doc] Fix graph doc issues

hughperkins:
Issue: #

### Brief Summary

- fix up graph.md doc
- also make trying to use streams from graph throw exception

copilot:summary

### Walkthrough

copilot:walkthrough


docs/source/user_guide/graph.md
```diff
@@ -161,11 +149,11 @@ Note that `qd.func`'s are inlined, so you can freely factorize these structures
 
 ### Restrictions
 
-- The counter ndarray may be swapped between calls: the cached graph reads each counter through an indirection slot that is refreshed on every launch, so passing a different ndarray (or alternating between several) replays the cached graph without rebuilding it.
+- The counter ndarray may be swapped between calls: passing a different ndarray (or alternating between several) replays the cached graph without rebuilding it.
```
duburcqa: I realise that the notion of "replaying graph" was never introduced. It is not very intuitive I feel, unless one already has a clear view of what a compute graph is concretely.

docs/source/user_guide/graph.md
```diff
@@ -161,11 +149,11 @@ Note that `qd.func`'s are inlined, so you can freely factorize these structures
...
+On platforms without native device-side conditional graph nodes — currently CUDA pre-SM 9.0 and **AMDGPU** ([HIP](https://rocm.docs.amd.com/projects/HIP/en/latest/what_is_hip.html) has no conditional / while node API as of [ROCm](https://www.amd.com/en/products/software/rocm.html) 7.2) — the value of the `graph_do_while` parameter will be copied from the GPU to the host each iteration, in order to check whether we should continue iterating. This causes a GPU pipeline stall. For nested loops this host round-trip happens once per iteration of each loop level, and each loop-body task is replayed individually, so deeply nested loops on these backends pay correspondingly more host overhead (they remain correct, just slower than the CUDA SM 9.0+ native path). At the end of each loop iteration:
```
duburcqa: > — currently CUDA pre-SM 9.0 and **AMDGPU** ([HIP](https://rocm.docs.amd.com/projects/HIP/en/latest/what_is_hip.html) has no conditional / while node API as of [ROCm](https://www.amd.com/en/products/software/rocm.html) 7.2) —

Just cross-reference 'Backend support' section. It is good enough.

duburcqa: > On platforms without [...]

On GPU backends [...] ?

docs/source/user_guide/graph.md
```diff
@@ -302,8 +304,8 @@ def k1(a: qd.type.NDArray, b: qd.type.NDArray, c: qd.type.NDArray):
 ```
 
 Results:
-- on hardware-accelerated platforms, we only launch a single graph from the host, rather than 3 kernels
-- on other platforms, there is no change: we still launch 3 gpu kernels: no change: not better, not worse
+- on hardware-accelerated platforms, we only launch a single graph from the host, rather than 3 separate kernels
+- on other platforms, there is no change: we still launch 3 separate kernels: no change: not better, not worse
```
duburcqa: > 3 separate kernels

3 separate gpu kernels ?

docs/source/user_guide/graph.md
```diff
@@ -349,8 +350,6 @@ Now:
 - on supported hardware, the cond evaluation takes place on the gpu
     - and we avoid the gpu pipeline stall
 - on unsupported hardware, we still incur the pipeline stall, as before
-    - note that there will be some small acceleration, because the condition evaluation and kernel launch will take place entirely from c++, bypassing python
-    - no worse, incrementally better
```
duburcqa, line 353: > because the condition evaluation and kernel launch will take place entirely from c++, bypassing python

Only this part is worth removing, the rest is relevant for users no?

docs/source/user_guide/graph.md
```diff
@@ -369,8 +368,8 @@ for _ in range(num_its):
 In this case, we have `num_its` launches of the three gpu kernels in k1
 - there is nothing on the host side that waits for anything to finish on the gpu-side
 - there is kernel launch latency associated with:
-    - running k1 from host-side python
-    - launching the gpu kernels for each of fn_1, fn_2, fn_3 from host-side c++
+    - running k1 from host-side
```
duburcqa, line 374: "host-side python" was more explicit I think.

docs/source/user_guide/graph.md
```diff
@@ -393,11 +392,9 @@ k1(a, count)
...
 
-If you do want fixed-size for loops to run optimally on unsupported hardware platforms, we could add a specializd `qd.graph_range_for` function. This would:
-- on graph-do-while-supported hardware: handle adding the additional increment kernel
-- on graph-do-while-unsupported hardware: handle running the loop entirely on the host-side, to avoid adding a gpu pipeline stall
+If you do want fixed-size for loops to run optimally on unsupported hardware platforms, please raise an issue, and we can look into this.
 
-In practice, for our own kernels, i.e. in genesis-world, they largely fall under the do while formulation, see the previous section. However, also have some that used to be do while, but have been migrated to an optimized fixed-size, see next section.
+In practice, for our own [genesis-world](https://github.com/Genesis-Embodied-AI/genesis-world) kernels, they largely fall under the do while formulation, see the previous section. However, also have some that used to be do while, but have been migrated to an optimized fixed-size, see next section.
```
duburcqa: What is the point of this sentence? Like, what Quadrants' users are supposed to do with this information?

docs/source/user_guide/graph.md
hughperkins: would fail doc check, because it's a forward reference.

docs/source/user_guide/graph.md
hughperkins: I think it's confusing without the explanation? We could put this in an 'advanced' section potentailly. A typical end-user is not going to notice the difference until they are heavily optimizing, at which time, they could read an advanced section.

docs/source/user_guide/graph.md
duburcqa: As I said, I think forward reference which are not necessary to read to understand what's going on should be considered fine. It should be reframed as (for details about the supported backend, see 'Backend support' section). At this point, we don't care about which backend is supported or not in practice.

docs/source/user_guide/graph.md
duburcqa: > I think it's confusing without the explanation?

Why confusing? This statements are very clear. Maybe raising more questions, but not confusing by itself I feel.

> We could put this in an 'advanced' section potentailly. A typical end-user is not going to notice the difference until they are heavily optimizing, at which time, they could read an advanced section.

Yes, sounds like a good idea.

duburcqa:
ok to merge

docs/source/user_guide/graph.md
hughperkins: replace "for our own [genesis-world](https://github.com/Genesis-Embodied-AI/genesis-world) kernels" with "for many real-world cases"?

docs/source/user_guide/graph.md
hughperkins: It was, but then we have to make a distinction btween python and c++, and talking about c++ is adding information that the user perhaps doesn't need to know.

docs/source/user_guide/graph.md
duburcqa: yes.

docs/source/user_guide/graph.md
```diff
@@ -268,7 +259,21 @@ while status.yielded:
...
+
+| Feature | `qd.cuda` SM 9.0+ | `qd.cuda` < SM 9.0 | `qd.amdgpu` | `qd.metal` | `qd.vulkan` | `qd.cpu` |
+| --- | --- | --- | --- | --- | --- | --- |
+| `graph=True` | hardware accelerated | hardware accelerated | hardware accelerated | runs (no acceleration) | runs (no acceleration) | runs (no acceleration) |
+| `qd.graph_do_while` | hardware accelerated | host fallback | host fallback | host fallback | host fallback | host fallback |
+| `qd.checkpoint` | GPU-side | GPU-side | GPU-side | GPU-side | GPU-side | host-side |
+
+AMDGPU `graph_do_while` falls back to the host-side loop because HIP does not currently expose conditional / while graph nodes (as of [ROCm](https://www.amd.com/en/products/software/rocm.html) 7.2).
```
duburcqa, line 274: Not clear if this is relevant for end-users.

docs/source/user_guide/graph.md
```diff
@@ -464,5 +464,38 @@ The effect in reality is situation dependent:
...
+
+### What a GPU kernel launch is
+
+When you call a `@qd.kernel` from Python, the numerical work does not run inline in the calling thread. Instead Quadrants *launches* the work onto the GPU: it hands a compiled kernel plus its arguments to the GPU driver, the driver enqueues it onto a stream (an ordered queue of GPU work), and the launch call returns to the host before the GPU has finished - usually before it has even started. Host and GPU run asynchronously.
+
+Each top-level `for`-loop in a kernel is a separate offloaded task, and each offloaded task becomes one GPU kernel. So the three-loop `k1` in the earlier **qd.kernel** section launches three GPU kernels every time you call it. Every launch carries a fixed overhead that is independent of how much data the kernel processes - that per-launch overhead is the "kernel launch latency" this guide keeps referring to. When a kernel crunches a lot of data the overhead is negligible; when it does little work, or when you relaunch small kernels many times in a loop, launch latency can dominate the total runtime.
+
```
duburcqa: This section is great, but I feel it should be moved somewhere else, in some introduction doc to Quadrants, instead of something that is graph-specific.

docs/source/user_guide/graph.md
hughperkins: It's making clear that it falls back to host-side because the AMD GPUs don't support the necessary conditional functionality, not just because we didn't get around to adding support for it.

docs/source/user_guide/graph.md
```diff
@@ -464,5 +464,38 @@ The effect in reality is situation dependent:
...
+### Latency hiding
+
+Because host and GPU run asynchronously, kernel launches overlap with kernel execution: the host can be issuing the next launch while earlier kernels are still running on the GPU. As long as the host issues launches at least as fast as the GPU drains them, the queue never empties, the GPU stays busy, and the launch latency is fully *hidden* behind execution.
+
+So reducing launch latency does not always improve overall throughput. When each kernel runs long enough that its execution time exceeds the launch overhead, the GPU is the bottleneck and shaving launch latency changes nothing - the launches were already hidden behind execution.
+
+Kernel launch latency matters when kernels are relatively small. Then the GPU can finish a kernel before the host has issued the next one, so the GPU sits idle waiting for the next launch and the launch latency is *exposed* on the critical path. In that regime, cutting launch latency - for example by collapsing many launches into a single graph - can directly increase throughput.
```
duburcqa: We already have some other part of the documentation mentioning this issue? If so, it would be great to add some cross reference.

docs/source/user_guide/graph.md
```diff
@@ -464,5 +464,38 @@ The effect in reality is situation dependent:
...
+
+### Why graph_do_while can help slightly even without hardware support
```
duburcqa: I would remove slightly here. it is already in the body. And just say:

'Why using graph_do_while is relevant even without hardware support'

duburcqa:
Great documentation improvements!

docs/source/user_guide/graph.md
duburcqa: As you wish.

docs/source/user_guide/graph.md
hughperkins: done

docs/source/user_guide/graph.md
hughperkins: done

docs/source/user_guide/performance.md
```diff
@@ -0,0 +1,34 @@
...
+Kernel launch latency matters when kernels are relatively small. Then the GPU can finish a kernel before the host has issued the next one, so the GPU sits idle waiting for the next launch and the launch latency is *exposed* on the critical path. In that regime, cutting launch latency can directly increase throughput.
+
+## Reducing launch latency
+
+Quadrants provides several tools that attack launch latency in different ways:
+
+- [Graphs](graph.md) capture a sequence of launches into a graph and replay it as a single launch, so on backends with hardware graph support you pay the Python and C++ per-task costs once when the graph is built, rather than on every call. This is the most direct way to cut launch latency.
+- [Streams](streams.md) let independent top-level for-loops run concurrently instead of serializing on the default stream, so the GPU can stay busy across otherwise-independent launches.
+- [Fastcache](fastcache.md) reduces the time to load cached kernels when a new Python process starts, cutting first-call latency.
+- [Performance dispatch](perf_dispatch.md) benchmarks multiple implementations of an operation at runtime and picks the fastest.
```
hughperkins, line 39: this has nothing to do with launc hlatency

docs/source/user_guide/performance.md
```diff
@@ -0,0 +1,34 @@
...
+
+Kernel launch latency matters when kernels are relatively small. Then the GPU can finish a kernel before the host has issued the next one, so the GPU sits idle waiting for the next launch and the launch latency is *exposed* on the critical path. In that regime, cutting launch latency can directly increase throughput.
+
+## Reducing launch latency
+
+Quadrants provides several tools that attack launch latency in different ways:
+
+- [Graphs](graph.md) capture a sequence of launches into a graph and replay it as a single launch, so on backends with hardware graph support you pay the Python and C++ per-task costs once when the graph is built, rather than on every call. This is the most direct way to cut launch latency.
+- [Streams](streams.md) let independent top-level for-loops run concurrently instead of serializing on the default stream, so the GPU can stay busy across otherwise-independent launches.
+- [Fastcache](fastcache.md) reduces the time to load cached kernels when a new Python process starts, cutting first-call latency.
```
hughperkins: this is not really related to launch latency

docs/source/user_guide/performance.md
```diff
@@ -0,0 +1,34 @@
...
+So reducing launch latency does not always improve overall throughput. When each kernel runs long enough that its execution time exceeds the launch overhead, the GPU is the bottleneck and shaving launch latency changes nothing - the launches were already hidden behind execution.
+
+Kernel launch latency matters when kernels are relatively small. Then the GPU can finish a kernel before the host has issued the next one, so the GPU sits idle waiting for the next launch and the launch latency is *exposed* on the critical path. In that regime, cutting launch latency can directly increase throughput.
+
+## Reducing launch latency
+
+Quadrants provides several tools that attack launch latency in different ways:
+
+- [Graphs](graph.md) capture a sequence of launches into a graph and replay it as a single launch, so on backends with hardware graph support you pay the Python and C++ per-task costs once when the graph is built, rather than on every call. This is the most direct way to cut launch latency.
+- [Streams](streams.md) let independent top-level for-loops run concurrently instead of serializing on the default stream, so the GPU can stay busy across otherwise-independent launches.
```
hughperkins: this is not related to launch latency either

hughperkins: Opus here: Agreed - moved to a new "Other performance tools" section; streams is about concurrency, not launch latency. Fixed in 886512484.

docs/source/user_guide/performance.md
hughperkins: Opus here: Agreed - moved to "Other performance tools"; fastcache is about kernel load time at process start, not launch latency. Fixed in 886512484.

docs/source/user_guide/performance.md
hughperkins: Opus here: Agreed - moved to "Other performance tools"; perf dispatch picks the fastest implementation, unrelated to launch latency. Fixed in 886512484.

hughperkins:
@codex review

docs/source/user_guide/graph.md
hughperkins: addressed

docs/source/user_guide/performance.md
```diff
@@ -0,0 +1,37 @@
...
+
+So reducing launch latency does not always improve overall throughput. When each kernel runs long enough that its execution time exceeds the launch overhead, the GPU is the bottleneck and shaving launch latency changes nothing - the launches were already hidden behind execution.
+
+Kernel launch latency matters when kernels are relatively small. Then the GPU can finish a kernel before the host has issued the next one, so the GPU sits idle waiting for the next launch and the launch latency is *exposed* on the critical path. In that regime, cutting launch latency can directly increase throughput.
+
+## Reducing launch latency
+
+[Graphs](graph.md) are the main tool for reducing launch latency: they capture a sequence of launches into a single graph launch, so on backends with hardware graph support you pay the Python and C++ per-task costs once when the graph is built, rather than on every call.
```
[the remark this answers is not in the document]
hughperkins: Opus here: Good catch, and confirmed against the code: `Kernel.__call__` -> `launch_kernel` -> `make_launch_context()` runs on every call, so the per-call Python overhead is not amortized to graph-build time. Reworded to say graphs collapse the per-task C++/driver launch cost, that the per-call Python cost is still paid every call, and that only `graph_do_while` cuts that by turning many iterations into one call. Fixed in 38c8ad1ec.

hughperkins:
@codex review

docs/source/user_guide/performance.md
```diff
@@ -0,0 +1,37 @@
+# Performance
+
+This page covers the execution model that the other performance guides build on: what happens when you launch a GPU kernel, where the time goes, and when reducing that time actually speeds up your program. The other pages in this section - graphs, streams, fastcache, and performance dispatch - are specific tools for reducing that cost.
```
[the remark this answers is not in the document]
hughperkins: Opus here: Agreed - this was inconsistent with the "Other performance tools" section, which says streams/fastcache/perf dispatch do not reduce launch latency. Reworded the intro to describe the four as performance tools built on this execution model, rather than tools for reducing launch latency. Fixed in ddb03e6af.

hughperkins:
@codex review

docs/source/user_guide/graph.md
```diff
@@ -44,18 +30,22 @@ my_kernel(x, y)  # first call: builds and caches the graph
...
 
 ### Restrictions
 
 - **No struct return values.** Kernels that return values (e.g. `-> qd.i32`) cannot use graphs. An error is raised if `graph=True` is set on such a kernel.
-- **Primal kernels only.** The `graph=True` flag is applied to the primal (forward) kernel only, not its adjoint. Autodiff kernels use the normal launch path.
+- **Primal kernels only.** The `graph=True` flag is applied to the primal (forward) kernel only, not its adjoint (backward). [Autodiff](autodiff.md) kernels use the normal launch path.
 - **Device-resident ndarrays.** Graph mode bakes device pointers into the cached graph, so all ndarray arguments must be on the GPU. Passing a host-resident ndarray raises an error.
-- **`qd_stream` is incompatible** with `graph=True`. Choose one or the other.
+
+- [streams](streams.md) are **incompatible** with graph.
```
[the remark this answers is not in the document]
hughperkins: Opus here: Thanks - you correctly spotted that qd.stream_parallel() was not being rejected in graph kernels (it was silently serialized inside the graph, per the comment in quadrants/runtime/amdgpu/graph_manager.cpp). However, the intended design is that streams are fully incompatible with graphs, not just the qd_stream launch arg. So rather than narrowing the doc, we closed the gap: qd.stream_parallel() inside a graph=True kernel now raises a QuadrantsSyntaxError at compile time (cb8bca0bd), and streams.md documents both qd_stream and qd.stream_parallel as incompatible with graphs.

hughperkins:
@codex review

python/quadrants/lang/ast/ast_transformer.py
```diff
@@ -1621,6 +1621,12 @@ def build_With(ctx: ASTTransformerFuncContext, node: ast.With) -> None:
             )
         if not ctx.is_kernel:
             raise QuadrantsSyntaxError("qd.stream_parallel() can only be used inside @qd.kernel, not @qd.func")
+        kernel = ctx.global_context.current_kernel
+        if kernel is not None and kernel.use_graph:
```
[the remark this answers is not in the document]
hughperkins: I think it should be fine 🤞

duburcqa:
ok to merge

PR #763, opened 2026-06-24
[CI] Exempt advanced-section terms from doc-quality undefined-term rule

hughperkins:
Rule 2 already allows internal/implementation content inside clearly-marked Advanced / Under the hood / Internals / Implementation sections, but Rule 1 still demanded end-user definitions for every term file-wide. That is self-contradictory: we let such a section discuss compiler internals, then flag it for using compiler vocabulary (e.g. bitcode, IR, instruction selection). Extend the advanced-section exemption to Rule 1 so terms first introduced inside such a section do not need a definition.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


docs/source/user_guide/contributing.md
```diff
@@ -180,9 +180,19 @@ The agent reports up to 5 violations, each annotated with the host file's hotnes
...
+Uses an AI agent to review documentation changes for an end-user audience (someone writing Quadrants kernels in Python, not a compiler engineer).
 
-The following do not count as violations: references to public APIs that the author links to their docs and/or labels as public; brief reader-directed pointers suggesting the reader could contribute upstream or file an issue; the core public API vocabulary a user already knows (e.g. `@qd.kernel`, `@qd.func`, `qd.Template`, fields, ndarrays). The agent reports up to 10 violations. This check is delayed by 30 minutes, to avoid running repeatedly if multiple commits pushed with a short delay between each.
+
+Let's first define an *advanced / internal section* as a clearly-marked section whose heading contains "Advanced", "Under the hood", "Internals", or "Implementation"; it targets a more advanced reader rather than the typical end user.
+
+
```
duburcqa, line 188: This formulation is weird, there is no need to "define" something that is mentioned only once. And I would rather move it at the end. It is weird to introduce the exception before the rules.

duburcqa:
ok to merge

docs/source/user_guide/contributing.md
hughperkins: It's mentioned twice, which is why I facotrized out. It's used by both rule 1, and by rule 2.

PR #768, opened 2026-07-03
[Bug] Fix DLPack field offset truncation and two allocator bugs.

duburcqa:
Issue: N/A

### Brief Summary

copilot:summary

Three host-side memory-safety fixes, found while root-causing a Genesis simulation NaN that appeared once a SNode tree grew past 2 GiB:

- **DLPack export truncated in-tree field byte offsets to signed `int`** (`field_to_dlpack`). For a field placed past 2^31 bytes in its tree, the wrapped offset failed the `>= 0` guard and was silently dropped, so the exported view aliased the tree base: reads returned unrelated fields and writes corrupted them. Fixed by keeping the offset `size_t`/`uint64_t` end to end (`DLTensor::byte_offset` is `uint64_t`). Regression test: `test_dlpack_field_offset_past_2gib` (marked `slow` + `run_in_serial`, 2.24 GB allocation), verified to fail before the fix and pass after.
- **`UnifiedAllocator` overstated the chunk tail** (`tail = data + size + allocation_size` instead of `data + allocation_size`), so a follow-up allocation could be handed an address past the real end of the mmap'd chunk. Regression test added in `host_memory_pool_test.cpp`.
- **`CachingAllocator::allocate` dereferenced an erased iterator** (`it_blk->second` after `mem_blocks_.erase(it_blk)`); erase order swapped.

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/python/dlpack_funcs.cpp
+++ b/quadrants/python/dlpack_funcs.cpp
@@ -215,8 +215,8 @@ pybind11::capsule field_to_dlpack(Program *program, SNode *snode, int element_nd
         "'to_dlpack'.")
   }
 
-  // A SNode tree packs every field of a program stage, so in-tree byte offsets routinely exceed
-  // 2^31; the offset must stay 64-bit end to end (DLTensor::byte_offset is uint64_t).
+  // A SNode tree packs every field of a program stage, so in-tree byte offsets routinely exceed 2^31; the offset must
+  // stay 64-bit end to end (DLTensor::byte_offset is uint64_t).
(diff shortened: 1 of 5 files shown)


hughperkins:
@codex review

hughperkins:
checklist:
- codex review passing (earlier, rechecking now)
- test added for 2GB dlpack
- no user-facing doc changes required
- all CI checks passing, including agent checks of factorization etc

=> ok to merge (once codex review comes back ok)

hughperkins:
codex review passed 

=> ok to merge

PR #773, opened 2026-07-10
[AMDGPU] Keep fdiv correctly rounded under fast_math (fixes modulo bug #749)

### LINKED ISSUE #749: AMDGPU modulo bug
To reproduce:

```
"""Standalone quadrants reproducer for AMDGPU modulo bug.

Bug: `float_x % field_int` where the int is loaded from a scalar field
returns the unreduced value instead of applying the modulo, on AMDGPU only.

  float x = 180.0
  int   m = 180           (literal)        →   floor(x) % m      = 0   ✓ (OK on AMDGPU)
  int   m = 180           (scalar field)   →   floor(x) % m      = 180 ✗ (BUG on AMDGPU)
  int   m = 180           (scalar field)   →   int(floor(x)) % m = 0   ✓ (workaround)

Run on both backends:
  python qd_modulo_bug.py --backend cpu
  python qd_modulo_bug.py --backend gpu
"""
import argparse
import quadrants as qd

ap = argparse.ArgumentParser()
ap.add_argument("--backend", choices=["cpu", "cuda", "amdgpu"], required=True)
args = ap.parse_args()

backend_map = {"cpu": qd.cpu, "cuda": qd.cuda, "amdgpu": qd.amdgpu}
qd.init(arch=backend_map[args.backend], default_fp=qd.f32, default_ip=qd.i32)

# Scalar field holding the divisor (180) — mirrors how Genesis stores support_res.
mod = qd.field(dtype=qd.i32, shape=())
mod[None] = 180

out = qd.field(dtype=qd.i32, shape=(6,))


@qd.kernel
def repro(mod: qd.template(), out: qd.template()):
    x = qd.f32(180.0)
    m_const = qd.i32(180)
    m_field = mod[None]

    # (A) float % int-literal              — OK on both
    out[0] = qd.i32(qd.math.floor(x) % m_const)
    # (B) float % int-from-field           — BUG on AMDGPU
    out[1] = qd.i32(qd.math.floor(x) % m_field)
    # (C) int-cast-first, then % field     — workaround
    out[2] = qd.i32(qd.math.floor(x)) % m_field
    # (D) int-cast-first, then % literal   — OK
    out[3] = qd.i32(qd.math.floor(x)) % m_const
    # (E) sanity: field value itself
    out[4] = m_field
    # (F) pure int % int-from-field
    out[5] = qd.i32(180) % m_field


repro(mod, out)
r = out.to_numpy().tolist()
print(f"backend={args.backend}")
print(f"  (A) i32(floor(180.0) % 180_literal)       = {r[0]}   expected 0")
print(f"  (B) i32(floor(180.0) % 180_field)         = {r[1]}   expected 0  <-- BUG if 180")
print(f"  (C) i32(floor(180.0)) % 180_field         = {r[2]}   expected 0")
print(f"  (D) i32(floor(180.0)) % 180_literal       = {r[3]}   expected 0")
print(f"  (E) 180_field                             = {r[4]}   expected 180")
print(f"  (F) 180_literal % 180_field               = {r[5]}   expected 0")
assert r[0] == 0
assert r[2] == 0 and r[3] == 0
assert r[4] == 180
assert r[5] == 0
if r[1] != 0:
    print(f"\n!!! BUG CONFIRMED on {args.backend}: (B) returned {r[1]}, should be 0")
else:
    print(f"\nOK on {args.backend}: no bug, (B) returned 0 as expected")
```

ZJLi2013:
Fixes: #749

## Summary

On AMDGPU, `float % (int loaded from a field)` returns the *unreduced* dividend instead of the remainder (reported in #749 as the "AMDGPU modulo bug"). The root cause is not modulo-specific: **AMDGPU float division is being lowered to an approximate reciprocal**, so exactly-divisible cases like `180.0 / 180.0` return `0.99999994`. Any `floor(a / b)` — spatial hashing, grid indexing (`support_res`), bucketization — silently lands one cell off. Modulo just makes it obvious because it is lowered as `a - floor(a/b)*b`.

Fix: clear the `afn` (ApproxFunc) fast-math flag for the emitted `FDiv` so division stays correctly rounded. `afn` is meant for transcendentals, not basic arithmetic; the change is a no-op on CPU/CUDA (their `fdiv` is already correctly rounded).

## Root cause

`compile_config.fast_math` defaults to `true`. `TaskCodeGenLLVM::initialize_context()` sets global fast-math flags on the IRBuilder, including `setApproxFunc()`:

```cpp
llvm::FastMathFlags fast_flags;
fast_flags.setNoInfs();
fast_flags.setNoSignedZeros();
fast_flags.setAllowReassoc();
fast_flags.setApproxFunc();          // afn
builder->setFastMathFlags(fast_flags);
```

Every instruction, including `BinaryOpType::div` → `CreateFDiv`, inherits `afn`. The AMDGPU backend uses `afn` on `fdiv` to select the fast `v_rcp_f32`-based reciprocal path (~2.5 ULP) instead of a correctly-rounded division. x86 and NVPTX ignore `afn` for `fdiv` (their hardware divide is correctly rounded regardless), so the bug is AMDGPU-only.

Why `float % literal` looks fine but `float % field` does not: a constant divisor is folded at compile time (`constant_fold.cpp`, exact), while a runtime field value takes the `fdiv` path and hits the approximate reciprocal.

Trace of `floor(180.0) % mod[None]` (mod=180), storing every intermediate as f32:

| step | AMDGPU (before) | CPU |
|---|---|---|
| `x / mf`   (float / field-int) | `0.99999994` | `1.0` |
| `floor(x / mf)`                | **`0.0`**    | `1.0` |
| `floor(x / const_180)`         | `1.0`        | `1.0` |
| `x - floor(x/mf)*mf`  (mod)    | **`180.0`**  | `0.0` |

Setting `fast_math=False` makes `floor(x/mf) = 1.0` and `mod = 0` on AMDGPU, confirming the trigger is `afn`-on-`fdiv`.

## Fix

In `TaskCodeGenLLVM::visit(BinaryOpStmt)` for `BinaryOpType::div` (real branch), clear `afn` for the emitted `FDiv` via a `FastMathFlagGuard`, leaving the builder's other fast-math flags untouched. `mod` and `floordiv` are lowered to `BinaryOpType::div` in `demote_operations.cpp`, so this one spot covers `float / int`, `floordiv`, and `%` alike.

Two non-obvious points worth flagging for review:

- The flag must be set on the **builder** (so it applies as `CreateFDiv` builds the instruction). A post-hoc `inst->setFastMathFlags(...)` after `CreateFDiv` did *not* take effect on AMDGPU in practice.
- Emitted AMDGPU IR for `180.0 / field`, before → after:

```
- %12 = fdiv reassoc ninf nsz afn float 1.800000e+02, %11   ; approximate -> 0.99999994
+ %12 = fdiv reassoc ninf nsz     float 1.800000e+02, %11   ; correctly rounded -> 1.0
```

## Testing

Standalone quadrants reproducer from #749 (`qd_modulo_bug.py`, cases A–F) plus a step-by-step intermediate-value diagnostic. Built from source and validated on:

- **R9700 (gfx1201 / RDNA4)**, quadrants 1.1.1 (`main` @ `2e2b43b40`)
- **MI300 (gfx942 / CDNA3)**, quadrants 1.0.3

| case | before (amdgpu) | after (amdgpu) | cpu |
|---|---|---|---|
| `float % int-literal`     | 0   | 0 | 0 |
| `float % int-from-field`  | **180** | **0** | 0 |
| `int(float) % field`      | 0   | 0 | 0 |
| `floor(180.0 / field_180)`| **0.0** | **1.0** | 1.0 |

No change to non-`floor` float results; `fast_math` approximation for transcendentals (sin/cos/exp/…) is retained.

## Notes / scope

- The fix is in the shared LLVM codegen, not AMDGPU-specific, but is only observable on AMDGPU (CPU/CUDA `fdiv` already ignores `afn`). Clearing the flag keeps behavior consistent across backends.
- Independent of the shuffle correctness / DPP work on other AMD branches; this is a pure arithmetic-precision fix.


hughperkins:
Could we gate this change on using AMD please? That would make me feel more comfortable.

hughperkins:
Could we add some unit tests for this please?

hughperkins:
@paveltc Thoughts on this PR? Does this PR relate to any of your own changes?

hughperkins:
(checked iwth agent: #773 relates to #868, but independent/non-overlapping)

hughperkins:
@codex review

hughperkins:
(agent adding a unit test 🤞 )

hughperkins:
To do: figure out if this affects non-amd platforms

hughperkins:
Unclear if this affects cpu/cuda. The worst case is worse performacne ,not correctness.

=> lets run genesis benchmarks

hughperkins:
running Genesi benchmarks on bench1

hughperkins:
Genesis benchmarks dont look good:

20260827_fdiv_1132

I guess I can run before/afer script to check.

hughperkins:
running before/after script `git pull && python bench_before_after.py --ref 20260828_fdiv_1147 --before main --after ZJLi2013-amd-fdiv-correctly-rounded --num-runs 3 --partition rtx-high --filter dex_hand,g1_fall,g1_fall_accessors`

hughperkins:
Results of running interlevaed on a single node, 3 times per test:
```
=== before/after interleaved (3 runs each, single node) ===
before qd=1dfe12d78  after qd=5ed356618  genesis=56e4aa5d

dex_hand                 4096  cuda                 2        32677.0   0.16        25802.5   3.32   -21.04      2.63  <-- SIG
g1_fall                  4096  cuda                 2      1888174.5   0.49      1505975.0   0.81   -20.24      0.81  <-- SIG
g1_fall_accessors        4096  cuda                 2      1280905.5   1.83      1155385.0   0.32    -9.80      1.86  <-- SIG
[09:17:28] Wrote /home/hugh/git/tmp/20260828_fdiv_1147/meta.yaml
[09:17:28] Done. SIGNIFICANT diffs: dex_hand(-21.0%), g1_fall(-20.2%), g1_fall_accessors(-9.8%)
```
=> seems not good. Clear regression.

paveltc:
On regression risk: clearing afn forces the slower correctly-rounded divide, so I'd expect AMD to take a hit on the same division-heavy kernels that regressed on CUDA (correctly-rounded fdiv is a multi-instruction sequence vs. a single v_rcp_f32). Worth measuring before merge — I can run a before/after A/B on gfx942 for dex_hand/g1_fall/g1_fall_accessors + a couple of rigid scenes. If the AMD hit is large, there's a more surgical option: clear afn only for divisions that feed floor/floordiv/% (the only contexts where the 2.5-ULP reciprocal is actually wrong), rather than all fdiv. If it's small, the current blunt-but-consistent fix is fine and matches how we settled #868.

paveltc:
> ⚠️ **Correction / retraction — please disregard the numbers below.**
>
> After posting, I ran an ISA/IR sanity check and found my before/after build harness never actually compiled this change into the wheels: both arms emitted byte-identical divide IR **and** ISA (both retained the `afn` flag) and both still reproduced #749. The build used a git *worktree* + cached build tree that compiled the original, unpatched `codegen_llvm.cpp` for both arms, so the A/B compared two identical binaries — the `.so` hash differences I relied on were just build nondeterminism, not real code changes. I confirmed this three ways: an unconditional `afn`-clear didn't change the emitted `fdiv`, and a deliberate "multiply the divide by 2" marker in that branch had no effect on the output.
>
> **The performance numbers in this comment are therefore invalid.** I'm redoing the measurement from a clean (non-worktree) checkout and will verify at the ISA level that the before arm lowers to the approximate `v_rcp_f32` and the after arm to the correctly-rounded divide sequence **before** reporting any numbers. Apologies for the noise.

---

<sub>~~Original (retracted) comment below.~~</sub>

Ran an AMD before/after A/B for this change on gfx942 (MI308X) to check whether clearing `afn` on fdiv costs us on AMD the way it did on CUDA. Short version: **it doesn't** — the division-heavy scenes that regressed on CUDA are flat on AMD.

### Setup
- Hardware: AMD Instinct MI308X (gfx942), ROCm, single GPU pinned.
- Built two AMDGPU wheels from **one** source tree so the only difference is this PR's fdiv hunk:
  - **before** = plain `CreateFDiv`
  - **after** = `afn` cleared on fdiv, gated to `Arch::amdgpu`
  - same version string, verified different sha256 (the hunk does change the emitted binary).
- Paired A/B: arms interleaved per trial with alternating order to cancel thermal/clock drift; each measurement is test_rigid.py's unmodified scene factory + `run_benchmark()` (45s warmup / 15s record) at the batch sizes below.

### Results (Δ = after/before)

| Scene | n_envs | mean Δ | verdict |
|---|---|---|---|
| dex_hand | 4096 | +0.16% | no sig diff |
| g1_fall | 4096 | +0.38% | no sig diff |
| franka_accessors | 30000 | +0.02% | no sig diff |
| anymal_zero | 4096 | +0.70% | no sig diff |
| franka | 30000 | +0.09% | no sig diff |
| go2 | 4096 | −2.04% (median) | small, borderline (not significant) |

`dex_hand`, `g1_fall`, `franka_accessors` are the division-heavy path; all three are flat. `go2` (hardened with n=12 + warmup discard) shows a small ~2% dip that agrees across full/trimmed mean and median but is not statistically significant (t≈−1.6). Everything else is within noise.

### Takeaway
The correctly-rounded divide is essentially free on gfx942 for these kernels — unlike NVPTX, where `afn` selects the fast approximate divide and clearing it forced the slower path (~10–20% on division-heavy CUDA kernels). So gating the `afn`-clear to `Arch::amdgpu` gets the #749 correctness fix with negligible AMD perf cost.

### Notes
- `g1_fall_accessors` isn't defined in my genesis-world checkout, so I used `franka_accessors` as the accessor/division-path stand-in (also flat).
- Re: #868 — related in theme (both make AMD `fast_math` FP match CUDA/CPU) but independent in code (this is `codegen_llvm.cpp` fdiv; #868 is `codegen_amdgpu.cpp` pow). No overlap.

paveltc:
Redid the AMD A/B for this change properly (my earlier comment was invalid — details at the bottom). Short version: on gfx942 this fixes #749 at the ISA level, and the AMD perf cost is small — geomean ≈ **−0.7%**, worst case **−1.5%** — i.e. roughly an order of magnitude smaller than the ~10–20% CUDA regression that motivated gating the `afn`-clear to `Arch::amdgpu`.

### Setup
- HW: AMD Instinct MI308X (gfx942), ROCm, single GPU pinned (`HIP_VISIBLE_DEVICES=0`).
- Two AMDGPU wheels built from one standalone checkout; the **only** difference is this PR's fdiv hunk. Each arm committed before building (this build compiles committed HEAD, not the working tree) and the two arms use **separate offline-cache dirs** so a cached kernel from one arm can never be reused by the other.
- Paired A/B: arms interleaved per trial with alternating order to cancel thermal/clock drift; each measurement is test_rigid.py's unmodified scene factory + `run_benchmark()` (45s warmup / 15s record). 6 trials/arm.

### Codegen is actually different (ISA-verified, offline cache disabled)
Minimal `q = a/b`, `r = a % b` kernel with the divisor loaded from a runtime field (so it isn't constant-folded), divisor = 180.0:

| arm | fdiv IR | divide ISA | `180.0/180.0` | `a % b` |
|---|---|---|---|---|
| before | `fdiv … afn` | approximate `v_rcp_f32` | `0.9999999` | `180.0` ❌ (#749) |
| after | `fdiv …` (no `afn`) | correctly-rounded `v_div_scale/v_div_fmas/v_div_fixup` | `1.0` | `0.0` ✅ |

### Results (Δ = after/before − 1, n=6 paired, t_crit=2.57)

| Scene | n_envs | before fps | after fps | mean Δ | verdict |
|---|---|---|---|---|---|
| dex_hand | 4096 | 2,395 | 2,381 | −0.60% | significant (t=−2.7) |
| g1_fall | 4096 | 155,529 | 153,205 | −1.49% | significant (t=−47) |
| franka_accessors | 30000 | 2,454,105 | 2,459,302 | +0.21% | significant (t=+3.0) |
| anymal_zero | 4096 | 916,045 | 904,665 | −1.22% | not sig (t=−1.9) |
| franka | 30000 | 2,174,470 | 2,171,677 | −0.13% | not sig |
| go2 | 4096 | 709,087 | 701,181 | −1.10% | not sig (t=−1.1) |

**Geomean ≈ −0.72%, worst case −1.49% (g1_fall).**

### Takeaway
The correctly-rounded divide has a small but real cost on gfx942 (tightest/most-significant on `g1_fall` at −1.5%), not the "flat" I incorrectly reported before. But it's far below the CUDA hit, so gating the `afn`-clear to `Arch::amdgpu` buys the #749 correctness fix on AMD for ≤~1.5% while leaving CUDA on the fast approximate path.

### Notes
- `g1_fall_accessors` isn't defined in my genesis-world checkout, so `franka_accessors` stands in for the accessor/division path (essentially flat / slightly positive).
- Re: #868 — same theme (make AMD `fast_math` FP match CUDA/CPU) but independent code (this is `codegen_llvm.cpp` fdiv; #868 is `codegen_amdgpu.cpp` pow). No overlap.

---
<sub>Why the earlier comment was wrong: the before/after wheels in that run were byte-identical because the build compiles committed git HEAD and the patch was only applied to the (uncommitted) working tree, and the quadrants offline-cache key ignores the quadrants version so probes reused each other's cached kernels. Both are fixed above (commit-before-build + per-arm cache isolation + ISA re-verification).</sub>

paveltc:
I'm in favor of merging this PR. It fixes a genuine AMD-only correctness bug (#749), and it's gated to Arch::amdgpu so CPU/CUDA lowering is unchanged — the CUDA regression that prompted the gating is fully avoided. My ISA-verified A/B on gfx942 (MI308X) measured the AMD cost at geomean ≈ −0.72%, worst case −1.5% (g1_fall), which is a reasonable price for correctness and ~10× smaller than the CUDA hit. Regression tests for the runtime-divisor path are in, and CI is green. I'd keep the current blunt-but-AMD-gated afn-clear rather than the surgical floor/%-only variant — it's consistent with how we settled #868 and adds no new surface. No objections from me.

hughperkins:
Seems cannot affect anything other than amd

hughperkins:
so we have
- [x] cannot affect non-amd (in latest incarnation)
- [x] check doc
- [x] check comments/code ratio
- [x] decide on whether to use this PR CI, or parallel PR cI
- [x] check CI passes

hughperkins:
I dont think we need to create parallel PR:
- nothing much to factorize
- line wrapping looks ~correct to me
- there are tests
- test coverage, skimming, appears plausible
- no comments are deleted

hughperkins:
no doc change required

hughperkins:
comments/code ratio isnt great, but I think I'll just leave it as-is.

hughperkins:
So I'll just wait for ci on this pr to go green (ignoring agent tests, prediminatnly things starting with 'check'), and then merge

hughperkins:
Thanks!

PR #774, opened 2026-07-10
[AMDGPU] Feat: per-kernel LLVM function attributes via @qd.kernel(fn_attrs=...)

paveltc:
## Summary

Backport of ROCm/quadrants#11 onto the upstream `main` branch.

Lets users override AMDGPU codegen attributes per kernel without editing
the JIT pipeline. Attributes must be pre-registered in
`quadrants/program/fn_attrs_registry.h`; unknown backend or attribute names
raise `QuadrantsSyntaxError` at decoration time.

**Currently registered attributes (`amdgpu` backend):**
- `amdgpu-max-num-workgroups`
- `amdgpu-agpr-alloc`
- `amdgpu-waves-per-eu`
- `amdgpu-flat-work-group-size`
- `amdgpu-sched-strategy`

**Example usage:**
\`\`\`python
@qd.kernel(fn_attrs={"amdgpu": {"amdgpu-max-num-workgroups": "128,1,1"}})
def k(...): ...

@qd.kernel(fn_attrs={"amdgpu": {"amdgpu-waves-per-eu": "1,2"}})
def k(...): ...
\`\`\`

## Changes

- `quadrants/program/fn_attrs_registry.h` (new): central registry of allowed backend/attr pairs
- `quadrants/program/kernel.h`: adds `fn_attrs` field to `Kernel`
- `quadrants/python/export_lang.cpp`: exposes `set_fn_attrs` + `get_fn_attrs_registry` pybindings
- `quadrants/codegen/llvm/codegen_llvm.cpp`: applies per-kernel fn_attrs via `addFnAttr` at codegen
- `quadrants/analysis/offline_cache_util.cpp`: includes fn_attrs in the offline cache key (deterministic serialisation)
- `quadrants/runtime/amdgpu/jit_amdgpu.cpp`: default AMDGPU attributes (`unsafe-fp-math`, `amdgpu-waves-per-eu`, `amdgpu-ieee`, `amdgpu-dx10-clamp`, flat-work-group-size inheritance) gated by `hasFnAttribute` so user-supplied values always win
- `quadrants/runtime/llvm/llvm_context.{h,cpp}`: `mark_function_as_amdgpu_kernel` gains optional `block_dim` parameter to set `amdgpu-flat-work-group-size` at codegen time
- Python layer (`kernel.py`, `kernel_impl.py`, `src_hasher.py`): `fn_attrs` plumbed through `@qd.kernel()` decorator, validated at decoration time, included in fastcache key
- `tests/python/test_fn_attrs.py` (new): validation, cache-key differentiation, and AMDGPU end-to-end IR plumbing tests

## Merge notes

Backported on top of commit `2e2b43b40` (same base as PR #1 and PR #2). Conflicts resolved by combining HEAD's `graph`/`checkpoints`/`use_graph`/`use_checkpoints` additions with this commit's `fn_attrs` additions — all parameters coexist.

## Test results

Full suite run in Docker image `quadrants:amd-upstream-pr3` with `QD_AMDGPU_FORCE_PERMLANE64_FALLBACK=1`:

```
21 failed, 4992 passed, 142 skipped, 28 xfailed  (1:46:01)
```

All 21 failures are pre-existing baseline flaky tests in `test_simt.py::test_block_reduce` and `test_algorithms.py::test_reduce_composition` — identical failure set as the unpatched `amd-upstream` baseline. Zero new regressions.

`tests/python/test_fn_attrs.py` (all 5 new tests) passed.

Co-authored-by: kevinjosephamd <kevin.joseph@amd.com>

### DIFF
--- a/docs/source/user_guide/optimization_passes.md
+++ b/docs/source/user_guide/optimization_passes.md
@@ -81,6 +81,38 @@ All of these are fields of `CompileConfig`, so you set them at `qd.init(...)` (o
 
 For everyday use, leave them at their defaults. The most common deliberate change is `cfg_optimization=False` when iterating on a kernel whose compile time is in your way. Note that, in general, changing these options is relatively fragile since the Quadrants tests run assuming the default values.
 
+## Per-kernel LLVM function attributes (`fn_attrs`)
+
+For fine-grained control over AMDGPU codegen on a per-kernel basis, `@qd.kernel` accepts an optional `fn_attrs` argument. This lets you pass LLVM function attributes directly to the backend JIT without editing the JIT pipeline.
+
+```python
+@qd.kernel(fn_attrs={"amdgpu": {"amdgpu-waves-per-eu": "1,4"}})
+def my_kernel():
+    ...
+
+@qd.kernel(fn_attrs={"amdgpu": {"amdgpu-max-num-workgroups": "128,1,1"}})
+def another_kernel():
+    ...
+```
+
+**How it works:** attributes are applied inside `codegen_llvm.cpp` before the JIT sees the module. The JIT's own defaults (e.g. `amdgpu-waves-per-eu`, `amdgpu-ieee`) are set only when the kernel does *not* already carry the attribute, so `fn_attrs` values always win.
+
+**Supported backends and attribute names** are registered in `quadrants/program/fn_attrs_registry.h`. Passing an unknown backend name or an unregistered attribute name raises `QuadrantsSyntaxError` at decoration time (not at compile time), so mistakes surface early.
+
+Currently registered attributes for the `amdgpu` backend:
+
+| Attribute | Effect |
+|-----------|--------|
+| `amdgpu-waves-per-eu` | Target waves-per-execution-unit range, e.g. `"1,2"` |
+| `amdgpu-flat-work-group-size` | Flat work-group size range, e.g. `"64,256"` |
+| `amdgpu-max-num-workgroups` | Maximum number of workgroups per dispatch, e.g. `"128,1,1"` |
+| `amdgpu-agpr-alloc` | AGPR allocation hint |
+| `amdgpu-sched-strategy` | Instruction scheduling strategy |
+
+**Cache interaction:** changing `fn_attrs` produces a different fastcache key, so kernels with different attribute sets are compiled and cached independently.
+
+**Other backends:** only `amdgpu` is currently registered. Adding a new backend means adding it to `fn_attrs_registry.h` and ensuring the consuming codegen path picks it up.
+
 ## Inspecting what the compiler did
 
 These environment variables dump the IR so you can see the effect of each pass. Files are written to `debug_dump_path` (default `/tmp/ir/`):
--- a/python/quadrants/lang/kernel_impl.py
+++ b/python/quadrants/lang/kernel_impl.py
@@ -181,6 +181,14 @@ def _kernel_impl(
     cuda_graph: bool = False,
     fn_attrs: dict[str, dict[str, str]] | None = None,
 ) -> QuadrantsCallable:
+    if cuda_graph and not graph:
+        import warnings
+        warnings.warn(
+            "@qd.kernel(cuda_graph=True) is deprecated; use graph=True instead.",
+            DeprecationWarning,
+            stacklevel=3,
+        )
+        graph = True
     if fn_attrs:
         _validate_fn_attrs(fn_attrs, _func.__name__)
     # Can decorators determine if a function is being defined inside a class?
@@ -194,7 +202,6 @@ def _kernel_impl(
     primal.use_graph = graph
     primal.use_checkpoints = checkpoints
     adjoint.use_checkpoints = checkpoints
-    primal.use_cuda_graph = cuda_graph
     primal.fn_attrs = fn_attrs
     adjoint.fn_attrs = fn_attrs
     # Having |primal| contains |grad| makes the tape work.
--- a/python/quadrants/lang/kernel_impl.py
+++ b/python/quadrants/lang/kernel_impl.py
@@ -5,6 +5,7 @@ import typing
 from functools import update_wrapper, wraps
 from typing import Any, Callable, TypeVar, cast, overload
 
+from quadrants._lib import core as _qd_core
 from quadrants.lang import impl
 from quadrants.lang.exception import (
     QuadrantsCompilationError,
@@ -153,13 +154,35 @@ def _inside_class(level_of_class_stackframe: int) -> bool:
     return False
 
 
+def _validate_fn_attrs(fn_attrs: dict[str, dict[str, str]], func_name: str) -> None:
+    registry = _qd_core.get_fn_attrs_registry()
+    for backend, attrs in fn_attrs.items():
+        if backend not in registry:
+            raise QuadrantsSyntaxError(
+                f"@qd.kernel(fn_attrs=...) on {func_name!r}: unknown backend "
+                f"{backend!r}. Registered backends: {sorted(registry)}."
+            )
+        allowed = registry[backend]
+        for key in attrs:
+            if key not in allowed:
+                raise QuadrantsSyntaxError(
+                    f"@qd.kernel(fn_attrs=...) on {func_name!r}: unknown "
+                    f"{backend} attribute {key!r}. Registered attributes: "
+                    f"{sorted(allowed)}."
+                )
+
+
 def _kernel_impl(
     _func: Callable,
     level_of_class_stackframe: int,
     verbose: bool = False,
     graph: bool = False,
     checkpoints: bool = False,
+    cuda_graph: bool = False,
+    fn_attrs: dict[str, dict[str, str]] | None = None,
 ) -> QuadrantsCallable:
+    if fn_attrs:
+        _validate_fn_attrs(fn_attrs, _func.__name__)
     # Can decorators determine if a function is being defined inside a class?
     # https://stackoverflow.com/a/8793684/12003165
     is_classkernel = _inside_class(level_of_class_stackframe + 1)
@@ -171,6 +194,9 @@ def _kernel_impl(
     primal.use_graph = graph
     primal.use_checkpoints = checkpoints
     adjoint.use_checkpoints = checkpoints
+    primal.use_cuda_graph = cuda_graph
+    primal.fn_attrs = fn_attrs
+    adjoint.fn_attrs = fn_attrs
     # Having |primal| contains |grad| makes the tape work.
     primal.grad = adjoint
 
@@ -213,7 +239,13 @@ def _kernel_impl(
 # TODO: This callable should be Callable[[F], F].
 # See comments below.
 def kernel(
-    _fn: None = None, *, pure: bool = False, graph: bool = False, checkpoints: bool = False
+    _fn: None = None,
+    *,
+    pure: bool = False,
+    graph: bool = False,
+    checkpoints: bool = False,
+    cuda_graph: bool = False,
+    fn_attrs: dict[str, dict[str, str]] | None = None,
 ) -> Callable[[Any], Any]: ...
 
 
@@ -224,7 +256,15 @@ def kernel(
 # However, by making it return Any, we can make the pure parameter
 # change now, without breaking pyright.
 @overload
-def kernel(_fn: Any, *, pure: bool = False, graph: bool = False, checkpoints: bool = False) -> Any: ...
+def kernel(
+    _fn: Any,
+    *,
+    pure: bool = False,
+    graph: bool = False,
+    checkpoints: bool = False,
+    cuda_graph: bool = False,
+    fn_attrs: dict[str, dict[str, str]] | None = None,
+) -> Any: ...
 
 
 def kernel(
@@ -234,6 +274,8 @@ def kernel(
     fastcache: bool = False,
     graph: bool = False,
     checkpoints: bool = False,
+    cuda_graph: bool = False,
+    fn_attrs: dict[str, dict[str, str]] | None = None,
 ):
     """
     Marks a function as a Quadrants kernel.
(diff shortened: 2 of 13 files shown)


quadrants/runtime/amdgpu/jit_amdgpu.cpp
```diff
@@ -31,6 +31,90 @@ std::string JITSessionAMDGPU::compile_module_to_hsaco(std::unique_ptr<llvm::Modu
...
   function_pass_manager_addrcast.doFinalization();
 
+  for (auto &F : *llvm_module) {
+    // Match CUDA parity: jit_cuda.cpp:332-335 unconditionally applies
+    // unsafe-fp-math to ALL functions via hardcoded kFTZDenorms=1.
+    // Enables FMA contraction, reciprocal for division, and operation
+    // reordering. Applied to all functions (not just kernels) because
+    // internal body functions contain the actual FP compute.
+    F.addFnAttr("unsafe-fp-math", "true");
+    F.addFnAttr("no-signed-zeros-fp-math", "true");
```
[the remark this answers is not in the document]
paveltc: Fixed in commit 9cb65ab3e. `unsafe-fp-math` and `no-signed-zeros-fp-math` are now gated inside `if (this->config_.fast_math)` in `jit_amdgpu.cpp`, matching the TargetOptions block below and preserving IEEE semantics when `qd.init(fast_math=False)` is set.

quadrants/runtime/llvm/llvm_context.cpp
```diff
@@ -1017,8 +1017,16 @@ void QuadrantsLLVMContext::mark_function_as_cuda_kernel(llvm::Function *func, in
...
-void QuadrantsLLVMContext::mark_function_as_amdgpu_kernel(llvm::Function *func) {
+void QuadrantsLLVMContext::mark_function_as_amdgpu_kernel(llvm::Function *func, int block_dim) {
   func->setCallingConv(llvm::CallingConv::AMDGPU_KERNEL);
+  if (block_dim > 0) {
+    // Wavefront size of 64 matches CDNA3. RDNA in wave32 mode would need a
+    // runtime query; revisit if Quadrants ever supports RDNA AMDGPU targets.
+    constexpr int kAmdgpuWavefrontSize = 64;
+    int clamped = std::max(block_dim, kAmdgpuWavefrontSize);
+    std::string size_str = std::to_string(clamped) + "," + std::to_string(clamped);
+    func->addFnAttr("amdgpu-flat-work-group-size", size_str);
```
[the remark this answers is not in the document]
paveltc: Fixed in commit 9cb65ab3e. The wavefront-size clamping is removed; `mark_function_as_amdgpu_kernel` now uses `block_dim` verbatim (`block_dim,block_dim`) so tasks with `block_dim=1` correctly produce `amdgpu-flat-work-group-size="1,1"` instead of `"64,64"`.

python/quadrants/lang/kernel_impl.py
```diff
@@ -234,6 +274,8 @@ def kernel(
     fastcache: bool = False,
     graph: bool = False,
     checkpoints: bool = False,
+    cuda_graph: bool = False,
```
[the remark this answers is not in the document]
paveltc: Fixed in commit 9cb65ab3e. `cuda_graph=True` now sets `use_graph=True` with a `DeprecationWarning` directing users to `graph=True`. The now-unused `use_cuda_graph` field is removed from `Kernel`.

python/quadrants/lang/kernel_impl.py
```diff
@@ -234,6 +274,8 @@ def kernel(
...
+    fn_attrs: dict[str, dict[str, str]] | None = None,
```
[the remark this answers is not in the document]
paveltc: Fixed in commit 9cb65ab3e. Added a 'Per-kernel LLVM function attributes' section to `docs/source/user_guide/optimization_passes.md` covering usage, the registered AMDGPU attribute table, and cache interaction.

python/quadrants/lang/kernel_impl.py
```diff
@@ -153,13 +154,43 @@ def _inside_class(level_of_class_stackframe: int) -> bool:
...
+                )
+
+
 def _kernel_impl(
     _func: Callable,
     level_of_class_stackframe: int,
     verbose: bool = False,
     graph: bool = False,
     checkpoints: bool = False,
+    cuda_graph: bool = False,
```
hughperkins: why are you adding 'cuda_graph'? We don't have 'cuda_graph'. We have 'graph'.

docs/source/user_guide/optimization_passes.md
```diff
@@ -81,6 +81,38 @@ All of these are fields of `CompileConfig`, so you set them at `qd.init(...)` (o
 
 For everyday use, leave them at their defaults. The most common deliberate change is `cfg_optimization=False` when iterating on a kernel whose compile time is in your way. Note that, in general, changing these options is relatively fragile since the Quadrants tests run assuming the default values.
 
+## Per-kernel LLVM function attributes (`fn_attrs`)
+
+For fine-grained control over AMDGPU codegen on a per-kernel basis, `@qd.kernel` accepts an optional `fn_attrs` argument. This lets you pass LLVM function attributes directly to the backend JIT without editing the JIT pipeline.
```
hughperkins, line 86: Platform-specific configuration conflicts with the ideal that Quadrants code will run unmodified on all platforms.

The standard approach is to set appropriate defaults - or use appropriate heuristics - within each platform-specific pipeline.

hughperkins (changes requested):
Platform-specific configuration conflicts with the ideal that Quadrants code will run unmodified on all platforms.

The standard approach is to set appropriate defaults - or use appropriate heuristics - within each platform-specific pipeline.

python/quadrants/lang/kernel_impl.py
paveltc: I removed this. It was something that was left over from the internal PR.

docs/source/user_guide/optimization_passes.md
paveltc: Thanks — I agree with the principle, and I think this PR is consistent with it. The sensible AMDGPU defaults already live in the backend pipeline (`jit_amdgpu.cpp`, each gated by `hasFnAttribute`); `fn_attrs` is just an *optional override* on top for when a user knows better. Portable code never needs it.

Portability is also preserved: attributes are only applied under `arch == Arch::amdgpu` (`codegen_llvm.cpp:3244`), so `@qd.kernel(fn_attrs={"amdgpu": {...}})` still runs unmodified on CPU/CUDA/Metal — the block is simply ignored there. It's namespaced by backend for exactly that reason, much like CUDA `__launch_bounds__` or Triton `num_warps`: an optional hint that only tunes the backend that understands it.

To land the right scope, which is your concern:

1. the concept of any user-facing backend override, or
2. specifically exposing it on `@qd.kernel(...)`?

If (1), I'll narrow this PR to just the backend-default infrastructure and drop the public `fn_attrs=` for now. If (2), I'll move it behind a clearly-marked advanced API. Just say which unblocks the merge.

hughperkins: My concern is anything platform-specific in the public API.

What I want to avoid is code that runs on one platform, and not on other; or on all platforms except one.

This includes optimizations, ideally. Code should ideally run optimally on each platform without the user needing platform-specific knowledge.

paveltc: Agreed as the default, and it holds for the common case. For peak perf I don't think it's attainable, though — no production GPU stack manages it. CUDA and HIP ship __launch_bounds__ precisely so the user can state an occupancy/register tradeoff the compiler can't infer; Triton has num_warps/num_stages/@autotune; Halide/TVM make the schedule explicit. The vendors ship these knobs because the "sufficiently smart compiler" doesn't exist for this.

On the "AMD-only" objection: I'd argue that's expected, not a defect. Tuning is vendor-owned — AMD owns the AMD backend and its lowering and can't be expected to implement the NVIDIA/Intel/Apple equivalents; each vendor adds their own. That's how the ecosystem is already structured (torch.backends.cuda/.cudnn/.mps, LLVM's namespaced amdgpu-*/nvvm.* attributes).

And crucially, the current design already answers the "runs on one platform, breaks on another" worry:

- It's namespaced by backend — fn_attrs={"amdgpu": {...}}, applied only under arch == Arch::amdgpu (codegen_llvm.cpp:3244). @qd.kernel(fn_attrs={"amdgpu": {...}}) runs unmodified on CPU/CUDA/Metal; the block is just ignored. Portable code is unaffected; only the backend that understands the hint tunes on it — same model as __launch_bounds__.

- It's registry-validated, not an open escape hatch — _validate_fn_attrs rejects unknown backends and unknown attributes against get_fn_attrs_registry(), so the surface is curated and enumerable rather than "any LLVM attribute goes."

So it's an optional, explicitly-scoped, validated per-backend hint — opt-in, ignored elsewhere, and owned by the vendor whose backend it tunes. That seems to satisfy the underlying concern (no silent cross-platform breakage, no unbounded surface) while accepting that vendor-specific tuning is a legitimate and necessary thing to expose.

hughperkins: Interesting. Currently, in genesis-world, we are not - I think - doing such tuning, for either CPU or CUDA. Perhaps we should. Or perhaps we are doing so implicitly.

I would much prefer to ground design and needs on real-world needs, ideally.

Do you see anywhere in genesis-world where we would be able to improve our performance, concretely, on our rigid benchmarks https://github.com/Genesis-Embodied-AI/genesis-world/blob/main/tests/benchmarks/test_rigid.py , for CUDA and AMD, by doing such tuning? And crucially, do you see anywhere where such tuning would use different numbers for each of CUDA and AMD?

This would give a concrete example to inspect and discuss, rather than discussing in the abstract.

(You could also provide a non genesis-world example, but, if it's about as much work for you to use genesis-world vs some other example, genesis-world will be a lot easier for me personally to understand and gain intuition and insight from).

paveltc: ### Benchmark: `amdgpu-waves-per-eu` via this PR's per-kernel `fn_attrs`

I benchmarked the knob this PR exposes on a Genesis workload — the `anymal_random`
benchmark in `tests/benchmarks/test_rigid.py`, 30,000 parallel envs, MI308X GPU.
I applied `amdgpu-waves-per-eu="2,4"` to the rigid-body solver kernels via
`@qd.kernel(fn_attrs={"amdgpu": {"amdgpu-waves-per-eu": "2,4"}})` and compared against the
same kernels with no attribute (compiler default occupancy):

| Workload | speedup vs default | consistency |
|---|---|---|
| `anymal_random` (30k envs, GPU) | **+48%** (geomean) | faster in 5/5 rounds |

This node runs the CG constraint solver (the parametrization passes `solver=None`). Each
round runs default and tuned back-to-back and takes the ratio within the round, averaged
with a geometric mean (30 s warmup + 30 s record, 5 rounds).

What was modified: the feature under test is this PR's `fn_attrs=` API in Quadrants. To
exercise it, the only change on the Genesis side is passing `fn_attrs=...` on the
rigid-solver kernel decorators — the `anymal_random` workload and the `runtime_fps` metric
are the stock upstream `test_rigid.py`. The 5-round interleaved averaging is an external
wrapper around repeated stock runs (to cancel GPU clock drift); the longer warmup/record
windows are an optional tweak and not required to see the effect.

`amdgpu-waves-per-eu` directly tells the register allocator the target occupancy
(resident wavefronts per EU), trading latency-hiding against registers-per-wave. This is
a more direct handle than `__launch_bounds__` (which HIP and CUDA already expose):
`__launch_bounds__` constrains occupancy only indirectly, as a side effect of a block-size
cap and an optional min-blocks hint, whereas `amdgpu-waves-per-eu` sets the occupancy
target itself. That LLVM attribute isn't otherwise reachable from the JIT kernel path —
this PR is what makes it settable per kernel. The optimal point is kernel- and
workload-specific, so per-kernel granularity is the right level rather than a single
global flag, and the results show it has first-order performance impact on a real, public
workload — so exposing it is clearly worthwhile.

### Reproduction

Requirements: Quadrants built from this PR, an AMD gfx942 GPU (MI3xx), and a Genesis build. `test_rigid.py` is run **unmodified** — the only source edit is adding the attribute to the rigid-solver kernel decorators (that's the feature under test).

**Step 1.** Add `fn_attrs={"amdgpu": {"amdgpu-waves-per-eu": "2,4"}}` to these `@qd.kernel(...)` decorators — `kernel_step_1` and `kernel_step_2` in `genesis/engine/solvers/rigid/rigid_solver.py`, and `func_solve_init` and `_kernel_solve_monolith` in `genesis/engine/solvers/rigid/constraint/solver.py`. For example, `@qd.kernel(fastcache=True)` becomes `@qd.kernel(fastcache=True, fn_attrs={"amdgpu": {"amdgpu-waves-per-eu": "2,4"}})`.

**Step 2.** Run the stock benchmark node once **with** the edit (tuned) and once **without** it (baseline):

```
python -m pytest -m benchmarks --backend gpu "tests/benchmarks/test_rigid.py::test_speed[anymal_random-None-None-30000-gpu]"
```

**Step 3.** Compare `runtime_fps`. A single run of each already shows the gain; averaging a few back-to-back rounds just tightens it against GPU clock noise.

hughperkins: Note: as discussed, the requirement is NOT that by using this you can show a performance gain on AMD.

The requirement is that:
- you can demosntrate that you have done due diligence to achieve similar speedups without needing to change the user-facing API
     - report experiments and results from this due diligence
- in the case that an API change cannot be avoided, show that you have done due diligence to come up with an API change that is cross-platform, not AMD specific

Note that I would recommend doing the above for each attribute you are proposiing to add, one by one, using a separate PR for each attribute you wish to add. Also, preference to have these attributes as standard parameters, rather than a dict of parameters.

paveltc:
Closing as this goes against Quadrants policy.

PR #775, opened 2026-07-10
[AMDGPU] Fix: apply amdgpu-ieee/amdgpu-dx10-clamp to all functions

paveltc:
## Summary

Backend-only AMDGPU codegen change. No public API, no new user-tunable knobs, solver-agnostic. Self-contained; does not depend on #774.

Sets `amdgpu-ieee` and `amdgpu-dx10-clamp` uniformly on **all** functions (not just `AMDGPU_KERNEL` entries). LLVM's inliner refuses to inline a callee into a caller when the two carry **different values** for these mode attributes; if only kernels carry them, internal device functions (e.g. `gpu_parallel_range_for` and the body functions it dispatches) fail to inline, and `InferAddressSpaces` can no longer follow the pointer chain from kernel params to field data to promote `flat_load`/`flat_store`/`flat_atomic` to `global_*`.

## Perf context (please read — the ~4% is not against `main`)

This change was ported from our AMD integration branch, where these attributes were previously set **kernel-only**. Against *that* kernel-only baseline, moving them to all functions removed the caller/callee mismatch and recovered **~4% throughput (301k → 314k)**.

Upstream `main` sets **neither** attribute anywhere, so there is no mismatch to fix on `main` and no perf delta to show. I verified this on gfx942 (MI308X, ROCm 7.2.4, LLVM 22.1.0): builds with and without this change produce **byte-identical AMDGCN and identical throughput** (saxpy ~1159 GB/s, atomic reduction ~126 GB/s, both directions). I also confirmed the mechanism at the IR level with the pinned LLVM 22.1.0 `opt` — the inliner blocks **only** on a present-vs-present *value* mismatch; uniform (either value), absent/absent, and absent/present all inline.

So on `main` this PR is best understood as a **consistency + correctness** change, not a standalone perf win:
- It guarantees these mode attributes can never be asymmetric, so the mismatch-induced regression cannot appear if kernel-only mode attributes are ever introduced.
- Per review, it now **preserves strict IEEE semantics**: `amdgpu-ieee` follows `fast_math` (stays at the IEEE-compliant default `true` unless `fast_math` is requested).

Full mechanism write-up and A/B data are in the follow-up comment below.

## Change

Single file (`quadrants/runtime/amdgpu/jit_amdgpu.cpp`): a small loop over every non-declaration function setting both attributes uniformly. `amdgpu-ieee` is gated on `fast_math` (IEEE-compliant `true` by default, `false` only under `fast_math`); `amdgpu-dx10-clamp=false` (the normal compute value). Attributes are written unconditionally (no `hasFnAttribute` guard) so a value preset on a linked ROCm device-lib function cannot re-introduce the mismatch. ~13 lines, no other files touched.

## Independence from #774

An earlier version of this fix was stacked on #774 and reused its per-kernel `fn_attrs` registry. That coupling has been removed — this PR applies against `main` directly and can merge with or without #774.

### DIFF
--- a/quadrants/runtime/amdgpu/jit_amdgpu.cpp
+++ b/quadrants/runtime/amdgpu/jit_amdgpu.cpp
@@ -31,25 +31,20 @@ std::string JITSessionAMDGPU::compile_module_to_hsaco(std::unique_ptr<llvm::Modu
     function_pass_manager_addrcast.run(*func);
   function_pass_manager_addrcast.doFinalization();
 
-  // Apply amdgpu-ieee and amdgpu-dx10-clamp to ALL functions, not just
-  // AMDGPU_KERNEL entries. LLVM's inliner refuses to inline a callee into a
-  // caller when they carry mismatching target-specific attributes, so setting
-  // these on kernels only prevents internal runtime device functions (e.g.
-  // gpu_parallel_range_for and the body functions it dispatches) from being
-  // inlined into the kernel entry. Without that inlining, InferAddressSpaces
-  // cannot see the full pointer chain from kernel params to field data and
-  // cannot promote flat_load/flat_store/flat_atomic to global_*, which causes
-  // flat-atomic coherency issues and a throughput regression on gfx942
-  // (MI300X). Applying them uniformly keeps every function's attribute set
-  // consistent so inlining and address-space inference run as intended. The
-  // hasFnAttribute guard keeps each write idempotent.
+  // Give every function the same amdgpu-ieee / amdgpu-dx10-clamp mode attributes. LLVM's inliner refuses to inline a
+  // callee whose target-mode attributes differ from the caller, so a kernel-only setting would leave the runtime device
+  // functions (gpu_parallel_range_for and the body functions it dispatches) un-inlined; without that inlining
+  // InferAddressSpaces cannot follow the pointer chain from kernel params to field data and cannot promote
+  // flat_load/flat_store/flat_atomic to global_*, costing a ~4% throughput regression on gfx942 (MI300X). Uniformity is
+  // what restores inlining, not the specific value, so honor strict math: amdgpu-ieee stays at the IEEE-compliant
+  // default (true) unless fast_math is requested. We overwrite unconditionally (rather than guarding on hasFnAttribute)
+  // so a value preset on a linked ROCm device-lib function cannot re-introduce the mismatch.
+  const char *ieee_mode = this->config_.fast_math ? "false" : "true";
   for (auto &F : *llvm_module) {
-    if (!F.hasFnAttribute("amdgpu-ieee")) {
-      F.addFnAttr("amdgpu-ieee", "false");
-    }
-    if (!F.hasFnAttribute("amdgpu-dx10-clamp")) {
-      F.addFnAttr("amdgpu-dx10-clamp", "false");
-    }
+    if (F.isDeclaration())
+      continue;
+    F.addFnAttr("amdgpu-ieee", ieee_mode);
+    F.addFnAttr("amdgpu-dx10-clamp", "false");
   }
 
   if (llvm::verifyModule(*llvm_module, &llvm::errs())) {
--- a/quadrants/runtime/amdgpu/jit_amdgpu.cpp
+++ b/quadrants/runtime/amdgpu/jit_amdgpu.cpp
@@ -31,6 +31,27 @@ std::string JITSessionAMDGPU::compile_module_to_hsaco(std::unique_ptr<llvm::Modu
     function_pass_manager_addrcast.run(*func);
   function_pass_manager_addrcast.doFinalization();
 
+  // Apply amdgpu-ieee and amdgpu-dx10-clamp to ALL functions, not just
+  // AMDGPU_KERNEL entries. LLVM's inliner refuses to inline a callee into a
+  // caller when they carry mismatching target-specific attributes, so setting
+  // these on kernels only prevents internal runtime device functions (e.g.
+  // gpu_parallel_range_for and the body functions it dispatches) from being
+  // inlined into the kernel entry. Without that inlining, InferAddressSpaces
+  // cannot see the full pointer chain from kernel params to field data and
+  // cannot promote flat_load/flat_store/flat_atomic to global_*, which causes
+  // flat-atomic coherency issues and a throughput regression on gfx942
+  // (MI300X). Applying them uniformly keeps every function's attribute set
+  // consistent so inlining and address-space inference run as intended. The
+  // hasFnAttribute guard keeps each write idempotent.
+  for (auto &F : *llvm_module) {
+    if (!F.hasFnAttribute("amdgpu-ieee")) {
+      F.addFnAttr("amdgpu-ieee", "false");
+    }
+    if (!F.hasFnAttribute("amdgpu-dx10-clamp")) {
+      F.addFnAttr("amdgpu-dx10-clamp", "false");
+    }
+  }
+
   if (llvm::verifyModule(*llvm_module, &llvm::errs())) {
     llvm_module->print(llvm::errs(), nullptr);
     QD_WARN("Module broken");


paveltc:
> [!NOTE]
> **Outdated (Aug 20, 2026):** This comment no longer applies to the current PR. Keeping it for historical reference.

**Depends on #774** — please review/merge that first.

This branch is stacked on #774: of its three commits, the first two (`0f4410279`, `9cb65ab3e`) belong to #774's per-kernel `fn_attrs` work, and only the last one (`23f2bb97b`) is unique to this PR. That's why the diff currently shows ~+499/-8 — most of it is #774's code, not this change.

The actual change here is a single-file, +16/-6 edit to `quadrants/runtime/amdgpu/jit_amdgpu.cpp`: it moves the `amdgpu-ieee` / `amdgpu-dx10-clamp` function attributes out of the `AMDGPU_KERNEL`-only block so they apply to *all* functions (so the inliner will inline device body functions into kernels, letting `InferAddressSpaces` promote flat_* to global_*). It modifies the fn-attribute loop that #774 introduces in `jit_amdgpu.cpp`, so it cannot land on `main` independently.

Plan: once #774 merges, I'll rebase this onto `main`; it will then collapse to just the small `jit_amdgpu.cpp` change for a clean review.

paveltc:
@codex review

paveltc:
Thanks @codex. Quick disposition of the three comments:

**P1 &mdash; "Split unrelated AMDGPU behavior out of this fix" (`jit_amdgpu.cpp`)** and
**P1 &mdash; "Document the new `cuda_graph` decorator argument" (`kernel_impl.py`)**

Both of these are from the **inherited #774 commits**, not this PR's actual change. This branch is stacked on #774: of its three commits, the first two (`0f4410279`, `9cb65ab3e`) are #774's per-kernel `fn_attrs` work (fast-math gating, scheduling/work-group heuristics, the `cuda_graph` alias, bindings, public API, cache keys), and only the last commit (`23f2bb97b`) is unique to this PR &mdash; a single-file **+16/&minus;6** edit to `jit_amdgpu.cpp` that moves `amdgpu-ieee`/`amdgpu-dx10-clamp` out of the `AMDGPU_KERNEL`-only block. That's why the diff currently reads ~+499/&minus;8.

So the fix is already localized to the two attributes as requested; the extra surface Codex is flagging lives in #774. Once #774 merges, I'll rebase this onto `main` and the diff collapses to just the `jit_amdgpu.cpp` change (and the `cuda_graph` docs point is tracked on #774, where that change actually lives).

**P2 &mdash; "Allow the guarded AMDGPU attributes through validation" (`fn_attrs_registry.h`)**

Good catch &mdash; this one is real and specific to this PR's claim. Fixed in `d62feedfb`: `amdgpu-ieee` and `amdgpu-dx10-clamp` are now registered, so `_validate_fn_attrs` accepts them and the `!F.hasFnAttribute(key)` guards (and the "user-supplied values win" contract) are actually reachable.

hughperkins:
Given the instructions to merge/ review #774 first, will wait for 774 to be merged/reviewed before reviewing this PR.

paveltc:
@hughperkins This PR actually no longer depends on 774.

hughperkins:
@codex review

quadrants/runtime/amdgpu/jit_amdgpu.cpp
```diff
@@ -31,6 +31,27 @@ std::string JITSessionAMDGPU::compile_module_to_hsaco(std::unique_ptr<llvm::Modu
     function_pass_manager_addrcast.run(*func);
   function_pass_manager_addrcast.doFinalization();
 
+  // Apply amdgpu-ieee and amdgpu-dx10-clamp to ALL functions, not just
```
hughperkins: could we wrap this at 120c, rather than 80c please

hughperkins: also, lets remove ' Applying them uniformly keeps every function's attribute set
  // consistent so inlining and address-space inference run as intended. The
  // hasFnAttribute guard keeps each write idempotent.', to try to keep the comment concise, please.

hughperkins:
Review from Opus:

Screenshot 2026-08-24 at 13 40 22

Screenshot 2026-08-24 at 13 43 05

Please could you address the concerns that Opus raises.

quadrants/runtime/amdgpu/jit_amdgpu.cpp
```diff
@@ -31,6 +31,27 @@ std::string JITSessionAMDGPU::compile_module_to_hsaco(std::unique_ptr<llvm::Modu
...
+  // flat-atomic coherency issues and a throughput regression on gfx942
+  // (MI300X). Applying them uniformly keeps every function's attribute set
+  // consistent so inlining and address-space inference run as intended. The
+  // hasFnAttribute guard keeps each write idempotent.
+  for (auto &F : *llvm_module) {
+    if (!F.hasFnAttribute("amdgpu-ieee")) {
+      F.addFnAttr("amdgpu-ieee", "false");
+    }
+    if (!F.hasFnAttribute("amdgpu-dx10-clamp")) {
+      F.addFnAttr("amdgpu-dx10-clamp", "false");
```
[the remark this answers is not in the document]
paveltc: Fixed in 4ec180d. `amdgpu-ieee` is now gated on `config_.fast_math`: it keeps the IEEE-compliant default (`true`) unless `fast_math` is requested, in which case it relaxes to `false`. `amdgpu-dx10-clamp` stays `false` (the normal compute value). The important part — verified on gfx942 (MI308X) — is that the flat_*→global_* promotion depends only on the values being *uniform* across caller and callee, not on the specific value, so strict-math users now keep full IEEE semantics with no loss of the promotion. Details in the summary comment below.

quadrants/runtime/amdgpu/jit_amdgpu.cpp
paveltc: Done in 4ec180d — comment is now wrapped at 120c.

paveltc: Done in 4ec180d. Dropped the 'Applying them uniformly...' / hasFnAttribute-idempotency sentence. (The guard itself is also gone now — see the summary comment.)

paveltc:
Thanks for the thorough review (and for relaying Opus's notes, @hughperkins). Pushed 4ec180d addressing all points. Summary + the empirical answer to the crux question:

### The crux: value vs. uniformity

Opus's core question — *"would setting both to `true` uniformly also restore inlining and the `global_*` promotion?"* — is exactly right, and the answer is **yes**. I confirmed it on gfx942 (MI308X, ROCm 7.2.4) by dumping the emitted AMDGCN for a SAXPY-with-atomic kernel under both settings. The `global_*` instruction counts are **identical**:

| ISA op | `amdgpu-ieee=false` (fast_math) | `amdgpu-ieee=true` (strict) |
|---|---|---|
| `global_load_dwordx2` | 75 | 75 |
| `global_store_dwordx2` | 109 | 109 |
| `global_atomic_add_x2` | 6 | 6 |
| `global_load_dwordx4` | 15 | 15 |
| `global_store_dwordx4` | 31 | 31 |

So the throughput win comes from **uniformity restoring inlining** (→ `InferAddressSpaces` promotion), **not** from the `false` value. Choosing the value is therefore a separable FP/NaN-semantics decision, exactly as Opus laid out.

### What changed

1. **Gate `amdgpu-ieee` on `fast_math`.** It now stays at the IEEE-compliant default (`true`) and only relaxes to `false` when the user opts into `fast_math`. Verified in the optimized IR: `"amdgpu-ieee"="true"` uniformly across all functions under `fast_math=False`, `"false"` uniformly under `fast_math=True`. Strict-math users keep full IEEE semantics **and** the promotion. (Addresses Codex P1 + Opus "semantics regardless of fast_math".) `amdgpu-dx10-clamp` stays `false` — the normal compute value.

2. **Dropped the `hasFnAttribute` guard; overwrite unconditionally.** As Opus noted, the "user-supplied values win" contract lives in #774, not here, so it didn't justify the guard in this standalone PR — and worse, if a linked ROCm device-lib function arrived with a preset `amdgpu-ieee`/`dx10-clamp`, the guard would preserve it and re-introduce the very mismatch we're eliminating. Since `addFnAttr(StringRef, StringRef)` overwrites same-key string attrs, no guard is needed for idempotency, and overwriting unconditionally more robustly guarantees uniformity.

3. **Skip declarations** (`if (F.isDeclaration()) continue;`) so target attrs aren't placed on intrinsic/external decls.

4. **Nits:** comment rewrapped at 120c; removed the redundant trailing "Applying them uniformly…/hasFnAttribute…" sentence.

### On the ~4% number

Agree the headline number deserves a full repro (commit hash, batch size, before/after ISA). The table above is the before/after `global_*` vs `flat_*` evidence on a microkernel; I'll follow up with the end-to-end benchmark repro. Note a handful of `flat_*` ops legitimately remain (scratch/shared-addressed traffic that can't be promoted) — identical in both modes.

paveltc:
### Benchmark / mechanism follow-up (gfx942 MI308X, ROCm 7.2.4, pinned LLVM 22.1.0)

Following up on the ~4% claim. Rather than just quote a number, I isolated the mechanism at the IR level and ran an A/B on hardware. Summary: **the inliner mechanism this PR relies on is real and confirmed on the current toolchain. On this specific LLVM/ROCm combination the microbenchmarks show no delta (clean `main` already inlines and promotes), so the measurable win depends on the toolchain/config that produces a mode-attribute mismatch; this PR is a self-contained correctness/consistency + IEEE-preservation change that makes that mismatch impossible regardless.**

#### 1. Mechanism, proven with the pinned LLVM 22.1.0 `opt`

Minimal `amdgcn-amd-amdhsa` kernel calling a device function, varying only the mode attributes, run through `inline`:

| callee | caller (kernel) | result |
|---|---|---|
| `ieee=true` | `ieee=true` | inlined |
| `ieee=false` | `ieee=false` | inlined |
| *(none)* | *(none)* | inlined |
| *(none)* | `ieee=false` | inlined |
| `ieee=false` | *(none)* | inlined |
| **`ieee=true`** | **`ieee=false`** | **NOT inlined** |
| **`dx10-clamp=true`** | **`dx10-clamp=false`** | **NOT inlined** |

The inliner blocks **only when both sides carry the attribute with different values**. Absent-vs-present is compatible; uniform (either value) inlines. This confirms the PR's core rationale — and confirms Opus's crux point: **uniformity is what restores inlining, not the specific value** (both `true/true` and `false/false` inline).

#### 2. Hardware A/B (baseline = sweep stripped, = `main`; vs this PR)

| kernel (64 MB, f32) | baseline | this PR | delta |
|---|---|---|---|
| saxpy | 1159.0 GB/s | 1157.6 GB/s | ~0% |
| atomic_add reduction | 126.0 GB/s | 126.2 GB/s | ~0% |

Emitted AMDGCN is **byte-for-byte identical** between the two builds. The reason: on this LLVM 22.1.0 / ROCm 7.2.4 combination, clean `main` sets these attributes on *nothing* → the "neither set" row above → the kernels already inline and `InferAddressSpaces` already promotes `flat_*`→`global_*` without any help. So there is simply no mismatch present for the sweep to fix on this toolchain, which is why the microbenchmark delta is ~0% here.

#### 3. Where the ~4% came from

The ~4% (**301k → 314k**) was measured on our AMD integration branch, where `amdgpu-ieee`/`dx10-clamp` were previously set **kernel-only**. That kernel-only placement created the caller/callee value mismatch (the BLOCKED row), and moving the attributes to all functions removed it. Upstream `main` sets neither attribute anywhere, so that mismatch does not exist here — which is why the A/B above is ~0% on `main`. The value of this PR upstream is guaranteeing the attributes can never be asymmetric (plus the IEEE-preservation fix), not a standalone perf win against `main`.

#### 4. Device-lib guard check (Opus)

Checked the ROCm 7.2.4 bitcode (`ocml`, `ockl`, `oclc_*`): **none carry explicit `amdgpu-ieee`/`dx10-clamp`**, so on this ROCm they're the harmless "absent" case. But Opus's warning holds in principle — *if* a linked function carried a differing explicit value, the old `hasFnAttribute` guard would preserve it and re-create the BLOCKED row. Dropping the guard (4ec180d) is the robust choice and is what guarantees uniformity.

#### Repro
- Inliner truth-table: hand-written IR + `opt -passes='inline,instcombine'` on the pinned LLVM 22.1.0.
- A/B: same kernels, kernel-profiler timing, 7 trials x 1000 iters, built with and without the sweep on the same MI308X.

hughperkins:
For the benchmark, I'm not sure I follow. I was imagining that the benchmark would use some scenario that shows a benefit on this PR, compared to main?

paveltc:
@hughperkins Good instinct, and after tracing this all the way back the honest answer is: **there isn't a scenario that beats `main`, and here's why.**

The `flat_*`→`global_*` promotion is unlocked by *inlining*, and the inliner only refuses to inline when caller and callee carry the mode attribute with **different values**. Upstream `main` sets `amdgpu-ieee`/`dx10-clamp` on **nothing**, so there's no mismatch — the runtime device functions already inline into kernels and `InferAddressSpaces` already promotes. I confirmed this two ways on gfx942 (MI308X):

- **A/B on `main`**: built with vs. without this change → byte-identical AMDGCN and identical throughput (~0% delta).
- **IR-level inliner probe** (pinned LLVM 22.1.0 `opt`): uniform values inline; absent/absent inlines; absent/present inlines; only present-vs-*different*-present blocks.

The ~4% (301k → 314k) in the original description was measured on our AMD integration branch, where these attributes were set **kernel-only** — that kernel-only placement is what created the mismatch, and this change fixed it *there*. Upstream never had the kernel-only setting, so the regression scenario doesn't exist to reproduce against `main`.

So I'd reframe this PR (and I've updated the description accordingly): on `main` it is **not** a perf win but a **consistency + correctness** change — it makes the mode attributes impossible to set asymmetrically (so the regression can't reappear if kernel-only mode attributes are ever introduced), and it now preserves strict IEEE semantics by gating `amdgpu-ieee` on `fast_math`.

Totally fair if you'd prefer to hold this until there's an upstream consumer that actually sets these attributes (at which point the benefit becomes demonstrable) — your call on whether the forward-looking consistency + the IEEE-gating fix are worth landing now on their own.

paveltc:
Note for reviewers / future upstreaming (scoping caveat).

In our AMD integration branch, the `jit_amdgpu.cpp` attribute loop that this PR draws from also sets two **FP-relaxation** attributes on *all* functions alongside the mode attributes:

```
F.addFnAttr("unsafe-fp-math", "true");
F.addFnAttr("no-signed-zeros-fp-math", "true");
```

**This PR deliberately does _not_ bring those upstream** — it ports only `amdgpu-ieee`/`dx10-clamp` (and now gates `amdgpu-ieee` on `fast_math` to preserve strict IEEE). I've confirmed neither of the other in-flight AMDGPU PRs (#843, #866) sets any FP-semantics attributes either, so nothing in the current batch changes FP results.

Flagging it because `unsafe-fp-math` / `no-signed-zeros-fp-math` change numerical results **unconditionally** (even at `fast_math=False`). If anyone later ports more of that loop wholesale, those two lines are the thing to catch in review and gate on `fast_math` — they should not ride along silently.

paveltc:
Closed as the only use of this PR came from PR 774, which has been closed.

PR #759, opened 2026-06-24
[Misc] Migrate to nanobind

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

PR #779, opened 2026-07-13
[Nanobind] Silence interpreter-shutdown leak warnings

hughperkins:
nanobind's shutdown leak checker reports the primitive DataType_* module attributes (registered via the PER_TYPE X-macro in export_lang.cpp), along with the DataTypeCxx/Type class objects and their bound methods they keep alive, as leaked instances/types/functions. These are long-lived module-global singletons rather than a growing leak (RSS is flat across tens of thousands of churn iterations), so the messages are pure noise for downstream users such as Genesis at process exit. Disable them via nb::set_leak_warnings(false) in NB_MODULE.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/python/export.cpp
+++ b/quadrants/python/export.cpp
@@ -12,6 +12,12 @@ namespace quadrants {
 NB_MODULE(quadrants_python, m) {
   m.doc() = "quadrants_python";
 
+  // Disable nanobind's interpreter-shutdown leak checker. It reports the primitive DataType_* module
+  // attributes (registered via the PER_TYPE X-macro in export_lang.cpp) plus the class/method objects
+  // they keep alive as "leaks". These are long-lived module-global singletons, not a growing leak, so
+  // the warnings are pure noise for downstream users (e.g. Genesis) at process exit.
+  nb::set_leak_warnings(false);
+
   register_py_exception_translator();
 
   for (auto &kv : InterfaceHolder::get_instance()->methods) {
--- a/cmake/QuadrantsCore.cmake
+++ b/cmake/QuadrantsCore.cmake
@@ -336,7 +336,10 @@ if(QD_WITH_PYTHON)
     # NOMINSIZE: keep the project optimization level (-O3) for the binding layer instead of nanobind's
     # default -Os, matching the previous pybind11 NO_EXTRAS build and avoiding per-call overhead in the
     # hot launch path. NB_STATIC links the nanobind core statically (single extension module).
-    nanobind_add_module(${CORE_WITH_PYBIND_LIBRARY_NAME} NB_STATIC NOMINSIZE ${QUADRANTS_PYBIND_SOURCE})
+    # NB_DOMAIN: isolate our nb_internals (type registry + the leak-warning flag) from any other
+    # nanobind extension sharing the process. Without it we build in the default domain, so our
+    # nb::set_leak_warnings(false) would also silence unrelated default-domain extensions' diagnostics.
+    nanobind_add_module(${CORE_WITH_PYBIND_LIBRARY_NAME} NB_DOMAIN quadrants NB_STATIC NOMINSIZE ${QUADRANTS_PYBIND_SOURCE})
 
     # Remove symbols from static libs: https://stackoverflow.com/a/14863432/12003165
     if (LINUX)


hughperkins:
tested using prerelease 1.1.1b1.

with 1.1.0:

Screenshot 2026-07-13 at 14 25 26

with 1.1.1b1:

Screenshot 2026-07-13 at 14 25 34

hughperkins:
@codex review

hughperkins:
With 1.1.1b2:

Screenshot 2026-07-13 at 17 27 38

quadrants/python/export.cpp
duburcqa, line 1: I cannot confidently approve this change. It sounds dubious. Can you confirm you know what you are doing here?

hughperkins: - documented at https://nanobind.readthedocs.io/en/latest/refleaks.html
- specifically silences the leak warnings
- do you mean, do. I confirm that hte warnigns are benign?

duburcqa: > do you mean, do. I confirm that hte warnigns are benign?

yes. and that it is bounded to your package, without altering the behaviour of nanobind for other packages that also use nanobind under the hood.

hughperkins: We've changed the NB_DOMAIN to quadrants, so the change only applies to quadrants. This is handled by the cmake changes.

hughperkins: Opus write-up of whether the warnings matter (tldr; they dont)

# nanobind leak warnings after the quadrants migration: do they matter?

## Context

After migrating the `quadrants` Python bindings from pybind11 to nanobind (PR #759, now on `main`), running
Genesis prints a wall of messages at interpreter shutdown, e.g.:

```
nanobind: leaked 402 instances!
 - leaked instance 0x... of type "quadrants._lib.core.quadrants_python.DataTypeCxx"
 - ...
nanobind: leaked 19 types!
nanobind: leaked 257 functions!
```

pybind11 never printed anything like this. The question this doc answers: **are these warnings a real memory
leak we need to fix, or benign noise?**

TL;DR: **benign.** They are long-lived, bounded module-global singletons (plus the class/method objects they
transitively keep alive), reported once at process exit. They do not grow with simulation steps. The fix shipped
on branch `hp/nanobind-suppress-leak-warnings` (PR #779) is to silence the checker with
`nb::set_leak_warnings(false)` and isolate quadrants in its own `NB_DOMAIN` so the suppression only affects us.

All experiments below were run CPU-only in a container on a cluster node, against a locally built nanobind
`quadrants` wheel.

## Background: what nanobind's leak checker does

nanobind installs a `Py_AtExit` handler that, at interpreter shutdown, walks its internal registry and reports
any Python objects (instances, bound types, bound functions) that are still alive. It is meant to catch genuine
reference leaks in bindings. Two important properties:

- It fires **only at shutdown**, and only reports objects still reachable from nanobind's internals at that
  point. It says nothing about steady-state memory during the run.
- nanobind instances that lack cyclic-GC support are **not** tracked by Python's `gc`, so `gc.get_objects()` /
  `gc.get_referrers()` cannot see or locate them (this misled an early holder-scan attempt).

The flag it checks (`print_leak_warnings`) lives in nanobind's per-domain `nb_internals` struct:

```1194:1203:.venv/.../nanobind/src/common.cpp
bool leak_warnings() noexcept {
    return internals->print_leak_warnings;
}
...
void set_leak_warnings(bool value) noexcept {
    internals->print_leak_warnings = value;
```

## Experiments

### 1. Clean quadrants does not leak (bindings are correct)

Using a standalone probe (`tmp/leak_probe.py`) that churns and accumulates the objects Genesis exercises:

| Scenario                                             | Result                                     |
|------------------------------------------------------|--------------------------------------------|
| Churn `DataTypeCxxWrapper` instances, N up to 1000   | 0 leaks reported                           |
| Accumulate then drop, `gc.collect()`                 | 0 leaks reported                           |
| `@qd.kernel` closures (cycle-prone), N up to 100     | 0 leaks reported                           |
| 40,000 churn iterations                              | **RSS flat** (no growth)                   |

Conclusion: the migrated bindings themselves have no reference leak, and nothing grows over time. Whatever the
shutdown warning is reporting, it is **not** an accumulating leak.

### 2. Positive control: the checker works, and the count is just "objects alive at exit"

To confirm the detector is functional, we deliberately keep objects alive past shutdown by parking them in a
live daemon thread (`leak_probe.py thread N`):

```python
holds = [make(i) for i in range(N)]
def worker(objs): time.sleep(3600)
t = threading.Thread(target=worker, args=(holds,), daemon=True)
t.start(); del holds
```

This reliably prints `nanobind: leaked N instances!`, and the number scales exactly with `N`. So the warning
simply counts "nanobind objects still reachable when the interpreter tears down" — a daemon thread's stack frame
is enough to trigger it. This is a known, benign category.

### 3. Genesis, minimal scene: reproduces the warning at small scale

Running a minimal Genesis scene (`tmp/gs_leak.py baseline`: build + step a trivial scene) reproduced the report
at scaled-down numbers:

```
nanobind: leaked 12 instances!   (all of type DataTypeCxx)
nanobind: leaked 2 types!        (DataTypeCxx, Type)
nanobind: leaked 13 functions!   (to_string, __hash__, __getstate__, __call__, shape,
                                  ptr_removed, __str__, get_ptr, __setstate__, ...)
```

Observations:

- Only `MainThread` was alive at exit — so unlike the positive control, background/daemon threads are **not**
  the cause here.
- `gc.collect()` did not clear them; neither did `qd.reset()` + `gc.collect()`.
- The 12→2→13 shape is a single cascade: the 12 live `DataTypeCxx` instances keep their 2 class objects
  (`DataTypeCxx`, `Type`) alive, which in turn keep their bound methods alive. It is really **one** root: 12
  instances.

### 4. It is NOT the dtypes Genesis stores in `__init__.py`

Genesis's `_init()` stashes a pile of quadrants dtypes as module globals in `genesis/__init__.py`:

```python
qd_float = qd.f32; qd_int = qd.i32; qd_bool = qd.u1
qd_vec2 = qd.types.vector(2, qd_float)   # ... vec3/4/6/7/11, ivec2/3/4
qd_mat3 = qd.types.matrix(3, 3, qd_float); qd_mat4 = qd.types.matrix(4, 4, qd_float)
```

Decisive test (`gs_leak.py nullglobals`): enumerate **all 45** `qd_*` globals Genesis defines, print their
types, set every one to `None`, then `qd.reset()` + `gc.collect()`, and re-check at exit.

- The globals are: 3 bare `DataTypeCxx` (`qd_float/qd_int/qd_bool`), 11 `VectorType`/`MatrixType` wrappers
  (`qd_vecN`, `qd_matN`, `qd_ivecN` — note these are Python wrapper objects, *not* the bare `DataTypeCxx` being
  leaked), and 31 `QuadrantsCallable` helpers.
- **Result: still `leaked 12 instances`, unchanged.**

So Genesis's `__init__` globals are not what pins the leaked objects. Genesis stores those dtypes by
reference/derivation (`qd_float = qd.f32`), so dropping Genesis's copy doesn't free the canonical object — some
other holder owns it.

### 5. Root cause: primitive-type attributes on the quadrants extension module

The leaked instances are bare `DataTypeCxx` (not the `DataTypeCxxWrapper` subclass from `quadrants/lang/util.py`,
nor the `VectorType`/`MatrixType` wrappers). The only place bare, module-level `DataTypeCxx` objects are created
is the `PER_TYPE` X-macro in the module init:

```815:817:quadrants/quadrants/python/export_lang.cpp
#define PER_TYPE(x) m.attr(("DataType_" + data_type_name(PrimitiveType::x)).c_str()) = PrimitiveType::x;
#include "quadrants/inc/data_type.inc.h"
#undef PER_TYPE
```

`data_type.inc.h` lists 14 primitives (`f16 f32 f64 i8 i16 i32 i64 u1 u8 u16 u32 u64 gen unknown`), so at import
this creates ~14 bare `DataTypeCxx` instances and stores them as attributes (`quadrants_python.DataType_f32`,
`DataType_i32`, `DataType_u1`, …) directly on the C++ extension module object. These live exactly as long as the
extension module object lives. ~12 of these being alive at nanobind's check is the entire report; the 2 types and
13 functions cascade from them.

### 6. Why they only show up as "leaks" after real work (not on bare import)

Empirically:

| Workload                                 | Leak warnings at exit |
|------------------------------------------|-----------------------|
| `import quadrants`                       | none                  |
| `import genesis; gs.init()` (no scene)   | none                  |
| Genesis `scene.build()` + step           | 12 (minimal) …        |
| Full Genesis run                         | 402 (user's report)   |

At a clean shutdown, Python clears the extension module's `__dict__` before nanobind's atexit check runs, so the
`DataType_*` attributes are freed first → 0 leaks. After a real workload, quadrants' C++ type-system / program
singletons (created during kernel compilation and never torn down) keep the module — or nanobind's internals —
alive past that point, so the module-level `DataType_*` attributes are still present when the check fires and get
reported. The count scales with how much gets compiled (12 for a trivial scene; 402 for a full sim, which also
leaves `Expr`/kernel/`Ndarray` instances pinned by the same never-finalized statics).

## Conclusion

The warnings are **benign**:

- They are bounded module-global singletons + downstream class/method objects, not per-step growth (RSS is flat
  across 40k iterations).
- They are reported only at interpreter exit, when the process is going away anyway.
- pybind11 simply never had a shutdown leak checker, so the same long-lived state was silently present before.

They are, however, noisy and alarming for downstream users (Genesis), so silencing them is worthwhile.

## Fix (branch `hp/nanobind-suppress-leak-warnings`, PR #779)

1. **Silence the checker** (`7c07a61b8`) — `nb::set_leak_warnings(false)` in `NB_MODULE`
   (`quadrants/python/export.cpp`).

2. **Isolate the domain** (`387415398`) — add `NB_DOMAIN quadrants` to `nanobind_add_module` in
   `cmake/QuadrantsCore.cmake`. This was raised in review: `set_leak_warnings` writes to the per-domain
   `nb_internals` flag, which is shared by every nanobind extension with the same ABI tag in the *default*
   domain (the internals are coordinated via an interpreter-level capsule keyed
   `__nb_internals_<abi_tag>_<NB_DOMAIN>__`; `NB_STATIC` does **not** isolate this). Without a private domain, our
   suppression would also silence the shutdown diagnostics of any unrelated default-domain nanobind extension in
   the same process. A private domain scopes the flag (and the type registry) to quadrants, and is the
   nanobind-recommended practice for standalone extensions anyway.

## Verification (cluster, CPU-only)

Incrementally rebuilt the branch (both commits) → compiles and links cleanly with `NB_DOMAIN` +
`set_leak_warnings`; `import quadrants; qd.init(arch=qd.cpu)` OK.

| Test                                                        | Before fix                     | After fix          |
|-------------------------------------------------------------|--------------------------------|--------------------|
| Positive control (`leak_probe.py thread 50`)               | `leaked 50 instances!`         | **silent**         |
| Genesis minimal scene — `baseline` mode                    | `12 instances / 2 / 13`        | **no warnings**    |
| Genesis minimal scene — `gc` mode                          | warnings                       | **no warnings**    |
| Genesis minimal scene — `teardown` mode                    | warnings                       | **no warnings**    |

All modes ran to completion (`EXITING` printed) with the leak warnings gone.

## Appendix: probe scripts

- `tmp/leak_probe.py` — standalone quadrants leak/RSS probe. Modes: `count`, `rss`, `kernel`, `poscontrol`,
  `thread` (daemon-thread positive control).
- `tmp/gs_leak.py` — minimal Genesis scene probe. Modes: `baseline`, `gc`, `teardown`, `findholder`,
  `nullglobals` (null all `qd_*` Genesis module globals and re-check).
- `tmp/gs_setup.sh` — creates `/venv`, installs CPU torch + Genesis, then force-reinstalls the locally built
  nanobind `quadrants` wheel to override Genesis's pinned version.

duburcqa:
Ok to merge

PR #783, opened 2026-07-14
[CI] Render PR change report in job summary instead of a standalone check

hughperkins:
A check run created via checks.create({head_sha}) cannot be bound to this workflow's check suite, so GitHub grouped the "PR change report" check under an arbitrary workflow (e.g. "Check non-ASCII characters / PR change report") and produced a duplicate entry. Write the full report to the job step summary instead, and link the PR comment to the run summary page, so it is always attributed to the PR change report workflow. Drop the now-unused checks:write permission.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/.github/workflows/pr_change_report.yml
+++ b/.github/workflows/pr_change_report.yml
@@ -12,7 +12,6 @@ concurrency:
 permissions:
   contents: read
   pull-requests: write
-  checks: write
 jobs:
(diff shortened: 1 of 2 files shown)


hughperkins:
@codex review

PR #778, opened 2026-07-13
[CI] Add CI non ascii check

hughperkins:
Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

PR #784, opened 2026-07-14
[CI] Fix non-ASCII scanner treating "+++" added lines as file headers

hughperkins:
find_violations recognized "+++ "/"--- " as file headers before classifying the diff marker, so an added source line whose content starts with "++ " (git emits it as "+++ ...") was skipped -- letting its non-ASCII characters through and corrupting the tracked path and line numbers for following violations. Track hunk boundaries via the line counts in the @@ header and only treat "+++ "/"--- " as headers outside a hunk body. Adds regression tests. Ref PR #778 comment r3571557923.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/python/tools/check_non_ascii.py
+++ b/python/tools/check_non_ascii.py
@@ -24,9 +24,10 @@ import sys
 import unicodedata
 from dataclasses import dataclass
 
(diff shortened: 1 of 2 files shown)


hughperkins:
@codex review

PR #799, opened 2026-07-20
[Bug] Fix the reverse-mode gradient of a loop-carried vector rebuilt across a dynamic loop.

duburcqa:
Issue: #797

Stacked on #796 (the scalarize `size_expr` propagation): promoting a loop-carried tensor to an adstack relies on that fix to size the scalar component stacks. Review/merge #796 first; this PR's base will retarget to `main` once #796 lands.

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

PR #800, opened 2026-07-20
[Test] Run the ndarray autodiff tests on Metal and restrict f64 to where it is required.

duburcqa:
Issue: #

Stacked on #799 (the loop-carried-vector adstack fix); review/merge that first. Base retargets to `main` once #799 lands.

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/tests/python/test_ad_dataclass.py
+++ b/tests/python/test_ad_dataclass.py
@@ -23,7 +23,7 @@ import quadrants as qd
 
 from tests import test_utils
 
-archs_support_ndarray_ad = [qd.cpu, qd.cuda, qd.amdgpu]
+archs_support_ndarray_ad = [qd.cpu, qd.cuda, qd.amdgpu, qd.metal]
 
 
 # ----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


tests/python/test_ad_dataclass.py
```diff
@@ -23,33 +23,33 @@
 
 from tests import test_utils
 
-archs_support_ndarray_ad = [qd.cpu, qd.cuda, qd.amdgpu]
+archs_support_ndarray_ad = [qd.cpu, qd.cuda, qd.amdgpu, qd.metal]
```
hughperkins, line 26: vulkan is still excluded?

duburcqa: Yes. Vulkan is half-broken at the moment. Exploring...

hughperkins: Ok. Not pushing to fix it. (was just asking on the off-chance we could simply drop the list of platforms altogether, and run on all)

duburcqa: Here it is: https://github.com/Genesis-Embodied-AI/quadrants/pull/801

hughperkins:
- test only change ✅ 
- ad only change ✅ 
- no user doc-facing changes required ✅ 

what is changing:
- broadening tests to include metal too (still missing vulkan, but it's strictly wider coverage, as far as platforms)
- swapping many tests from f64 to f32 (different coverage (f64 vs f32), but strictly same number of tests running)

ok to merge

Issue #810, opened 2026-07-24
stale in-kernel read (a[0] == 0.0) in serialize=True loop


hughperkins:
```
"""Quadrants CPU miscompile repro (quadrants commit d42a9ab8, llvm 22.1.0, arm64).

Inside a kernel whose outer loop is serialized via qd.loop_config(serialize=True), a load through a
LITERAL index (a[0]) is hoisted above an aliasing store through the LOOP-VAR index (a[i_c]), so the
load returns the pre-kernel value even when i_c == 0. Memory itself ends up correct after the kernel;
only the in-kernel read is stale.

All four ingredients are required - removing any one of them makes the kernel correct:
  1. qd.loop_config(serialize=True) on the outer loop
  2. an inner loop wrapping the body (a single-iteration dynamic `for _ in range(1)` suffices)
  3. a store through the outer loop variable index (a[i_c] = ...)
  4. a read of the same tensor through a literal index, inside a branch comparing the loop var to
     that literal (if i_c == 0: ... a[0])

Expected output: a[0] = 1.0. Buggy output: a[0] = 0.0 (stale), while a[i_c] = 1.0 and a after = [1, 1].
"""

import quadrants as qd

qd.init(arch=qd.cpu, default_fp=qd.f64)

a = qd.tensor(dtype=qd.f64, shape=(2,), backend=qd.Backend.NDARRAY)
out = qd.tensor(dtype=qd.f64, shape=(2,), backend=qd.Backend.NDARRAY)


@qd.kernel
def run(a: qd.Tensor, out: qd.Tensor):
    qd.loop_config(serialize=True)
    for i_c in range(2):
        for i_iter in range(1):
            a[i_c] = 1.0
            if i_c == 0:
                out[0] = a[0]
                out[1] = a[i_c]


run(a, out)
reads = out.to_numpy()
print("[REPRO] a[0] =", reads[0], "| a[i_c] =", reads[1], "| a after:", a.to_numpy())
assert reads[0] == 1.0, "BUG: literal-index load hoisted above aliasing loop-var store"
```

PR #811, opened 2026-07-24
[Bug] Don't cache may-aliasing buffers in CacheLoopInvariantGlobalVars (#810)

hughperkins:
CacheLoopInvariantGlobalVars could scalarize a loop-invariant global access into a local slot and defer its write-back until after the loop, while a distinct may-aliasing access to the same buffer read/wrote global memory directly. When the two addresses coincide at runtime (e.g. a store through a loop-variable index a[i] and a guarded literal-index read a[0] in a serialized loop) the direct access observed a stale value, so the in-kernel read returned the pre-write-back value even though post-kernel memory was correct.

The serial-offload path took an unconditional is_offload_unique()==true short-circuit, bypassing the may-alias exclusion the parallel path already gets via gather_uniquely_accessed_pointers.

Exclude from caching any buffer accessed within the offload through two distinct pointers P1, P2 with alias_analysis(P1, P2) == uncertain. Buffers whose accesses are all definitely-same or all definitely-different remain cacheable, so read-modify-write accumulators and disjoint accesses are unaffected.

Add a regression test (field + ndarray) reproducing the four ingredients: a serialized outer loop, an inner loop, a store through the outer loop variable index, and a guarded literal-index read of the same buffer.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/transforms/cache_loop_invariant_global_vars.cpp
+++ b/quadrants/transforms/cache_loop_invariant_global_vars.cpp
@@ -34,6 +34,88 @@ void gather_atomic_dests(IRNode *root,
   }
 }
 
+// Return the buffer that |ptr| addresses: a field by SNode (out_snode) or an
+// ndarray by arg_id (out_arr_id). Returns true if |ptr| resolves to either.
+bool resolve_access_buffer(Stmt *ptr, const SNode *&out_snode, std::vector<int> &out_arr_id) {
+  Stmt *origin = ptr;
+  if (auto *mptr = origin->cast<MatrixPtrStmt>()) {
+    origin = mptr->origin;
+  }
+  if (auto *gptr = origin->cast<GlobalPtrStmt>()) {
+    out_snode = gptr->snode;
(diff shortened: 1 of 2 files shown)


hughperkins:
@codex review

hughperkins:
thoughts on agent on perf:

Screenshot 2026-07-24 at 11 16 17

hughperkins:
Comparison with @duburcqa 's AI's version. Verdict is that his AI's output is better, on this occasion:

Screenshot 2026-07-24 at 11 24 00

hughperkins:
Some nuance:

Screenshot 2026-07-24 at 11 50 47

hughperkins:
Moving towards a hybrid of both solutions:

Screenshot 2026-07-24 at 12 27 35

hughperkins:
closing this in favor of https://github.com/Genesis-Embodied-AI/quadrants/pull/812

PR #825, opened 2026-07-30
[DONOTMERGE] Mac CI: within-VM A/B (--forked vs in-process teardown)

hughperkins:
## Do not merge - diagnostic only

Isolating the Mac teardown collapse from bimodal runner-lottery variance.

### Why
Every prior experiment compared a fix leg vs a control leg on **separate** runner VMs. The baseline is bimodal (~50 min on a good VM, 150-350 min on a bad one), so a single fix-vs-control pair cannot distinguish "the fix helped" from "the fix leg drew a good VM". Daemon-quiescing (SIGSTOP / deprio / disable) all failed to beat control under that noisy method; only more cores (xlarge) or fewer tests (split-arch) helped.

### What
Run **both arms back-to-back in one job (one VM)** so the VM factor cancels:
- **Candidate**: `pytest --forked` - each test in its own forked subprocess, so teardown is process exit rather than the in-process `qd.reset()` accumulation the collapse lives in. `genesis-world` runs `--forked` and never collapses.
- **Baseline**: the current in-process invocation (reproduces the collapse).

Candidate runs first so it is always captured even if the baseline arm collapses and runs long; the two pytest sessions are independent (fresh offline cache each) so order does not contaminate.

### Artifacts
Per-arm wall time + pytest `--durations`, `resource_sampler.log`, deep captures (spindump/thread dumps on load spike), and a summary comparing the two arms.


hughperkins:
Closing: conclusively negative. The `--forked` candidate arm failed in 99s with 22x 'Fatal Python error: Aborted' (159 fork-related frames) - the classic fork() + Apple Metal/Objective-C incompatibility (forking a process that has initialized the Metal/GPU runtime aborts in the child). --forked is not viable for quadrants' GPU backends on macOS. Pursuing thread-caps (#826) and the per-backend split (#827) instead.

PR #839, opened 2026-08-05
[perf_dispatch] Default repeat_after_count/seconds to None

hughperkins:
Previously repeat_after_seconds defaulted to 1.0 while repeat_after_count defaulted to 0. Since the two re-benchmark triggers are OR'd, setting only repeat_after_count still left a hidden per-second re-evaluation active, which was unintuitive.

Both now default to None ("unset"); a value <= 0 means explicitly disabled. When both are left unset, perf_dispatch picks an internal default of repeat_after_count=300 with no time-based trigger. Passing only one leaves the other off, so exactly the requested trigger is used.

Update the docstring, the perf_dispatch user guide, and add tests covering the new default and None semantics.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
@codex review

duburcqa (approved):

PR #840, opened 2026-08-05
[Docs] Fix doc-quality issues in tile user guide

hughperkins:
## Summary

Fixes the pre-existing doc-quality violations flagged by the "Check doc quality" CI job in `docs/source/user_guide/tile.md`:

- **Undefined term "subgroup"** (first used at line 5): introduce and link it to [`subgroup.md`](https://github.com/Genesis-Embodied-AI/quadrants/blob/main/docs/source/user_guide/subgroup.md) with a "(warp / wavefront)" gloss in the intro sentence, before first use.
- **Internal implementation detail out of scope**: drop "their slice-dispatch caches are independent" (internal jargon + implementation detail), keeping the user-facing statement that the two tile types can be mixed within one kernel.
- **Undefined jargon "FMA"**: expand to "fused multiply-add (FMA)".
- **Non-plain-English "N ≳ 32"**: reword to "roughly N >= 32".

These are independent of the per-lane `eps` docs change in #838; splitting them out so that PR isn't blocked on unrelated pre-existing issues (the doc-quality job reviews each changed doc file in full).

## Test plan

- [ ] Docs-only change.
- [ ] "Check doc quality" and "Check non-ASCII characters" jobs pass (all added lines are ASCII; no new undefined terms).


docs/source/user_guide/tile.md
```diff
@@ -1,6 +1,6 @@
 # Register-resident tiles: `Tile16x16` and `Tile32x32`
 
-Quadrants provides two register-resident matrix tile types:
+Quadrants provides two register-resident matrix tile types, each spread across the threads of one [subgroup](subgroup.md) (warp / wavefront):
```
hughperkins: technically that link doesnt define subgroup, but it's the natural place for us to define it, and since this PR is only scoped to fixing tile.md, I feel like fixing subgroup,md is out of the scope of this PR.

docs/source/user_guide/tile.md
```diff
@@ -128,7 +128,7 @@ rhs[0:N, 0:N] = B
 
 ## SharedArray support
 
-Tiles can load from and store to `qd.simt.block.SharedArray` using the same slice syntax as device arrays:
+Tiles can load from and store to `qd.simt.block.SharedArray` (block-scoped shared memory; see the [block execution guide](block.md)) using the same slice syntax as device arrays:
```
hughperkins: block.md neither defines shared memory, nor do I feel it is a natural place to define it. Perhaps we should simply define it here for now, in tile.md.

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,25 @@
+# GPU execution model
+
+This page defines the core GPU execution-model terms used across the Quadrants user guide: thread, block, subgroup, lane, and shared memory. Other pages link here the first time they use one of these words. If you already know them from CUDA, HIP, or another GPU framework you can skip this page; the only Quadrants-specific content is the API name attached to each concept.
+
+Quadrants compiles the same Python kernel to several GPU backends (CUDA, AMD, Vulkan, and Metal), and each vendor uses a different word for the same piece of hardware. This page picks one name per concept and lists the vendor synonyms, so the rest of the guide can stay vendor-neutral.
+
+## Thread
+
+A thread is a single instance of your kernel body, with its own registers and its own local variables. Launching a kernel starts many threads that run in parallel, each with an index it can read to decide which data to work on. This is the ordinary GPU "thread" you already use; the terms below just describe how threads are grouped and how they cooperate.
```
hughperkins: This is not very accurate

GPU kernel and qd.kernel are two different concepts, and should probably be clarified in this doc.

A qd.kernel does not 'start many threads'.

A hardware kernel starts 1 or more t hreads, that run in parallel. This is true for all platforms (GPU, CPU).

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Block
+
+A block (also called a thread block, a CUDA cooperative thread array or CTA, or a workgroup on AMD, Vulkan, and Metal) is a set of threads that launch together, can wait for each other at a barrier (a point every thread in the block must reach before any thread moves past it), and can cooperate through [shared memory](#shared-memory). You set how many threads a block has with `qd.loop_config(block_dim=N)`. Threads in different blocks cannot share shared memory or synchronize with one another.
```
hughperkins: Whilst technically true, I believe some barriers/fences work between blocks?

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Subgroup
+
+A subgroup (a warp on NVIDIA, a wavefront or wave on AMD, and a subgroup on Vulkan and Metal) is a fixed-size group of threads inside a block that the hardware runs in lockstep, meaning they advance through the same instruction at the same time. Common sizes are 32 threads (NVIDIA, and AMD in wave32 mode) and 64 threads (AMD in wave64 mode). Because a subgroup runs in lockstep, its threads can exchange values held in registers directly, without going through shared memory, using [subgroup operations](subgroup.md). Quadrants' register-resident [tiles](tile.md) are built on this.
```
hughperkins: I think it's important to state that the subgroup size is defined by the hardware. It's I believe the number of threads that runs on the SM at one time, though this is nuanced a lot:
- a warp actually runs on the SMP as half-warps I think?
- an SM contains typically 4 paritions, each of which can run a warp, so 4 warps can run in parallel on a single SM

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Lane
+
+A lane is the position of a thread within its subgroup, numbered from 0 to the subgroup size minus 1. A global thread index tells you which thread you are among all the launched threads; the lane index tells you where you sit inside your own subgroup. Subgroup operations that move a value from one thread to another (a shuffle, for example) name the source or destination thread by its lane. In Quadrants you read the calling thread's lane with `qd.simt.subgroup.invocation_id()`.
```
hughperkins: I think a lane is per warp, rather than per block?

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Shared memory
+
+Shared memory is fast on-chip memory that every thread in one [block](#block) can read and write, used to pass data between threads that would otherwise only see their own registers and the slower global memory. It is much faster than global memory but small in size, private to a single block, and gone once the block finishes. Vendors call it shared memory (CUDA), local data share or LDS (AMD), or threadgroup memory (Metal). In Quadrants you allocate it with `qd.simt.block.SharedArray(shape, dtype)` and coordinate access to it with the block barrier `qd.simt.block.sync()`; see [block primitives](block.md).
```
hughperkins: I think it is important to point out that shared memory is faster than global memory, but slower than registers.

docs/source/user_guide/tile.md
hughperkins: Repointed this to a new glossary entry, [`gpu_execution_model.md#subgroup`](gpu_execution_model.md#subgroup), which actually defines the term - so subgroup.md stays untouched / out of scope. (0fa84d7)

docs/source/user_guide/tile.md
hughperkins: Agreed block.md isn't the right home for this. Now linked to a new glossary entry, [`gpu_execution_model.md#shared-memory`](gpu_execution_model.md#shared-memory), that defines shared memory, instead of block.md.

docs/source/user_guide/gpu_execution_model.md
hughperkins: Reworded to separate the two concepts: the `@qd.kernel` you write is a Python program, which Quadrants compiles into a hardware kernel that the device then runs as one or more threads in parallel (GPU and CPU backends alike). It no longer says the kernel itself "starts many threads".

docs/source/user_guide/gpu_execution_model.md
hughperkins: Good catch - dropped the "cannot synchronize" overclaim. It now says a block's shared memory and block barrier only reach that one block, and that coordinating across blocks needs coarser tools such as a device-scope memory fence or a separate kernel launch (links grid.md).

docs/source/user_guide/gpu_execution_model.md
hughperkins: Added that the subgroup size is fixed by the hardware and backend, not something you choose (32 on NVIDIA / wave32, 64 on wave64). I kept the SM-partition / half-warp scheduling nuance out to keep this at end-user level - happy to add an "Advanced" note on how warps map onto SM partitions if you'd like.

docs/source/user_guide/gpu_execution_model.md
hughperkins: The definition was already per-subgroup; I've made that explicit: "Lane numbers are per subgroup, not per block: two threads in the same block but in different subgroups can carry the same lane number."

docs/source/user_guide/gpu_execution_model.md
hughperkins: Added the ordering: shared memory "sits between the two: slower than a thread's own registers, but much faster than global memory".

docs/source/user_guide/gpu_execution_model.md
hughperkins: Followed up: "kernel" now has its own dedicated glossary section (placed first), spelling out the two senses explicitly - a Quadrants `@qd.kernel` (the Python function you write) vs the hardware kernel Quadrants compiles it into and the device launches as threads. (6cc1fe04a)

hughperkins: Also captured the one-to-many mapping in the Kernel section: "Quadrants splits each `@qd.kernel` into one or more hardware kernels. Every top-level `for` loop becomes its own hardware kernel whose iterations run in parallel across threads, and every contiguous run of serial statements outside any top-level `for` loop becomes its own hardware kernel as well." (a8e8af470)

hughperkins: Fixed the Thread definition too - you're right that it was backwards. Now: "A thread is one of the parallel executions the device spawns when it launches a hardware kernel ... A single launch of a hardware kernel spawns one or more threads that run at the same time." (kept "one or more" rather than strictly "many" because a hardware kernel compiled from a contiguous serial block runs as a single thread.)

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Subgroup
+
+A subgroup (a warp on NVIDIA, a wavefront or wave on AMD, and a subgroup on Vulkan and Metal) is a group of threads inside a block that the hardware runs together, advancing through the same instruction at close to the same time. Its size is fixed by the hardware and backend, not something you choose: it is typically 32 threads (NVIDIA, and AMD in wave32 mode) or 64 threads (AMD in wave64 mode). Because the threads of a subgroup run together like this, they can exchange values held in registers directly, without going through shared memory, using [subgroup operations](subgroup.md). Quadrants' register-resident [tiles](tile.md) are built on this.
```
hughperkins: remove 'Quadrants' register-resident [tiles](tile.md) are built on this.'

docs/source/user_guide/gpu_execution_model.md
hughperkins: Right - dropped the "run at the same time" claim. Now: "The hardware schedules and runs these threads independently and does not necessarily run them all at the same time: GPUs time-slice and interleave groups of threads, and CPU threads are commonly interleaved too." (The lockstep-within-a-subgroup property is stated separately in the Subgroup entry.)

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,34 @@
...
+
+The two are not one-to-one: Quadrants splits each `@qd.kernel` into one or more hardware kernels. Every top-level `for` loop becomes its own hardware kernel whose iterations run in parallel across [threads](#thread), and every contiguous run of serial statements outside any top-level `for` loop becomes its own hardware kernel as well. Running a `@qd.kernel` launches these hardware kernels in order; the `@qd.kernel` itself does not run on the device (this compilation model holds on GPU and CPU backends alike).
+
+## Thread
+
+When the device launches a [hardware kernel](#kernel) it spawns one or more threads; a thread is one such execution of the kernel's code, with its own registers and its own local variables. The hardware schedules and runs these threads independently and does not necessarily run them all at the same time: GPUs time-slice and interleave groups of threads, and CPU threads are commonly interleaved too. Each thread reads an index identifying it so it can decide which data to work on. This is the ordinary "thread" you already use; the terms below just describe how threads are grouped and how they cooperate.
```
hughperkins: remove 'independently and does not necessarily run them all at the same time: GPUs time-slice and interleave groups of threads, and CPU threads are commonly interleaved too.'

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Subgroup
+
+A subgroup (a warp on NVIDIA, a wavefront or wave on AMD, and a subgroup on Vulkan and Metal) is a group of threads inside a block that the hardware runs together, advancing through the same instruction at close to the same time. Its size is fixed by the hardware and backend, not something you choose: it is typically 32 threads (NVIDIA, and AMD in wave32 mode) or 64 threads (AMD in wave64 mode). Because the threads of a subgroup run together like this, they can exchange values held in registers directly, without going through shared memory, using [subgroup operations](subgroup.md). Quadrants' register-resident [tiles](tile.md) are built on this.
```
hughperkins: "at close to the same time" => "at the same time"

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Block
+
+A block (also called a thread block, a CUDA cooperative thread array or CTA, or a workgroup on AMD, Vulkan, and Metal) is a set of threads that launch together and can cooperate closely: they can wait for each other at a barrier (a point every thread in the block must reach before any thread moves past it) and share fast on-chip [shared memory](#shared-memory). You set how many threads a block has with `qd.loop_config(block_dim=N)`. A block's shared memory and its block barrier only reach the threads of that one block; coordinating across blocks needs coarser tools such as a device-scope memory fence or a separate kernel launch (see [grid](grid.md)).
+
+## Subgroup
```
hughperkins: move subgroup in front of block.

docs/source/user_guide/gpu_execution_model.md
hughperkins: Removed.

docs/source/user_guide/gpu_execution_model.md
hughperkins: Removed - the Thread entry no longer describes scheduling/interleaving; it just says the hardware schedules and runs the threads.

docs/source/user_guide/gpu_execution_model.md
hughperkins: Changed to "advancing through the same instruction at the same time".

docs/source/user_guide/gpu_execution_model.md
hughperkins: Reordered: Subgroup now comes before Block. I dropped "inside a block" from the subgroup definition (so it no longer forward-references block) and moved the relationship into the Block entry: "The threads of a block are divided into subgroups."

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Block
+
+A block (also called a thread block, a CUDA cooperative thread array or CTA, or a workgroup on AMD, Vulkan, and Metal) is a set of threads that launch together and can cooperate closely: they can wait for each other at a barrier (a point every thread in the block must reach before any thread moves past it) and share fast on-chip [shared memory](#shared-memory). The threads of a block are divided into [subgroups](#subgroup). You set how many threads a block has with `qd.loop_config(block_dim=N)`. A block's shared memory and its block barrier only reach the threads of that one block; coordinating across blocks needs coarser tools such as a device-scope memory fence or a separate kernel launch (see [grid](grid.md)).
```
hughperkins: A block comprises one or more subgroups. I don think a block can be a non-integer multiple of the number of threads in a subgroup?

hughperkins: Agreed. Reworded: "A block is made up of one or more whole subgroups, so its thread count is a multiple of the subgroup size (a requested size that is not a multiple is rounded up, with the extra lanes left idle)."

hughperkins: Moved it to the front - the Block paragraph now opens with "A block comprises one or more whole subgroups, so its thread count is a multiple of the subgroup size ...".

docs/source/user_guide/gpu_execution_model.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Block
+
+A block comprises one or more whole [subgroups](#subgroup), so its thread count is a multiple of the subgroup size (a requested size that is not a multiple is rounded up, with the extra lanes left idle). It (also called a thread block, a CUDA cooperative thread array or CTA, or a workgroup on AMD, Vulkan, and Metal) is a set of threads that launch together and can cooperate closely: they can wait for each other at a barrier (a point every thread in the block must reach before any thread moves past it) and share fast on-chip [shared memory](#shared-memory). You set how many threads a block has with `qd.loop_config(block_dim=N)`. A block's shared memory and its block barrier only reach the threads of that one block; coordinating across blocks needs coarser tools such as a device-scope memory fence or a separate kernel launch (see [grid](grid.md)).
```
hughperkins: rmeove 'whole'. remove 'so its thread count is a multiple of the subgroup size (a requested size that is not a multiple is rounded up, with the extra lanes left idle)'. 'It (also called ' => "A block can also be referred to as '

hughperkins: Done. Now: "A block comprises one or more subgroups. A block can also be referred to as a thread block, a CUDA cooperative thread array or CTA, or a workgroup on AMD, Vulkan, and Metal. It is a set of threads that launch together and can cooperate closely: ..." Dropped "whole" and the thread-count-multiple clause.

duburcqa (approved):

PR #843, opened 2026-08-07
[AMDGPU] Use native atomics for reductions instead of runtime reduce_* helpers

paveltc:
## Summary
On AMDGPU, `optimized_reduction()` lowered every reduction (`+=`, `atomic_min`, `atomic_max`, ...) to a call into a runtime `reduce_*` helper (`reduce_add_i32`, `reduce_add_f32`, `reduce_min_f32`, ...). Those helpers take `addrspace(0)` pointers, so a global SNode destination was routed through a flat pointer for the call, and the f32-add helper path did not lower to the hardware float atomic.

This PR removes the AMDGPU `optimized_reduction()` override. Reduction atomics now fall through the standard atomic visitor to `real_type_atomic()` / `integral_type_atomic()`, which already emit native LLVM atomics (`AtomicRMW` / `FAdd` / `FMin` / `FMax`) at `"agent"` syncscope via `kernel_atomic_syncscope()`. On gfx942 these lower to hardware `global_atomic_*` instructions and preserve the destination address space.

This is a simplification / dependency removal — it deletes AMDGPU-specific reduction codegen in favor of the shared LLVM atomic path already used by the CPU and CUDA backends. This also aligns with the project's broader move toward native float atomics (cf. the separate Metal-backend migration in #788), and is a cleaner base for upcoming AMDGPU address-space work.

## Changes
- `quadrants/codegen/amdgpu/codegen_amdgpu.cpp`: remove `optimized_reduction()` override (+9/-27), replaced by an explanatory comment. Single-file change.

## Validation (gfx942 / MI300X)
- **IR:** f32 add now emits native `atomicrmw fadd ... syncscope("agent")`; i32 add/min/max likewise; no `@reduce_*` helper call.
- **Correctness:** i32 add/min/max/and/or/xor and f32 add/min/max match NumPy across 1M+ elements.
- **Performance:** perf-neutral vs. the helper path (f32 add ~0.72 ms, i32 add ~0.73 ms / 16M elems, matching baseline). Emitting the same reduction as a *default-syncscope* `atomicrmw fadd` regresses ~1000x due to software CAS lowering; this path correctly uses `"agent"` scope and avoids that.

### DIFF
--- a/quadrants/codegen/amdgpu/codegen_amdgpu.cpp
+++ b/quadrants/codegen/amdgpu/codegen_amdgpu.cpp
@@ -224,16 +224,6 @@ class TaskCodeGenAMDGPU : public TaskCodeGenLLVM {
 #undef UNARY_STD
   }
 
-  // AMDGPU deliberately does not override optimized_reduction(). Reduction atomics
-  // fall through to real_type_atomic() / integral_type_atomic(), which emit native
-  // LLVM atomics (AtomicRMW / FAdd / FMin / FMax) at "agent" syncscope (see
-  // kernel_atomic_syncscope()); on gfx942 these lower to hardware global_atomic_*.
-  // The previous override routed reductions through the runtime reduce_* helpers,
-  // which take addrspace(0) pointers and so forced a flat-pointer addrspace cast at
-  // the call site. In particular the f32-add helper path defeated the hardware
-  // float atomic, making a contended f32 sum reduction ~1000x slower than the
-  // native agent-scoped atomicrmw fadd.
-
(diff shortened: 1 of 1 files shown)


paveltc:
@codex review

hughperkins:
Agent appears good with your change:

Screenshot 2026-08-24 at 09 18 20

Check whether we need something similar for CUDA: no (not that I would ask you to apply a similar CUDA change, but just checking anyway).
Screenshot 2026-08-24 at 09 17 35

hughperkins:
Intend to merge once CI passes.

hughperkins:
Thanks!

PR #850, opened 2026-08-07
[fastcache] Exclude non-codegen config keys from the fast-cache hash

hughperkins:
## Summary

`hash_compile_config` builds the fast-cache config hash by walking every pybind-exported `CompileConfig` field and
filtering only by prefix (`EXCLUDE_PREFIXES`). That sweeps in knobs which have no influence on generated code, so
changing one of them forces a full Python AST transform and lowering for kernels whose GPU code is byte-identical.

By contrast the C++ offline cache key (`get_offline_cache_key_of_compile_config` in
`quadrants/analysis/offline_cache_util.cpp`) is an explicit allowlist of compilation-relevant fields. This PR moves
three fields on the Python side into an exact-name `EXCLUDE_KEYS` set so the two keys agree:

- `external_metal_command_queue` is a borrowed `MTLCommandQueue*`. Its address is process-local, so hashing it handed
  every restart a different key and no shared-queue Metal process could reuse another's entry. It is read live from the
  config in two places, both at runtime: `metal_program.cpp` when creating the device, and
  `GfxRuntime::submit_current_cmdlist_if_timeout` when deciding whether to force a flush. It is never baked into a
  compiled artifact.
- `debug_dump_path` only chooses the destination directory for IR dumps. Whether anything is dumped at all is gated on
  the `QD_DUMP_IR` env var, which is not part of the cache key either, so the path cannot affect the generated kernel.
- `cuda_stack_limit` is read in exactly one place, `llvm_runtime_executor.cpp`, to call
  `context_set_limit(CU_LIMIT_STACK_SIZE, ...)` at runtime. It never reaches codegen.

Also removes `CompileConfig::extra_flags`, which is declared in `compile_config.h` and never read or written anywhere
in the repo. It is not initialised by the `CompileConfig` constructor and not exported to Python, so nothing can set it.

Matching by exact name rather than by prefix is deliberate: `debug_`, `cuda_` and `external_metal_` happen to match only
these fields today, but a prefix would silently drop any codegen-relevant field added under the same prefix later.
`external_` in particular would wrongly drop `external_optimization_level`, which is a genuine compiler input.

## Relationship to #849

#849 fixes the same `external_metal_command_queue` bug by a narrower route: it rewrites the pointer to a sentinel string
while hashing, but only when `external_metal_command_queue_is_torch_queue` is set, and keeps arbitrary non-PyTorch queue
addresses distinct on the grounds that they may belong to a different Metal device.

This PR excludes the field for all values instead. Device identity does not need the pointer: it is already covered by
`DeviceCapabilityConfig::devcaps`, which is hashed separately into the C++ key. Non-PyTorch external queues (which
`docs/source/user_guide/metal_shared_queue.md` explicitly supports) then get the same cross-process reuse as PyTorch
ones, rather than remaining broken.

The two PRs touch the same lines and will conflict. #849's `test_config_hasher_normalizes_torch_mps_queue_address` also
asserts that two different non-torch queue addresses hash differently, which is the behaviour this PR deliberately
removes, so whichever lands second needs updating.

## Safety

Removing something from a cache key risks stale hits, so worth stating why there is none here. The persisted
`CacheValue` stores source-info hashes, dataclass parameters, `graph_do_while` levels and checkpoint metadata, plus a
nested `frontend_cache_key` which is the C++ key. None of the excluded fields is captured in any of that, and none is in
the C++ key, so no cached artifact can depend on them. All three are read live from the config at runtime, so behaviour
still follows the current process's settings. The direction of the change is over-hashing to correct hashing, where the
failure mode being fixed is a spurious miss rather than a wrong result.

## Test plan

- [ ] New `test_config_hasher_ignores_debug_and_runtime_only_keys` in
      `tests/python/quadrants/lang/fast_caching/test_config_hasher.py` asserts the hash is stable across changes to all
      three excluded fields, and still changes for a genuine compiler input (`fast_math`) so the exclusions are not
      masking everything. It pokes `qd.cfg` directly, so it runs on every backend without special hardware.
- [ ] Full build, to confirm dropping `extra_flags` breaks nothing.
- [ ] End-to-end shared-queue cache reuse needs Metal hardware; the repro in #849 covers it.

### DIFF
--- a/docs/source/user_guide/fastcache.md
+++ b/docs/source/user_guide/fastcache.md
@@ -121,7 +121,7 @@ Each compiled artifact is stored under a key derived from all of the following:
 - The **Quadrants version** (`quadrants.__version__`).
 - The **source code** of the kernel function or any `@qd.func` it calls.
 - The **argument types** (e.g. switching from `f32` to `f64`, or changing ndarray dimensionality).
-- The **compiler configuration** (e.g. `arch`, `debug`, `opt_level`, `fast_math`).
+- The **compilation-relevant parts of the compiler configuration** (e.g. `arch`, `debug`, `opt_level`, `fast_math`). Options that cannot change the generated code are deliberately excluded, so that toggling them does not force a recompile: the `print_*` and `verbose_*` debug options, the `offline_cache*` options, `debug_dump_path` (only chooses the directory IR dumps are written to) and `cuda_stack_limit` (only sets a CUDA context limit at runtime).
 - **Template parameter values** (since they are baked into the compiled kernel).
 
 When any of these change, the resulting key is different, so a new compilation occurs and a new entry is stored. Previous entries remain on disk - multiple cached versions coexist. You do not need to manually clear the cache when making code changes - the hash mismatch causes a transparent recompilation.
--- a/docs/source/user_guide/fastcache.md
+++ b/docs/source/user_guide/fastcache.md
@@ -122,6 +122,7 @@ Each compiled artifact is stored under a key derived from all of the following:
 - The **source code** of the kernel function or any `@qd.func` it calls.
 - The **argument types** (e.g. switching from `f32` to `f64`, or changing ndarray dimensionality).
 - The **compilation-relevant parts of the compiler configuration** (e.g. `arch`, `debug`, `opt_level`, `fast_math`).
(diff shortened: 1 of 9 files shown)


hughperkins:
Relates to / inspired by https://github.com/Genesis-Embodied-AI/quadrants/pull/849

hughperkins:
@codex review

hughperkins:
Depends on https://github.com/Genesis-Embodied-AI/quadrants/pull/853 to merge first

hughperkins:
@codex review

hughperkins:
@codex review

duburcqa:
ok to merge

PR #851, opened 2026-08-07
[Caching] Per-construct frontend split + in-memory construct cache

hughperkins:
Stacked on #837. Second of three; the top of the stack carries the headline result.

## What

Runs the pre-offload frontend **per top-level construct** instead of once over the whole kernel, then reassembles the
offloaded tasks in source order, and caches each construct's output in memory.

The motivation is that the dominant frontend pass is super-linear in block size: on the genesis `_step_kernel`,
`merge_global_ptrs` is **11.1 s whole-kernel but ~14 ms per construct**. A one-line edit re-ran all of it.

## Correctness work this required

- **Segment-based constructs + backward-slice isolation.** The first definition stripped a store's own pointer operand,
  producing a dangling operand and a segfault.
- **Recompute-safety gate** with whole-kernel fallback: a local mutated in one construct and read in another cannot be
  recomputed, so those kernels do not split.
- **Graph region tags in the cache key.** The IR printer omits `gdw_level` for pre-offload constructs, so constructs at
  different `graph_do_while` levels collided; the cached clone carried the wrong level and the kernel never terminated.
- **Cache invalidation on `destroy_snode_tree()`.** Tree ids are recycled and the cache holds cloned IR with raw
  `SNode*` into freed trees — a use-after-free.
- Per-task key widened with the kernel ABI and task index (aliased tasks previously gave silently wrong results).

## Safety

Entirely behind opt-in env gates; default code paths are unchanged and verified byte-identical.
Correctness is anchored on a BDF1 freefall oracle: `q_sum = 1.7492255125`, identical across every variant.

### DIFF
--- a/python/quadrants/lang/_kernel_types.py
+++ b/python/quadrants/lang/_kernel_types.py
@@ -27,6 +27,20 @@ class FeLlCacheObservations:
 
 
 @dataclass
+class PerOffloadCacheObservations:
+    """Per-construct (per-offloaded-task) compilation cache stats for one kernel compile.
+
+    A "construct" is one top-level offloaded region of the kernel. On a warm compile where only one construct changed,
+    the expected result is ``constructs_recompiled == 1`` and ``constructs_cache_hit == constructs_total - 1``. Counts
+    (not wall time) so the behavior can be asserted deterministically in tests.
+    """
+
+    constructs_total: int = 0
+    constructs_cache_hit: int = 0
+    constructs_recompiled: int = 0
+
+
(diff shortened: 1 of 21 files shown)


hughperkins:
Superseded. Re-splitting this work by what is actually being changed rather than by the branches that happened to exist, and dropping the env gates and diagnostic scaffolding entirely (everything defaults on). Plan: `perso_hugh/doc/per_offload_cache/per_offload_prs.md`. Branches stay pushed; nothing is lost.

PR #852, opened 2026-08-07
[Caching] Per-offload cache

hughperkins:
Stacked on #851 (which is stacked on #837). This is the top of the stack and carries the headline result.

## Result

Editing **one offloaded task** of the giant qipc nested-graph `_step_kernel` and re-running in a fresh process
(freefall NB=1, CUDA, caches on, same harness for every cell):

| scenario | baseline | this stack |
|---|---|---|
| cold | 22.24 s | **20.75 s** |
| warm, unchanged | 0.163 s | 0.190 s |
| **warm, one offload edited** | 23.82 s | **5.69 s** |

`compile_kernel` for the edit case: **18.92 s -> 1.87 s**. No regression on cold or warm-unchanged.
`q_sum = 1.7492255125` byte-identical throughout.

## What this branch adds

1. **Per-task artifact cache (disk).** A cubin alone cannot be launched — the launcher and CUDA graph builder run off
   `OffloadedTask` metadata — so the record is `{cubin, tasks, used_tree_ids, struct_for_tls_sizes}`. Probed *before*
   `compile_task`, so a hit skips CHI->LLVM, link, optimize, PTX and ptxas entirely.
2. **Construct manifests (disk).** CHI IR has no round-trippable codec, so instead of persisting a construct's IR we
   persist what it *produced*: the ordered per-task artifact keys. A hit skips that construct's whole frontend.
3. **Cubin cache re-keyed by the per-task IR key** (was a SHA of the sub-module's LLVM text, which could only be
   computed *after* codegen — so a hit could only ever skip ptxas). Directory namespaced by SM.
4. **Whole-module link skipped** when tasks are code-only, removing the prototype's double-build.
5. **`clone_block_subset`** — the split cloned the whole 2685-stmt block per construct, O(constructs x block).
   `compile_kernel` 7.73 s -> 2.13 s.
6. **Parallel per-task cubin build** — 242 serial `ptxas -c` shell-outs were 9.8 s of a 13.7 s cubin build (LLVM->PTX
   was only 1.6 s). ptxas is a subprocess on private temp files so it parallelises; LLVM->PTX stays serialised because
   those modules may share an `LLVMContext`. This removed a cold regression of ~7 s.
7. Memoised AST source-position formatting (see caveat below).

## Reviewer notes

- **This does not enable anything by default.** Every change is behind an opt-in env gate; de-gating is deliberately
  left as follow-up so this can land without changing default behaviour. The numbers above are with the gates on.
- **Overlap with #804.** The AST source-position memo here targets the same hot spot as #804's second change
  (`get_pos_info` formatting a caret hint per AST node). #804's approach is better — it avoids the work rather than
  caching it — and its blocking review comments are all on its *other* change (`_inside_class`/`linecache`). Happy to
  drop the memo here in favour of #804's approach, or to have #804 split so its second half can land.
- Known follow-ups tracked in `perso_hugh/doc/per_offload_cache/`: eviction limits for the new cache tiers, task-index
  shift handling, `ptxas` -> `libnvptxcompiler_static`, and de-gating.

### DIFF
--- a/python/quadrants/lang/ast/ast_transformer_utils.py
+++ b/python/quadrants/lang/ast/ast_transformer_utils.py
@@ -217,6 +217,14 @@ class ASTTransformerGlobalContext:
         self.caller_in_non_static_control_flow: bool = False
 
 
+# Memo for ASTTransformerFuncContext.get_pos_info. Process-global because the win comes from reuse ACROSS transforms
+# of the same function (each is transformed once per inlined call site, ~44x on the genesis step kernel), which is
+# exactly what a per-context cache would miss. Bounded in practice by (#functions x #AST nodes) and holds only short
+# strings; entries stay valid for the life of the process because a given file+function's source cannot change under
+# a running interpreter.
+_POS_INFO_CACHE: dict[tuple, str] = {}
+
+
 class ASTTransformerFuncContext:
     def __init__(
         self,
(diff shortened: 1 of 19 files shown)


hughperkins:
Superseded. Re-splitting this work by what is actually being changed rather than by the branches that happened to exist, and dropping the env gates and diagnostic scaffolding entirely (everything defaults on). Plan: `perso_hugh/doc/per_offload_cache/per_offload_prs.md`. Branches stay pushed; nothing is lost.

PR #853, opened 2026-08-07
[Bug] Stop the caches from silently defeating the IR dump and load env vars

hughperkins:
## Summary

`QD_LOAD_IR` and `QUADRANTS_LOAD_PTX` make codegen throw away the module it just built and use IR / PTX read back from
`debug_dump_path` instead:

- `quadrants/codegen/llvm/codegen_llvm.cpp` reparses `<debug_dump_path>/<task>_llvm.ll` and replaces `module` with it.
- `quadrants/runtime/cuda/jit_cuda.cpp` reads `<debug_dump_path>/<name>.ptx` and uses it as the PTX.

Codegen only runs on a cache miss, because `KernelCompilationManager::load_or_compile` is
`cached_kernel ? *cached_kernel : compile_and_cache_kernel(...)`. So with caching enabled, the second run of an edited
kernel is handed the previously compiled artifact and the edited file is never read, with no diagnostic at all. The
kernel runs, it just is not the code that was edited.

It does work on a cold cache, which makes the failure worse rather than better: the workflow appears to function once and
then silently stops, so the natural conclusion is that the IR edit had no effect.

`qd.init` now raises instead of letting that happen.

## Scoped to the backends that read each variable

The check runs after `adaptive_arch_select`, since that is where the effective arch becomes known, and it may fall back
to an arch other than the one requested. Each variable is then gated to the backends that can actually consume it:

- `QD_LOAD_IR` is read only in `codegen_llvm.cpp`, so it is gated on `arch_uses_llvm` (CPU, CUDA, AMDGPU). The SPIR-V
  backends, Vulkan and Metal, never read it, and neither does the python backend.
- `QUADRANTS_LOAD_PTX` is read only in `jit_cuda.cpp`, so it is gated on `cuda`.

Without that gating, a shell with `QD_LOAD_IR` exported would fail every Metal or Vulkan run at init over a variable that
could never have been read.

## Why both caches have to be off

The error requires `offline_cache=False` *and* `src_ll_cache=False`. `offline_cache` gates the C++ lookup, but the Python
fastcache entry point (`Kernel._try_load_fastcache` in `python/quadrants/lang/kernel.py`) is gated on
`self.runtime.src_ll_cache` alone, not on `offline_cache`, so turning off only the offline cache is not sufficient to
guarantee codegen runs.

## Making the two variables agree on how they are read

The two were read differently. `QD_LOAD_IR` goes through the shared `get_environ_config` helper, which parses the value
with `std::stoi`, so `QD_LOAD_IR=0` leaves it off. `QUADRANTS_LOAD_PTX` was only tested for presence in `jit_cuda.cpp`,
so `QUADRANTS_LOAD_PTX=0` still loaded PTX.

Rather than encode that asymmetry in the check and then document it as a gotcha, `jit_cuda.cpp` now reads both of its
variables through `get_environ_config`, as `codegen_llvm.cpp` already does; it was the only site bypassing the helper for
these. That also fixes a second inconsistency inside the same function: `QD_DUMP_IR` was compared against `"1"` for the
`.ll` dump but only tested for presence for the `.ptx` dump, so `QD_DUMP_IR=0` wrote PTX anyway.

`QUADRANTS_LOAD_PTX` also gets a `LOAD_PTX_ENV` constant in `ir_dump.h`, next to `DUMP_IR_ENV` and `LOAD_IR_ENV`, rather
than a bare literal at the use site.

One consequence worth naming: a non-numeric value now throws from `std::stoi` instead of counting as enabled. That is
already the behavior of `QD_LOAD_IR` and `QD_SHOW_COMPILING` today, so this makes the family consistent rather than
introducing a new failure mode, and these are all dev-only variables.

## Docs

`QD_LOAD_IR` and `QUADRANTS_LOAD_PTX` were not mentioned anywhere under `docs/`, so the dump-edit-rerun workflow was
undiscoverable and its dependency on disabling caching was invisible. It is now documented in the `## Advanced` section
of `docs/source/user_guide/contributing.md`, next to "Building LLVM for debugging it".

That page rather than `debug.md`: hand-editing compiler output to try a codegen change is work on Quadrants itself, not
debugging your own kernels, which is what `debug.md` covers. `contributing.md` is also titled "Advanced: Contributing to
quadrants", so the content sits in a doc already aimed at that reader.

## The dump path had the same bug, with a warning instead of a fix

`print_ir` and `QD_DUMP_IR` hit the same wall from the other side, and a `DUMP_IR_CACHE_MISMATCH` warning a few lines
above already said so: a cached kernel is returned without running codegen, so nothing gets printed or dumped for it.
Warning had two problems.

It covered 2 of the 12 options with this problem. `CompileConfig` has nine `print_*` booleans
(`print_preprocessed_ir`, `print_ir`, `print_accessor_ir`, `print_ir_dbg_info`, `print_struct_llvm_ir`,
`print_kernel_llvm_ir`, `print_kernel_llvm_ir_optimized`, `print_kernel_asm`, `print_kernel_amdgcn`) and there are three
dump env vars (`QD_DUMP_IR`, `QD_DUMP_CFG`, `QD_DUMP_SIMPLIFY`), all of which only produce output while a kernel is
being compiled. The check also only looked at `offline_cache`, so the fastcache could swallow the output with no warning
at all.

And it left the fix to the user in a situation with only one sensible outcome: if you asked for the IR, you want the IR.

Both caches are now turned off for any of the twelve, and the warning stays, because every kernel then recompiles and
that is worth saying out loud. This overrides `offline_cache=True` even when it is passed explicitly: asking for the
output is the more specific intent, and honoring both would drop the output for precisely the kernels that are already
cached.

The trigger now also matches how the C++ reads these. The old condition was `os.getenv("QD_DUMP_IR") == "1"` while the
C++ int-parses the value, so `QD_DUMP_IR=2` dumped without warning.

## Related

This is the load-path counterpart of the existing `DUMP_IR_CACHE_MISMATCH` warning a few lines above, which covers
`print_ir` / `QD_DUMP_IR` against `offline_cache`. That one is a warning because a missing dump is self-evident; a
silently stale kernel is not, so this one is an error.

Orthogonal to #850, which excludes non-codegen fields from the fastcache key: no overlapping files. It does firm up one
of #850's justifications, though. `debug_dump_path` is excluded from the fastcache key there, and this PR is what makes
that sound, by guaranteeing the field cannot be a codegen input in any process where a fastcache key is computed.

## Test plan

- [ ] `tests/python/test_error_load_ir_with_caching.py` covers: error for `QD_LOAD_IR` with `offline_cache`, error for
      `QD_LOAD_IR` with `src_ll_cache`, and no error for caching disabled, for `QD_LOAD_IR=0`, for
      `QUADRANTS_LOAD_PTX=0`, for neither variable set, for `QD_LOAD_IR` on a non-LLVM arch, and for
      `QUADRANTS_LOAD_PTX` on a non-CUDA arch.
- [ ] The CUDA-specific cases call the helper directly with a stub config, so they assert the gating without needing CUDA
      on the test machine. Going through `qd.init` would not work: `adaptive_arch_select` falls back to the CPU arch when
      a backend is unavailable, which would turn a non-LLVM arch back into an LLVM one.
- [ ] Env vars are set and restored by a context manager so nothing leaks between tests.
- [ ] The `jit_cuda.cpp` change is mechanical (four branches switched to two hoisted bools) and is covered by CI
      building the CUDA backend, but the end-to-end load path still wants a manual run on a CUDA box.
- [ ] `tests/python/test_warn_print_dump_ir_with_caching.py` keeps its three original cases, and adds: both caches off
      over an explicit request, each of the three dump env vars, a `print_*` option other than `print_ir`,
      `QD_DUMP_IR=0` leaving caching alone, no request leaving caching alone, and a guard that every watched config key
      still exists so a rename fails there rather than inside `qd.init`.

### DIFF
--- a/docs/source/user_guide/contributing.md
+++ b/docs/source/user_guide/contributing.md
@@ -228,6 +228,21 @@ Quadrants comprises at least three important parts:
 2. `quadrants` device runtime (bitcode): C++ code compiled using `clang++` from the distribution/OS. Using `clang++` is required as it has to support the same targets as `LLVM`.
 3. `LLVM` libraries used by host runtime: statically or dynamically linked, used to lower the kernel's final IR to machine code on the host. The CI uses an LLVM version compiled from source.
 
+### Running hand-edited IR or PTX
+
+A dump written by `QD_DUMP_IR` (see [Dumping compiled IR](./debug.md#dumping-compiled-ir)) can be edited and fed back in, which is a quick way to try an instruction sequence by hand before changing codegen. `QD_LOAD_IR=1` makes the LLVM backends reparse `<debug_dump_path>/<task>_llvm.ll` instead of using the module they just built, and `QUADRANTS_LOAD_PTX=1` makes the CUDA backend read `<debug_dump_path>/<name>.ptx` instead of the PTX it just compiled. Both are read as integers, so `0` leaves them off.
+
+Both caches have to be off in the script under test, since the replacement files are only read while a kernel is being compiled and a cached kernel is handed back without compiling. `qd.init` raises a `ValueError` rather than let an already-cached kernel keep silently running its old code. `offline_cache` has an environment variable, `src_ll_cache` does not, so the latter has to be passed to `qd.init`:
+
+```python
+qd.init(arch=qd.cuda, offline_cache=False, src_ll_cache=False)
+```
+
+```bash
+QD_DUMP_IR=1 QD_DEBUG_DUMP_PATH=/tmp/ir python my_script.py  # dump, then edit the files under /tmp/ir
+QD_LOAD_IR=1 QD_DEBUG_DUMP_PATH=/tmp/ir python my_script.py  # run what you edited
+```
+
 ### Building LLVM for debugging it
 
 Sometimes, it could be useful to have a `LLVM` version that allows to print intermediate passes or with debug symbols to find out where and why LLVM fails (for example, when Instruction Selection fails). To do so you would have to build LLVM by yourself. If so, you should take some inspiration from our [CI pipeline to build LLVM](https://github.com/Genesis-Embodied-AI/quadrants-sdk-builds/blob/main/.github/workflows/llvm-ci.yml) to tweak a little bit to your liking (and not enable/disable options that would create discrepancies).
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -38,9 +38,7 @@ a = x[-1]     # AssertionError in debug mode
 
 #### Adstack overflow
 
-The adstack is the per-thread stack that reverse-mode autodiff uses to save the intermediate values a kernel produces, so that the backward pass can read them back. Overflowing it is checked always, on every backend, regardless of `debug`, and raises `QuadrantsAssertionError("[Aa]dstack overflow")`.
-
-The error surfaces at the first Quadrants call after the kernel that overflowed: a kernel launch, a host-side field / ndarray read, or `qd.sync()`. On CPU that is the offending launch itself, but on GPU it can be one or more calls later, so the traceback may point at innocent code and the kernel to look at is one launched earlier. The message names the likely cause, either an untracked tensor mutation between launches or Quadrants under-estimating the capacity it needed to reserve, which is a bug in Quadrants, along with the recovery flow; see [Autodiff -> What can go wrong](autodiff.md) for the full description.
+The adstack overflow check on reverse-mode autodiff runs always, on every backend, regardless of `debug`. A push past the per-stack capacity raises `QuadrantsAssertionError("[Aa]dstack overflow")` at the next Quadrants Python entry that polls the overflow flag after the offending kernel has executed - kernel launch, host-side field / ndarray read, or `qd.sync()`. On CPU this is the entry of the offending launch itself; on GPU it can be one or more entries later, since the GPU may not have run the offending push by the time the poll at end-of-launch fires. The error message describes the cause (untracked tensor mutation between launches, or sizer under-estimate caused by a bug in Quadrants) and the recovery flow; see [Autodiff -> What can go wrong](autodiff.md) for the full description.
 
 ### Assertions in kernels
 
@@ -112,10 +110,6 @@ Per-backend support:
 
 **Important.** Avoid kernel `print()` calls in production code where you can. Quadrants synchronizes the compute queue after every dispatch of a kernel that contains a `print()` so the output appears as close as possible to the call site. The synchronization happens unconditionally on every launch of that kernel, even when the surrounding control flow leaves the `print()` unreached at runtime; the cost is the full per-launch sync overhead, not just the cost of the `print()` itself.
 
-## Under the hood: compiler and runtime internals
-
-The tools below expose Quadrants' own compiler and runtime internals. They are for understanding exactly what Quadrants generated, or for debugging Quadrants itself, rather than for debugging the logic of your own kernels.
-
 ### Dumping compiled IR
 
 To inspect the compiled intermediate representation, use the `QD_DUMP_IR` environment variable:
@@ -125,21 +119,6 @@ QD_DUMP_IR=1 QD_OFFLINE_CACHE=0 python my_script.py
 ```
 
 Compiled kernels will be written to `/tmp/ir` by default. Use `QD_DEBUG_DUMP_PATH=` to redirect to a custom directory.
-
-### Running hand-edited IR or PTX
-
-A dump can be edited and fed back in, which is a quick way to try an instruction sequence by hand before changing the compiler. `QD_LOAD_IR=1` makes the CPU, CUDA and AMDGPU backends read LLVM IR (the representation they compile through) back from the dump directory instead of generating it; `QUADRANTS_LOAD_PTX=1` does the same for PTX (the assembly the NVIDIA driver consumes) on CUDA. Both caches have to be off in the script under test:
-
-```python
-qd.init(arch=qd.cuda, offline_cache=False, src_ll_cache=False)
-```
-
-```bash
-QD_DUMP_IR=1 QD_DEBUG_DUMP_PATH=/tmp/ir python my_script.py  # dump, then edit the files under /tmp/ir
-QD_LOAD_IR=1 QD_DEBUG_DUMP_PATH=/tmp/ir python my_script.py  # run what you edited
-```
-
-**Important.** The replacement files are only read while a kernel is being compiled, and a cached kernel is handed back without compiling, so with either cache left on, an already-cached kernel keeps running its old code and nothing says so. `qd.init` raises a `ValueError` instead of letting that happen. Unlike `offline_cache`, [`src_ll_cache`](./fastcache.md) has no environment variable, so it has to be passed to `qd.init`.
 
 ### Tracing adstack heap allocations
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -38,7 +38,9 @@ a = x[-1]     # AssertionError in debug mode
 
 #### Adstack overflow
 
-The adstack is the per-thread stack that reverse-mode autodiff uses to save the intermediate values a kernel produces, so that the backward pass can read them back. Its overflow check runs always, on every backend, regardless of `debug`. A push past the per-stack capacity raises `QuadrantsAssertionError("[Aa]dstack overflow")` at the next Quadrants Python entry that polls the overflow flag after the offending kernel has executed - kernel launch, host-side field / ndarray read, or `qd.sync()`. On CPU this is the entry of the offending launch itself; on GPU it can be one or more entries later, since the GPU may not have run the offending push by the time the poll at end-of-launch fires. The error message describes the cause (untracked tensor mutation between launches, or Quadrants under-estimating the capacity it needed to reserve, which is a bug in Quadrants) and the recovery flow; see [Autodiff -> What can go wrong](autodiff.md) for the full description.
+The adstack is the per-thread stack that reverse-mode autodiff uses to save the intermediate values a kernel produces, so that the backward pass can read them back. Overflowing it is checked always, on every backend, regardless of `debug`, and raises `QuadrantsAssertionError("[Aa]dstack overflow")`.
+
+The error surfaces at the first Quadrants call after the kernel that overflowed: a kernel launch, a host-side field / ndarray read, or `qd.sync()`. On CPU that is the offending launch itself, but on GPU it can be one or more calls later, so the traceback may point at innocent code and the kernel to look at is one launched earlier. The message names the likely cause, either an untracked tensor mutation between launches or Quadrants under-estimating the capacity it needed to reserve, which is a bug in Quadrants, along with the recovery flow; see [Autodiff -> What can go wrong](autodiff.md) for the full description.
 
 ### Assertions in kernels
 
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -120,6 +120,23 @@ QD_DUMP_IR=1 QD_OFFLINE_CACHE=0 python my_script.py
 
 Compiled kernels will be written to `/tmp/ir` by default. Use `QD_DEBUG_DUMP_PATH=` to redirect to a custom directory.
 
+### Running hand-edited IR or PTX
+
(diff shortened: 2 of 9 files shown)


python/quadrants/lang/misc.py
```diff
@@ -452,6 +454,26 @@ def init(
...
+    # files are never read. The kernel still runs, it just is not the code that was edited, which is silent and easy to
+    # lose hours to. Refuse the combination rather than warn.
+    active_load_envs = []
+    # codegen_llvm.cpp reads QD_LOAD_IR through get_environ_config, which parses the value as an int, so "0" is off.
+    if os.getenv("QD_LOAD_IR", "0") not in ("", "0"):
+        active_load_envs.append("QD_LOAD_IR")
+    # jit_cuda.cpp only checks that QUADRANTS_LOAD_PTX is present, so any value at all enables the PTX load path.
+    if os.getenv("QUADRANTS_LOAD_PTX") is not None:
+        active_load_envs.append("QUADRANTS_LOAD_PTX")
+    if active_load_envs and (cfg.offline_cache or src_ll_cache):
```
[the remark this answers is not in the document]
hughperkins: Agent said:

Good catch, fixed in f0eeee4a1.

Confirmed the scoping: `QD_LOAD_IR` is read only in `codegen_llvm.cpp` (so Vulkan, Metal and the python backend never see it) and `QUADRANTS_LOAD_PTX` only in `jit_cuda.cpp`. And you were right that the check could not see the arch where it sat - `cfg.arch` is not assigned until `adaptive_arch_select`, well below it, which can also fall back to a different arch than requested.

The check now runs after arch selection, gated on `arch_uses_llvm(cfg.arch)` for `QD_LOAD_IR` and `cfg.arch == cuda` for `QUADRANTS_LOAD_PTX`. Tests cover both non-applicable cases. This mattered more than a normal false positive because the diagnostic is a hard error rather than a warning.

python/quadrants/lang/misc.py
```diff
@@ -452,6 +454,26 @@ def init(
...
+        names = " and ".join(active_load_envs)
+        raise ValueError(
+            f"Caching must be disabled when using {names}: replacement IR/PTX is read from debug_dump_path only for a "
+            "kernel that is actually compiled, and cached kernels are returned without running codegen, so the files "
+            f"would be silently ignored. Pass offline_cache=False and src_ll_cache=False to qd.init, or unset {names}."
+        )
```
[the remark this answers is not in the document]
hughperkins: Agent said:

Agreed on the AGENTS.md requirement, docs added in f0eeee4a1.

One correction to the reasoning: users could not have been learning this from the user guide, because neither `QD_LOAD_IR` nor `QUADRANTS_LOAD_PTX` appeared anywhere under `docs/` at all. That makes the case stronger rather than weaker, so rather than only documenting the new error I documented the workflow itself.

`docs/source/user_guide/optimization_passes.md` now has a "Running hand-edited IR or PTX" subsection next to the existing `QD_DUMP_IR` / `QD_DUMP_SIMPLIFY` / `QD_DUMP_CFG` list, covering what each variable reads, which backends read it, the requirement to disable both caches, and the fact that `QUADRANTS_LOAD_PTX` is presence-only so `0` still enables it.

docs/source/user_guide/debug.md
```diff
@@ -38,7 +38,9 @@ a = x[-1]     # AssertionError in debug mode
 
 #### Adstack overflow
 
-The adstack overflow check on reverse-mode autodiff runs always, on every backend, regardless of `debug`. A push past the per-stack capacity raises `QuadrantsAssertionError("[Aa]dstack overflow")` at the next Quadrants Python entry that polls the overflow flag after the offending kernel has executed - kernel launch, host-side field / ndarray read, or `qd.sync()`. On CPU this is the entry of the offending launch itself; on GPU it can be one or more entries later, since the GPU may not have run the offending push by the time the poll at end-of-launch fires. The error message describes the cause (untracked tensor mutation between launches, or sizer under-estimate caused by a bug in Quadrants) and the recovery flow; see [Autodiff -> What can go wrong](autodiff.md) for the full description.
+The adstack is the per-thread stack that reverse-mode autodiff uses to save the intermediate values a kernel produces, so that the backward pass can read them back. Overflowing it is checked always, on every backend, regardless of `debug`, and raises `QuadrantsAssertionError("[Aa]dstack overflow")`.
```
hughperkins: Note: I believe this was rewritten because it was failing doc quality check CI job.

python/quadrants/lang/misc.py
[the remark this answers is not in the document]
hughperkins: Agent said:

Update on where this landed, since the above no longer describes the PR.

The docs moved from `optimization_passes.md` to `docs/source/user_guide/debug.md`, under a new "Under the hood: compiler and runtime internals" heading, directly after the existing `QD_DUMP_IR` subsection that produces the files being edited. That is a better fit: a reader who just dumped IR is already on that page, and the dump example there already had to disable the offline cache.

The presence-only note is also gone, because the asymmetry it described is fixed rather than documented. `jit_cuda.cpp` was the only site reading these variables with a bare `getenv` presence check, so it now goes through the shared `get_environ_config` helper like `codegen_llvm.cpp` already did for `QD_DUMP_IR` and `QD_LOAD_IR`. `QUADRANTS_LOAD_PTX=0` is therefore off, matching `QD_LOAD_IR=0`, and the check treats them identically.

That also fixed a second inconsistency in the same function: `QD_DUMP_IR` was compared against `"1"` for the `.ll` dump but only tested for presence for the `.ptx` dump, so `QD_DUMP_IR=0` wrote PTX anyway.

hughperkins:
@codex review

docs/source/user_guide/contributing.md
```diff
@@ -228,6 +228,21 @@ Quadrants comprises at least three important parts:
...
 3. `LLVM` libraries used by host runtime: statically or dynamically linked, used to lower the kernel's final IR to machine code on the host. The CI uses an LLVM version compiled from source.
 
+### Running hand-edited IR or PTX
+
+A dump written by `QD_DUMP_IR` (see [Dumping compiled IR](./debug.md#dumping-compiled-ir)) can be edited and fed back in, which is a quick way to try an instruction sequence by hand before changing codegen. `QD_LOAD_IR=1` makes the LLVM backends reparse `<debug_dump_path>/<task>_llvm.ll` instead of using the module they just built, and `QUADRANTS_LOAD_PTX=1` makes the CUDA backend read `<debug_dump_path>/<name>.ptx` instead of the PTX it just compiled. Both are read as integers, so `0` leaves them off.
+
+Both caches have to be off in the script under test, since the replacement files are only read while a kernel is being compiled and a cached kernel is handed back without compiling. `qd.init` raises a `ValueError` rather than let an already-cached kernel keep silently running its old code. `offline_cache` has an environment variable, `src_ll_cache` does not, so the latter has to be passed to `qd.init`:
```
duburcqa, line 235: This behaviour is not consistent with how other environment variables are handled. User-specified values that are incompatible are simply ignored with a warning.

hughperkins: - for the existing behavior, give me an example?
- in this case, I would prefer it to simply fail out than warn me. I'm explicilty using the env var to request a specific action, and if it's going to not do that thing, then there is no point in it running further I feel

duburcqa: > for the existing behavior, give me an example?

Every single qd init option supports this, see [official documentation](https://genesis-embodied-ai.github.io/quadrants/user_guide/init_options.html#qd-init-options), eg `QD_OFFLINE_CACHE` or `QD_ARCH`.

```python
QD_ARCH=metal ipython
Python 3.12.8 (main, Apr  3 2025, 16:45:27) [Clang 17.0.0 (clang-1700.0.13.3)]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.30.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import quadrants as qd
[Quadrants] version 1.1.4, llvm 22.1.0, commit 9578f164, osx, python 3.12.8

In [2]: qd.init(arch=qd.cpu)
[I 08/11/26 15:48:19.591 80043228] [misc.py:init@503] Following QD_ARCH setting up for arch=metal
[Quadrants] Starting on arch=metal
```

> in this case, I would prefer it to simply fail out than warn me. I'm explicilty using the env var to request a specific action, and if it's going to not do that thing, then there is no point in it running further I feel

I'm not ok with introducing inconsistency. Either change the behaviour everywhere or nowhere.

hughperkins: your example shows the env var being honored. You request metal. You get metal.

Do you have an example of an env var being not honored?

duburcqa: > your example shows the env var being honored. You request metal. You get metal.

Exactly. That is my point. It does not raise any exception and env vars are honoured.

hughperkins: Ok. But in this case we can NOT honor the env var, when cache is enabled.

duburcqa: Exactly. Which means that it should automatically disable cache.

hughperkins: This is primarily used for dev purposes. Having to make any env vars that are primarily used for dev purposes have user-facing semantics, ie automated toggling of cache, introduces a fairly high dev burden I feel, with limited reward; and very likely complicates the codebase in undesirable ways.

duburcqa: This does not change my opinion. You could say the same for all env vars. All of them exist primarily for dev purposes.

hughperkins: for example QD_ARCH is used by quadrants users, not quadrants internal developers.

duburcqa: I don't think it is, but anyway, you can take QD_FAST_MATH or QD_DEBUG as an example if you prefer.

hughperkins: QD_DEBUG is used by quadrants users. For example, I believe that Yiling used to use it for debugging Genesis.

I've never used QD_FAST_MATH, and I'm a quadrants internal developer. It sounds to me like a user facing option, to enable/disable fast maths.

QD_LOAD_IR is ONLY targeted at quadratns internal developers. There is no obvious reason why a quadrants user would want to load IR from disk, instead of running the actual compiled quadrants kernel I feel. It is only used for debugging quadrants itself, not quadrants kernels.

This is why it is documented in contributing.md, not in debug.md.

duburcqa: `QD_DEBUG` is not supposed to be used by users. If a user of quadrants  or genesis-world wants to enable debug, then just pass it as input argument of the corresponding init. Env var are always for dev / debugging purpose by definition. This is not sufficient to introduce inconsistency. You can raise a warning if the specified option cannot be honoured. This is what is done for other env var overwriting python-side options.

duburcqa, line 235: `src_ll_cache` not having an env variable is problematic as it makes it impossible to use this feature in Genesis without having to edit `gs.init`.

hughperkins: Can I make a ticket for that as something to address in the future?

This PR itself does not actually directly achieve anything I need. It was just to placate agent review in https://github.com/Genesis-Embodied-AI/quadrants/pull/850, ie this codex comment:
- https://github.com/Genesis-Embodied-AI/quadrants/pull/850#discussion_r3738127499

And that PR was just intended to be a quick generalization and simplification of the PR at https://github.com/Genesis-Embodied-AI/quadrants/pull/849#issuecomment-5220677940

Definitely getting Hal changing a lightbulb vibes here 😓 https://www.youtube.com/watch?v=8fnfeuoh4s8

duburcqa: > Can I make a ticket for that as something to address in the future?

Let's do this

duburcqa:
ok to merge

EDIT: Ignore this. It does not apply anymore.

docs/source/user_guide/contributing.md
hughperkins: Had a stab at turning off caching automatically, but tons of knock-on effects. *much* would prefer to simply fail for unsupported combos, rather than tryign to automtically deicde what the user really intended pelase.

Screenshot 2026-08-14 at 09 21 34

As a general rule, I find that 'magic' in programs tends to be brittle, and hard to understand. Vastly prefer just failing wht a clear message, on teh whole.

hughperkins: created cmp-349 to address the src_ll_cache not having env var.

hughperkins: ceated cmp-350 to address removing magic from quadrants env vars

duburcqa:
I'm no longer ok to merge.

duburcqa:
ok to merge

PR #868, opened 2026-08-14
[AMDGPU] Fix f32 pow precision (compute in f64) to match CUDA/CPU

paveltc:
## Summary

Reworked per @hughperkins's review: rather than loosening `test_binary_f` on AMDGPU, this **tightens AMDGPU's f32 `pow` precision** so tolerances stay uniform across all backends.

AMDGPU's `__ocml_pow_f32` computes `exp2(z * log2(y))`, so the `log2` rounding error is scaled by the exponent (~1e-3 relative error), which exceeds the default `1e-6` tolerance. This change computes f32 `pow` in **double precision** (`__ocml_pow_f64`) and rounds the result back to f32 — effectively correctly-rounded, matching CPU/CUDA within the default tolerance. It mirrors the existing i32 `pow` path a few lines below in the same function.

The earlier arch-gated test-tolerance change has been **reverted**; `test_binary_f` now uses the uniform default tolerance on every backend (net test diff vs `main` is zero).

## Tradeoff

f64 `pow` is slower than a native f32 `pow`, but `pow` is rarely perf-critical and gfx942 has strong FP64 throughput. There is no correctly-rounded f32 `pow` variant or OCML config flag that would fix the accuracy without the f64 path.

## Tests — AMD Instinct MI308X (`gfx942`, ROCm 7.2.4), at the default `1e-6` tolerance

- **amdgpu:** `test_element_wise.py::test_binary_f` 3/3 passed; `test_pow.py` 11 passed / 6 pre-existing xfail
- **cpu:** `test_element_wise.py::test_binary_f` 3/3 passed (unchanged)

The two commits can be squashed on merge; the net change is a single-hunk codegen fix.

> Note: the earlier `runtime.cpp` `__builtin_trap` assert change remains split out of this PR (it needs a host-side fault-translation path before it can land), as discussed above.


paveltc:
Thanks @chatgpt-codex-connector — both points were valid. I've split this PR to address them:

**P2 (relaxed tolerance leaking to other backends):** fixed. The tolerance is now gated on the active arch, so only AMDGPU uses `rel=1e-3` and CPU/CUDA/Metal keep their tighter defaults:

```python
pow_kwargs = {"rel": 1e-3} if qd.lang.impl.current_cfg().arch == qd.amdgpu else {}
assert test_utils.allclose(x[6], y**z, **pow_kwargs)
```

(same `current_cfg().arch` pattern already used in e.g. `test_abs.py`). Re-validated on MI308X: `test_binary_f` passes on `amdgpu` (3/3) and still passes on `cpu` (3/3) under the default tolerance; reverting to the default tolerance on `amdgpu` fails (1/3), confirming the relaxation is both necessary and correctly scoped.

**P1 (assert contract broken by the trap):** correct — `__builtin_trap()` faults the whole dispatch, so `check_runtime_error()`'s first `synchronize()` returns `hipErrorLaunchFailure` and `AMDGPUFunction::operator()` throws before `runtime_retrieve_and_reset_error_code` runs, losing the `QuadrantsAssertionError` and formatted message. Rather than ship a regression, I've **removed the `runtime.cpp` change from this PR**. This PR is now scoped to the pow-tolerance test fix only. The assert-hang fix will come as a separate PR once it has a host-side path that translates the fault back into the assertion contract.

tests/python/test_element_wise.py
```diff
@@ -61,7 +61,11 @@ def func():
     assert test_utils.allclose(x[3], y / z)
     assert test_utils.allclose(x[4], y // z)
     assert test_utils.allclose(x[5], y % z)
-    assert test_utils.allclose(x[6], y**z)
+    # AMDGPU __ocml_pow_f32 uses log2->mul->exp2, giving ~0.06% relative error
+    # vs x86 pow; loosen the tolerance for AMDGPU only so other backends keep
+    # their default (tighter) tolerance.
+    pow_kwargs = {"rel": 1e-3} if qd.lang.impl.current_cfg().arch == qd.amdgpu else {}
```
hughperkins: Not ok with this. Tolerance should be ~similar on all platforms I feel.

The correct fix is to tighten the AMDGPU precision I feel, so it is ~similar to CUDA.

paveltc: Agreed — reworked to tighten the AMDGPU precision instead of loosening the test.

AMDGPU's f32 `pow` now computes in double precision (`__ocml_pow_f64`) and rounds back to f32, so it's effectively correctly-rounded and matches CPU/CUDA within the default `1e-6` tolerance. I reverted the arch-gated tolerance change, so `test_binary_f` uses the uniform default tolerance on every backend (net test diff vs `main` is zero).

Validated on MI308X (`gfx942`, ROCm 7.2.4) at the default tolerance: `test_binary_f` 3/3 on both `amdgpu` and `cpu`, and `test_pow.py` 11 passed / 6 pre-existing xfail on `amdgpu`. Pushed as `b7f047dad`.

hughperkins:
@codex review

hughperkins:
Opus seems ok with this PR 🎉

quadrants/codegen/amdgpu/codegen_amdgpu.cpp
```diff
@@ -496,7 +496,15 @@ class TaskCodeGenAMDGPU : public TaskCodeGenLLVM {
       if (ret_quadrants_type->is_primitive(PrimitiveTypeID::f16)) {
         llvm_val[stmt] = call("__ocml_pow_f16", {lhs, rhs});
       } else if (ret_quadrants_type->is_primitive(PrimitiveTypeID::f32)) {
-        llvm_val[stmt] = call("__ocml_pow_f32", {lhs, rhs});
+        // __ocml_pow_f32 is exp2(z * log2(y)); the log2 rounding error is
```
hughperkins: long comment, but all seems necessary. I'll just get agent to erwap to 120c.

hughperkins:
once agent has erwarped the comment, I'llj ust wait for the amd ci job to pass, then merge

hughperkins:
Thanks!

PR #870, opened 2026-08-14
[AMDGPU] Surface QuadrantsAssertionError after in-kernel assert (fix barrier hang)

paveltc:
## Summary

On AMDGPU, a failed in-kernel `assert` previously emitted `asm("S_ENDPGM")`, which only terminates the *faulting wavefront*. Peer wavefronts still waiting on `s_barrier` then deadlock, and the host hangs forever in `hipStreamSynchronize`. This PR replaces that with a dispatch-wide `__builtin_trap()` and translates the resulting fault back into a proper `QuadrantsAssertionError` on the host, preserving the debug-mode assertion contract without hanging.

CUDA / CPU / Metal paths are unchanged. All new behavior is gated on `debug` + `Arch::amdgpu`.

## Approach

`__builtin_trap()` faults the whole dispatch, so the host gets `hipErrorLaunchFailure` (719) rather than a hang — but the context is then dead, so the usual device-side error-retrieval kernels can no longer run. To preserve the error message:

1. **Pinned host-coherent mirror.** In `materialize_runtime` (debug + amdgpu only) we `hipHostMalloc(...Coherent)` an `AmdgpuAssertErrorState` and publish its device-mapped address into the runtime. This mirrors the existing `adstack_overflow_flag_dev_ptr` precedent and survives a device fault.
2. **Device publish before trap.** In `quadrants_assert_format`, the faulting wavefront (serialized under the existing `error_message_lock`) copies the message template + arguments into the pinned buffer, issues a system-scope fence (`amdgpu_system_mem_fence`, patched to an LLVM seq_cst fence in `llvm_context.cpp`), stores `error_code` last, then `__builtin_trap()`s.
3. **Host translation.** `AMDGPUFunction::operator()` intercepts `hipErrorLaunchFailure`, and a debug-only hook reads the pinned state and raises `QuadrantsAssertionError` (a subclass of `AssertionError`). Subsequent 719s on the now-dead context are ignored so `Program` teardown does not `terminate()` from a destructor.

## Testing

Validated on an AMD Instinct MI308X (gfx942), ROCm 7.2.4, base `upstream/main`:

- `tests/python/test_assert.py::test_amdgpu_assert_raises` — a failed assert raises `QuadrantsAssertionError` with the formatted message; `isinstance(e, AssertionError)` holds.
- `tests/python/test_assert.py::test_amdgpu_assert_barrier_no_hang` — one thread asserts while siblings hit `block.sync()`; raises instead of hanging (the original bug).
- CPU sanity (`test_assert_*`) unchanged.

Both new tests run each case in an isolated child subprocess (the HIP context is dead after a trap; HIP is unsafe after `fork`) with a wall-clock timeout that fails on the barrier-hang regression.

## CI notes

Upstream AMDGPU CI (`test_gpu.yml` → `test_linux_amdgpu`, `runs-on: amdgpu`) runs **bare-metal** on the self-hosted runner — no container — which matches the environment where the trap returns a catchable `hipErrorLaunchFailure`. Some ROCm/HSA configs (notably inside Docker) instead escalate the trap to an uncatchable `SIGABRT`; the tests treat a `SIGABRT`-killed child as `pytest.skip` (environment limitation) while still failing on timeout or on a wrong/absent exception, so no runner goes spuriously red.

## Known limitations / possible follow-ups

- After an assert, the HIP context is dead — one assert per process (tests isolate per subprocess). Accepted debug-mode limitation.
- The pinned struct is duplicated as a layout-compatible host view (`AmdgpuAssertErrorStateHostView`) in `llvm_runtime_executor.cpp`; a `static_assert` on size/offsets would harden this.
- The launch-failure hook and surfaced-flag are process-global singletons (assume a single active `Program`).

### DIFF
--- a/quadrants/rhi/amdgpu/amdgpu_driver.cpp
+++ b/quadrants/rhi/amdgpu/amdgpu_driver.cpp
@@ -6,6 +6,32 @@
 
 namespace quadrants {
 namespace lang {
+namespace {
+
+AmdgpuLaunchFailureHook g_amdgpu_launch_failure_hook = nullptr;
+bool g_amdgpu_device_assert_already_surfaced = false;
+
+}  // namespace
+
+void set_amdgpu_launch_failure_hook(AmdgpuLaunchFailureHook hook) {
+  g_amdgpu_launch_failure_hook = hook;
+}
+
+AmdgpuLaunchFailureHook get_amdgpu_launch_failure_hook() {
+  return g_amdgpu_launch_failure_hook;
+}
+
+bool amdgpu_device_assert_already_surfaced() {
+  return g_amdgpu_device_assert_already_surfaced;
+}
+
+void amdgpu_reset_device_assert_surfaced_flag() {
+  g_amdgpu_device_assert_already_surfaced = false;
+}
+
+void amdgpu_mark_device_assert_surfaced() {
+  g_amdgpu_device_assert_already_surfaced = true;
+}
 
 std::string get_amdgpu_error_message(uint32 err) {
   auto err_name_ptr = AMDGPUDriver::get_instance_without_context().get_error_name(err);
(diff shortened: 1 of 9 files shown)


paveltc:
Superseded by #871, which raises the same change from the AMD-Ecosystem organization fork per policy (topic branch on an AMD org fork → upstream). Closing this one to avoid duplicates.

Issue #856, opened 2026-08-10
Support optional non-template kernel arguments that specialize away the absent code path


duburcqa:
**Concisely describe the proposed feature**

Quadrants should support optional non-template kernel arguments spelled `T | None`, defaulting to `None` when unset, and should specialize the compiled kernel on presence versus absence so that the unused code path is never traced. Concretely, this should be valid and supported:

```python
@qd.kernel
def apply_at(out: qd.types.NDArray[qd.f32, 2], point: qd.Tensor | None = None):
    for i in range(out.shape[0]):
        if qd.static(point is not None):
            ...  # use point[i]
        else:
            ...  # use the default anchor
```

The motivating case is the application point of a force applied to a geometry. The point can be left unspecified (the center of mass is used), given as an absolute position, or given as an offset relative to the center of mass. Each variant wants a different arithmetic path, and the unspecified variant wants no arithmetic at all rather than adding zero or multiplying by one. Today the choice is between duplicating the kernel once per variant, which multiplies the maintenance burden and lets the copies drift, or keeping a single kernel that pays for arithmetic it does not need in the common case.

The real-world instance of this is Genesis-Embodied-AI/genesis-world#3143, which adds an application point to the external force API and hits exactly this trade-off in the kernel behind it.

**What works today**

The underlying mechanism already exists, but it is not spellable the way the code above reads, and it is not covered by any test.

A `qd.template()` slot accepts `None` and specializes on it, since a template argument is part of the specialization key by construction. That is not a usable answer here: the same slot rejects an ndarray outright with `Ndarray shouldn't be passed in via 'qd.template()'` (`python/quadrants/lang/_template_mapper_hotpath.py:268`), so an argument that is a tensor when present cannot go through it. The only route that accepts both forms today is `qd.Tensor`.

`qd.Tensor` is a value-dispatched argument slot: `_extract_arg` routes ndarray-shaped values through the ndarray feature path and everything else through the template path (`python/quadrants/lang/_template_mapper_hotpath.py:237`), so a `None` argument lands in the specialization key. `FunctionDefTransformer._decl_and_create_variable` then binds the kernel-scope name to the literal Python `None` (`python/quadrants/lang/ast/ast_transformers/function_def_transformer.py:82`), and `qd.static` folds the branch at AST build time.

So the following runs correctly on CPU at `47e9dfb5`:

```python
import numpy as np

import quadrants as qd

qd.init(arch=qd.cpu)


@qd.kernel
[...]

PR #877, opened 2026-08-20
[Lang] W1 Cache spec keys for None kernel arguments

hughperkins:
## Summary

A kernel argument of `None` (an absent optional `qd.Tensor` / `qd.template()` slot) is ~21% slower per launch than it should be, because its template-mapper spec-key cache entry is never stored.

`TemplateMapper.lookup` weakref-tracks every argument whose type is not in `_primitive_types`, so it can evict the spec-key cache entry when the argument is garbage-collected. `weakref.ref(None)` raises `TypeError`, which was caught and downgraded to a `warn_once`, leaving the entry unstored - so every launch passing `None` re-ran full spec-key extraction instead of hitting the cache.

This adds `NoneType` to `_primitive_types`. `None` is an immortal singleton that never needs lifetime tracking, so excluding it from the weakref loop is both correct and removes the overhead.

## Measurements

CPU, 20000 iterations, microseconds per launch, optional `qd.Tensor` slot:

| variant | before | after |
| --- | --- | --- |
| `b=None` | 148.6 | 122.8 |
| `b=ndarray` (present) | 123.9 | 123.9 |
| no optional slot (control) | 122.2 | 122.2 |

The absent branch does strictly less work than the present branch yet was slower than the control; the gap closes from +20.8% to ~0.3%. Specialization counts and results are unchanged. The spec-key cache now populates for a `None` argument (0 -> 1 entries).

## Test plan

- [x] New `test_none_argument_populates_spec_key_cache` passes on x64 and cuda
- [x] Full CPU suite via the CI harness (`tests/run_tests.py -r 3`, x64, split by `needs_torch` exactly as `linux/4_test.sh`): this branch's unique failing set is **identical to unmodified `main`** and the `needs_torch` phase passes clean, so this PR introduces **no new failures**.

The only shared failures are three tests that cannot pass in the offline cluster container regardless of branch (they fail the same way on `main`), all unrelated to this change:
  - `test_committed_fatbin_layout` - the container's `cuobjdump` predates CUDA 13 and misreads the **committed** cubin's build toolkit as `0.2` instead of `13.0+`;
  - `test_pyi_stubs` - `python -m pyright` cannot fetch its Node runtime on an offline compute node;
  - `test_ipython.ipynb` - `io.UnsupportedOperation` from quadrants' stdout wrapping under nbmake's ipykernel capture.

(An earlier revision of this checklist reported raw `pytest` counts; those were inflated because that run omitted CI's `-r 3` flaky-rerun, so load-induced flakes and xdist worker-kill collateral surfaced as hard failures. Under the CI harness they do not.)

Context: first of two prerequisite fixes for optional (`None`) kernel arguments, motivated by #856.

### DIFF
--- a/python/quadrants/lang/_template_mapper_hotpath.py
+++ b/python/quadrants/lang/_template_mapper_hotpath.py
@@ -73,7 +73,11 @@ AnnotationType = Union[
 
 _ExprCxx = _qd_core.ExprCxx
 _composite_mutable_types = {list, dict, set}
-_primitive_types = {int, float, bool}
+# ``type(None)`` belongs here so ``None`` arguments are treated like other primitives and excluded from the
+# weakref-based spec-key cache tracker in ``TemplateMapper.lookup``: ``weakref.ref(None)`` raises ``TypeError``, and
+# ``None`` is an immortal singleton that never needs lifetime tracking. Without it, the spec-key cache entry for a
+# ``None`` argument is dropped and every such launch re-runs full spec-key extraction.
+_primitive_types = {int, float, bool, type(None)}
 
 
 # Per-instance ndarray-path cache, stored OFF-instance in a module-level ``id(arg) -> list[paths]`` dict and cleaned
--- a/tests/python/test_tensor_annotation.py
+++ b/tests/python/test_tensor_annotation.py
@@ -244,6 +244,42 @@ def test_tensor_slot_later_wrapper_uses_impl_identity_cache():
     assert len(copy._primal.mapper._mapping_cache) == cache_size
 
 
+@test_utils.test()
+def test_none_argument_populates_spec_key_cache():
+    """A ``None`` in a ``qd.Tensor`` slot must land in the template mapper's spec-key cache.
+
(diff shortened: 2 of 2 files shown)


hughperkins:
@codex review

hughperkins:
No user-facing doc changes required.

hughperkins:
Running Genesis bench and unit tests

hughperkins:
some varaince on accessors perhaps. I'll run with before/after script

20260821_W1_0836

hughperkins:
Genesis unit tests passing ok:

Screenshot 2026-08-21 at 09 23 04

hughperkins:
Before/after results:

```
=== before/after interleaved (5 runs each, single node) ===
before qd=52df3b168  after qd=ee5766ba1  genesis=d00c93b3

anymal_random           20000  cuda                 4      6970628.2   0.88      6960386.8   1.62    -0.15      1.61
g1_fall_accessors        4096  cuda                 4      1037617.2   2.67      1041105.0   4.21     0.34      4.04
[06:35:34] Wrote /home/hugh/git/tmp/20260821_W1_0853/meta.yaml
[06:35:34] Done. NO significant before/after difference
```

hughperkins:
Seesm ok => lets merge.

Issue #831, opened 2026-07-31
Make the quadrants API type-checkable: annotation spellings, kernel decorator overloads, element-typed containers


duburcqa:
**Concisely describe the proposed feature**

Quadrants cannot currently be type checked by a downstream project, for several independent reasons. This issue covers all of them; the annotation spellings below are the largest and the one that has to be fixed first, and a second set of blockers in kernel bodies, decorators and call sites is described further down.

Kernel parameter annotations that are built by *calling* a factory - `qd.types.vector(3)`, `qd.types.matrix(3, 3)` - are not valid Python type expressions. A call is not part of the type-expression grammar, so it is rejected regardless of what the callee returns, and the rejection is not tool-specific: mypy 1.15 reports it as `[valid-type]`, pyright 1.1.407 as `reportInvalidTypeForm`, ty 0.0.65 as `invalid-type-form`. Any downstream project that wants to run a type checker over code containing quadrants kernels therefore cannot, and this blocks Genesis-Embodied-AI/Genesis#1271.

Quadrants already solves this for two of its three annotation families. Measuring on Genesis at `54b9ca2e5`, `ty check genesis` reports 5376 diagnostics, of which 1716 are `invalid-type-form` across 54 files. Grouped by the exact annotation expression:

| annotation form | sites | type-expression-valid spelling available today |
| --- | --- | --- |
| `qd.types.ndarray()` | 813 | yes - `qd.types.NDArray` |
| `qd.template()` | 588 | yes - `qd.Template` |
| `qd.types.ndarray(ndim=N)` and other keyword variants | 53 | yes - `qd.types.NDArray[None, N]` |
| `qd.types.vector(3)`, `qd.types.vector(4)` | 257 | **no** |

So 1454 of the 1716 are already fixable downstream today, and the remaining 257 are blocked here.

The reason the first three work is that `qd.types.ndarray` and `qd.template` are *aliases for classes* (`NdarrayType`, `Template`), so the bare alias is a plain class reference - a valid type expression - and `NdarrayType` additionally defines `__class_getitem__`, so the subscripted form carries the parameters through. By contrast `qd.types.vector` and `qd.types.matrix` are factory functions returning `VectorType` / `MatrixType` instances; neither class is exported from `quadrants.types` nor subscriptable, so for those two there is no valid spelling at all.

[...]

PR #878, opened 2026-08-20
[Lang] W3 Accept numpy/torch external arrays at qd.Tensor slots

hughperkins:
## Summary

- **W3** of the nullable-tensor / [#856](https://github.com/Genesis-Embodied-AI/quadrants/issues/856) plan: a `qd.Tensor` kernel-argument slot now accepts **external arrays** (`numpy.ndarray`, `torch.Tensor`), routed through the same ndarray feature path as a bare `qd.ndarray` and keyed by dtype/rank. Previously torch produced one specialization *per tensor instance* and numpy raised `TypeError: unhashable type: 'numpy.ndarray'`.
- The predicate is a **positive numpy/torch allowlist**, not a duck-typed `.shape`/`.dtype` check: those are exactly the external types the launch path (`FuncBase._recursive_set_args`) can bind, and a duck-typed check would wrongly divert quadrants-native `Field`/`SNode` (which also expose `.shape`/`.dtype`) off the template path.
- Two call sites change: spec-key extraction (`_template_mapper_hotpath._extract_arg`) and the launch path (`_func_base`). `tensor.md` documents the accepted types.

This matches the set of external arrays a `qd.types.ndarray()` slot already accepts, so `qd.Tensor` and `qd.types.NDArray` now support the same external types end-to-end. Independent of the `None` half of #856 (W1/W2/W4/W5) and of #831.

## Test plan

- [x] New tests in `test_tensor_annotation.py`: numpy and torch each collapse to a single specialization (not one per instance); numpy no longer raises `unhashable`; a `Field` through the same `qd.Tensor` slot still takes the template path.
- [x] `test_tensor_annotation.py` full file: **30 passed** on cpu, from a clean branch checkout.
- [x] Full CPU suite (`run_tests.py --arch cpu`): **4340 passed**, 4 failures that are **pre-existing** (verified identical on clean `main`: 3x `test_committed_fatbin_layout` under a CUDA-OFF build, 1x `test_pyi_stubs`) - **0 regressions from this change**.

Built CPU-only (`QD_WITH_CUDA=OFF`) on the cluster; W3 is backend-agnostic Python dispatch. A perf-tuned fast path for the predicate on the hot `qd.Tensor` template branch can be layered later if profiling warrants.


docs/source/user_guide/tensor.md
```diff
@@ -207,6 +207,8 @@ The kernel argument is unwrapped to the bare impl before the template-mapper / A
 
 For a parameter annotated `qd.Tensor`, calls to the same kernel may alternate among a wrapper, its bare field or ndarray implementation, and `None`. If callers can pass `None`, guard tensor operations with [`qd.static()`](static.md), for example `if qd.static(x is not None):`; the guarded operations are not traced for the `None` specialization. The parameter remains annotated `qd.Tensor`, and `None` must be passed explicitly.
 
+A `qd.Tensor` parameter also accepts **external arrays** - `numpy.ndarray` and `torch.Tensor` - dispatched through the same ndarray path as a bare `qd.ndarray`, i.e. specialized by dtype and rank rather than by object identity. Distinct arrays of the same dtype and rank therefore reuse one compiled kernel. This is the same set of external array types a `qd.types.ndarray()` slot accepts. Other Python objects that merely expose `.shape` / `.dtype` (a `qd.field`, for instance) are not treated as external arrays and keep following the template path.
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Good catch - fixed in 33359ecd8. The doc now states that reuse of a compiled specialization requires matching dtype, rank *and* gradient state, and explicitly calls out that a torch tensor's `requires_grad` is folded into the specialization key (so two tensors differing only in `requires_grad` get separate specializations).

python/quadrants/lang/_template_mapper_hotpath.py
```diff
@@ -239,7 +262,7 @@ def _extract_arg(raise_on_templated_floats: bool, arg: Any, annotation: Annotati
         if type(arg) in _TENSOR_WRAPPER_TYPES:
             arg = arg._unwrap()
         arg_type = type(arg)
-        if issubclass(arg_type, (Ndarray, AnyArray)):
+        if issubclass(arg_type, (Ndarray, AnyArray)) or _is_external_array(arg):
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Verified empirically on CPU (quadrants `33359ecd8`). This is a pre-existing property of the shared `TemplateMapper.lookup()` identity cache (`args_hash = tuple(id(arg) for arg in args)`), not something this PR introduces - a `qd.types.ndarray()` slot behaves identically.

Probe: launch with a float32 numpy array, then reinterpret the *same* object in place as int32 (same itemsize/shape/id) and launch again; then a fresh-id int32 array as a control:

```
[qd.Tensor]           specs 1 -> 1 (stale), out stays [1 2 3 4];  fresh-id int32 -> specs 2 (recompiles)
[qd.types.ndarray()]  specs 1 -> 1 (stale), out stays [1 2 3 4];  fresh-id int32 -> specs 2
IDENTICAL = True
```

So widening `qd.Tensor` to external arrays inherits exactly the caching semantics `qd.types.ndarray()` already has. (And before this PR, numpy at a `qd.Tensor` slot raised `TypeError: unhashable type: 'numpy.ndarray'`, so there is no regression - only strictly more works now.)

Re-keying mutable external arrays or bypassing the identity cache would have to change the shared, perf-critical `lookup()` path and would affect the existing ndarray path equally; scoping it to only external arrays at `qd.Tensor` slots would re-introduce the exact `qd.Tensor`-vs-`qd.types.NDArray` inconsistency this PR is removing. Better tracked as a separate issue against `lookup()`. Not changing it in this PR.

hughperkins:
@codex review

python/quadrants/lang/_template_mapper_hotpath.py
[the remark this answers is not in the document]
hughperkins: Agent says:

Verified on CPU (quadrants `33359ecd8`). This is a pre-existing, global property of `stringify_obj_type()`, not introduced by this PR: it omits gradient state for *every* array repr, and the string is annotation-independent (a torch tensor at a `qd.types.ndarray()` slot hashes identically to one at a `qd.Tensor` slot). Direct calls into `stringify_obj_type`:

```
[qd.Tensor slot        ] requires_grad=False -> '[pt-torch.float32-1]'
[qd.Tensor slot        ] requires_grad=True  -> '[pt-torch.float32-1]'   collide=True
[qd.types.NDArray slot ] requires_grad=False -> '[pt-torch.float32-1]'
[qd.types.NDArray slot ] requires_grad=True  -> '[pt-torch.float32-1]'   collide=True
ndarray needs_grad=False -> '[nd-f32-1]'
ndarray needs_grad=True  -> '[nd-f32-1]'   collide=True
```

So the same key collision already exists for the pre-existing `qd.types.NDArray` torch path and even for quadrants-native `qd.ndarray` (`[nd-...]`). A correct fix therefore belongs in the shared hasher and must cover all array reprs (`pt`/`nd`/`ndv`/`ndm`) while mirroring `_extract_arg`'s exact `needs_grad` derivation - and it changes the persistent key format for all array-taking kernels, invalidating existing offline caches. That is out of scope for this consistency PR (which only widens `qd.Tensor` to accept the external-array types `qd.types.NDArray` already accepts, and does not change caching behavior relative to that path).

Tracked separately as #882. Not changing it in this PR.

hughperkins:
@codex review

hughperkins:
Will ignore doc quality chekcs for tensor.md, since handled separately by https://github.com/Genesis-Embodied-AI/quadrants/pull/881 (unelss that pr merges first).

hughperkins:
Running genesis benchamrks and unit tests

hughperkins:
Genesis bnechmarks look okish:

20260821_W3_1042

hughperkins:
Agent addressing doc quality

hughperkins:
Genesis unit tests passing:

Screenshot 2026-08-21 at 16 08 54

docs/source/user_guide/tensor.md
```diff
@@ -207,6 +207,8 @@ Inside the kernel the argument is the bare impl rather than the wrapper, so kern
 
 For a parameter annotated `qd.Tensor`, calls to the same kernel may alternate among a wrapper, its bare field or ndarray implementation, and `None`. If callers can pass `None`, guard tensor operations with [`qd.static()`](static.md), for example `if qd.static(x is not None):`; the guarded operations are not traced for the `None` specialization. The parameter remains annotated `qd.Tensor`, and `None` must be passed explicitly.
 
+A `qd.Tensor` parameter also accepts **external arrays** - `numpy.ndarray` and `torch.Tensor` - dispatched through the same ndarray path as a bare `qd.ndarray`, i.e. specialized by dtype, rank and gradient state rather than by object identity. Distinct arrays that agree on all three therefore reuse one compiled kernel; differing on any of them (including a torch tensor's `requires_grad`, which is folded into the specialization) produces a separate one. Other Python objects that merely expose `.shape` / `.dtype` (a `qd.field`, for instance) are not treated as external arrays and keep following the template path.
```
hughperkins: not clear waht "dispatched through the same ndarray path as a bare `qd.ndarray" means. Do we need this phrase at all?

What does it mean 'follow the template path'?

python/quadrants/lang/_template_mapper_hotpath.py
```diff
@@ -78,6 +81,26 @@
...
 # ``None`` argument is dropped and every such launch re-runs full spec-key extraction.
 _primitive_types = {int, float, bool, type(None)}
+_np_ndarray = np.ndarray
+
+
+def _is_external_array(arg: Any) -> bool:
+    """Whether ``arg`` is an external array (numpy / torch) that a ``qd.Tensor`` slot should route through the
+    ndarray feature path instead of the template path.
+
+    Positive allowlist of exactly ``numpy.ndarray`` and ``torch.Tensor``: these are the only external array types
```
hughperkins: this whole paragraph "Positive allowlist of exactly ``numpy.ndarray`` and ``torch.Tensor``: these are the only external array types
   the launch path (``FuncBase._recursive_set_args``) can actually bind, so anything else would specialize here
    only to raise there. A looser duck-typed check on ``.shape`` / ``.dtype`` is unsafe because quadrants-native
    ``Field`` and ``SNode`` both expose those attributes and must keep taking the template path. ``torch`` is
    resolved via ``sys.modules`` so this stays free of a hard torch dependency and costs nothing when torch was
    never imported." seems superflous?

python/quadrants/lang/_template_mapper_hotpath.py
```diff
@@ -78,6 +81,26 @@
 # ``None`` is an immortal singleton that never needs lifetime tracking. Without it, the spec-key cache entry for a
 # ``None`` argument is dropped and every such launch re-runs full spec-key extraction.
 _primitive_types = {int, float, bool, type(None)}
+_np_ndarray = np.ndarray
+
+
+def _is_external_array(arg: Any) -> bool:
+    """Whether ``arg`` is an external array (numpy / torch) that a ``qd.Tensor`` slot should route through the
```
hughperkins: this comment seems superfluous. We know from the function name what it does I feel. Remove "Whether ``arg`` is an external array (numpy / torch) that a ``qd.Tensor`` slot should route through the  ndarray feature path instead of the template path."

tests/python/test_tensor_annotation.py
```diff
@@ -481,3 +481,97 @@ def fill_2d(x: qd.Tensor):
     fill_2d(a_nd)
     np.testing.assert_array_equal(a_nd.to_numpy(), expected)
     # The autouse fixture's teardown will reset() again on the way out.
+
+
+# ----------------------------------------------------------------------------
+# External arrays (numpy / torch) at a qd.Tensor slot (issue #856, W3).
```
hughperkins: why are we documenting this in tests? Seems superfluous. Also, dont write history, like 'before this' etc. Lets just remove "# A ``qd.Tensor`` slot routes numpy arrays and torch tensors through the ndarray feature path (keyed by dtype/ndim),
# exactly like ``qd.types.ndarray()`` already does, instead of falling through to the template path. Before this,
# torch produced one specialization per tensor instance and numpy raised ``TypeError: unhashable type:
# 'numpy.ndarray'``. Field and scalar template values must be unaffected: Field also exposes ``.shape`` / ``.dtype``,
# so the predicate is a positive numpy/torch allowlist rather than a duck-typed check."

hughperkins:
Need to get agent to address some commetns I've left.

docs/source/user_guide/tensor.md
hughperkins: Agent says:

Reworded in 4092837e9 to drop both internal phrases. It now describes the behavior directly: an external array is specialized by its dtype, rank and gradient state (not by the array object's identity), so arrays sharing those reuse one compiled kernel; and objects that merely expose `.shape` / `.dtype` (like a `qd.field`) are handled like any other `qd.Tensor` argument. No more "ndarray path" / "template path".

python/quadrants/lang/_template_mapper_hotpath.py
hughperkins: Agent says:

Agreed - removed the whole `_is_external_array` docstring in 4092837e9. The one load-bearing point (allowlist numpy/torch only, because `Field`/`SNode` also expose `.shape` / `.dtype` so a duck-typed check would misroute them) is covered by the design doc and this PR discussion rather than inline. Happy to add a one-line comment back if you'd prefer it kept near the code.

python/quadrants/lang/_template_mapper_hotpath.py
hughperkins: Agent says:

Done - removed in 4092837e9. The entire docstring is gone (per the sibling comment on the same function).

tests/python/test_tensor_annotation.py
hughperkins: Agent says:

Removed in 4092837e9 - dropped the design/history block (including the "before this" wording), leaving just a one-line section header.

hughperkins:
Genesis unit tests passing earlier:

Screenshot 2026-08-24 at 14 51 43

hughperkins:
running Genesis unit tests again

hughperkins:
running Genesis bechmarks again

hughperkins:
Genesis benchmaks look ok-ish:

20260824_W3_1454

hughperkins:
Genesis unit tests pass:

Screenshot 2026-08-24 at 15 39 44

docs/source/user_guide/tensor.md
```diff
@@ -207,6 +207,8 @@ Inside the kernel the argument is the bare impl rather than the wrapper, so kern
 
 For a parameter annotated `qd.Tensor`, calls to the same kernel may alternate among a wrapper, its bare field or ndarray implementation, and `None`. If callers can pass `None`, guard tensor operations with [`qd.static()`](static.md), for example `if qd.static(x is not None):`; the guarded operations are not traced for the `None` specialization. The parameter remains annotated `qd.Tensor`, and `None` must be passed explicitly.
 
+A `qd.Tensor` parameter also accepts **external arrays** - `numpy.ndarray` and `torch.Tensor`. The kernel is specialized by an external array's dtype, rank and gradient state rather than by the identity of the array object, so two distinct arrays that share those properties reuse the same compiled kernel, while any difference between them - including a torch tensor's `requires_grad` - compiles a separate one. Only `numpy.ndarray` and `torch.Tensor` count as external arrays; other objects that merely expose `.shape` / `.dtype` (such as a `qd.field`) do not, and are handled like any other `qd.Tensor` argument.
```
hughperkins: That seems really complicated. Is there anything other than qd.field that is not handled like an external array? what do you mean by 'handled like any other qd.Tensor' argument? Isnt a field the one thing that is NOT handled like any other qd.Tensor argument?

hughperkins: Agent says:

You're right - simplified in 267f66986. Answering the questions directly:

- No, `qd.field` is not special here, and it is not the only non-external-array value. A `qd.Tensor` slot accepts `qd.field`, `qd.ndarray`, a `qd.Tensor` wrapper, and `None` as its native inputs; this PR just adds `numpy.ndarray` and `torch.Tensor` on top. "External array" means only those two - everything else is a native input.
- You're also right that "handled like any other `qd.Tensor` argument" was wrong/confusing for a field (a field is a core native input, not some fallback case), so I dropped that sentence entirely.
- The `.shape` / `.dtype` bit was internal duck-typing rationale - the same detail you had me remove from the code comment - so it doesn't belong in the user doc either.

The paragraph now just says: external arrays (`numpy.ndarray` / `torch.Tensor`) are accepted in addition to the `qd.field` / `qd.ndarray` tensors and `None` a slot already takes, and are specialized by dtype, rank and gradient state.

docs/source/user_guide/tensor.md
```diff
@@ -207,6 +207,8 @@ Inside the kernel the argument is the bare impl rather than the wrapper, so kern
 
 For a parameter annotated `qd.Tensor`, calls to the same kernel may alternate among a wrapper, its bare field or ndarray implementation, and `None`. If callers can pass `None`, guard tensor operations with [`qd.static()`](static.md), for example `if qd.static(x is not None):`; the guarded operations are not traced for the `None` specialization. The parameter remains annotated `qd.Tensor`, and `None` must be passed explicitly.
 
+A `qd.Tensor` parameter also accepts **external arrays** - `numpy.ndarray` and `torch.Tensor` - in addition to the `qd.field` / `qd.ndarray` tensors and `None` it already takes. An external array is specialized by its dtype, rank and gradient state rather than by the identity of the array object, so two distinct arrays that share those properties reuse the same compiled kernel, while any difference between them - including a torch tensor's `requires_grad` - compiles a separate one.
```
hughperkins: ` it already takes` is confusing. Remove.

hughperkins: Agent says:

Removed in 794103c4f. The sentence now reads: "...accepts external arrays - numpy.ndarray and torch.Tensor - in addition to the qd.field / qd.ndarray tensors and None."

hughperkins:
doc looks ok to me

hughperkins:
I guess I'll wait for CI to go green, then merge

PR #879, opened 2026-08-20
[Cache] W2 Make None arguments fast-cacheable

hughperkins:
## Summary

An absent optional argument (a `qd.Tensor` or `qd.template()` slot passed `None`) silently disabled fast cache for the entire kernel call and logged a warning.

`stringify_obj_type` had no branch for `None`, so it fell through to the `PARAM_INVALID` catch-all, which set the warn flag and made `hash_args` return `FastcacheSkip.WARN`. That hit exactly the absent-branch specialization that #856 wants to be the cheap common case.

`None` is a singleton, so its type fully determines its value: this tags it with a constant. It now hashes to a stable key, does not poison the other arguments in the same call, and appears in the fastcache supported-parameter table.

> Stacked on #877 (base branch `hp/w1-nonetype-primitive-types`). Review/merge that first.

## Verification

`@qd.kernel(fastcache=True)` with an optional `qd.Tensor` slot, fresh cache then reused cache:

| call | before | after run 1 | after run 2 (same cache) |
| --- | --- | --- | --- |
| `b=None` | `cache_key_generated=False`, warns | `cache_key_generated=True` | `cache_validated=True` |
| `b=ndarray` | `cache_validated=True` | `cache_key_generated=True` | `cache_validated=True` |

The `None` specialization now gets a genuine cross-process fastcache hit, and no `PARAM_INVALID` / `INVALID_FUNC` warnings are emitted. Results and autodiff gradients are unchanged.

## Test plan

- [x] New `test_args_hasher_none_is_cacheable` passes on x64
- [x] `None` fastcache validated across processes end-to-end (`fastcache=True` kernel)
- [x] Full CPU suite via the CI harness (`tests/run_tests.py -r 3`, x64, split by `needs_torch` exactly as `linux/4_test.sh`): this branch's unique failing set is **identical to unmodified `main`** (and to W1 #877) and the `needs_torch` phase passes clean, so this PR introduces **no new failures**.

The only shared failures are three tests that cannot pass in the offline cluster container regardless of branch (they fail the same way on `main`), all unrelated to this change:
  - `test_committed_fatbin_layout` - the container's `cuobjdump` predates CUDA 13 and misreads the **committed** cubin's build toolkit as `0.2` instead of `13.0+`;
  - `test_pyi_stubs` - `python -m pyright` cannot fetch its Node runtime on an offline compute node;
  - `test_ipython.ipynb` - `io.UnsupportedOperation` from quadrants' stdout wrapping under nbmake's ipykernel capture.

(An earlier revision of this checklist reported raw `pytest` counts; those were inflated because that run omitted CI's `-r 3` flaky-rerun.)

Context: second of two prerequisite fixes for optional (`None`) kernel arguments, motivated by #856.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -65,3 +65,132 @@ autodoc_typehints_description_target = "all"
 autodoc_default_flags = ["members", "show-inheritance", "undoc-members"]
 autodoc_member_order = "bysource"
 autosummary_generate = True
+
+
+# --- Surface CompileConfig option types and defaults in autodoc --------------
+# nanobind knows each property's type (via __nb_signature__) and writes it into the .pyi stub, but Sphinx's autodoc
+# only reads standard __annotations__, which nanobind leaves empty on its nb_method getters. So autodoc shows neither
+# the type nor a default for CompileConfig's def_rw properties. We bridge that here: the type is read straight from
+# nanobind's own signature metadata, and the default is read from a single default-constructed CompileConfig() (with a
+# small override map for the few defaults that are machine-derived or set Python-side).
+import re as _re  # noqa: E402
+
+_QD_MODULE_PREFIXES = (
+    "quadrants._lib.core.quadrants_python.",
+    "quadrants._lib.core.",
+)
+
+# Defaults that a default-constructed CompileConfig() reports incorrectly for an end user: machine-derived values, or
+# values the Python frontend overrides. A value of None means "do not emit an auto default" (it is covered in prose).
+_CONFIG_DEFAULT_OVERRIDES = {
+    "offline_cache": "``True``",  # frontend sets this; the C++ default is False
+    "arch": None,  # machine-derived; documented in the description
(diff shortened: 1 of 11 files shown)


hughperkins:
@codex review

hughperkins:
No user-facing doc changes required beyond the single row addition to the table in fastcache.md

hughperkins:
I will address the doc quality check in separate PR

hughperkins:
running genesis benchmarks (dont see how this could affect htem, but anwyay)

hughperkins:
Genesis benchmarks look ok:

20260821_W2_1508

hughperkins:
Need to run genesis unit tests next (waiting for a node first...)

hughperkins:
running genesis unit tests

hughperkins:
Genesis unit tests ok:

Screenshot 2026-08-21 at 17 30 04

PR #884, opened 2026-08-24
[Debug] Make QD_DUMP_CFG no longer change cfg behavior

hughperkins:
QD_DUMP_CFG=1 forced cfg_optimization back onto the whole-kernel path, so the graph it dumped for debugging was not the graph the compiler actually builds and optimizes: post-offload the compiler optimizes each offloaded task's CFG independently, but with the dump on it ran (and dumped) one whole-kernel CFG. Enabling a debug dump thus changed which optimizations ran.

Decouple the dump from path selection: dump_cfg no longer gates which path runs, only whether the CFG is written. The graph is dumped at the granularity actually used -- per offloaded task post-offload (files suffixed _task<N> so sibling tasks don't overwrite), whole-kernel before offload and for kernels that are never offloaded. In the pre-offload phases where cfg_optimization is deliberately ditched, QD_DUMP_CFG builds a whole-kernel CFG only to dump the before graph, running no optimization. build_cfg/dump don't mutate IR, so the compiled kernel is identical with the dump on or off.

Adds test_dump_cfg_per_task asserting per-task dump files are produced for a two-task kernel, and updates optimization_passes.md.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough

### DIFF
--- a/quadrants/ir/statements.h
+++ b/quadrants/ir/statements.h
@@ -1427,6 +1427,11 @@ class OffloadedStmt : public Stmt {
   // launch-context level table to rebuild nested graph_do_while loops.
   int graph_do_while_level_id{-1};
 
+  // Kernel-wide index of this task among the kernel's offloaded tasks (`-1` until numbered). Set unconditionally
+  // by `compile_kernel_to_module` before per-task codegen; currently read only to give QD_DUMP_CFG per-task CFG
+  // dumps a stable, collision-free `_task<N>` suffix. Pure metadata: never affects the compiled result.
+  int task_index{-1};
+
   // Pre-chunking loop trip-count `SizeExpr` captured by `determine_ad_stack_size`. Set on adstack-bearing
   // range-for tasks before `make_cpu_multithreaded_range_for` rewrites the loop into per-thread chunks, so the
   // SizeExpr still describes the original user-loop bound (handles both compile-time constants and
--- a/quadrants/ir/transforms.h
+++ b/quadrants/ir/transforms.h
@@ -43,19 +43,6 @@ bool cfg_optimization(const CompileConfig &config,
                       const std::string &kernel_name = "unknown",
                       const std::string &phase = "");
 
-// RAII: records this thread's task id while KernelCodeGen::compile_kernel_to_module lowers one task in isolation
-// (a one-task block), so cfg_optimization can name QD_DUMP_CFG per-task dumps by real index instead of colliding
-// on "_task0". No effect unless QD_DUMP_CFG is set.
-class ScopedTaskCodegenId {
- public:
-  explicit ScopedTaskCodegenId(int task_id);
-  ~ScopedTaskCodegenId();
-  ScopedTaskCodegenId(const ScopedTaskCodegenId &) = delete;
-  ScopedTaskCodegenId &operator=(const ScopedTaskCodegenId &) = delete;
-
- private:
-  std::optional<int> prev_;
-};
 bool alg_simp(IRNode *root, const CompileConfig &config);
 bool demote_operations(IRNode *root, const CompileConfig &config);
 bool binary_op_simplify(IRNode *root, const CompileConfig &config);
--- a/quadrants/ir/transforms.h
+++ b/quadrants/ir/transforms.h
@@ -42,6 +42,22 @@ bool cfg_optimization(const CompileConfig &config,
                       const std::optional<ControlFlowGraph::LiveVarAnalysisConfig> &lva_config_opt = std::nullopt,
                       const std::string &kernel_name = "unknown",
                       const std::string &phase = "");
+
+// Records, for the current thread, the kernel-wide index of the offloaded task being compiled in isolation.
+// KernelCodeGen::compile_kernel_to_module lowers each task on its own worker thread with the IR being a one-task
+// block, so cfg_optimization cannot derive the task's kernel-wide index itself. QD_DUMP_CFG reads this to name
+// per-task CFG dumps by real index, so concurrent tasks don't all collide on the "_task0" file. No effect unless
+// QD_DUMP_CFG is set; nullopt (its default) in whole-kernel contexts, where the index is taken from the block.
+class ScopedTaskCodegenId {
+ public:
+  explicit ScopedTaskCodegenId(int task_id);
+  ~ScopedTaskCodegenId();
+  ScopedTaskCodegenId(const ScopedTaskCodegenId &) = delete;
+  ScopedTaskCodegenId &operator=(const ScopedTaskCodegenId &) = delete;
+
+ private:
+  std::optional<int> prev_;
+};
 bool alg_simp(IRNode *root, const CompileConfig &config);
 bool demote_operations(IRNode *root, const CompileConfig &config);
 bool binary_op_simplify(IRNode *root, const CompileConfig &config);
--- a/quadrants/ir/transforms.h
+++ b/quadrants/ir/transforms.h
@@ -43,11 +43,9 @@ bool cfg_optimization(const CompileConfig &config,
                       const std::string &kernel_name = "unknown",
                       const std::string &phase = "");
 
-// Records, for the current thread, the kernel-wide index of the offloaded task being compiled in isolation.
-// KernelCodeGen::compile_kernel_to_module lowers each task on its own worker thread with the IR being a one-task
-// block, so cfg_optimization cannot derive the task's kernel-wide index itself. QD_DUMP_CFG reads this to name
-// per-task CFG dumps by real index, so concurrent tasks don't all collide on the "_task0" file. No effect unless
-// QD_DUMP_CFG is set; nullopt (its default) in whole-kernel contexts, where the index is taken from the block.
+// RAII: records this thread's task id while KernelCodeGen::compile_kernel_to_module lowers one task in isolation
+// (a one-task block), so cfg_optimization can name QD_DUMP_CFG per-task dumps by real index instead of colliding
+// on "_task0". No effect unless QD_DUMP_CFG is set.
 class ScopedTaskCodegenId {
  public:
   explicit ScopedTaskCodegenId(int task_id);
--- a/quadrants/transforms/cfg_optimization.cpp
+++ b/quadrants/transforms/cfg_optimization.cpp
@@ -33,6 +33,22 @@ std::vector<OffloadedStmt *> collect_offloaded_tasks(IRNode *root) {
   return tasks;
 }
 
+// Suffix for a CFG dump filename. dump_graph_to_file writes "<kernel>_CFG<suffix>.txt", so this yields
+// "_<phase>_before_cfg_opt" / "_<phase>_post_cfg_opt" (the phase segment is dropped when |phase| is empty). For
+// per-task dumps, |task_index| adds a "_task<N>" segment so sibling tasks of the same kernel do not overwrite
+// each other's files.
+std::string cfg_dump_suffix(const std::string &phase, bool post, std::optional<int> task_index = std::nullopt) {
+  std::string suffix;
+  if (!phase.empty()) {
+    suffix += "_" + phase;
+  }
+  if (task_index.has_value()) {
+    suffix += "_task" + std::to_string(task_index.value());
+  }
+  suffix += post ? "_post_cfg_opt" : "_before_cfg_opt";
+  return suffix;
+}
+
 // Build and optimize a control-flow graph for a SINGLE offloaded task, scoped to that task alone.
 //
 // The task is temporarily moved into a throwaway wrapper block and run through the normal Block ->
@@ -51,11 +67,28 @@ std::vector<OffloadedStmt *> collect_offloaded_tasks(IRNode *root) {
 // the CFG spanning only one task, every global address -- fields, external tensors, and the global-temporary
 // buffer that carries scalars between tasks -- is therefore treated as live-in and live-out of the task, so no
 // store a sibling task may read is eliminated and no value is forwarded across a task (device-launch) boundary.
+//
+// QD_DUMP_CFG is observation-only here: when |dump_cfg| is set, the task's scoped CFG is written out before (and,
+// when the optimization runs, after) its per-task analyses, so the dump reflects the exact per-task graph the
+// compiler works on -- it never changes which optimizations run. The |real_matrix_enabled| path runs no per-task
+// analyses (matching the whole-kernel path); when it is set we only build+dump the "before" graph if dumping, and
+// skip entirely when not dumping so the no-op real-matrix behavior is preserved.
 bool optimize_one_task(Block *parent,
                        OffloadedStmt *off,
                        bool after_lower_access,
                        bool autodiff_enabled,
-                       const std::optional<ControlFlowGraph::LiveVarAnalysisConfig> &lva_config_opt) {
+                       bool real_matrix_enabled,
+                       const std::optional<ControlFlowGraph::LiveVarAnalysisConfig> &lva_config_opt,
+                       bool dump_cfg,
+                       const CompileConfig &config,
+                       const std::string &kernel_name,
+                       const std::string &phase,
+                       int task_index) {
+  // Nothing to do when the per-task optimization is disabled (real-matrix path) and we are not dumping: skip the
+  // CFG build so the no-op behavior of the real-matrix post-offload path is preserved exactly.
+  if (real_matrix_enabled && !dump_cfg) {
+    return false;
+  }
   const int location = parent->locate(off);
   QD_ASSERT(location != -1);
   Block wrapper;
@@ -65,12 +98,20 @@ bool optimize_one_task(Block *parent,
     // |cfg| holds raw pointers into |wrapper| (its container nodes) and into the task's own sub-blocks; keep
     // both alive until the analyses are done, then move the task back before |wrapper| leaves scope.
     auto cfg = analysis::build_cfg(&wrapper);
-    cfg->simplify_graph();
-    if (cfg->store_to_load_forwarding(after_lower_access, autodiff_enabled)) {
-      modified = true;
+    if (dump_cfg) {
+      cfg->dump_graph_to_file(config, kernel_name, cfg_dump_suffix(phase, /*post=*/false, task_index));
     }
-    if (cfg->dead_store_elimination(after_lower_access, lva_config_opt)) {
-      modified = true;
+    if (!real_matrix_enabled) {
+      cfg->simplify_graph();
+      if (cfg->store_to_load_forwarding(after_lower_access, autodiff_enabled)) {
+        modified = true;
+      }
+      if (cfg->dead_store_elimination(after_lower_access, lva_config_opt)) {
+        modified = true;
+      }
+      if (dump_cfg) {
+        cfg->dump_graph_to_file(config, kernel_name, cfg_dump_suffix(phase, /*post=*/true, task_index));
+      }
     }
   }
   parent->insert(wrapper.extract(off), location);
@@ -92,6 +133,11 @@ bool cfg_optimization(const CompileConfig &config,
   const char *dump_cfg_env = std::getenv(DUMP_CFG_ENV.data());
   const bool dump_cfg = dump_cfg_env != nullptr && std::string(dump_cfg_env) == "1";
 
+  // QD_DUMP_CFG is observation-only: it must NOT change which path runs or which optimizations run -- it only
+  // controls whether the CFG is written out. So the path selection below is identical with or without the dump,
+  // and the graph is dumped at the granularity the compiler actually uses (per offloaded task post-offload,
+  // whole-kernel otherwise). The dump therefore reflects the real compilation instead of forcing it.
+
   // Per-offloaded-task scoping. Once the kernel is offloaded we optimize each task's CFG independently; and we
   // deliberately DITCH the expensive whole-kernel cfg_optimization in the pre-offload phase, relying on the
   // post-offload per-task cfg below to do the store-to-load forwarding + dead-store elimination once tasks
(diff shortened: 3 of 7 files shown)


quadrants/transforms/cfg_optimization.cpp
```diff
@@ -55,7 +67,18 @@ bool optimize_one_task(Block *parent,
...
+                       bool real_matrix_enabled,
+                       const std::optional<ControlFlowGraph::LiveVarAnalysisConfig> &lva_config_opt,
+                       bool dump_cfg,
+                       const CompileConfig &config,
+                       const std::string &kernel_name,
+                       const std::string &phase,
+                       int task_index) {
+  // Nothing to do when the per-task optimization is disabled (real-matrix path) and we are not dumping: skip the
+  // CFG build so the no-op behavior of the real-matrix post-offload path is preserved exactly.
+  if (real_matrix_enabled && !dump_cfg) {
```
hughperkins: why is dump_cfg changing the control flow?

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -92,52 +123,62 @@ bool cfg_optimization(const CompileConfig &config,
...
-    // store-to-load forwarding + dead-store elimination once tasks exist. So for exactly those phases we ditch
-    // cfg, keeping only the cheap dead-alloca cleanup. For ANY other caller of full_simplify on non-offloaded
-    // IR (unit tests, standalone blocks / function bodies that are never offloaded), we must still run the
-    // whole-kernel cfg below, or its forwarding/DSE would be silently lost -- so we fall through.
-    const bool pre_offload_compile_phase =
-        phase == "simplify_I" || phase == "simplify_II" || phase == "pre_autodiff" || phase == "post_autodiff";
-    if (pre_offload_compile_phase) {
-      die(root);
-      return false;
+    // TODO: implement cfg->dead_instruction_elimination()
```
hughperkins, line 152: what is this todo? why wasnt it here before? why is it here now?

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -92,52 +133,62 @@ bool cfg_optimization(const CompileConfig &config,
...
-      die(root);  // remove unused allocas across the whole kernel
-      return result_modified;
+  auto tasks = collect_offloaded_tasks(root);
+  if (!tasks.empty()) {
+    // Post-offload: per-task store-to-load forwarding + dead-store elimination (skipped for the real-matrix
+    // path, matching the whole-kernel path which runs no analyses there). With QD_DUMP_CFG each task's CFG is
+    // dumped before/after its own optimization; the "_task<N>" suffix keeps sibling tasks from overwriting.
+    bool result_modified = false;
+    auto *block = root->as<Block>();
+    int task_index = 0;
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Good catch, fixed in c3b31150. You're right: during LLVM codegen each offload is compiled in its own one-task Block on a worker thread, so for the phases that run there (before_lower_access, simplify_IV, scalarize, and make_block_local's nested full_simplify) the local block index is always 0 -- they would all collide on the `_task0` files and race across workers.

Instead of threading `task_codegen_id` through offload_to_executable -> FullSimplifyPass::Args -> cfg_optimization (and every nested full_simplify call site, which would also miss make_block_local), I record the task's kernel-wide index per worker thread with a small RAII guard `ScopedTaskCodegenId`, set in `compile_kernel_to_module`'s worker. cfg_optimization reads it for the dump name (`t_task_codegen_id.value_or(local_index)`), falling back to the block-order index in whole-kernel contexts (post-offload simplify_III, SPIR-V whole-kernel). This needs no hot-path signature changes and covers every per-task codegen phase. `test_dump_cfg_per_task` now requires `task1` dumps from more than one phase, which fails without the fix.

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

Agreed -- it should not, and no longer does (fixed in 71f0ff2). That line was `if (real_matrix_enabled && !dump_cfg) return false;` inside optimize_one_task, so enabling the dump changed whether the CFG got built/optimized on the real-matrix path -- exactly the coupling this PR removes. optimize_one_task no longer takes `real_matrix_enabled` or gates on `dump_cfg` in its control flow: the real-matrix guard moved to the caller, and `dump_cfg` now only gates the two dump_graph_to_file calls. build_cfg / simplify_graph / store_to_load_forwarding / dead_store_elimination run identically with or without QD_DUMP_CFG.

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

It is not new -- it is pre-existing on main. `// TODO: implement cfg->dead_instruction_elimination()` appears twice on main (lines 115 and 159); the line-115 one was inside the old `if (!dump_cfg)` per-task branch. My refactor lifted that post-offload per-task branch out from under `if (!dump_cfg)`, so git renders the same TODO as an added line (main 115/159 -> HEAD 151/194). It sits right before the `die(root)` dead-alloca cleanup and notes that a real CFG-level dead-instruction pass could eventually replace/augment `die()`. I kept it verbatim rather than silently drop an existing note -- happy to remove both if you would prefer not to carry it.

hughperkins:
@codex review

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -92,52 +122,58 @@ bool cfg_optimization(const CompileConfig &config,
...
+  // cfg_optimization, whose super-linear reaching-definition / forwarding analyses on the monolithic IR dominate
+  // compile time. Safe because cfg_optimization is an optimization, not correctness, and cross-task forwarding/DSE
+  // on the monolithic IR is invalid across separate device launches anyway; the post-offload per-task cfg redoes
+  // the intra-task forwarding + DSE once tasks exist.
+  auto tasks = collect_offloaded_tasks(root);
+  if (!tasks.empty()) {
+    // Per-task store-to-load forwarding + dead-store elimination, skipped for the real-matrix path (matching the
+    // whole-kernel path, which runs no analyses there).
+    bool result_modified = false;
+    if (!real_matrix_enabled) {
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in a3162630. You're right -- with real_matrix_enabled=true the post-offload branch skipped the whole per-task loop, so QD_DUMP_CFG dumped nothing there, regressing main (which dumped a whole-kernel "before" graph) and contradicting the "one graph per offloaded task" doc.

I split the per-task loop's optimization from its dump. `run_optimization` (= !real_matrix_enabled) gates only the store-to-load forwarding / dead-store elimination; `dump_cfg` independently gates the dumps. So on the real-matrix path each task's "before" graph is still built and written -- there is no "after" since no optimization runs -- now at the same per-task granularity as the optimized path. The CFG build is still skipped entirely when we neither optimize nor dump (real-matrix and QD_DUMP_CFG off), so the flag stays zero-cost when off and never changes the compiled result.

hughperkins:
@codex review

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -92,52 +127,63 @@ bool cfg_optimization(const CompileConfig &config,
...
+  // optimizations run. The graph is dumped at the granularity actually used (per offloaded task post-offload,
+  // whole-kernel otherwise), so it reflects real compilation instead of forcing it.
+
+  // Post-offload we optimize each task's CFG independently; pre-offload (no tasks yet) we DITCH the whole-kernel
+  // cfg_optimization, whose super-linear reaching-definition / forwarding analyses on the monolithic IR dominate
+  // compile time. Safe because cfg_optimization is an optimization, not correctness, and cross-task forwarding/DSE
+  // on the monolithic IR is invalid across separate device launches anyway; the post-offload per-task cfg redoes
+  // the intra-task forwarding + DSE once tasks exist.
+  auto tasks = collect_offloaded_tasks(root);
+  if (!tasks.empty()) {
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in 6d1933f. You're right: make_block_local's full_simplify passes a bare OffloadedStmt (not a Block), so collect_offloaded_tasks() returns empty and it lands on the whole-kernel fallback, which was not applying the per-task suffix -- so concurrent BLS + matrix-scalarize struct-for workers all wrote the same make_block_local_CFG file.

I took the "apply the scoped task suffix on the fallback path" option: the fallback (and the pre-offload) dump now pass t_task_codegen_id into cfg_dump_suffix. On a codegen worker that is the task's kernel-wide id (unique per worker); for genuine whole-kernel / never-offloaded function-body roots off the codegen threads it is nullopt, so those filenames are unchanged. I did not route the bare offload through the per-task path because optimize_one_task needs the offload's parent Block (parent->extract/insert), which a bare-offload root does not have.

hughperkins:
@codex review

hughperkins:
I guess we should:
- wait for CI
- run Genesis benchmarks
- run Genesis unit tests
(since CFG very sensitive)

hughperkins:
running Genesis benchmarks

hughperkins:
CI green

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -92,52 +128,62 @@ bool cfg_optimization(const CompileConfig &config,
...
+  // is written -- at the granularity actually used (per task post-offload, whole-kernel otherwise).
+
+  // Post-offload we optimize each task's CFG independently; pre-offload (no tasks yet) we DITCH the whole-kernel
+  // cfg_optimization, whose super-linear reaching-definition / forwarding analyses on the monolithic IR dominate
+  // compile time. Safe because cfg_optimization is an optimization, not correctness, and cross-task forwarding/DSE
+  // on the monolithic IR is invalid across separate device launches anyway; the post-offload per-task cfg redoes
+  // the intra-task forwarding + DSE once tasks exist.
+  auto tasks = collect_offloaded_tasks(root);
+  if (!tasks.empty()) {
+    // Per-task store-to-load forwarding + dead-store elimination, skipped on the real-matrix path (like the
```
hughperkins, line 138: "-- but QD_DUMP_CFG still dumps each task's "before" graph there." seems superfluous

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -92,52 +128,62 @@ bool cfg_optimization(const CompileConfig &config,
...
-        auto *block = root->as<Block>();
-        for (auto *off : tasks) {
-          result_modified |= optimize_one_task(block, off, after_lower_access, autodiff_enabled, lva_config_opt);
-        }
+  // QD_DUMP_CFG is observation-only: it never changes the path or which optimizations run, only whether the CFG
+  // is written -- at the granularity actually used (per task post-offload, whole-kernel otherwise).
+
+  // Post-offload we optimize each task's CFG independently; pre-offload (no tasks yet) we DITCH the whole-kernel
+  // cfg_optimization, whose super-linear reaching-definition / forwarding analyses on the monolithic IR dominate
+  // compile time. Safe because cfg_optimization is an optimization, not correctness, and cross-task forwarding/DSE
```
hughperkins: remove 'Safe because cfg_optimization is an optimization, not correctness, and cross-task forwarding/DSE
  // on the monolithic IR is invalid across separate device launches anyway; the post-offload per-task cfg redoes
  // the intra-task forwarding + DSE once tasks exist.'

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -33,29 +37,45 @@ std::vector<OffloadedStmt *> collect_offloaded_tasks(IRNode *root) {
...
-// reaching_definition_analysis seeds the start node with all global pointers ("may already hold data") and
-// live_variable_analysis seeds the final node with all global store destinations ("may be read later"). With
-// the CFG spanning only one task, every global address -- fields, external tensors, and the global-temporary
-// buffer that carries scalars between tasks -- is therefore treated as live-in and live-out of the task, so no
-// store a sibling task may read is eliminated and no value is forwarded across a task (device-launch) boundary.
+// Scoping to one task is safe because each task is a separate device launch and CFG boundary seeding is
+// conservative across it: reaching-definition seeds every global pointer live-in and live-variable seeds every
+// global store destination live-out, so nothing is forwarded or eliminated across the launch boundary.
+//
+// |run_optimization| gates the optimization, |dump_cfg| the dumps, independently: with optimization off
```
hughperkins: remove '// |run_optimization| gates the optimization, |dump_cfg| the dumps, independently: with optimization off
// (real-matrix path) the "before" graph is still dumped, and QD_DUMP_CFG never changes what runs.'

quadrants/ir/transforms.h
```diff
@@ -42,6 +42,20 @@ bool cfg_optimization(const CompileConfig &config,
                       const std::optional<ControlFlowGraph::LiveVarAnalysisConfig> &lva_config_opt = std::nullopt,
                       const std::string &kernel_name = "unknown",
                       const std::string &phase = "");
+
+// RAII: records this thread's task id while KernelCodeGen::compile_kernel_to_module lowers one task in isolation
+// (a one-task block), so cfg_optimization can name QD_DUMP_CFG per-task dumps by real index instead of colliding
+// on "_task0". No effect unless QD_DUMP_CFG is set.
```
hughperkins: why couple it to QD_DUMP_CFG?

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -92,52 +128,62 @@ bool cfg_optimization(const CompileConfig &config,
...
-    if (!tasks.empty()) {
-      // Post-offload: per-task store-to-load forwarding + dead-store elimination (skipped for the real-matrix
-      // path, matching the whole-kernel path which runs no analyses there).
-      bool result_modified = false;
-      if (!real_matrix_enabled) {
-        auto *block = root->as<Block>();
-        for (auto *off : tasks) {
-          result_modified |= optimize_one_task(block, off, after_lower_access, autodiff_enabled, lva_config_opt);
-        }
+  // QD_DUMP_CFG is observation-only: it never changes the path or which optimizations run, only whether the CFG
```
hughperkins: remove ' // QD_DUMP_CFG is observation-only: it never changes the path or which optimizations run, only whether the CFG
  // is written -- at the granularity actually used (per task post-offload, whole-kernel otherwise).'

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -92,52 +128,62 @@ bool cfg_optimization(const CompileConfig &config,
...
+  auto tasks = collect_offloaded_tasks(root);
+  if (!tasks.empty()) {
+    // Per-task store-to-load forwarding + dead-store elimination, skipped on the real-matrix path (like the
+    // whole-kernel path) -- but QD_DUMP_CFG still dumps each task's "before" graph there.
+    const bool run_optimization = !real_matrix_enabled;
+    bool result_modified = false;
+    auto *block = root->as<Block>();
+    int task_index = 0;
+    for (auto *off : tasks) {
+      // Kernel-wide task id: t_task_codegen_id during isolated per-task codegen (one-task block), else the
```
hughperkins: remove '      // Kernel-wide task id: t_task_codegen_id during isolated per-task codegen (one-task block), else the
      // loop index (whole-kernel, tasks in order).'

quadrants/transforms/cfg_optimization.cpp
```diff
@@ -92,52 +128,62 @@ bool cfg_optimization(const CompileConfig &config,
...
+      // loop index (whole-kernel, tasks in order).
+      const int dump_task_id = t_task_codegen_id.value_or(task_index);
+      ++task_index;
+      // Nothing to optimize or dump: skip the CFG build (keeps QD_DUMP_CFG zero-cost when off).
+      if (!run_optimization && !dump_cfg) {
```
hughperkins: why do we have a branch conditional on dump_cfg?

hughperkins:
Genesis benchmarks ok:

20260824_dump_cfg_1733

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

Removed in 84a748dc. The per-task loop still dumps each task's "before" graph on the real-matrix path, but that now follows from the run_optimization / dump_cfg split without needing a note here.

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

Removed in 84a748dc.

Heads up: that "safe because ... optimization not correctness ... invalid across separate device launches" rationale was carried over from main, so dropping it may re-trip the "Check deleted comments" CI gate. The core reason it's safe to ditch (the whole-kernel cfg is redundant once the post-offload per-task cfg redoes the intra-task forwarding + DSE) still lives on the "No offloaded tasks yet" block below. If the gate complains, I can dismiss it or restore a one-line version -- your call.

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

Removed in 84a748dc.

quadrants/ir/transforms.h
hughperkins: Agent says:

It isn't a runtime coupling -- ScopedTaskCodegenId exists solely to give per-task CFG dumps a stable kernel-wide filename. The codegen worker records t_task_codegen_id unconditionally (a cheap thread-local optional write, so no per-task getenv), but the only code that reads it is cfg_optimization's dump-naming path. When QD_DUMP_CFG is off the recorded id is simply never read, hence "no effect unless set." So it's coupled by purpose (a debug-only aid), not by adding any dependency to normal compilation. If you'd rather not record it when dumping is off, I can gate the guard on a cached QD_DUMP_CFG bool instead.

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

Removed in 84a748dc.

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

Removed in 84a748dc.

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

It's a pure short-circuit, not a dump_cfg-driven behavior change. When we will neither optimize (run_optimization is false on the real-matrix path) nor dump (QD_DUMP_CFG off), building the per-task CFG is wasted work -- nothing consumes it, and build_cfg doesn't mutate IR -- so we skip it. This matches main, where the whole per-task loop was skipped for real matrices. The compiled IR is identical whether we skip or build-and-discard; dump_cfg only decides whether we do that observation work. I can move the early-out inside optimize_one_task so dump_cfg doesn't appear in the loop condition, but it's equivalent.

hughperkins:
CI green

quadrants/ir/transforms.h
hughperkins: Agent says:

Done -- removed the coupling entirely in 95d56e14; there's no more thread-local / `ScopedTaskCodegenId`.

`OffloadedStmt` now carries an unconditional `task_index` field (its kernel-wide position among the kernel's offloaded tasks), set in `compile_kernel_to_module` before each task is lowered and copied by `clone()` onto the per-task block. `cfg_optimization` just reads `off->task_index` for the dump suffix, falling back to block order before offload.

So the numbering is decoupled from `QD_DUMP_CFG`: it's inert metadata that never affects compilation, is always populated, and the dump path is its only reader. Built clean and `test_dump_cfg_per_task` / `test_cfg_continue` pass on cuda + x64.

hughperkins:
@codex review

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

Restored a one-line version in 0df630f0. I kept only the load-bearing part -- cross-task forwarding/DSE on the monolithic IR is invalid across the separate device launches anyway -- since that's the "why it's safe to ditch" that the redundancy note below doesn't cover.

quadrants/transforms/cfg_optimization.cpp
hughperkins: Agent says:

Moved in 0df630f0. The `!run_optimization && !dump_cfg` short-circuit now lives at the top of `optimize_one_task` (it returns before touching the IR), so `dump_cfg` no longer appears in the caller loop's control flow -- it only gates the dump calls. Behavior is identical.

hughperkins:
@codex review

quadrants/ir/statements.h
```diff
@@ -1427,6 +1427,11 @@ class OffloadedStmt : public Stmt {
   // launch-context level table to rebuild nested graph_do_while loops.
   int graph_do_while_level_id{-1};
 
+  // Kernel-wide index of this task among the kernel's offloaded tasks (`-1` until numbered). Set unconditionally
+  // by `compile_kernel_to_module` before per-task codegen; currently read only to give QD_DUMP_CFG per-task CFG
+  // dumps a stable, collision-free `_task<N>` suffix. Pure metadata: never affects the compiled result.
+  int task_index{-1};
```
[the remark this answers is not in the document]
hughperkins: it removes a conditional. I asked for this. The concept of a task-id is not clearly itself specific to qd_dump_cfg, even if it's only being used by qd_dump_cfg currently.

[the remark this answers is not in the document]
hughperkins: curious about 'special handling' and 'prodcution-code' mutation though.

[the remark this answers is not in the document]
hughperkins: Agent says:

Thanks -- considered this, but keeping the field deliberately; declining the revert.

`task_index` isn't really dump-only metadata: it's the task's kernel-wide launch position among the kernel's offloaded tasks, a real, inert property of an `OffloadedStmt`. It sits alongside the naming/debug metadata `OffloadedStmt` already carries and `clone()` already copies (`loop_name`, `range_hint`), and like those it never participates in codegen or the IR hash -- so the "production mutation" is a single inert `int` assignment that cannot change the compiled result. The dump path is just its first reader.

On contact area specifically: the partitioned alternative (`ScopedTaskCodegenId`) isn't smaller. It puts a new class into `transforms.h` (a very central, widely-included header), plus a `thread_local` global in `cfg_optimization.cpp` and an RAII guard constructed in the per-task codegen hot loop. The field replaces all of that with one struct member + one `clone()` line + one assignment, and removes a thread-local from the codebase. We judge that the smaller and more honest surface, so this is the maintainer's preferred design here rather than an oversight.

hughperkins:
runnng genesis unit-tests on tests0

hughperkins:
Once genesis unit tests pass, lets merge

hughperkins:
Lets try running Genesis unit tests again (were crashing before)

hughperkins:
Genesis unit tests running on tests0

hughperkins:
*finally* Genesis unit tests passing 😌 

Screenshot 2026-08-25 at 15 46 00

PR #886, opened 2026-08-25
[Lang] Make kernel source capture resilient to transient linecache misses

hughperkins:
## Summary

- Kernel materialization re-reads each kernel's `.py` source at runtime via `inspect.getsourcelines` -> stdlib `linecache.getlines`. Under memory or filesystem pressure `getlines` can transiently return an empty list (a `MemoryError` makes it call `linecache.clearcache()` and return `[]`; a transient `os.stat`/`open` failure in `updatecache` has the same effect). Because every `_wrap_inspect` fallback (stdlib, dill, bpy) is linecache/source based, one such transient turned a perfectly readable file into a fatal `OSError: Cannot find source code for Object: <function ext_arr_to_ndarray ...>` (`_wrap_inspect.py:161`), aborting compilation.
- Adds a last-resort `_direct_file_findsource` that bypasses linecache and re-reads the source file directly (honoring the PEP 263 encoding cookie), wired in as the final fallback in `_custom_findsource`. Also flattens the fallback chain; the original error message is preserved verbatim for the genuinely-unrecoverable case.
- Adds a CPU-only regression test that injects the transient empty `getlines` and asserts source capture still succeeds.

## Root cause (why it was flaky / environment-gated)

This surfaced as intermittent failures of Genesis IPC tests (e.g. `test_contact_pair_friction_resistance`) under heavy concurrent test load (dozens of xdist workers + torch/CUDA + other jobs on the node). The failing kernel `ext_arr_to_ndarray` is materialized by an explicit `from_numpy` deep in the rigid constraint solver, on the first `sim.step()` during `scene.build()`.

Instrumentation of `get_source_info_and_src` showed the failing kernel is source-captured an identical number of times on the affected and unaffected commits, and injecting a single empty `getlines` at that capture reproduces the exact error on both. So this is a preexisting fragility in quadrants' runtime source-inspection path (not a specific-branch regression), and it can affect any kernel. The `Lock ... failed` offline-cache warnings seen alongside it are a symptom of the same contended environment, not the cause.

## Test plan / evidence

Red/green against the `main` wheel (`3d97d7c32`), CPU-only (no device init):

- Unpatched: `2 failed` -- `OSError: Cannot find source code for ... ext_arr_to_ndarray` at `_wrap_inspect.py:161`.
- Patched:   `2 passed`.

- [ ] CI green on the full suite.

### DIFF
--- a/python/quadrants/lang/_wrap_inspect.py
+++ b/python/quadrants/lang/_wrap_inspect.py
@@ -150,10 +150,9 @@ def _REPL_findsource(obj):
 
 
 def _direct_file_findsource(obj):
-    # Fallback for when linecache.getlines() transiently returns [] for a readable
-    # file (MemoryError -> linecache.clearcache(), or a transient os.stat/open
-    # failure), which would otherwise abort kernel compilation. Returns
-    # (lines, lineno) like inspect.findsource.
+    # Fallback for when linecache.getlines() transiently returns [] for a readable file (MemoryError ->
+    # linecache.clearcache(), or a transient os.stat/open failure), which would otherwise abort kernel compilation.
+    # Returns (lines, lineno) like inspect.findsource.
     code = getattr(obj, "__code__", None)
     if code is None:
         raise IOError(f"No __code__ to locate source for {obj}")
--- a/python/quadrants/lang/_wrap_inspect.py
+++ b/python/quadrants/lang/_wrap_inspect.py
@@ -19,6 +19,7 @@ import atexit
 import inspect
 import os
 import tempfile
+import tokenize
 from typing import Callable
 
 import dill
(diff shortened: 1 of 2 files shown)


hughperkins:
@codex review

hughperkins:
Alsirhgt so, we want:
- CI green
- genesis unit tests <=
- genesis benchamrsk

hughperkins:
runing genesis unit tests on tests0

hughperkins:
running genesis benchmarks on bench1

hughperkins:
Genesis benchmarks look ok-ish:

20260825_linecache_1239

hughperkins:
CI green except for line wrapping

hughperkins:
gen unit tests passing:

Screenshot 2026-08-25 at 14 23 46

PR #891, opened 2026-08-26
[Misc] Make qd.init arch optional and stop QD_ARCH overriding arch arg

hughperkins:
## Summary

- `qd.init`'s `arch` argument now takes priority over the `QD_ARCH` env var, mirroring how `default_fp` / `default_ip` already behave. Previously `QD_ARCH` unconditionally overrode whatever `arch` was passed to `qd.init`.
- `QD_ARCH` is only consulted when no `arch` argument is passed; when both are set, a warning is emitted and the argument wins.
- `arch` stays optional: with neither an argument nor `QD_ARCH`, `adaptive_arch_select` falls back to `cpu`.

This also sidesteps a gotcha where `arch_from_name` only accepts the names in `archs.inc.h` (`x64/arm64/js/python/cuda/metal/opencl/amdgpu/vulkan`), so an explicit `arch=` argument is no longer discarded in favor of parsing the env var.

## Test plan

- Added `test_init_arch_optional_defaults_to_cpu` and `test_init_arch_arg_overrides_env` to `tests/python/test_runtime.py`.
- Built quadrants (CUDA + x64) on an rtx-mid node and ran the arch tests: 5 passed, 4 skipped (unavailable backends). The `test_init_arch_arg_overrides_env[Arch.cuda]` case confirms `qd.init(arch=cuda)` wins over `QD_ARCH=opencl` (would fail under the old override behavior).
- Full `test_runtime.py`: 48 passed, 0 failed.

### DIFF
--- a/python/quadrants/lang/misc.py
+++ b/python/quadrants/lang/misc.py
@@ -490,9 +490,7 @@ def init(
         runtime.print_non_pure = print_non_pure
         _logging.set_logging_level(spec_cfg.log_level.lower())
 
-    # select arch (backend). The qd.init argument takes priority over the QD_ARCH env var (mirroring default_fp/ip);
-    # the env var is only consulted when no arch argument was passed. When neither is set, adaptive_arch_select falls
-    # back to cpu, so arch remains optional.
+    # select arch (backend):
     env_arch = os.environ.get("QD_ARCH")
     if env_arch is not None:
         if arch is not None:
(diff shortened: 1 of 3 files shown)


hughperkins:
running gen tests on tests2

hughperkins:
Genesis unit tests passed:

Screenshot 2026-08-26 at 15 08 11

hughperkins:
runnning Genesis benchamrs on bench1

hughperkins:
Genesi benchmark runs ok (and looks ok, but Im ostly cared whether it actually runs)

20260827_arch_1509

hughperkins:
I'll wait for AMD to pass too I think.

PR #896, opened 2026-08-30
Support CUDA 13 cuBLAS libraries

jeongseok-meta:
**Why:** CUDA 13 provides `libcublas.so.13` / `cublas64_13.dll`, while the runtime loader currently tries only cuBLAS 11, 12, and 10. As a result, Quadrants' optional CUDA linear-algebra paths cannot load cuBLAS from a CUDA 13-only environment. This surfaced while packaging Quadrants 1.3 in [conda-forge/staged-recipes#34661](https://github.com/conda-forge/staged-recipes/pull/34661).

**What:** Accept the CUDA 13 cuBLAS SONAME while preserving the existing CUDA 11/12 preference used for PyTorch-compatible environments.

**How:**

- **CUDA runtime loader** -- add version `13` after versions `11` and `12` in `CUBLASDriver::load_cublas()`.
- **Compatibility** -- leave the existing version order and all other CUDA library loading unchanged.

**Evidence:** The identical source patch was built into a conda package and exercised on an NVIDIA RTX 3080 with `cuda-version 13.0`, cuBLAS 13.1.1, cuSPARSE 12.6.3, and cuSOLVER 12.0.4. A real sparse conjugate-gradient solve completed and matched NumPy; details are recorded in the [staged-recipes validation comment](https://github.com/conda-forge/staged-recipes/pull/34661#issuecomment-5463768193).

**Test plan:**

```bash
git diff --check upstream/main...HEAD
```
Run URL: local only (no run URL emitted)

No public API changes.

### DIFF
--- a/quadrants/rhi/cuda/cuda_driver.cpp
+++ b/quadrants/rhi/cuda/cuda_driver.cpp
@@ -244,7 +244,7 @@ bool CUBLASDriver::load_cublas() {
    * it would confict with torch's cublas. When using libcublas.so.11, the
    * torch's cublas will be loaded.
    */
-  cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 10});
+  cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 13, 10});
   if (!cublas_loaded_) {
     return false;
   }
--- a/quadrants/rhi/cuda/cuda_driver.cpp
+++ b/quadrants/rhi/cuda/cuda_driver.cpp
@@ -239,13 +239,9 @@ CUBLASDriver &CUBLASDriver::get_instance() {
 }
 
 bool CUBLASDriver::load_cublas() {
-  /* Load a versioned cuBLAS SONAME (e.g. libcublas.so.12) rather than the
-   * unversioned libcublas.so. In a PyTorch environment torch has already loaded
-   * its own cuBLAS; matching a versioned SONAME reuses torch's copy, whereas
-   * libcublas.so would pull in a second, conflicting system cuBLAS. The first
-   * already-loaded match wins, so in practice we bind to whichever cuBLAS torch
-   * loaded. torch currently ships CUDA 12/13 builds (older torch shipped CUDA
-   * 11), so we try 11, 12, and 13.
+  /* Load a versioned cuBLAS SONAME (e.g. libcublas.so.12) rather than the unversioned libcublas.so. In a PyTorch
+   * environment torch has already loaded its own cuBLAS; matching a versioned SONAME reuses torch's copy, whereas
+   * libcublas.so would pull in a second, conflicting system cuBLAS.
    */
   cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 13});
   if (!cublas_loaded_) {
--- a/quadrants/rhi/cuda/cuda_driver.cpp
+++ b/quadrants/rhi/cuda/cuda_driver.cpp
@@ -242,6 +242,10 @@ bool CUBLASDriver::load_cublas() {
   /* Load a versioned cuBLAS SONAME (e.g. libcublas.so.12) rather than the unversioned libcublas.so. In a PyTorch
    * environment torch has already loaded its own cuBLAS; matching a versioned SONAME reuses torch's copy, whereas
    * libcublas.so would pull in a second, conflicting system cuBLAS.
+   *
+   * FIXME: the version list is hardcoded. Ideally we would both derive fresh-load candidates from the detected
+   * driver version (like load_cusolver()/load_cusparse()) and still scan a broad set for torch's already-loaded
+   * cuBLAS. That needs try_load_lib_any_version() to take separate "already-loaded scan" and "fresh-load" lists.
    */
   cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 13});
   if (!cublas_loaded_) {
--- a/quadrants/rhi/cuda/cuda_driver.cpp
+++ b/quadrants/rhi/cuda/cuda_driver.cpp
@@ -243,9 +243,8 @@ bool CUBLASDriver::load_cublas() {
    * environment torch has already loaded its own cuBLAS; matching a versioned SONAME reuses torch's copy, whereas
    * libcublas.so would pull in a second, conflicting system cuBLAS.
    *
-   * FIXME: the version list is hardcoded. Ideally we would both derive fresh-load candidates from the detected
-   * driver version (like load_cusolver()/load_cusparse()) and still scan a broad set for torch's already-loaded
-   * cuBLAS. That needs try_load_lib_any_version() to take separate "already-loaded scan" and "fresh-load" lists.
+   * FIXME: the version list is hardcoded. Ideally we would both derive fresh-load candidates from the detected driver
+   * version (like load_cusolver()/load_cusparse()) and still scan a broad set for torch's already-loaded cuBLAS.
    */
   cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 13});
   if (!cublas_loaded_) {
--- a/quadrants/rhi/cuda/cuda_driver.cpp
+++ b/quadrants/rhi/cuda/cuda_driver.cpp
@@ -239,12 +239,15 @@ CUBLASDriver &CUBLASDriver::get_instance() {
 }
 
 bool CUBLASDriver::load_cublas() {
-  /* To be compatible with torch environment, please libcublas.so.11 other than
-   * libcublas.so. When using libcublas.so, the system cublas will be loaded and
-   * it would confict with torch's cublas. When using libcublas.so.11, the
-   * torch's cublas will be loaded.
+  /* Load a versioned cuBLAS SONAME (e.g. libcublas.so.12) rather than the
+   * unversioned libcublas.so. In a PyTorch environment torch has already loaded
+   * its own cuBLAS; matching a versioned SONAME reuses torch's copy, whereas
+   * libcublas.so would pull in a second, conflicting system cuBLAS. The first
+   * already-loaded match wins, so in practice we bind to whichever cuBLAS torch
+   * loaded. torch currently ships CUDA 12/13 builds (older torch shipped CUDA
+   * 11), so we try 11, 12, and 13.
    */
-  cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 13, 10});
+  cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 13});
   if (!cublas_loaded_) {
     return false;
   }


jeongseok-meta:
Hi @hughperkins, would you mind taking a look when you have a chance? Thanks!

hughperkins:
@codex review

quadrants/rhi/cuda/cuda_driver.cpp
```diff
@@ -244,7 +244,7 @@ bool CUBLASDriver::load_cublas() {
    * it would confict with torch's cublas. When using libcublas.so.11, the
    * torch's cublas will be loaded.
    */
-  cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 10});
+  cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 13, 10});
```
hughperkins: @ agent (or @jeongseok-meta perhaps you know the answers too?), some questions:
- why do we even need to give versions here at all? Like why doesnt it just accept any version?
- why don't we just provide a range of versions?
- how do we decide the order?
    - like, why isnt the later version first?
    - why is 10 last, even though other versions are ordered lowest version first?

hughperkins: Agent says:

**Why give versions at all / why not accept any version?**

`dlopen`/`LoadLibrary` take a concrete filename, not a wildcard - there is no "load any libcublas" call. On Linux, runtime packages ship the versioned SONAME (`libcublas.so.12`) but usually not the bare `libcublas.so` symlink (that lives in the `-dev`/devel package). So to load at runtime you have to reconstruct the versioned filename, which requires knowing the major version. There is a bare-`.so` fallback at the end of the Linux path, but it cannot be relied on, and there is a deliberate reason to avoid it (see below).

**Why not just provide a range (e.g. 8..20)?**

- A range forces a single monotonic order, but this list is a preference order, not a numeric one (11, 12 before 13, and 10 dead last). A range cannot express that.
- The explicit list also documents which versions are actually known/supported, and avoids probing a bunch of nonexistent SONAMEs on every startup.

**How is the order decided? Why not newest-first? Why is 10 last?**

The order is preference, not recency, and it is driven mainly by the torch-compatibility concern in the comment above the call:

```
/* To be compatible with torch environment, please libcublas.so.11 other than
 * libcublas.so. When using libcublas.so, the system cublas will be loaded and
 * it would confict with torch's cublas. When using libcublas.so.11, the
 * torch's cublas will be loaded.
 */
```

- The first loop in `try_load_lib_any_version` reuses an already-loaded cuBLAS if one is present. In a PyTorch env, torch has already loaded its own cuBLAS (currently 11 or 12), so listing 11/12 first makes us detect and reuse torch's copy instead of pulling in a second, conflicting one. That is the whole point of the ordering.
- So "newest first" is not the goal - "most-likely-to-already-be-present-and-compatible first" is. torch ships 11/12 today, 13 is not yet common there, so 13 sits after 12.
- 10 is last because it is legacy/deprecated: you only want it as a last resort if nothing modern is available. So it is "ascending among the modern supported versions, with the legacy one demoted to the back," rather than a strict numeric sort.

On this PR specifically, putting 13 after 12 but before 10 (`{11, 12, 13, 10}`) is consistent with that logic - it keeps the torch-preferred 11/12 ahead and leaves legacy 10 at the back.

One caveat: the "already loaded" ordering only matters when multiple versions are simultaneously loaded (rare); for a fresh `dlopen`, the second loop's order is a soft preference and mostly moot in torch envs because the first loop catches torch's copy.

hughperkins: @ agent:
- do we still need 10? I think 10 is deprecated now right? maybe 11 too???
- can you check what are the actual current torch versions please

hughperkins: Agent says:

**Current torch versions (as of Sep 2026)**

Latest stable is torch 2.13.0 (2.14 is branch-cut). Per PyTorch's `RELEASE.md` support matrix, the CUDA builds for current releases are:

| torch | CUDA builds shipped | cuBLAS SONAME |
| --- | --- | --- |
| 2.12 - 2.14 | 12.6 (legacy), 13.0 (stable/default on PyPI), 13.2 (experimental) | `.so.12`, `.so.13` |
| 2.11 | 12.6, 12.8, 13.0 | `.so.12`, `.so.13` |
| 2.9 - 2.10 | 12.6, 12.8, (13.0) | `.so.12`, `.so.13` |
| 2.8 | 12.6, 12.8, 12.9 | `.so.12` |
| 2.7 | **11.8**, 12.6, 12.8 | `.so.11`, `.so.12` |
| 2.0 - 2.6 | **11.8** + various 12.x | `.so.11`, `.so.12` |

Key takeaways:
- The default `pip install torch` wheel from PyPI now targets **CUDA 13.0**, so torch's default cuBLAS is now `libcublas.so.13`.
- CUDA **12.6** is deliberately kept as the "legacy" build for older GPU archs (Maxwell/Pascal/Volta) that CUDA 13.x dropped, so `libcublas.so.12` is still very common.
- **No current torch ships CUDA 11** - the last CUDA 11.8 wheels were in torch 2.7 (Apr 2025); dropped from 2.8 onward.
- No torch has shipped CUDA 10 since ~1.11/1.12 (2022).

**Do we still need 10?**

Effectively not for torch - nothing remotely recent uses it. It would only ever match a non-torch user with an ancient standalone CUDA 10 toolkit. Note Quadrants' own `CUDADriver` already sets the floor at CUDA >= 10.0, so 10 is the absolute minimum it will accept. Keeping `10` in the list costs one entry and one harmless failed `dlopen`; dropping it only affects truly ancient setups. I'd lean toward dropping it (or keeping it purely as a harmless last-resort) - either is defensible.

**Maybe 11 too?**

I would **not** drop 11 yet. Current torch no longer ships CUDA 11, but a large number of pinned/real-world environments still run torch 2.0 - 2.7 on the cu118 wheels, which link `libcublas.so.11`. Removing 11 would break the "reuse torch's already-loaded cuBLAS" path for all of those. 11 is aging but still widely deployed, so keep it for now.

**Ordering implication**

Since torch's default is now CUDA 13, `libcublas.so.13` is now the most common torch cuBLAS, not 11/12. That weakens the original rationale for putting 11/12 first. In practice it barely matters, because the first loop in `try_load_lib_any_version` detects whatever cuBLAS torch already loaded regardless of list position - so `{11, 12, 13, 10}` is fine functionally. But the code comment above the call still says "please libcublas.so.11 ... torch's cublas will be loaded", which is now stale; worth updating it to reflect that torch defaults to 12/13 today.

Suggested net change: keep `{11, 12, 13}` (drop or demote `10`), and refresh the comment.

hughperkins: @ agent:
- why not 13, 12, 11? Wouldnt we prefer a later version, if available?

hughperkins: Agent says:

Short answer: `13, 12, 11` would be almost entirely harmless, but "prefer the latest" isn't actually the goal here, and in the one scenario where order matters it can be slightly *worse*. Details:

**1. For the torch case (the whole reason this list exists), order is irrelevant.**

The first loop in `try_load_lib_any_version` only matches libraries *already loaded into the process*, not merely installed on disk. In a torch env exactly one cuBLAS is loaded (torch's), so the loop returns torch's version regardless of whether the list is `11,12,13` or `13,12,11`. The goal is "bind to whatever cuBLAS torch already has," not "bind to the newest" - so reordering changes nothing here.

**2. Where order does matter, latest-first can defeat the goal.**

Order only matters when two+ cuBLAS SONAMEs are simultaneously present. Two sub-cases:

- *Two loaded at once* (e.g. torch loaded the legacy `.so.12` build, but some other lib pulled in `.so.13`): we still want torch's `.so.12`, because we share a process with torch. A `13`-first preference would pick the non-torch `.so.13` and reintroduce exactly the dual-cuBLAS conflict the comment warns about. So latest-first is actively wrong here.

- *Standalone, two installed on disk, none loaded* (no torch): the second loop probes in order and loads the first that exists. Here `13,12,11` would grab 13, `11,12,13` grabs 12. Newer cuBLAS needs a newer driver, so probing lower-first biases toward a version the (possibly older) driver can actually back; a `dlopen` of `.so.13` can succeed but then fail when called if the driver is CUDA 12-era. So lower-first is the safer default, not the latest.

**3. So the ordering is a compatibility heuristic, not a recency preference.**

If you genuinely want "prefer the version that matches this machine," the more principled approach is what `load_cusolver()`/`load_cusparse()` already do - derive the candidates from the detected driver version (`{cuda_version, cuda_version - 1}`) rather than hardcoding a list. cuBLAS is the odd one out using a fixed list, precisely because it prioritizes matching *torch's* loaded lib over the driver's version.

Net: I wouldn't flip to `13,12,11`. Either keep the current `{11, 12, 13}` (drop/demote 10), or, if you want latest-preference done safely, switch cuBLAS to a driver-derived list like the other two libs - the already-loaded first loop keeps torch-compat intact either way.

hughperkins: @ agent:
- please remove 10
- it sounds like there is a stale comment that needs updating too?

hughperkins: Agent says:

Done - pushed to `support-cublas-13` (commit e89b2db49).

**1. Removed `10`:** the list is now `{11, 12, 13}`. No torch has shipped CUDA 10 since ~1.12 (2022), and `CUDADriver` already refuses anything below CUDA 10.0, so `libcublas.so.10` was effectively dead weight.

**2. Refreshed the stale comment.** It previously said torch's build is `libcublas.so.11`, which is no longer true (torch defaults to CUDA 12/13 now). New comment:

```cpp
/* Load a versioned cuBLAS SONAME (e.g. libcublas.so.12) rather than the
 * unversioned libcublas.so. In a PyTorch environment torch has already loaded
 * its own cuBLAS; matching a versioned SONAME reuses torch's copy, whereas
 * libcublas.so would pull in a second, conflicting system cuBLAS. The first
 * already-loaded match wins, so in practice we bind to whichever cuBLAS torch
 * loaded. torch currently ships CUDA 12/13 builds (older torch shipped CUDA
 * 11), so we try 11, 12, and 13.
 */
cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 13});
```

I kept the order `{11, 12, 13}` rather than reversing it, for the reasons discussed above (the already-loaded first loop makes torch-compat order-independent, and lower-first is the safer default for the standalone case).

hughperkins: @ agent:
- remove 'The first
   * already-loaded match wins, so in practice we bind to whichever cuBLAS torch
   * loaded. torch currently ships CUDA 12/13 builds (older torch shipped CUDA
   * 11), so we try 11, 12, and 13.', because the former is redundant, and the latter will get stale
- wrap the comment to 120c, not 80c

hughperkins: Agent says:

Done - pushed 154654709.

- Removed the redundant "first already-loaded match wins" sentence and the stale "torch currently ships CUDA 12/13" sentence.
- Rewrapped the remaining comment to 120c (lines are now 113/114 chars).

```cpp
  /* Load a versioned cuBLAS SONAME (e.g. libcublas.so.12) rather than the unversioned libcublas.so. In a PyTorch
   * environment torch has already loaded its own cuBLAS; matching a versioned SONAME reuses torch's copy, whereas
   * libcublas.so would pull in a second, conflicting system cuBLAS.
   */
  cublas_loaded_ = try_load_lib_any_version("cublas", "64_", {11, 12, 13});
```

hughperkins: @ agent:
- you say ' the more principled approach is what load_cusolver()/load_cusparse() already do - derive the candidates from the detected driver version ({cuda_version, cuda_version - 1}) rather than hardcoding a list. cuBLAS is the odd one out using a fixed list, precisely because it prioritizes matching torch's loaded lib over the driver's version.'
- does this mean we cannot both use the more principled appraoch of getting candidates from detect driver version AND match torch's loaded lib?

hughperkins: Agent says:

No - they are not mutually exclusive. You can do both. The current code only makes it look either/or because a single list feeds two different jobs inside `try_load_lib_any_version`:

1. **Phase 1 - scan for an already-loaded SONAME** (this is the torch-match). It iterates the *whole* list and returns the first version already loaded in the process, regardless of order.
2. **Phase 2 - fresh `dlopen`** in list order until one loads (this is where a driver-derived preference belongs), then a bare `libcublas.so` fallback.

Both phases read the same `versions_to_try`, so today you can't give them different sets.

**Why you can't just copy the cusolver/cusparse pattern verbatim:** those use `{cuda_version, cuda_version - 1}`, a narrow 2-wide window around the *driver's* max CUDA version (`cuDriverGetVersion` reports the newest CUDA the driver supports, not what torch pinned). The driver is usually newer than torch's cuBLAS. Concretely: datacenter driver supports CUDA 13, but the user has torch 2.6 on cu118, so torch has `libcublas.so.11` loaded. Driver-derived candidates `{13, 12}` don't include `11`, so Phase 1 misses torch's cuBLAS and Phase 2 loads a second, conflicting one - exactly the failure the broad fixed list avoids.

**How to get both:**

- *Minimal, works with the current function:* pass a union list, driver-derived first (for the fresh-load preference) followed by the broad torch set (so Phase 1 still matches torch). E.g.
  ```cpp
  int v = CUDADriver::get_instance().get_version_major();
  cublas_loaded_ = try_load_lib_any_version("cublas", "64_", dedup({v, v - 1, 11, 12, 13}));
  ```
  Phase 1 scans every entry, so torch is always matched; Phase 2 tries the driver version first.
- *Cleaner, small refactor:* split `try_load_lib_any_version` into two lists - a broad "already-loaded scan" set (torch match) and a driver-derived "fresh-load preference" set. This also fixes the one rare corner the union list gets slightly wrong: if *both* the driver-version cuBLAS and torch's cuBLAS are loaded at once, a union list would prefer the driver one; a dedicated torch-scan list wouldn't.

So the "odd one out" wording was about the current single-list API, not a fundamental incompatibility. If you want, I can implement the union-list version (minimal contact area) or the two-list refactor - your call.

hughperkins: @ agent:
- add a FIXME comment fox this change

hughperkins: Agent says:

Done - pushed acc5357a5. Added a FIXME to the cuBLAS comment block capturing the driver-derived + torch-match idea:

```cpp
  /* Load a versioned cuBLAS SONAME (e.g. libcublas.so.12) rather than the unversioned libcublas.so. In a PyTorch
   * environment torch has already loaded its own cuBLAS; matching a versioned SONAME reuses torch's copy, whereas
   * libcublas.so would pull in a second, conflicting system cuBLAS.
   *
   * FIXME: the version list is hardcoded. Ideally we would both derive fresh-load candidates from the detected
   * driver version (like load_cusolver()/load_cusparse()) and still scan a broad set for torch's already-loaded
   * cuBLAS. That needs try_load_lib_any_version() to take separate "already-loaded scan" and "fresh-load" lists.
   */
```

Wrapped to 120c and ASCII-clean.

hughperkins: - remove ' That needs try_load_lib_any_version() to take separate "already-loaded scan" and "fresh-load" lists'.

hughperkins: Agent says:

Done - pushed 6870e58f9. Dropped the "That needs try_load_lib_any_version() to take separate ... lists" sentence and rewrapped to 120c:

```cpp
   * FIXME: the version list is hardcoded. Ideally we would both derive fresh-load candidates from the detected driver
   * version (like load_cusolver()/load_cusparse()) and still scan a broad set for torch's already-loaded cuBLAS.
```

hughperkins:
Thanks!

PR #898, opened 2026-08-31
[AMDGPU] Non-terminating in-kernel assert + OOB index clamp (arch-independent)

paveltc:
## API impact: none
Pure runtime/codegen fix. No new Python API, kwargs, config toggles, or env vars. The out-of-bounds index clamp lives entirely inside the already debug-gated `check_out_of_bound` pass and is scoped to AMDGPU, so non-debug runs and CPU/CUDA codegen are unchanged.

## Problem
The AMDGPU in-kernel assert used `asm("S_ENDPGM")`, which terminates only the faulting wavefront. Peers parked on `s_barrier` then deadlock, hanging the host on `hipStreamSynchronize`. Replacing it with `__builtin_trap()` fixes the CDNA hang but escalates to an **uncatchable SIGABRT on some AMD architectures (e.g. RDNA1 / gfx1011)**, taking the process down instead of raising a catchable error — so that approach can't be made uniform across AMD arches.

## Approach (architecture-independent, cooperative)
- **Non-terminating assert** (`runtime.cpp`): the error is already recorded before the arch-specific tail, so the AMDGPU branch now simply returns. The faulting wave runs to the kernel's natural end and reaches every barrier alongside its peers — no deadlock, no trap. The host raises `QuadrantsAssertionError` through the normal post-sync error check, and the **HIP context stays alive** (no more one-assert-per-process / dead-context limitation).
- **OOB index clamp** (`check_out_of_bound.cpp`, AMDGPU-gated): because a failed bounds check now falls through to the access, the offending index is clamped into `[0, size-1]` (`min(max(idx,0), size-1)`) for `ExternalPtrStmt` (ndarray), `GlobalPtrStmt` (field), and `MatrixPtrStmt`. This keeps the continued execution in bounds so there's no real out-of-bounds access (and thus no GPU memory fault / HSA exception). The assert **predicate and message still use the original index**; only the pointer's index operand is redirected to the clamped value.

No trap semantics, pinned host buffers, launch-failure hooks, or fences are needed, so there's nothing that diverges by architecture.

## Tests
`tests/python/test_assert.py` gains three inline AMDGPU tests (no subprocess/SIGABRT/timeout harness needed anymore):
- `test_amdgpu_assert_barrier_no_hang` — thread 0 asserts while peers hit `block.sync`; must raise, not hang.
- `test_amdgpu_assert_context_survives` — after catching an assert, a follow-up kernel runs to completion with correct results.
- `test_amdgpu_assert_out_of_bound_no_fault` — OOB field + ndarray accesses raise the correct bounds error with no fault; context stays alive.

## Local validation (MI308X, gfx942 / CDNA3)
- Full `test_assert.py`: 22 passed on cpu+amdgpu, no hang.
- Smoke checks: assert raises with correct message; barrier case raises (not a timeout); context survives; field/ndarray OOB raise the correct bounds message with no fault.

RDNA1 (gfx1011) validation needs the secret-gated AMD runner in upstream CI (see note below).

## Note on RDNA1 CI
The gfx1011/V520 job (`test_gpu.yml`) requires repo secrets that aren't exposed to fork PRs, so this PR alone won't exercise it. A companion branch on the upstream repo (as was done for the trap approach in #895) is needed to run the RDNA1 validation. This PR intentionally leaves #871 untouched.

### DIFF
--- a/docs/source/user_guide/debug.md
+++ b/docs/source/user_guide/debug.md
@@ -57,6 +57,8 @@ Assertions support constant strings and f-strings for the error message.
 
 Note: `assert` is compiled into the kernel but only checked at runtime when `debug=True`. On Linux ARM64, assertions are not currently supported.
 
+Note: a failed assertion is reported at the next synchronization point (kernel completion, a field/ndarray read, or `qd.sync()`), which raises `QuadrantsAssertionError`, rather than at the exact failing statement. On the AMDGPU backend the faulting GPU thread continues to the end of the kernel before the host raises (this keeps all threads reaching shared barriers together and avoids a hang), so statements after a failed assertion in the same kernel may still execute. Do not rely on `assert` to halt a kernel mid-execution; use it to detect and report invalid states.
+
 ## Performance impact
 
 Debug mode adds runtime checks to every field access and assertion, which can significantly slow down kernel execution. It is intended for development and debugging, not production use.
--- a/quadrants/runtime/amdgpu/kernel_launcher.cpp
+++ b/quadrants/runtime/amdgpu/kernel_launcher.cpp
@@ -531,8 +531,7 @@ void KernelLauncher::launch_llvm_kernel(Handle handle, LaunchContextBuilder &ctx
     const auto &parameter = kv.second;
     if (parameter.is_array) {
       const auto arr_sz = ctx.array_runtime_sizes[arg_id];
-      const bool is_host_external =
-          ctx.device_allocation_type[arg_id] == LaunchContextBuilder::DevAllocType::kNone;
+      const bool is_host_external = ctx.device_allocation_type[arg_id] == LaunchContextBuilder::DevAllocType::kNone;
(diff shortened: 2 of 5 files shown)


paveltc:
Hi @hughperkins — I pushed one more commit (e22df010) that's a one-line clang-format fix — unwraps a line that fit within the 120-col limit, whitespace-only, no behavior change. It fixes the Linters failure from the previous commit.

Since it's a fork PR, the Actions for that commit are sitting in "action_required" — could you hit "Approve and run workflows" when you have a moment? That'll turn Linters green and let the Test Linux AMD GPU job pick it up off a warm runner.

hughperkins:
triggered. (not that linters are not blocking for starting the amd runner, AFAIK. The dependency is:
- manylinux builds need to *all* finish
- => amd job will run, when:
     - there is an amd runner started
     - it is available (not being used by other jobs)

PR #900, opened 2026-08-31
[Caching] Enable per-offload caching for certain global stores

hughperkins:
## Summary
- The per-construct frontend split's recompute-safety guard (condition 3) rejected a recomputed field/ndarray load whenever **any** intervening construct performed a global write, regardless of address. In a multi-array kernel the intervening write almost always targets a *different* field, so the guard fired spuriously and disabled the split for the whole kernel. Concretely, qipc `_step_kernel` fell back to whole-kernel compile, so 6a's per-task artifact reuse never engaged on the actual target workload.
- This narrows clause (3) to fall back only when an intervening write **may alias** a recomputed read, using the shared alias analysis (`maybe_same_address`). A write is mapped to its target pointer (`GlobalStoreStmt::dest` / `AtomicOpStmt::dest`; null => structural/opaque => may-alias-all). Container statements are excluded from the write set (their leaves carry the real effects), and memory-free runtime intrinsics (barriers / shuffles / thread-index, via an allowlist) are dropped since they report a side effect but write no addressable memory.
- Conditions (1) and (2) are unchanged. The refined check is a strict **subset** of what the old check rejected, so it can only add splits, never change the result of an already-allowed one.

## Why this is safe
Clause (3) exists to stop a construct from recomputing a load whose value an intervening construct has since overwritten. That can only happen if some intervening writer's address may equal the read's address. We fall back whenever aliasing is `same` OR `uncertain` OR the writer is not a single comparable pointer (structural snode ops, external / real-func calls). Only a provably-`different` pointer clears the guard, preserving the "unknown implies unsafe" invariant.

## Validation (CUDA sm_120)
- **Correctness:** `_step_kernel` now splits (130 constructs / 242 tasks) with `q_sum=1.7492255125`, byte-identical to the whole-kernel path (forced fallback via `QD_SPLIT_MAX_COST_RATIO=0` gives the same `q_sum` to the last digit); `max_y_err=5.125e-07`.
- **Reuse:** on a warm one-offload edit, all 242 tasks hit the 6a artifact cache (`tasks[hit=242 recompiled=0]`); `compile_kernel` drops from 17.2s (whole-kernel fallback) to 4.9s. The residual is entirely frontend (`frontend_constructs[hit=0 recompiled=130]`) -- the cost 6b's construct manifest is designed to remove.
- **Tests:** `test_per_offload_cache.py` 48 passed (incl. two new disjoint-write kernels: different ndarray / different atomic -> SPLIT; existing same-field case still falls back). `test_offline_cache.py` 55 passed.

## Test plan
- [x] `test_per_offload_cache.py` (CPU + CUDA)
- [x] `test_offline_cache.py` (CUDA)
- [x] freefall byte-identical `q_sum` vs whole-kernel
- [x] warm-edit per-task artifact reuse (242/242 hit)


quadrants/transforms/split_frontend_per_construct.cpp
```diff
@@ -563,15 +617,19 @@ bool split_is_recompute_safe(Block *block) {
...
-        return false;  // an intervening construct mutates global memory the recomputed load would re-read
+    // (3) A recomputed read is unsafe only if a MAY-ALIAS write runs between the read's original slot and this
+    // construct -- recomputing the read there would then observe that write. A write to a provably different address
+    // (different SNode / ndarray, via the shared alias analysis) is harmless, which is what lets a multi-array kernel
+    // split. A write we cannot pin to one address, or a null read pointer, may-aliases everything.
+    for (const RecomputedRead &r : recomputed_reads)
+      for (const GlobalWrite &w : global_writes)
+        if (w.pos > r.pos && w.pos < b_lo)
+          if (w.dest == nullptr || r.src == nullptr || irpass::analysis::maybe_same_address(r.src, w.dest))
+            return false;
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed by adding a runtime no-alias guard with a graceful whole-kernel fallback, so aliasing stays a supported calling pattern (no new rule) while qipc keeps the split fast path.

- The split still assumes cross-arg ndarray disjointness where `alias_analysis()` reports "different args", but it now *records* that assumption as a single bit (`assumed_ndarray_disjoint`) on the compiled kernel. The bit is threaded through `PerConstructCache` -> codegen -> `LLVMCompiledKernel` / SPIR-V metadata and serialized to the offline cache, so a cache-loaded kernel re-arms its own guard.
- At launch, for a kernel whose bit is set, Python compares the backing-buffer identity (`Ndarray.device_alloc_id()`, i.e. the `DeviceAllocation` alloc_id) of every ndarray argument (including those nested in `@qd.data_oriented` structs). If any two share a buffer -- e.g. `kernel(a, a, y)` -- it runs a whole-kernel variant compiled with the split disabled, which snapshots the read before the intervening write and is correct under aliasing. Disjoint calls (the common case, and every real qipc launch) stay on the split.
- The whole-kernel variant is a normal compile under a distinct cache key (memoized base key + `.nosplit` suffix), built lazily on the first aliasing call and cached thereafter.

So `kernel_disjoint(a, a, y)` no longer reloads the mutated value: it transparently takes the whole-kernel path. Covered by `test_per_construct_frontend_split_alias_guard_falls_back` (asserts the aliased call returns the pre-write value on x64 + CUDA); full `per_offload_cache` suite green.

hughperkins:
@codex review

quadrants/transforms/split_frontend_per_construct.cpp
```diff
@@ -487,10 +488,123 @@ std::unordered_set<Stmt *> compute_construct_needed(
...
+  ExternalPtrStmt *ea = as_ndarray_ptr(a);
+  ExternalPtrStmt *eb = as_ndarray_ptr(b);
+  if (ea == nullptr || eb == nullptr)
+    return false;
+  auto *arg_a = ea->base_ptr->cast<ArgLoadStmt>();
+  auto *arg_b = eb->base_ptr->cast<ArgLoadStmt>();
+  if (arg_a == nullptr || arg_b == nullptr || arg_a->arg_id.empty() || arg_b->arg_id.empty())
+    return false;
+  if (arg_a->arg_id == arg_b->arg_id || ea->is_grad != eb->is_grad)
+    return false;
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in 8cdb27b62. You're right: the `is_grad` early-return left cross-arg primal/gradient aliasing unguarded (e.g. `kernel(a, a.grad, out)` where the kernel reads `x.grad` and writes `y`). That alias is between one arg's gradient buffer and another's primal buffer, which the slot-based launch guard cannot detect - it compares primal `alloc_id`s only.

Rather than extend the guard to resolve gradient buffers, `split_is_recompute_safe` now refuses to split when a condition-(3) clearance rests only on cross-arg primal-vs-gradient disjointness (`grad_mismatch`): it returns false, so the whole-kernel path runs and the result is always correct under any aliasing. Same-`is_grad` cross-arg pairs are still recorded as guardable pairs, so the common (qipc / forward) path is unaffected. This is conservative only for autodiff kernels with that exact recompute pattern, which then compile whole-kernel exactly as they did before the split feature existed - no impact on the perf targets.

python/quadrants/lang/kernel.py
```diff
@@ -628,6 +653,48 @@ def materialize(self, key: "CompiledKernelKeyType | None", py_args: tuple[Any, .
...
+        """
+        prog = impl.get_runtime().prog
+        kernel = t_kernel
+        if key in self._parse_only_keys:
+            runtime = impl.get_runtime()
+            instance_id, arg_features = self.mapper.lookup(self.raise_on_templated_floats, py_args)
+            used = self.used_py_dataclass_parameters_by_key_enforcing.get(key)
+            pruning = Pruning(kernel_used_parameters=used)
+            if used is not None:
+                pruning.enforce()
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in 8cdb27b62, with a regression test in 19de63431. Correct: the fastcache-restore fallback built `Pruning` from only the cached kernel-root used set and enforced immediately, so a callee `@qd.func` with a dataclass arg had an empty used set and its required fields were pruned, failing the rebuild.

`compile_no_split_variant` now runs the same two-pass discovery+enforce that `materialize` uses for a fresh compile - the non-enforcing discovery pass populates every callee's used set before enforcement - forcing `only_parse_function_def=False` so the callee bodies are actually walked. `graph_do_while_levels` is snapshotted/restored around the rebuild so the fastcache key's own launch path stays untouched. The new `test_..._fastcache_fallback_prunes_callee` forwards a dataclass arg into a `@qd.func` that reads one of its ndarray fields; it fails without the discovery pass (field pruned, build error) and passes with it.

hughperkins:
@codex review

python/quadrants/lang/_split_alias_guard.py
```diff
@@ -0,0 +1,95 @@
...
+    """Return the CompiledKernelData to launch, swapping in the whole-kernel variant when this call's actual buffers
+    violate the split's cross-parameter disjointness assumption.
+
+    For a guarded key whose args alias, lazily compile (once) and cache the split-disabled whole-kernel variant -- it
+    never recomputes across a write, so it is correct under aliasing -- and reset the observations to the no-split
+    sentinel. Disjoint calls (the common case) return ``compiled_kernel_data`` unchanged, staying on the fast split.
+    """
+    if not (kernel._split_alias_guard_by_key.get(key) and launch_has_aliased_ndarrays(kernel, key, args)):
+        return compiled_kernel_data
+    no_split = kernel._compiled_no_split_by_key.get(key)
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in 8ae1f8962. `Kernel.reset()` now clears `_compiled_no_split_by_key`, `_split_alias_guard_by_key`, and `_explicit_ndarray_slot_info_by_key` alongside `materialized_kernels`/`compiled_kernel_data_by_key`. The whole-kernel fallback is owned by the Program that `qd.reset()` destroys, so leaving it cached would let a post-reset aliased launch hand it to the new Program. All three caches are re-armed on the next materialize (fresh compile or fastcache restore).

Added a regression test (`test_per_construct_frontend_split_alias_guard_cleared_on_reset`) that arms the guard, re-inits the same arch, and reuses the SAME kernel across the reset. It asserts the fallback cache is cleared and that the rebuilt fallback stays correct. Verified on the cluster: it fails without the fix and passes with it on both cpu and cuda.

hughperkins:
@codex review

quadrants/transforms/split_frontend_per_construct.cpp
```diff
@@ -563,15 +698,36 @@ bool split_is_recompute_safe(Block *block) {
...
+    // (3) A recomputed read is unsafe only if a MAY-ALIAS write runs between the read's original slot and this
+    // construct -- recomputing the read there would then observe that write. A write to a provably different address
+    // (different SNode / ndarray, via the shared alias analysis) is harmless, which is what lets a multi-array kernel
+    // split. A write we cannot pin to one address, or a null read pointer, may-aliases everything.
+    for (const RecomputedRead &r : recomputed_reads)
+      for (const GlobalWrite &w : global_writes) {
+        if (w.pos <= r.pos || w.pos >= b_lo)
+          continue;
+        if (w.dest == nullptr || r.src == nullptr || irpass::analysis::maybe_same_address(r.src, w.dest))
+          return false;
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Confirmed and fixed in 8b2a892c9. You're right: `alias_analysis` proves two `ExternalPtrStmt`s different purely by their raw index diff (`alias_analysis.cpp` ~L129-135), and `handle_external_ptr_boundary` runs after the split, so at `split_is_recompute_safe` time the indices are still clamp-oblivious. A `boundary="clamp"` access can then collapse "different" indices onto one in-bounds address, and the split would recompute the clamped load past the write.

Fix: a new `clamp_may_defeat_index_disjointness(read, write)` helper forces the whole-kernel path whenever a condition-(3) clearance is between two accesses on the SAME buffer (same arg_id + same is_grad) and either side clamps. Scoped to same-buffer so multi-array splits are unaffected; distinct buffers never alias regardless of clamp (cross-arg aliasing remains the launch guard's job).

Verified on the cluster with a regression test (`..._clamped_boundary_recompute_unsafe`). It uses an out-of-bounds-high index (`a[_N]` clamps to `a[_N-1]`) since the frontend rejects negative literals. Without the fix the split fires (`frontend_constructs_total=2`) and would read the post-write value; with the fix it falls back (`-1`) and returns the pre-write snapshot. Fails without the fix and passes with it on both cpu and cuda.

hughperkins:
@codex review

quadrants/transforms/split_frontend_per_construct.cpp
```diff
@@ -563,15 +718,38 @@ bool split_is_recompute_safe(Block *block) {
...
+        // Cleared: the read and write are provably different addresses. When that proof is cross-arg ndarray
+        // disjointness a caller can defeat it by aliasing the two params. A same-is_grad pair is checkable by the
+        // launch guard (compare the two args' primal alloc_ids) -- record it. A primal-vs-gradient cross-arg pair is
+        // not (the guard sees only primal alloc_ids), so refuse to split and let the whole-kernel path run instead.
+        int slot_a = 0, slot_b = 0;
+        bool grad_mismatch = false;
+        if (clearance_assumes_ndarray_disjoint(r.src, w.dest, slot_a, slot_b, grad_mismatch)) {
+          if (grad_mismatch)
+            return false;
+          disjoint_pairs.emplace(std::min(slot_a, slot_b), std::max(slot_a, slot_b));
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in f3af442e5. You're right - my earlier grad handling only rejected the is_grad *mismatch* case (one arg's grad vs another's primal) and still recorded gradient-vs-gradient cross-arg pairs. Since the launch guard resolves a slot to its arg's PRIMAL alloc_id only, it can't see two args' gradient companions sharing one allocation, so such a pair is unguardable.

I took the conservative option: `clearance_assumes_ndarray_disjoint` now sets `involves_grad = ea->is_grad || eb->is_grad`, and the caller refuses to split (whole-kernel path) whenever either side of the cleared cross-arg pair is a gradient access. Only primal-vs-primal pairs stay recorded/guardable. This subsumes the previous mismatch check.

Verified on the cluster with a regression test (`..._cross_arg_grad_pair_unsafe`: read `x.grad[0]` past a write to `y.grad[i]` on two distinct args). Without the fix the split fires and records the pair (`frontend_constructs_total=2`); with it the kernel falls back (`-1`) and records nothing. Fails without / passes with, on cpu and cuda. Full split/guard subset: 63 passed.

python/quadrants/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -220,6 +221,16 @@ def _transform_kernel_arg(
             if not result:
                 decl_type_func, type_args = obj
                 obj = decl_type_func(*type_args)
+            # Record the flat slot of a top-level ndarray param (positional_index set only from the kernel-arg loop, not
+            # dataclass-field recursion) for the launch-time no-alias guard. `obj` is the AnyArray from decl_ndarray_arg.
+            if positional_index is not None and type(argument_type) is ndarray_type.NdarrayType:
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in 74dae0753. Confirmed accurate: flattened dataclass ndarray fields get their own arg slots (via decl_ndarray_arg), but the guard only resolved top-level params (_explicit_ndarray_slot_info_by_key) and struct/template members (_struct_ndarray_launch_info_by_key), so a field slot resolved to None and every launch fell back to whole-kernel - safe, but it defeats the split for those kernels.

Fix: I thread the root positional arg index + attribute chain through the dataclass-field recursion in `_transform_kernel_arg` and record each ndarray field as `(slot, root_index, attr_chain)` in a new guard-only table (`_dataclass_ndarray_guard_info_by_key`). The guard resolves those slots with the existing `_resolve_struct_ndarray` walk (`args[root_index].<attr_chain>`), and `reset()` clears the table with the other guard caches. I kept it separate from `_struct_ndarray_launch_info_by_key` on purpose: that list also drives launch-arg *binding* via `_set_struct_ndarray_args`, and dataclass fields are already bound by the dataclass-arg walk, so reusing it would double-bind.

Verified on the cluster (cpu+cuda): a kernel with two dataclass ndarray fields the split assumes disjoint now keeps the split on a disjoint launch (no `_compiled_no_split_by_key` fallback) and correctly falls back when one buffer is bound to both fields; without the fix the disjoint launch always falls back. No regressions in function_def_transformer (6), py_dataclass (238 + 6 xfail), src_ll_cache (37), or the split subset (65).

hughperkins:
@codex review

python/quadrants/lang/ast/ast_transformers/function_def_transformer.py
```diff
@@ -220,6 +249,12 @@ def _transform_kernel_arg(
...
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in e9d84e00f. Correct - the recording gated on the annotation type (`type(argument_type) is NdarrayType`), so an ndarray-backed `qd.Tensor` param, which `_decl_and_create_variable` still turns into an external `AnyArray`, was excluded and its slot went unresolved -> always-fallback (safe, but loses the split).

I switched both the top-level and the dataclass-field recording to key off the generated object's external-tensor slot (`_external_tensor_arg_slot(obj) is not None`) rather than the annotation type. Non-ndarray args yield no slot and are skipped, so this only broadens to genuinely external-tensor-backed params. No resolution change was needed: `_alloc_of` / `_resolve_struct_ndarray` already unwrap `qd.Tensor` wrappers.

Regression test (`..._tensor_arg_slots`): two ndarray-backed `qd.Tensor` params the split assumes disjoint keep the split with the fix (no `_compiled_no_split_by_key`) and always fall back without it. Verified cpu+cuda; split subset now 67 passed, no regressions.

hughperkins:
@codex review

python/quadrants/lang/_split_alias_guard.py
```diff
@@ -0,0 +1,99 @@
...
+    that decides aliasing in the compiled kernel. ``None`` (unrecognized shape / unreadable id) means "cannot verify",
+    which the caller treats as a possible alias -> whole-kernel fallback (correct, just slower).
+    """
+    if nd is None:
+        return None
+    if type(nd) in _TENSOR_WRAPPER_TYPES:
+        nd = nd._unwrap()
+    arr = getattr(nd, "arr", None)
+    if arr is None:
+        return None
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Confirmed, and this is perf-only, not a correctness bug -- falling back to the whole-kernel variant is always sound. When a raw NumPy array / PyTorch tensor is passed directly to a `qd.Tensor` param, `_launch_alloc_id` has no `.arr`/`alloc_id` to read, so the guard cannot prove disjointness and conservatively selects the whole-kernel variant, defeating per-construct caching for that specialization.

Per the maintainer's direction we're deferring the code fix (resolving identity for directly-bound external arrays would mean threading the launcher's `set_arg_external_array_with_shape` binding identity into the guard). Instead this path is now:

- documented as an explicit fallback condition in the new user-facing doc `docs/source/user_guide/kernel_caching.md`, with the recommendation to pass `qd.ndarray` / ndarray-backed `qd.Tensor` objects to keep per-offload caching, and
- surfaced at runtime by a one-shot-per-specialization `[PER_OFFLOAD][FALLBACK]` warning that names the kernel (commit 3922dd64e).

So no functional change here; tracking the identity-resolution optimization as a follow-up.

hughperkins:
@codex review

quadrants/transforms/split_frontend_per_construct.cpp
```diff
@@ -487,10 +488,147 @@ std::unordered_set<Stmt *> compute_construct_needed(
...
+  ExternalPtrStmt *ea = as_ndarray_ptr(a);
+  ExternalPtrStmt *eb = as_ndarray_ptr(b);
+  if (ea == nullptr || eb == nullptr)
+    return false;
+  auto *arg_a = ea->base_ptr->cast<ArgLoadStmt>();
+  auto *arg_b = eb->base_ptr->cast<ArgLoadStmt>();
+  if (arg_a == nullptr || arg_b == nullptr || arg_a->arg_id.empty() || arg_b->arg_id.empty())
+    return false;
+  if (arg_a->arg_id == arg_b->arg_id)
+    return false;
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Good catch -- fixed in 981355f62. This is the same-argument analog of the cross-arg primal/grad P1, and I've taken the same conservative stance the maintainer chose there.

`clearance_assumes_ndarray_disjoint` returned `false` for a same-`arg_id` pair, treating "a param vs its own `.grad`" as true disjointness the caller can't defeat. But a caller can bind the grad companion onto the primal buffer (`a.grad = a` / `a._set_grad(a)`), so a same-index primal/grad pair that alias_analysis proves different only by `is_grad` then hits one address. The launch guard resolves a slot to its arg's PRIMAL alloc_id only, so it cannot see the grad companion -- the pair is unguardable.

Fix: added `grad_companion_may_alias` (same arg_id, differing `is_grad`), checked right after the clamp guard in the condition-(3) loop, and it refuses the split (whole-kernel fallback) -- mirroring the clamp and cross-arg-grad cases. Updated the stale comment on `clearance_assumes_ndarray_disjoint` accordingly.

Regression test `test_per_construct_frontend_split_same_arg_primal_grad_recompute_unsafe` verified on cluster (cpu+cuda): it fails without the check (split clears the pair, constructs>=2) and passes with it (constructs==-1).

hughperkins:
@codex review

quadrants/transforms/split_frontend_per_construct.cpp
```diff
@@ -694,7 +903,8 @@ bool maybe_split_frontend_per_construct(IRNode *ir,
     return false;
   if (block_has_concurrent_region(block))
     return false;  // concurrent constructs share one global-temp buffer; per-construct offload would alias offsets
-  if (!split_is_recompute_safe(block))
+  std::vector<int> assumed_disjoint_pairs;
+  if (!split_is_recompute_safe(block, assumed_disjoint_pairs))
     return false;
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in 4030db24f. You're right that the split was pure (bounded) frontend overhead where no per-task reuse tier exists.

`maybe_split_frontend_per_construct` now gates on an active reuse tier before splitting: the only such tier is the per-task artifact cache, which `artifact_tier` (codegen.cpp) keys on -- CUDA/AMDGPU with the offline cache on. I gate on `(arch == cuda || arch == amdgpu) && config.offline_cache`, which is equivalent by construction (the artifact dir, `pertask_artifact_dir_ref()`, is nonempty exactly when offline_cache was set at program construction -- llvm_program.cpp) and avoids coupling this transform to the LLVM codegen header. On CPU/Metal/Vulkan, and with offline_cache disabled, the kernel now stays on the whole-kernel path.

`QD_SPLIT_FORCE=1` overrides the gate (matching the existing `QD_SPLIT_MAX_COST_RATIO`/`QD_SPLIT_STATS` knobs) so the tests keep exercising the backend-agnostic split logic on CPU, and power users can opt in. A module fixture sets it for the existing tests, and a new `test_per_construct_frontend_split_gated_off_without_reuse_tier` covers the production gate.

Verified on cluster: CPU with offline_cache=False gates off (constructs==-1); CUDA with offline_cache=True splits (constructs>=2, correct results); full split subset 70 passed. User docs (kernel_caching.md, optimization_passes.md) now state per-offload is the default on CUDA/AMD GPUs and list the missing-reuse-tier case in the fallback table.

hughperkins:
@codex review

docs/source/user_guide/kernel_caching.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Which is better
+
+| | Advantage |
+|---|---|
+| Per-offload | Fast incremental recompiles. Editing one task in a large kernel (for example a big [`qd.graph.do_while`](graph.md) loop) reuses the other tasks' compiled code instead of recompiling the whole kernel. This is the win for large kernels you iterate on. |
+| Whole-kernel | Simplest and universal. It is always correct and is the only option for kernels the split cannot compile construct-by-construct equivalently. |
+
+Per-offload caching is **on by default on CUDA and AMD GPUs** with the [offline cache](init_options.md#offline_cache) enabled (itself the default), because that is where the per-task reuse tier lives. On other backends (CPU, Metal, Vulkan), or with the offline cache disabled, there is nowhere to reuse per-construct results, so Quadrants uses whole-kernel caching. Beyond that, Quadrants falls back to the whole-kernel path automatically and transparently whenever the split would not be equivalent, or cannot be verified as safe for a given launch. You never choose between them and the results are identical either way; the only observable difference is compile time.
```
hughperkins: delete ' because that is where the per-task reuse tier lives'

hughperkins: delete 'there is nowhere to reuse per-construct results, so'

docs/source/user_guide/optimization_passes.md
```diff
@@ -95,7 +95,7 @@ Once the kernel has been split into offloaded tasks, both CSE and the CFG optimi
 
 ## Under the hood: per-construct frontend compilation
 
-This section is for the curious; you never have to think about it to write kernels - the behavior below is transparent and on by default.
+This section is for the curious; you never have to think about it to write kernels - the behavior below is transparent, and on by default wherever a per-task cache can reuse the result (CUDA/AMD GPUs with the offline cache). For the user-facing view of what this buys you and when it falls back, see [Kernel caching](kernel_caching.md).
```
hughperkins: replace 'wherever a per-task cache can reuse the result (CUDA/AMD GPUs with the offline cache)' with 'for CUDA/AMD GPUs when offline cache is enabled'

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in fa043828b.

hughperkins: Agent says:

Done in fa043828b.

docs/source/user_guide/optimization_passes.md
hughperkins: Agent says:

Done in fa043828b.

docs/source/user_guide/kernel_caching.md
```diff
@@ -0,0 +1,43 @@
...
+
+## When it falls back to whole-kernel
+
+Some fallbacks are decided once, at compile time (the kernel never uses the split). Others are decided per launch (the kernel is split, but a particular call runs the whole-kernel variant instead).
+
+**Compile-time fallback** (this kernel stays whole-kernel):
+
+| Condition | Reason |
+|---|---|
+| No per-task reuse tier: CPU/Metal/Vulkan, or the offline cache disabled | There is nowhere to reuse per-construct results, so splitting would only add compile time. |
```
hughperkins: dont use the word `tier`.I don't think it's the right word here. Use some other more inutitive phrasing

docs/source/user_guide/kernel_caching.md
```diff
@@ -0,0 +1,43 @@
...
+| [Autodiff](autodiff.md) (gradient) kernels | Gradient computation is not construct-isolatable. |
+| A value carried between constructs | A local one construct builds up and another reads cannot be recomputed in isolation. |
```
hughperkins: This is not user-friendly. WRite it in a more intuitive way, and/or reference an example you give later in this doc.

docs/source/user_guide/kernel_caching.md
```diff
@@ -0,0 +1,43 @@
...
+| A snapshot read a later construct would re-read after an intervening write | Isolating it would read the overwritten value. Includes `boundary="clamp"` accesses, where an out-of-range index can collapse onto a written element. |
```
hughperkins: This is not user-friendly. WRite it in a more intuitive way, and/or reference an example you give later in this doc.

docs/source/user_guide/kernel_caching.md
```diff
@@ -0,0 +1,43 @@
...
+| Cross-argument gradient disjointness | Whether two gradient buffers alias cannot be checked at launch, so the split is refused. |
```
hughperkins: This is not user-friendly. WRite it in a more intuitive way, and/or reference an example you give later in this doc.

docs/source/user_guide/kernel_caching.md
```diff
@@ -0,0 +1,43 @@
...
+| Certain specialized kernels | For example task barriers or side-effecting constructs. |
```
hughperkins: list them exhaustively. 'for example' is not usable information.

docs/source/user_guide/kernel_caching.md
```diff
@@ -0,0 +1,43 @@
...
+
+**Runtime fallback** (a per-launch decision for a split kernel):
+
+| Condition | Reason |
+|---|---|
+| Two parameters the split assumed disjoint are bound to the same buffer | Recomputing a read across the aliased write could change the result, so the call uses the whole-kernel variant. |
+| A parameter's backing buffer cannot be read to confirm disjointness | For example a raw NumPy array or PyTorch tensor passed directly. The call conservatively uses the whole-kernel variant. Passing `qd.ndarray` / `qd.Tensor` objects avoids this. |
+
+A runtime fallback emits a one-time `[PER_OFFLOAD][FALLBACK]` warning naming the kernel. It is suppressible through the logging level (it uses the standard `warn` level).
```
hughperkins: I feel like we should at least have `debug` message, druing compilation for any of the compile-time fallback conditions.

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in 1cc168e9f. Dropped `tier` entirely; the row now reads "The backend has no per-task cache to reuse" and the intro paragraph no longer uses the word.

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in 1cc168e9f. Rewritten to "One construct produces a value a later construct consumes", with a concrete example inline: a running total built up in one top-level loop and read by the next.

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in 1cc168e9f. Rewritten to "A construct re-reads array data an earlier construct overwrote", explaining that compiling it in isolation would see the new value instead of the original (this also folds in the clamp case).

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in 1cc168e9f. Rewritten to "Safety would hinge on two array arguments not sharing memory, and one of them is a gradient", explaining that gradient overlap cannot be verified even at launch, so the split is refused.

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in 1cc168e9f. Dropped "for example"; the specialized row now lists the cases exhaustively: iterating an unstructured mesh (`qd.mesh`), containing a region explicitly scheduled to run concurrently, or compiled with a diagnostic mode enabled (line coverage or a compiler debug dump).

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in 1cc168e9f. Each compile-time fallback in maybe_split_frontend_per_construct now emits a debug-level line naming the kernel and the exact condition (autodiff / no reuse cache / mesh / concurrent / diagnostic mode / cost cap). split_is_recompute_safe threads the specific recompute reason out to the caller so those cases are named too (value carried, re-read-after-overwrite incl. clamp, primal/.grad companion, cross-arg gradient). Documented in kernel_caching.md next to the runtime warn note. (The per-launch runtime fallback already warns from Python.)

hughperkins:
@codex review

docs/source/user_guide/kernel_caching.md
hughperkins: When would this happen? You mean, if the cache hasn't been warmed up yet? I feel this is the nature of a cache. If this is the reason, we can drop this row.

docs/source/user_guide/kernel_caching.md
hughperkins: can we provide actual code xamples please. Link to the examples from these rows.

docs/source/user_guide/kernel_caching.md
hughperkins: ditto. provide example code please.

docs/source/user_guide/kernel_caching.md
hughperkins: provide a code example please.

docs/source/user_guide/kernel_caching.md
hughperkins: lets have one row for each of these.

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Good question, and it's not a warm-up thing. On CPU / Metal / Vulkan (or with the offline cache off) there is no per-task artifact cache at all, so per-offload caching is simply unavailable on those backends and they always use whole-kernel caching -- it never becomes available once things are 'warmed up'. Since the intro paragraph already states this ("On other backends ... Quadrants uses whole-kernel caching"), the row was redundant, so I dropped it. Done in d5979caad.

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in d5979caad. Added a runnable example (`running_total`: a running total accumulated in one top-level loop, read by the next) under 'Compile-time fallback examples', and the row now links to it. Verified on a GPU-class build that it falls back with exactly this reason (frontend_constructs_total == -1, debug line 'a value carried between constructs').

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in d5979caad. Added a runnable example (`stale_snapshot`: snapshot a[0], overwrite a, then reuse the snapshot) and linked the row to it. Verified it falls back (debug line 'a construct re-reads memory a later-ordered write may change').

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in d5979caad. Added a runnable example (`grad_snapshot`: two grad-enabled ndarray args a, b; snapshot a.grad[0], write b.grad, reuse) and linked the row to it. Verified it falls back with 'cross-argument gradient disjointness (uncheckable until launch)'. Note: grad params need the call-form annotation `qd.types.ndarray(dtype=qd.f32, ndim=1, needs_grad=True)` (the subscript form can't take kwargs).

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Done in d5979caad. Split into three separate rows: (1) mesh (`qd.mesh`) kernels, (2) a region explicitly scheduled to run concurrently, (3) a diagnostic mode enabled (line coverage or a compiler debug dump).

hughperkins:
@codex review

docs/source/user_guide/kernel_caching.md
```diff
@@ -0,0 +1,93 @@
...
+
+Each compile-time fallback logs a one-line `debug`-level message naming the kernel and the condition, so raising Quadrants' log level to `debug` tells you why a given kernel stayed whole-kernel. A runtime fallback instead emits a one-time `[PER_OFFLOAD][FALLBACK]` warning naming the kernel (at the standard `warn` level, so it is on by default and suppressible through the logging level).
+
+## Compile-time fallback examples
+
+Each kernel below trips one of the compile-time fallbacks above and stays whole-kernel. You can confirm it with [`per_offload_cache_observations`](optimization_passes.md#inspecting-the-split) (`frontend_constructs_total` is `-1`) or by reading the `debug`-level log line.
+
+### A value carried between constructs
+
+The first loop accumulates `total`; the second loop reads it. The second loop cannot be compiled on its own without also redoing the first.
```
hughperkins: could you, in a comment, not in the doc for now, summarize the reason why the second loop cannot be compiled on its own, and what would be involved in fixing this?

docs/source/user_guide/kernel_caching.md
```diff
@@ -0,0 +1,93 @@
...
+    total = 0.0
+    for i in range(x.shape[0]):    # construct 1: build up a running total
+        total += x[i]
+    for i in range(out.shape[0]):  # construct 2: reuse it
+        out[i] = total
+```
+
+### Re-reading overwritten data
+
+`first` snapshots `a[0]`, then the first loop overwrites `a`. Recomputing that snapshot inside the second loop would read the overwritten value instead of the original.
```
hughperkins, line 65: could you, in a comment, not in the doc for now, summarize the reason why re-reading overwritten data means the offload cannot be compiled on its own, and what would be involved in fixing this?

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

**Why the second loop can't be compiled on its own**

The split makes each top-level construct a self-contained offloaded task by *recomputing* the straight-line top-level values it reads, rather than routing them through a shared buffer. `total` breaks that model: it isn't a straight-line value, it's the running result of the first loop (a loop-carried reduction). The only way to reproduce `total` inside the second construct would be to re-run the entire first loop there -- which duplicates the first construct's work at runtime and breaks the 1:1 construct->cached-task reassembly. So `split_is_recompute_safe` condition (1) flags a loop-produced local shared across constructs as non-recomputable and falls back.

**What fixing it would involve**

Teach the split to *materialize* such carried values into a persisted cross-construct temp instead of recomputing them: the first construct writes `total` to a small temp buffer, the second reads it as an input. Each construct is then independently compilable, and editing one still recompiles only that one -- provided the temp is treated as a stable interface and its layout/type is folded into the per-task cache key so the two tasks stay compatible. The current design deliberately avoids persisted cross-construct buffers (pure recompute keeps tasks hermetic and cache keys simple), so it conservatively refuses instead. The same mechanism would also cover the re-read case.

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

**Why re-reading overwritten data blocks isolated compilation**

Same root cause as the loop-carried case, through the same recompute mechanism. `first = a[0]` is a straight-line top-level read, so the split recomputes it *inside* the second construct instead of routing it through a buffer. But at its original position `first` captured the ORIGINAL `a[0]`; the first loop then overwrites `a` (`a[i] = 0`). Recomputed inside the second construct -- which runs after the first loop's writes -- `a[0]` now reads the overwritten value (0), not the snapshot. `split_is_recompute_safe` condition (3) detects a may-alias write (the first loop, to the same array) sitting between the read's original slot and the consuming construct and refuses, because recomputing the read there would observe that write. (Clamped `boundary="clamp"` reads fold in here too: distinct indices can collapse onto one written element.)

**What fixing it would involve**

Same fix as the carried-value case: don't recompute the order-sensitive read -- *materialize* it at its original program point (before the overwrite) into a persisted temp the later construct consumes, or equivalently emit the snapshot as its own tiny construct that runs before the write and whose output is a temp. Either way it reintroduces a stable cross-construct buffer that must be keyed into the per-task cache -- exactly what the pure-recompute design avoids today.

docs/source/user_guide/kernel_caching.md
hughperkins: Agent says:

Added to the doc in bc6c2271b, as its own paragraph under the "A value carried between constructs" example:

> To support this, the split would need to let one construct hand the value to the next through a small buffer that persists between them, instead of rebuilding each construct from the kernel's arguments alone. That buffer becomes a shared interface the cache has to track (its type and layout) so that editing one construct can't silently break the other. It's a real amount of machinery for an uncommon case, so today Quadrants takes the simpler, always-correct route and caches the whole kernel.

I also corrected the intro sentence so it no longer implies the second loop is unfixable -- it now just says the split cannot compile it on its own (because `total` exists only once the first loop has run), which the new paragraph then qualifies.

python/quadrants/lang/_split_alias_guard.py
```diff
@@ -0,0 +1,186 @@
...
+    prog = impl.get_runtime().prog
+    compiled = t_kernel
+    if key in kernel._parse_only_keys:
+        runtime = impl.get_runtime()
+        instance_id, arg_features = kernel.mapper.lookup(kernel.raise_on_templated_floats, py_args)
+        fallback_name = f"{kernel.func.__name__}_c{kernel.kernel_counter}_{instance_id}_nosplit"
+        # The discovery pass rebuilds `graph_do_while_levels` in place (build_While appends as it walks); snapshot and
+        # restore so the fastcache key's own launches, which keep using the restored table, stay untouched.
+        saved_graph_do_while_levels = list(kernel.graph_do_while_levels)
+        pruning = Pruning(kernel_used_parameters=None)
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in 0041d2c16. `compile_no_split_variant` runs a real two-pass discovery, so it now sets `pruning.pass_0_ran = True` on pass 0 (mirroring `Kernel.materialize`), routing `_predeclare_struct_ndarrays` to the id-based prune branch instead of the fastcache flat-name branch.

Reproduced and verified on the cluster (cpu+cuda) with a new regression test, `test_per_construct_frontend_split_alias_guard_fastcache_fallback_predeclares_struct`: it fastcache-restores a kernel that reads an ndarray through a `@qd.data_oriented`/`qd.template()` arg, then aliases the top-level pair to force the whole-kernel fallback rebuild. Without the fix it fails with exactly `Ndarray ... not registered as a kernel parameter`; with the fix the full alias-guard suite (19 tests) passes.

hughperkins:
@codex review

quadrants/transforms/split_frontend_per_construct.cpp
```diff
@@ -562,16 +744,41 @@ bool split_is_recompute_safe(Block *block) {
...
-        return false;  // an intervening construct mutates global memory the recomputed load would re-read
+    // (3) A recomputed read is unsafe only if a MAY-ALIAS write runs between the read's original slot and this
+    // construct -- recomputing the read there would then observe that write. A write to a provably different address
+    // (different SNode / ndarray, via the shared alias analysis) is harmless, which is what lets a multi-array kernel
+    // split. A write we cannot pin to one address, or a null read pointer, may-aliases everything.
+    for (const RecomputedRead &r : recomputed_reads)
+      for (const GlobalWrite &w : global_writes) {
+        if (w.pos <= r.pos || w.pos >= b_lo)
+          continue;
+        if (w.dest == nullptr || r.src == nullptr || irpass::analysis::maybe_same_address(r.src, w.dest))
```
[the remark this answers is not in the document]
hughperkins: Agent says:

Fixed in 5698bf03d. You're right -- a whole-element read (`base = a[i]`, a bare `ExternalPtrStmt`) and a same-element component write (`a[j][c] = ...`, a `MatrixPtrStmt` over the same external buffer) were compared as `different`, so the split recomputed the snapshot after the write loop and observed the mutated value.

I took the normalization route rather than a blanket reject of the mixed case: a blanket reject would disable the split for common matrix/vector-ndarray kernels (e.g. qipc's step kernel). A new helper `whole_element_read_may_overlap_component_write` detects the mixed `ExternalPtr`/`MatrixPtr` pair, normalizes both to their external origins, and re-runs `maybe_same_address`. Matrix-vs-matrix and external-vs-external pairs keep their precise (component-aware) result, and cross-arg mixed pairs come back `different` here and remain the launch guard's concern.

Verified on the cluster (cpu+cuda) with a new regression test `test_per_construct_frontend_split_whole_element_vs_component_recompute_unsafe`: a `vec2` ndarray snapshots `base = a[0]`, a loop writes `a[i][0] = 2`, then a second loop reads `base[0]`. Without the fix the split runs (`frontend_constructs_total == 2`) and `base` recomputes to `2.0`; with the fix it falls back and `base` stays `7.0`. Full split+guard suite: 74 passed.

hughperkins:
@codex review

hughperkins:
ok, codex green ✅

hughperkins:
checked docs ok ✅

hughperkins:
so we have:
- [x] codex green
- [x] docs ok
- [x] comment/code ratio ok-ish
- [x] Genesis benchmarks ok
- [x] Genesis unit tests pass
- [x] CI green

hughperkins:
comment/code ratio ok-ish ✅

hughperkins:
runnning genesis unit tests on tests0

hughperkins:
Genesis unit tests passing:

Screenshot 2026-09-02 at 16 02 27

hughperkins:
Running gen bench on bench1

hughperkins:
CI green ✅

hughperkins:
Gen bench:

20260902_split_buf_1728

Seems positive, but suss that shows a benefit on accessors.

Since this is a new benchmark script, running on 8 shared gpus, rather than 1 exclusive 8-gpu node, will make a new bench_interleaved_shared.py script, that runs interleaved on 8 shared gpus, and see how that looks.

hughperkins:
running interleaved shared on bench3

hughperkins:
bench interleaved seem ok:

20260903_splitbug_inter_1014

PR #904, opened 2026-09-03
[Interop] Prototype Vulkan opaque-FD import for CUDA and AMDGPU

rglebovamd:
## Context

The Nyx AMD/ROCm integration needs an API-neutral way to import Vulkan external-memory allocations into the active Quadrants CUDA or AMDGPU backend. Nyx exports an opaque FD plus the full allocation size, logical payload size, and byte offset. The Genesis plugin should not select or call CUDA/HIP directly.

This is the quick prototype requested for API/design review. It is intentionally a draft, Linux-only, and not presented as the final production implementation.

## Prototype approach

- Adds `quadrants.interop.VkImport`.
- Imports `VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD` through CUDA Driver or HIP based on the active Quadrants backend.
- Maps only the requested logical range while importing the complete Vulkan allocation.
- Implements DLPack producer methods for PyTorch views.
- Provides `release_to_vulkan()` and `acquire_from_vulkan()` ownership-boundary methods.
- Uses transactional FD ownership: imports a duplicate and consumes the caller FD only after import and mapping both succeed.
- Rejects invalid ranges, shapes, dtypes, unsupported backends, and mismatched GPUs with actionable errors.
- Frees mapped memory and the imported external-memory object on `close()`.

## Validation performed

Using the Nyx integration on AMD and NVIDIA:

- CUDA and HIP opaque-FD import/read/write passed.
- 16-byte allocation / 12-byte logical payload passed.
- Nonzero byte offset mapping passed.
- DLPack tensor shape, dtype, and device passed.
- Failed-construction FD ownership passed.
- AMD-Vulkan-to-CUDA and NVIDIA-Vulkan-to-HIP mismatch rejection passed.
- Full Genesis -> Quadrants VkImport -> Nyx rendering passed on both backends.
- Unit metadata/ownership validation in this PR: 8 tests passed.

## Important prototype limitations / review questions

1. **Synchronization:** the prototype uses device-wide synchronization. What should the production stream/queue API be? Do we want imported/exported binary or timeline semaphores for Vulkan visibility?
2. **Implementation layer:** should the final external-memory import live in Quadrants C++ bindings rather than Python `ctypes`?
3. **DLPack lifetime:** the prototype requires the `VkImport` object to outlive every tensor view and uses a no-op DLPack deleter. What ownership model should the final API expose?
4. **Device selection:** CUDA currently uses device zero in the active process. The final version should use the exact initialized Quadrants device and ideally verify PCI/device identity against Vulkan.
5. **Platform scope:** this implements Linux opaque FDs only. Win32 external-memory handles need a separate ownership contract.
6. **API shape:** please review the constructor fields and the `release_to_vulkan()` / `acquire_from_vulkan()` names before the Nyx plugin pins them.

## Test

```bash
pytest tests/python/test_vk_import.py -q
```

The Nyx plugin PRs remain drafts until this API is finalized, released/pinned, and the real CUDA/AMD conformance matrix passes.

## Semaphore capability note

Both tested Vulkan devices report timeline-semaphore support and CUDA exposes the required external-semaphore functions. However, the ROCm 7.2 HIP headers explicitly document `hipImportExternalSemaphore`, `hipSignalExternalSemaphoresAsync`, and `hipWaitExternalSemaphoresAsync` as unsupported on Linux. The public API should therefore keep semaphores optional and retain a device/queue synchronization fallback for the initial AMD Linux path unless runtime probing shows the HIP documentation is stale.

### DIFF
--- a/docs/source/user_guide/interop.md
+++ b/docs/source/user_guide/interop.md
@@ -314,3 +314,50 @@ print(x.grad[0])  # 4.0
 | Direct pass-through | no | no | yes (as kernel arg) |
 
 The `copy` parameter is supported on `to_numpy()` and `to_torch()` for `ScalarField`, `MatrixField` (and `VectorField`), `StructField`, `qd.Tensor`, and all `Ndarray` types. See [Zero-copy interop via DLPack](#zero-copy-interop-via-dlpack) for the support matrix and lifetime rules.
+
+
+## Vulkan external-memory import
+
+`quadrants.interop.VkImport` imports a Linux Vulkan `VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD` allocation into the active CUDA or AMDGPU backend and exposes the mapped range through DLPack.
+
+```python
+import torch
+import quadrants as qd
+from quadrants.interop import VkImport
+
+qd.init(arch=qd.amdgpu)  # or qd.cuda
+
+mapping = VkImport(
+    handle=exported_fd,
+    allocation_size=allocation_size,
+    offset=byte_offset,
+    size=logical_size,
+    shape=(height, width, 3),
+    dtype="uint8",
+)
+tensor = torch.utils.dlpack.from_dlpack(mapping)
+```
+
+### Ownership contract
+
+- The constructor validates the logical range and shape before importing.
+- It imports an internal duplicate of the supplied FD.
+- The caller's original FD is consumed only after external-memory import and buffer mapping both succeed.
+- On constructor failure, the original FD remains caller-owned.
+- Keep the `VkImport` object alive for at least as long as every DLPack consumer.
+- Call `close()` only after all tensor views have been released.
+
+### Synchronization
+
+`release_to_vulkan()` completes compute-device work before Vulkan consumes the allocation. `acquire_from_vulkan()` completes the ownership boundary before compute reads Vulkan-written memory. The implementation uses device-wide synchronization and assumes the Vulkan queue is flushed by the producer/consumer call sequence.
+
+FIXME: Define specification-valid memory visibility in both directions and determine whether imported/exported binary or timeline semaphores are required instead of device-wide host synchronization.
+
+### Limitations
+
+- Linux opaque FDs only; Win32 handles are not implemented.
+- CUDA and AMDGPU only.
+- Vulkan and compute must select the same physical GPU.
+- CUDA device selection targets device zero in the active process.
+- The DLPack capsule uses a no-op deleter; the `VkImport` owner must outlive the tensor view.
+- This API is not yet a stable compatibility commitment.
--- a/python/quadrants/interop/_vk_import.py
+++ b/python/quadrants/interop/_vk_import.py
@@ -1,0 +1,352 @@
+"""Linux Vulkan opaque-FD import for API review and validation.
+
+The implementation uses ctypes and device-wide synchronization. It demonstrates the constructor, ownership, DLPack, and
+lifecycle contract; the synchronization FIXME is documented in the interop user guide.
+"""
+
+from __future__ import annotations
+
+import ctypes
+import os
+import sys
+from math import prod
+from typing import Any
+
+
+class _DLDevice(ctypes.Structure):
+    _fields_ = [("device_type", ctypes.c_int), ("device_id", ctypes.c_int)]
+
+
+class _DLDataType(ctypes.Structure):
+    _fields_ = [
+        ("code", ctypes.c_uint8),
+        ("bits", ctypes.c_uint8),
+        ("lanes", ctypes.c_uint16),
+    ]
+
+
+class _DLTensor(ctypes.Structure):
+    _fields_ = [
+        ("data", ctypes.c_void_p),
+        ("device", _DLDevice),
+        ("ndim", ctypes.c_int),
+        ("dtype", _DLDataType),
+        ("shape", ctypes.POINTER(ctypes.c_int64)),
+        ("strides", ctypes.POINTER(ctypes.c_int64)),
+        ("byte_offset", ctypes.c_uint64),
+    ]
+
+
+class _DLManagedTensor(ctypes.Structure):
+    pass
+
+
+_DL_DELETER = ctypes.CFUNCTYPE(None, ctypes.POINTER(_DLManagedTensor))
+_DLManagedTensor._fields_ = [
+    ("dl_tensor", _DLTensor),
+    ("manager_ctx", ctypes.c_void_p),
+    ("deleter", _DL_DELETER),
+]
+
+
+@_DL_DELETER
+def _noop_dlpack_deleter(_managed: ctypes.POINTER(_DLManagedTensor)) -> None:
+    # The Genesis plugin retains VkImport beside its tensor view.
+    return
+
+
+class _HandleUnion(ctypes.Union):
+    _fields_ = [("fd", ctypes.c_int), ("reserved", ctypes.c_void_p * 2)]
+
+
+class _ExternalMemoryHandleDesc(ctypes.Structure):
+    _fields_ = [
+        ("type", ctypes.c_int),
+        ("handle", _HandleUnion),
+        ("size", ctypes.c_ulonglong),
+        ("flags", ctypes.c_uint),
+        ("reserved", ctypes.c_uint * 16),
+    ]
+
+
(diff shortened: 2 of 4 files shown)


docs/source/user_guide/interop.md
```diff
@@ -314,3 +314,60 @@ print(x.grad[0])  # 4.0
 | Direct pass-through | no | no | yes (as kernel arg) |
 
 The `copy` parameter is supported on `to_numpy()` and `to_torch()` for `ScalarField`, `MatrixField` (and `VectorField`), `StructField`, `qd.Tensor`, and all `Ndarray` types. See [Zero-copy interop via DLPack](#zero-copy-interop-via-dlpack) for the support matrix and lifetime rules.
+
+
+## Prototype Vulkan external-memory import
```
hughperkins: please remove the word 'prototype' from the doc

docs/source/user_guide/interop.md
```diff
@@ -314,3 +314,60 @@ print(x.grad[0])  # 4.0
...
+
+`quadrants.interop.VkImport` is a Linux-only prototype for importing a Vulkan
+`VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD` allocation into the active CUDA or
```
hughperkins: .md docs should not be line-wrapped

docs/source/user_guide/interop.md
```diff
@@ -314,3 +314,60 @@ print(x.grad[0])  # 4.0
...
+allocation. `acquire_from_vulkan()` completes the current prototype's ownership
+boundary before compute reads Vulkan-written memory. The prototype currently
+uses device-wide synchronization and assumes the Vulkan queue is flushed by the
+producer/consumer call sequence.
+
+A production implementation must define specification-valid memory visibility
+in both directions and may require imported/exported binary or timeline
+semaphores rather than device-wide host synchronization.
+
+### Current limitations
```
hughperkins: remove 'current'

docs/source/user_guide/interop.md
```diff
@@ -314,3 +314,60 @@ print(x.grad[0])  # 4.0
...
+
+### Synchronization prototype
+
+`release_to_vulkan()` completes compute-device work before Vulkan consumes the
+allocation. `acquire_from_vulkan()` completes the current prototype's ownership
+boundary before compute reads Vulkan-written memory. The prototype currently
+uses device-wide synchronization and assumes the Vulkan queue is flushed by the
+producer/consumer call sequence.
+
+A production implementation must define specification-valid memory visibility
```
hughperkins: replace 'A production implementation' with 'FIXME:'

docs/source/user_guide/interop.md
```diff
@@ -314,3 +314,60 @@ print(x.grad[0])  # 4.0
...
+  buffer mapping both succeed.
+- On constructor failure, the original FD remains caller-owned.
+- Keep the `VkImport` object alive for at least as long as every DLPack consumer.
+- Call `close()` only after all tensor views have been released.
+
+### Synchronization prototype
+
+`release_to_vulkan()` completes compute-device work before Vulkan consumes the
+allocation. `acquire_from_vulkan()` completes the current prototype's ownership
+boundary before compute reads Vulkan-written memory. The prototype currently
```
hughperkins: remove 'currently'

python/quadrants/interop/_vk_import.py
```diff
@@ -0,0 +1,382 @@
+"""Prototype Linux Vulkan opaque-FD import for API review and validation.
+
+The implementation intentionally uses ctypes and device-wide synchronization.
```
hughperkins: comments should be 120c wrapped, not 80c

python/quadrants/interop/_vk_import.py
```diff
@@ -0,0 +1,352 @@
...
+class _ExternalMemoryBufferDesc(ctypes.Structure):
+    _fields_ = [
+        ("offset", ctypes.c_ulonglong),
+        ("size", ctypes.c_ulonglong),
+        ("flags", ctypes.c_uint),
+        ("reserved", ctypes.c_uint * 16),
+    ]
+
+
+_DTYPE: dict[str, tuple[int, int]] = {
```
hughperkins, line 81: can we add a comment saying what the two values are? I'm guessing:
- first number is a kind of virtual enum, where 0 is signed int, 1 is unsigned int, 2 is float?
- second number is number of bits?

python/quadrants/interop/_vk_import.py
```diff
@@ -0,0 +1,352 @@
...
+    "float16": (2, 16),
+    "float32": (2, 32),
+    "float64": (2, 64),
+}
+
+
+class VkImport:
+    """Import a Linux Vulkan opaque FD into the active CUDA or AMDGPU backend.
+
+    The constructor is transactional: it imports a duplicate of ``handle`` and
```
hughperkins, line 99: can we attach the constructor docstring to the __init__ rahter than the class?

python/quadrants/interop/_vk_import.py
```diff
@@ -0,0 +1,352 @@
...
+    "uint32": (1, 32),
+    "uint64": (1, 64),
+    "float16": (2, 16),
+    "float32": (2, 32),
+    "float64": (2, 64),
+}
+
+
+class VkImport:
+    """Import a Linux Vulkan opaque FD into the active CUDA or AMDGPU backend.
```
hughperkins, line 97: `into the active CUDA or AMDGPU backend` => `into the active backend.`

python/quadrants/interop/_vk_import.py
```diff
@@ -0,0 +1,352 @@
...
+        self._backend = ""
+        self._device_id = 0
+
+        if sys.platform != "linux":
+            raise NotImplementedError("VkImport prototype currently supports Linux opaque FDs only")
+
+        dtype = str(dtype).removeprefix("torch.")
+        if dtype not in _DTYPE:
+            raise TypeError(f"unsupported VkImport dtype: {dtype}")
+        shape = tuple(int(extent) for extent in shape)
```
hughperkins, line 126: looks like shape is thus only required to be `Iterable[int]`?

hughperkins: (so both list or tuple both ok for example, doesnt have to be tightly a tuple?)

python/quadrants/interop/_vk_import.py
hughperkins: would prefer the first int is an enum really I feel.

python/quadrants/interop/_vk_import.py
```diff
@@ -0,0 +1,352 @@
...
+        )
+
+    @staticmethod
+    def _close_if_open(fd: int) -> None:
+        try:
+            os.close(fd)
+        except OSError:
+            pass
+
+    def _initialize_cuda(self, handle: int, allocation_size: int, offset: int, size: int) -> None:
```
hughperkins, line 183: the method name sounds like it 'initialzies cuda', but in fact it is creating the mapped view, in CUDA, of the vulkan handle right?

In which case, let's rename to something like `_map_view_on_cuda`, or similar? (something that makes it clear we arent initializing cuda).

hughperkins: Agent says:

Agreed the current name is misleading - the method's real job is importing and mapping the Vulkan allocation, not "initializing CUDA."

One nuance: it does more than create the mapped view, so a plain `_map_view_on_cuda` would slightly undersell it. Today it:

1. `cuInit(0)` - actually initializes the CUDA driver
2. `cuDeviceGet(0)` + `cuDevicePrimaryCtxRetain` + `cuCtxSetCurrent` - retains/sets the primary context
3. `cuImportExternalMemory` - imports the external-memory object from the FD
4. `cuExternalMemoryGetMappedBuffer` - maps the buffer view
5. `os.close(handle)` - consumes the original FD

So "map view" covers step 4 but not the import (3) or the driver/context bring-up (1-2); import and map are two distinct CUDA calls.

Two options:

- Rename around the true purpose: `_import_external_memory_cuda` or `_import_and_map_cuda`.
- Or split it: a once-per-process `_ensure_cuda_context()` (cuInit + primary ctx) plus a per-import `_map_external_memory_cuda()`. That also surfaces that steps 1-2 are per-process, not per-import (cuInit is idempotent and the primary-ctx retain is refcounted, so running them on every construction is harmless but redundant).

Same treatment for `_initialize_hip`.

PR #905, opened 2026-09-03
[Caching] Fix per-construct split false fallback on ndarray-vs-non-ndarray mixed pointers

hughperkins:
## Problem

After #900 merged, qipc's giant `graph_do_while` `_step_kernel` stopped using the per-construct split: warm + single-edit time-to-first-step regressed from ~11s back to ~27s (a whole-kernel recompile), defeating the point of the split for the flagship kernel.

Root cause is the whole-element/component recompute-safety guard added in #900's final review round (`whole_element_read_may_overlap_component_write`). It correctly closes a real `alias_analysis` blind spot (a whole-element ndarray read `a[i]` vs a component write `a[j][c]=...` to the *same* ndarray, which `alias_analysis` wrongly treats as disjoint). But its conservative catch-all rejected **any** mixed `ExternalPtr`/`MatrixPtr` pair whose matrix side did not normalize to an external pointer:

```cpp
if (ea == nullptr || eb == nullptr)
  return true;  // <- wrong: treats field/alloca/global-temp writes as "may overlap"
```

Hard evidence from an instrumented build on a Blackwell node — the pair flagged for `_step_kernel`:

```
read  = extptr (ndarray arg 25, scalar)
write = matrixptr(origin=alloca)   <- a thread-local stack tensor, NOT an ndarray
```

A `MatrixPtr` over a local `alloca` / global temp / field is a different memory space than an ndarray read; `alias_analysis` already reports the pair `different` (checked before this guard runs). Overriding that to "may overlap" is a false positive.

## Fix

- `as_ndarray_ptr` now follows the matrix-ptr chain down to the underlying ndarray `ExternalPtrStmt` (handles higher-rank elements), returning null only when the base genuinely isn't an ndarray.
- The guard overrides `alias_analysis` **only when both sides are ndarray-backed**; otherwise it defers to `alias_analysis`'s verdict. The genuine whole-element-vs-component ndarray hazard still falls back.

## Validation (Blackwell node)

- New TDD regression test `..._whole_element_vs_non_ndarray_component_ok` **fails before / passes after**; the true-positive `..._whole_element_vs_component_recompute_unsafe` still falls back.
- `tests/python/test_per_offload_cache.py`: **78 passed, 3 skipped** (cpu+cuda).
- qipc `_step_kernel` warm + single-edit time-to-first-step: **26.8s -> 11.6s** (compile_kernel 21.7s -> 5.2s); cold 32.5s -> 19.8s.
- qipc `test_abd_freefall.py`: **4 passed** (split output correct).

### DIFF
--- a/quadrants/transforms/split_frontend_per_construct.cpp
+++ b/quadrants/transforms/split_frontend_per_construct.cpp
@@ -566,13 +566,19 @@ bool internal_func_is_memory_free(const std::string &name) {
 
 // The ndarray (external) access a load/store pointer resolves to, directly or through a MatrixPtr element; nullptr if
 // it isn't one.
+// The ndarray external pointer this pointer ultimately indexes into, following the matrix-ptr chain (a component
+// access `a[i][c]` is a MatrixPtrStmt over the element's ExternalPtrStmt; higher-rank elements nest further). Null when
+// the base is not an ndarray -- a local alloca, a global temp, or a field -- which lives in a different memory space.
 ExternalPtrStmt *as_ndarray_ptr(Stmt *p) {
-  if (p == nullptr)
+  while (p != nullptr) {
+    if (auto *e = p->cast<ExternalPtrStmt>())
+      return e;
+    if (auto *mp = p->cast<MatrixPtrStmt>()) {
+      p = mp->origin;
+      continue;
+    }
     return nullptr;
-  if (auto *e = p->cast<ExternalPtrStmt>())
-    return e;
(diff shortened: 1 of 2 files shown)


hughperkins:
@codex review

hughperkins:
- [x] code reivew ok
- [x] checked fixes qipc main split
- [x] ci green
- [ ] check doc
- [x] check ratio comments/code
- [ ] gen tests
- [ ] gen benchmarks

PR #908, opened 2026-09-10
[MISC] Revert "[Caching] Enable per-offload caching for certain global stores (#900)"

hughperkins:
This reverts commit 9e8bee91604d30eb9c9984db93ae78ac4a17f5cb.

Issue: #

### Brief Summary

copilot:summary

### Walkthrough

copilot:walkthrough


hughperkins:
Going to run this revert. Since re-architecting per-offload caching. We can always re-apply this PR later, if we need. The revert removes 1400 lines, and adds only 62, so seems a net win for concision.

